PR #1, opened 2025-02-09
Add support for RETURNING operator for ORCA

kooooootb:
At this time RETURNING operator is unsupported by GPDB-side query-to-dxl translator and ORCA
This PR adds this support

What was made:
- in case of nonempty returning, query's returningList was used to create in dxl query tree LogicalProject node that contains projection rules and Logical dml query node (LogicalInsert, etc)
- OutputColumns field in dxl query for dml queries was changed to contain expressions, requested under RETURNING operator, these expressions should be provided by previously created LogicalProject, in case of no RETURNING it will be empty as in this case it doesnt return anything
- PhysicalDML node's property computing methods were changed to return all modified columns (except CTAS case where behaviour was unchanged), currently it outputs only changed columns
- Another ProjList was added to generated plan in dml node which shows columns that dml node can return to parent node
- Table descriptor in dml node of generated plan was tightened to only include used columns, this was caused by MakeDXLTableDescr function, which was used to include in descriptor columns with the same ColIds as in output columns array
- nodeDML in executor was changed to return modified tuples and set returning projections, similarly to ModifyTable node
- also several edge cases (e.g. dml nodes for CTAS, triggers (which as i understood aren't supported in GPDB but are supported in ORCA), several minor optimizations for empty returning etc) were handled

test samples in gporca were updated, most of edits include: 
- clearing OutputColumns field in Query for dml queries as said above,
- adding ProjList in dml node in plan,
- updating values for colIds and costs (Im not quite sure that it is valid, maybe shuffling plan nodes changed id generator behaviour)

Changes are quite massive but i think it is the most effective way of adding RETURNING support to ORCA

### DIFF
--- a/src/backend/executor/nodeDML.c
+++ b/src/backend/executor/nodeDML.c
@@ -45,158 +45,168 @@ ExecDMLExplainEnd(PlanState *planstate, struct StringInfoData *buf)
 TupleTableSlot*
 ExecDML(DMLState *node)
 {
-
-	PlanState *outerNode = outerPlanState(node);
-	DML *plannode = (DML *) node->ps.plan;
-
-	Assert(outerNode != NULL);
-
-	TupleTableSlot *slot = ExecProcNode(outerNode);
-
-	if (TupIsNull(slot))
+	for (;;)
 	{
-		return NULL;
-	}
+		PlanState *outerNode = outerPlanState(node);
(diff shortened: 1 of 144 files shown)


kooooootb:
Most of adds and removes (~2500 adds) are in test data, as i said Im not quite sure why colids changed in some cases, i runned tests with instructions from readme.md

PR #3, opened 2025-06-16
New GitHub Actions CI added

magf:
## Changes in image assembly

The assembly of the final image is divided into two stages. Base images for different OS are separately assembled from different Dockerfiles named as `ci/Dockerfile.<OS><OS_Version>`.
The final image is assembled from a single `Dockerfile` located in the root of the repository, to which the assembly parameters are passed.

## New CI added

- Added the ability to manually start the assembly of one image in any branch when pushing to a branch `.github/workflows/greengage-build-manual.yml`
- Added a call (`.github/workflows/greengage-build.yml`) to external CI instructions from the repository `greengagedb/greengage-ci/.github/workflows/greengage-reusable-build.yml@main`, which is triggered by tagging and PR


.github/workflows/greengage-build-manual.yml
```diff
@@ -0,0 +1,101 @@
...
+      target_os: ${{ steps.parse_os.outputs.target_os }}
+      target_os_version: ${{ steps.parse_os.outputs.target_os_version }}
+      python3: ${{ steps.parse_os.outputs.python3 }}
+      version: ${{ steps.parse_os.outputs.version }}
+    steps:
+      - name: Parse OS combination
+        id: parse_os
+        run: |
+          # Set version based on target_branch
+          case "${{ inputs.target_branch }}" in
```
RekGRpth, line 50: It is [not safe to use external variables in shell scripts](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections) in GitHub actions.

magf: Good point!
But we don't have external variables that can take variety random strings (injection risk).

All external variables are predefined and have type `choice` or `boolean`:

```yaml
inputs:
  target_branch:
    description: 'Target branch to checkout and build'
    required: true
    default: 'main'
    type: choice
    options:
      - '7.x'
      - 'main'
  os_combination:
    description: 'Target OS (MUST be compatible with Version)'
    required: true
    default: 'Ubuntu 22'
    type: choice
    options:
      - 'CentOS 7 (for 6.x only)'
      - 'Ubuntu 22'
      - 'RockyLinux 8 (for 7.x only)'
  logLevel:
    description: 'Log level'
    required: true
    default: 'warning'
    type: choice
    options:
      - info
      - warning
      - debug
  debug_mode:
    description: 'Enable debug mode'
    type: boolean
    default: false
 ```

File: .github/workflows/greengage-build-manual.yml [hunk 2]
```diff
@@ -0,0 +1,101 @@
...
+        options:
+          - '7.x'
+          - 'main'
+      os_combination:
+        description: 'Target OS (MUST be compatible with Version)'
+        required: true
+        default: 'Ubuntu 22'
+        type: choice
+        options:
+          - 'CentOS 7 (for 6.x only)'
```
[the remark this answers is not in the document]
magf: You are right, GitHub Actions uses main branch for manual workflow_dispatch by default.
This file for main branch needed only, but should content all possible features and combinations for all branches.

ci/README.md
```diff
@@ -4,10 +4,13 @@
 
 Change directory to gpdb sources destination. Make sure that directry doesn't contain binary objects from previous builds. Then run:
 for Ubuntu:
+
```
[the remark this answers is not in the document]
magf: This edition does not include changes for manual testing.
All old scripts and descriptions are left unchanged for backward compatibility.
The old build and test model still works.
This README only includes small cosmetic changes to improve the syntax and rename from readme.md (capitalized filename)

.github/workflows/greengage-build-manual.yml
magf: file removed

Stolb27 (commented):
> This is the most important discussion.
It is assumed that we will draw a boundary along the OS line and bring the differences between images to the OS level and we will have a single Dockerfile in the root of the repository, which does not depend on the OS in which it will be built. All the different parameters will be passed from the outside as arguments.

Removing OS specific part may be solved by modification of existing Dockerfiles by moving OS specific actions to the appropriate scripts (what you already made in the duplicating files). 

> This is a strategically important change and I ask you to pay close attention to it.

PR description doesn't contain a motivation to split something here and now. Please add it.

> It is also assumed that this image will be built once and not rebuilt, but this is a moot point.

It's a bad practice. We should implement a separate jobs to build such images periodically to save several minutes. But it's a further task.

PR #33, opened 2025-07-31
Add completion for the DISTRIBUTED clause in CREATE TABLE

stanislavlyalin:
Add tab completion for the DISTRIBUTED keyword after CREATE TABLE with a column
list in parentheses. After DISTRIBUTED, provide completion options for
"BY(", "REPLICATED", and "RANDOMLY".

Introduce ends_with_paren() helper for detecting if a word ends with a closing
parenthesis, and use it where appropriate.

DISTRIBUTED

### DIFF
--- a/src/bin/psql/tab-complete.c
+++ b/src/bin/psql/tab-complete.c
@@ -813,6 +813,7 @@ static char *pg_strdup_keyword_case(const char *s, const char *ref);
 static PGresult *exec_query(const char *query);
 
 static void get_previous_words(int point, char **previous_words, int nwords);
+static bool ends_with_paren(const char *word);
 
 #ifdef NOT_USED
 static char *quote_file_name(char *text, int match_type, char *quote_pointer);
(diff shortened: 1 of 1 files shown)


RekGRpth:
Can you further improve the patch so that it suggests columns in the BY expression?

andr-sokolov:
@stanislavlyalin , can you add support for all GPDB-specific keywords which are used in the CREATE TABLE query?
https://techdocs.broadcom.com/us/en/vmware-tanzu/data-solutions/tanzu-greenplum/6/greenplum-database/ref_guide-sql_commands-CREATE_TABLE.html

src/bin/psql/tab-complete.c
```diff
@@ -4244,6 +4260,21 @@ get_previous_words(int point, char **previous_words, int nwords)
...
 }
 
+/*
+ * Returns true if the given word ends with a right parenthesis ')'.
+ * Returns false if the input is NULL or the word is empty.
+ * This function is typically used to check if a SQL token ends with ')',
+ * for context-sensitive tab completion.
+ */
+static bool
+ends_with_paren(const char *word)
```
andr-sokolov, line 4270: Do we really need this function? It seems that it is impossible that word is NULL or empty string

stanislavlyalin: This function provides a more readable alternative to checking `prev_wd[len - 1] == ')'`. Since this check is performed in several places, I decided to move it into a separate function. The NULL and non-empty string checks are necessary to safely evaluate the expression `word[len - 1] == ')'`.

PR #40, opened 2025-08-06
ADBDEV-7916: Fix double-cleanup SIGSEGV when coordinator encounters OOM

red1452:
Fix double-cleanup SIGSEGV when coordinator encounters OOM

When a coordinator backend encounters an OOM error during a distributed query,
it triggers a cleanup sequence that can execute twice or more times, causing a
use-after-free crash.

The sequence of events is as follows: during normal execution cleanup in
`standard_ExecutorEnd()`, the system calls `cdbdisp_destroyDispatcherState()` to
clean up the dispatcher state. This function frees the results array, including
the `segdbDesc` pointers that reference executor backend connections.

However, an OOM error may occur inside already occuring cleanup since we are
foolishly trying to allocate more memory, `AbortTransaction()`, then
`cdbcomponent_recycleIdleQE()` will be called later. This abort calls
`cdbdisp_cleanupDispatcherHandle()` to perform cleanup operations again. The
second cleanup attempt tries to read data from executor backends using the
`segdbDesc` pointers that were already freed during the first cleanup in
`checkDispatchResult()` to cancel the query, resulting in a segmentation fault
when dereferencing invalid memory.

Additionally, the `numActiveQEs` counter can become negative during reentrant
calls to `cdbcomponent_recycleIdleQE()`, causing assertion failures.

When a backend encounters an OOM error, it will throw an error and abort the
transaction. If the backend happens to be a coordinator, it would also attempt
to cancel the distributed query first, by reading results from executor
backends. It is futile, since the gang would be discarded anyways if
`ERRCODE_GP_MEMPROT_KILL` is encountered, which is a sign of OOM.

Fixes:

- Move `numActiveQEs` decrement after a `goto`.

- Prevent dispatcher handle cleanup from cancelling the dispatch in case of
  OOM. It is instead done by `DisconnectAndDestroyAllGangs()`, based on the
  new global variable.

- Avoid `ABORT` dispatch for distributed transactions when the new flag is
  active. Without a response from coordinator, the segments will eventually
  trigger `ABORT` themselves.

- Whether gang is destroyed depends on the error code. Specify error code for
  runaway cleaner cancellations, so they are treated equivalently to OOM
  conditions.

- Unit tests are modified to avoid mocking `in_oom_error_trouble()` for
  every `palloc()` call.

- The test makes sure there aren't any `palloc()` calls at all if the server
  is currently handling an OOM condition.


src/backend/cdb/dispatcher/cdbdisp.c
```diff
@@ -520,6 +521,19 @@ cleanup_dispatcher_handle(dispatcher_handle_t *h)
...
+	if (h->dispatcherState == NULL)
+	{
+		destroy_dispatcher_handle(h);
+		return;
+	}
+
+	h->dispatcherState->forceDestroyGang = true;
+	cdbdisp_destroyDispatcherState(h->dispatcherState);
+}
+
```
RekGRpth, line 536: Is it possible to combine similar non-public functions `cleanup_dispatcher_handle` and `destroy_dispatcher_state_and_handle`?

src/backend/cdb/dispatcher/cdbdisp.c
```diff
@@ -540,7 +554,10 @@ AtAbort_DispatcherState(void)
 	 * Cleanup all outbound dispatcher states belong to
 	 * current resource owner and its children
 	 */
-	CdbResourceOwnerWalker(CurrentResourceOwner, cdbdisp_cleanupDispatcherHandle);
+	if (ERRCODE_TO_CATEGORY(elog_geterrcode()) != ERRCODE_INSUFFICIENT_RESOURCES)
+		CdbResourceOwnerWalker(CurrentResourceOwner, cdbdisp_cleanupDispatcherHandle);
+	else
+		CdbResourceOwnerWalker(CurrentResourceOwner, cdbdisp_destroyDispatcherHandle);
```
RekGRpth, line 560: `pgindent` your new changes or maybe separate the callback into a separate variable that is set by the condition above

RekGRpth: or
```suggestion
	CdbResourceOwnerWalker(CurrentResourceOwner,
	ERRCODE_TO_CATEGORY(elog_geterrcode()) != ERRCODE_INSUFFICIENT_RESOURCES ?
	cdbdisp_cleanupDispatcherHandle :
	cdbdisp_destroyDispatcherHandle);
```

src/backend/cdb/dispatcher/cdbdisp.c
```diff
@@ -589,6 +606,25 @@ cdbdisp_cleanupDispatcherHandle(const struct ResourceOwnerData *owner)
...
+		curr = next;
+		next = curr->next;
+
+		if (curr->owner == owner)
+		{
+			destroy_dispatcher_state_and_handle(curr);
+		}
+	}
+}
+
```
RekGRpth, line 627: Is it possible to extract the common parts of similar functions `cdbdisp_cleanupDispatcherHandle` and `cdbdisp_destroyDispatcherHandle` into one?

src/backend/cdb/dispatcher/cdbgang.c
```diff
@@ -714,6 +714,34 @@ getCdbProcessesForQD(int isPrimary)
...
+
+	/* destroy cdb_component_dbs, disconnect all connections with QEs */
+	cdbcomponent_destroyCdbComponents();
+
+	if (resetSession)
+		resetSessionForPrimaryGangLoss();
+
+	ELOG_DISPATCHER_DEBUG("DestroyAllGangs done");
+}
+
```
RekGRpth, line 744: Is it possible to extract the common parts of similar functions `DestroyAllGangs` and `DisconnectAndDestroyAllGangs` into one?

src/backend/cdb/cdbutil.c
```diff
@@ -957,13 +954,26 @@ cdbcomponent_recycleIdleQE(SegmentDatabaseDescriptor *segdbDesc, bool forceDestr
 			 lastWriter = cell, cell = lnext(cell)) ;
 
 		if (lastWriter)
+		{
+#ifdef FAULT_INJECTOR
+			if (SIMPLE_FAULT_INJECTOR("cdb_freelist_append_oom") == FaultInjectorTypeSkip)
+			{
+				ereport(ERROR, (errcode(ERRCODE_GP_MEMPROT_KILL),
+						errmsg("Out of memory was emulated")
+				));
```
RekGRpth, line 963:
```suggestion
				ereport(ERROR, (errcode(ERRCODE_GP_MEMPROT_KILL),
						errmsg("Out of memory was emulated")));
```

src/backend/cdb/cdbutil.c
```diff
@@ -957,13 +954,26 @@ cdbcomponent_recycleIdleQE(SegmentDatabaseDescriptor *segdbDesc, bool forceDestr
...
 						 lastWriter, segdbDesc);
+		}
 		else
 			segdbDesc->segment_database_info->freelist =
 				lcons(segdbDesc, segdbDesc->segment_database_info->freelist);
 	}
 
+
+	/* update num of active and idle QEs */
+	DECR_COUNT(cdbinfo, numActiveQEs);
```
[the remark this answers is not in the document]
red1452: You may move decrement to the top and run the test. Test should fail.

src/backend/cdb/dispatcher/cdbdisp.c
```diff
@@ -540,7 +554,10 @@ AtAbort_DispatcherState(void)
 	 * Cleanup all outbound dispatcher states belong to
 	 * current resource owner and its children
 	 */
-	CdbResourceOwnerWalker(CurrentResourceOwner, cdbdisp_cleanupDispatcherHandle);
+	if (ERRCODE_TO_CATEGORY(elog_geterrcode()) != ERRCODE_INSUFFICIENT_RESOURCES)
```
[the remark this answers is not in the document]
red1452: Error category `ERRCODE_INSUFFICIENT_RESOURCES` contains some operation system errors:
```
# (PostgreSQL-specific error class)
53000    E    ERRCODE_INSUFFICIENT_RESOURCES                                 insufficient_resources
53100    E    ERRCODE_DISK_FULL                                              disk_full
53200    E    ERRCODE_OUT_OF_MEMORY                                          out_of_memory
53300    E    ERRCODE_TOO_MANY_CONNECTIONS                                   too_many_connections
53400    E    ERRCODE_CONFIGURATION_LIMIT_EXCEEDED                           configuration_limit_exceeded
53500    E    ERRCODE_GP_MEMPROT_KILL                                        gp_memprot_kill
```

I think, all of these error must be processed by the such way.

src/test/regress/sql/bfv_OOM.sql
```diff
@@ -0,0 +1,44 @@
...
+declare
+	rec1 record;
+begin
+	for rec1 in select * from test_table
+	loop
+		return next rec1;
+	end loop;
+end;
+$$
+language plpgsql;
```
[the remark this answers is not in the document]
red1452: segfault caused only when query contains select from view, view made select from function, function made select from table. I tryed to remove each of these steps, and segfault did not cause.

src/test/regress/sql/bfv_OOM.sql
```diff
@@ -0,0 +1,44 @@
...
+
+create view test_view as select t.* from test_function() t;
```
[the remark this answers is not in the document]
red1452: At normal case segfault does not cause, than table wiil be dropped at the end of the test.

src/test/regress/sql/bfv_OOM.sql
```diff
@@ -0,0 +1,44 @@
...
+from gp_segment_configuration
+where role = 'p' and content = -1;
+
+select * from test_view;
+
+select gp_inject_fault('cdb_freelist_append_oom', 'reset', dbid)
+from gp_segment_configuration
+where role = 'p' and content = -1;
+
+select * from test_view;
```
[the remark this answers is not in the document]
red1452: It needs to check, that after OOM connections will not be closed and query wiil be work.

src/test/regress/sql/bfv_OOM.sql
```diff
@@ -0,0 +1,41 @@
...
+
+-- start_matchsubs
+-- m/WARNING:.*Any temporary tables for this session have been dropped because the gang was disconnected/
+-- s/session id \=\s*\d+/session id \= DUMMY/gm
+-- end_matchsubs
+
+create temp table test_table (f1 text, f2 text, f3 text, f4 text)
+distributed by (f1);
+
+create function test_function() returns setof test_table as
```
andr-sokolov, line 18: All the objects created during the test should be dropped at the end

src/backend/cdb/cdbutil.c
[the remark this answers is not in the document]
red1452: Do you enable assertion?

src/test/regress/sql/bfv_OOM.sql
bandetto: Isn't

> If the table is temporary, automatical DROP at the session end will cascade to all dependent objects.

andr-sokolov: the table is dropped, but the function is not. But now I suggest to don't create a function
```
do $$                   
declare
  rec1 record;
begin
for rec1 in select * from test_table
loop
 null;
end loop;
end; $$;
```
To reproduce the error we need for loop over a table

src/test/regress/sql/bfv_OOM.sql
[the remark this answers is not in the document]
red1452: This function allows to use addtional connection at `segdbDesc`, which is tryed to add to Idle list as the last writer connection. It requires to reproduce the error.

andr-sokolov:
there is no re-entrance to cdbcomponent_recycleIdleQE when I run the test

src/backend/cdb/dispatcher/cdbgang.c
```diff
@@ -714,6 +714,43 @@ getCdbProcessesForQD(int isPrimary)
...
+	/* Destroy CurrentGangCreating before GangContext is reset */
+	if (CurrentGangCreating != NULL)
+	{
+		RecycleGang(CurrentGangCreating, true);
+		CurrentGangCreating = NULL;
+	}
+
+	/* cleanup all out bound dispatcher state */
+	if (needDisconnect)
+		CdbResourceOwnerWalker(CurrentResourceOwner, cdbdisp_destroyDispatcherHandle);
```
bandetto, line 729: `cdbdisp_cleanupDispatcherHandle()` changed to `cdbdisp_destroyDispatcherHandle()`, and this condition is triggered only when called from `DisconnectAndDestroyAllGangs()`. Is there any difference now?

Since we'll dropping the gang anyway, can this be unconditional? Why did we need `cdbdisp_cleanupDispatcherHandle()` here in the first place?

red1452: You are right, `cdbdisp_cleanupDispatcherHandle()` should be used here, because the `needDisconnect` flag propose to use it in case of normal disconnect and destroy gang.

src/backend/cdb/dispatcher/cdbdisp.c
```diff
@@ -365,8 +366,11 @@ cdbdisp_destroyDispatcherState(CdbDispatcherState *ds)
 	if (!ds)
 		return;
 #ifdef USE_ASSERT_CHECKING
-	/* Disallow reentrance. */
-	Assert (!ds->isGangDestroying);
+	/*
+	 * Disallow re-entrance. It may occur in case of OOM and recursive
```
andr-sokolov, line 370: We allow reentrance adding `elog_geterrcode() != 0`

andr-sokolov:
One of the purposes of the patch is to not allocate memory during OOM handling, but `lappend_int` is called in the case of the added test
```
#0  lappend_int (list=0x0, datum=3) at list.c:151
#1  0x000055a5640fb724 in cdbconn_termSegmentDescriptor (segdbDesc=0x55a56e095718) at cdbconn.c:110
#2  0x000055a5641696ba in cdbcomponent_recycleIdleQE (segdbDesc=0x55a56e095718, forceDestroy=1 '\001') at cdbutil.c:983
#3  0x000055a5641079d0 in RecycleGang (gp=0x55a56e0a2a80, forceDestroy=1 '\001') at cdbgang.c:1092
#4  0x000055a5640fdb3d in cdbdisp_destroyDispatcherState (ds=0x55a56e0a2a28) at cdbdisp.c:407
#5  0x000055a5640fde45 in cleanup_dispatcher_handle (h=0x55a56e0a2998, only_destroy=1 '\001') at cdbdisp.c:528
#6  0x000055a5640fdfad in cleanup_open_dispatcher_handles (owner=0x55a56e0a18b0, only_destroy=1 '\001') at cdbdisp.c:602
#7  0x000055a5640fdff5 in cdbdisp_destroyDispatcherHandle (owner=0x55a56e0a18b0) at cdbdisp.c:616
#8  0x000055a564088623 in CdbResourceOwnerWalker (owner=0x55a56e0a18b0, callback=0x55a5640fdfd8 <cdbdisp_destroyDispatcherHandle>) at resowner.c:1375
#9  0x000055a564088644 in CdbResourceOwnerWalker (owner=0x55a56cb22678, callback=0x55a5640fdfd8 <cdbdisp_destroyDispatcherHandle>) at resowner.c:1379
#10 0x000055a5640fdeb3 in AtAbort_DispatcherState () at cdbdisp.c:554
#11 0x000055a563a02286 in AbortTransaction () at xact.c:3308
#12 0x000055a563a02de8 in AbortCurrentTransaction () at xact.c:3904
#13 0x000055a563ea2a34 in PostgresMain (argc=1, argv=0x55a56c95c4e8, dbname=0x55a56c95c348 "postgres", username=0x55a56c95c328 "keremet") at postgres.c:5117
#14 0x000055a563e01005 in BackendRun (port=0x55a56c9847a0) at postmaster.c:4926
#15 0x000055a563e005ca in BackendStartup (port=0x55a56c9847a0) at postmaster.c:4590
#16 0x000055a563dfbde7 in ServerLoop () at postmaster.c:1975
#17 0x000055a563dfb1a4 in PostmasterMain (argc=6, argv=0x55a56c95a2f0) at postmaster.c:1544
#18 0x000055a563ce3782 in main (argc=6, argv=0x55a56c95a2f0) at main.c:245
```

src/backend/utils/mmgr/runaway_cleaner.c
```diff
@@ -194,12 +194,23 @@ RunawayCleaner_StartCleanup()
...
+				ereport(
+					ERROR,
+					(errcode(ERRCODE_GP_MEMPROT_KILL),
+					 errmsg(
+						 "Canceling query because of high VMEM usage. Used: %dMB, available %dMB, red zone: %dMB",
+						 VmemTracker_ConvertVmemChunksToMB(
+							 MySessionState->sessionVmem),
+						 VmemTracker_GetAvailableVmemMB(),
+						 RedZoneHandler_GetRedZoneLimitMB()),
+					 errprintstack(true)));
```
RekGRpth, line 213: This is not relevant to this patch, but if there is not enough memory, will the memory allocations for the error message still be successful?

RekGRpth:
Don't we need to cherry-pick the commit https://github.com/GreengageDB/greengage/commit/9cad39671258209b9f6c428d2c7779cb5705f234 first?

src/backend/cdb/dispatcher/cdbgang.c
```diff
@@ -1024,29 +1025,23 @@ GangOK(Gang *gp)
 void
 RecycleGang(Gang *gp, bool forceDestroy)
 {
-	int i;
+	int			i;
```
RekGRpth, line 1028: is this change needed?

src/backend/cdb/dispatcher/cdbgang.c
```diff
@@ -1070,6 +1066,7 @@ RecycleGang(Gang *gp, bool forceDestroy)
 
 		cdbcomponent_recycleIdleQE(segdbDesc, forceDestroy);
 	}
+
```
RekGRpth, line 1069: is this change needed?

src/backend/cdb/dispatcher/cdbdisp.c
```diff
@@ -364,15 +364,21 @@ cdbdisp_destroyDispatcherState(CdbDispatcherState *ds)
...
-	/* Disallow reentrance. */
-	Assert (!ds->isGangDestroying);
+	Assert(!ds->isGangDestroying || ds->forceDestroyGang);
 	ds->isGangDestroying = true;
 #endif
 
 	if (!ds->isExtendedQuery)
 	{
-		numNonExtendedDispatcherState--;	
+		numNonExtendedDispatcherState--;
```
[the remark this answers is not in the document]
bandetto: Why not? Ditto, for every similar comment.

src/backend/utils/mmgr/runaway_cleaner.c
bandetto: https://github.com/GreengageDB/greengage/blob/780483134615792d628581bf8626c418e69ca246/src/backend/utils/mmgr/mcxt.c#L163-L180

src/backend/cdb/dispatcher/cdbdisp.c
bandetto: Overhauled the solution.

Now there is a OOM flag, that is set on the first encounter of ERRCODE_GP_MEMPROT_KILL. It causes the dispatcher routines, mainly cdbdisp_destroyDispatcherState() and AtAbort_DispatcherState() to use shortcuts to make sure we never allocate any memory or throw elog(ERROR) while in OOM state.

We can't rely on elog_geterrcode(), since there are external BG worker code that does not rely on elog system. E.g. in ADCC. It executes queries in a PG_TRY() block, catches the errors, and only then calls AbortTransaction().

So, this is the most simple and convenient hack I could think of.

bandetto:
> Don't we need to cherry-pick the commit [9cad396](https://github.com/GreengageDB/greengage/commit/9cad39671258209b9f6c428d2c7779cb5705f234) first?

I don't think the commit has any relation to this patch.

src/backend/cdb/dispatcher/cdbdisp.c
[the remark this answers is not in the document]
RekGRpth: This patch is not about fixing formatting of code sections not related to the patch!
/ Then fix formatting in all code! /s=off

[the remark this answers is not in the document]
bandetto: But what's wrong with fixing the formatting around the changes introduces in this patch?

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -100,17 +100,25 @@ cdbconn_createSegmentDescriptor(struct CdbComponentDatabaseInfo *cdbinfo, int id
...
 	cdbconn_disconnect(segdbDesc);
 
+	cdbconn_freeSegmentDescriptor(segdbDesc);
+
+	/* Put QE identifier to free list for reuse. */
+	cdbs->freeCounterList = lappend_int(cdbs->freeCounterList, id);
+}								/* cdbconn_termSegmentDescriptor */
+
+void
+cdbconn_freeSegmentDescriptor(SegmentDatabaseDescriptor *segdbDesc)
```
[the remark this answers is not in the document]
bandetto: What do you mean by "no longer necessary to free memory"? The memory is freed regardless of the function used. The only change is lack of PQfinish() from cdbconn_disconnect() and the use of the freelist. Both of which are addressed by DisconnectAndDestroyAllGangs() in the end of AbortTransaction().

andr-sokolov:
palloc is called during OOM processing.
```
#0  malloc_and_store_metadata (size=65536) at memprot.c:409
#1  0x000055a138fa98fe in gp_malloc_internal (requested_size=65536) at memprot.c:455
#2  0x000055a138fa9a02 in gp_malloc (sz=65536) at memprot.c:486
#3  0x000055a138f978c5 in AllocSetAllocImpl (context=0x55a141328148, size=24, isHeader=0 '\000') at aset.c:1377
#4  0x000055a138f97e88 in AllocSetAlloc (context=0x55a141328148, size=24) at aset.c:1468
#5  0x000055a138fa3a2a in palloc (size=24) at mcxt.c:1373
#6  0x000055a138c128a5 in new_list (type=T_IntList) at list.c:75
#7  0x000055a138c12a6b in lappend_int (list=0x0, datum=0) at list.c:154
#8  0x000055a13902298e in cdbconn_termSegmentDescriptor (segdbDesc=0x55a14134bca0) at cdbconn.c:116
#9  0x000055a13908fc6e in cleanupComponentIdleQEs (cdi=0x55a14134a620, includeWriter=1 '\001') at cdbutil.c:631
#10 0x000055a13908fcfc in cdbcomponent_cleanupIdleQEs (includeWriter=1 '\001') at cdbutil.c:657
#11 0x000055a1390900cf in cdbcomponent_destroyCdbComponents () at cdbutil.c:785
#12 0x000055a13902e3ea in DisconnectAndDestroyAllGangs (resetSession=1 '\001') at cdbgang.c:741
#13 0x000055a138929514 in AbortTransaction () at xact.c:3432
#14 0x000055a138929dea in AbortCurrentTransaction () at xact.c:3907
#15 0x000055a138dc9a36 in PostgresMain (argc=1, argv=0x55a141221508, dbname=0x55a141221368 "postgres", username=0x55a141221348 "keremet") at postgres.c:5117
#16 0x000055a138d28007 in BackendRun (port=0x55a141249820) at postmaster.c:4926
#17 0x000055a138d275cc in BackendStartup (port=0x55a141249820) at postmaster.c:4590
#18 0x000055a138d22de9 in ServerLoop () at postmaster.c:1975
#19 0x000055a138d221a6 in PostmasterMain (argc=6, argv=0x55a14121f310) at postmaster.c:1544
#20 0x000055a138c0a784 in main (argc=6, argv=0x55a14121f310) at main.c:245
```
What about adding some inject fault to palloc or other function which allocates memory?

src/backend/access/transam/test/xact_test.c
```diff
@@ -16,6 +16,8 @@ mock_TransactionIdDidAbortForReader(TransactionId xid)
 	return xid > 100;
 }
 
+#define in_oom_error_trouble(...) (false)
```
andr-sokolov, line 19: Why do we need this macro and the `__wrap_in_oom_error_trouble` function below?

src/backend/access/transam/xact.c
```diff
@@ -3423,12 +3423,12 @@ AbortTransaction(void)
...
-	/*
-	 * If memprot decides to kill process, make sure we destroy all processes
-	 * so that all mem/resource will be freed
-	 */
-	if (elog_geterrcode() == ERRCODE_GP_MEMPROT_KILL)
+	/* It's an OOM error. Get rid of all gangs and their resources. */
+	if (in_oom_error_trouble())
+	{
 		DisconnectAndDestroyAllGangs(true);
+		reset_oom_flag();
```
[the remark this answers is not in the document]
andr-sokolov: this is a stack how we can get a second OOM before the error message is displayed by psql.
```
#1  0x0000555b278b0bd3 in AllocSetAllocImpl (context=0x555b2ea76178, size=24, isHeader=0 '\000') at aset.c:1169
#2  0x0000555b278b1f19 in AllocSetAlloc (context=0x555b2ea76178, size=24) at aset.c:1478
#3  0x0000555b278bdb87 in palloc0 (size=24) at mcxt.c:1392
#4  0x0000555b278bee63 in InitShortLivingMemoryAccounts () at memaccounting.c:617
#5  0x0000555b278c03f0 in InitMemoryAccounting () at memaccounting.c:1366
#6  0x0000555b278be406 in MemoryAccounting_Reset () at memaccounting.c:263
#7  0x0000555b276e3b96 in PostgresMain (argc=1, argv=0x555b2ea77a88, dbname=0x555b2ea778e8 "postgres", username=0x555b2ea778c8 "keremet") at postgres.c:5205
#8  0x0000555b27642007 in BackendRun (port=0x555b2eaa0800) at postmaster.c:4926
#9  0x0000555b276415cc in BackendStartup (port=0x555b2eaa0800) at postmaster.c:4590
#10 0x0000555b2763cde9 in ServerLoop () at postmaster.c:1975
#11 0x0000555b2763c1a6 in PostmasterMain (argc=6, argv=0x555b2ea75890) at postmaster.c:1544
#12 0x0000555b27524784 in main (argc=6, argv=0x555b2ea75890) at main.c:245
```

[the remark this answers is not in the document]
bandetto: The error message is printed at the line below, even before `AbortTransaction()` is called:
https://github.com/GreengageDB/greengage/blob/6f26515ac0cebd45ed52aa13847342309ca4ba66/src/backend/tcop/postgres.c#L5102
Which is before the line you sent.

src/backend/access/transam/test/xact_test.c
bandetto: To avoid mocking `in_oom_error_trouble()` for every `palloc()` call in unit tests.

silent-observer (approved):

PR #41, opened 2025-08-07
Add extra tuple length checks

dkovalev1:
We faced an issue when tuple length read from logical tape happened to be 1 followed by a SIGSEGV. It's not possible to trace down root cause for it, so adding more sanity checks for tuple length value. It should replace SIGSEGV with an error without the need to restart the executor.

Asserts are not enough in this case as the issue happened with a release build. This enhances existing checks in `readtup_heap()` and `writetup_heap()` and should not have measurable impact.

### DIFF
--- a/src/backend/utils/sort/tuplesort_mk.c
+++ b/src/backend/utils/sort/tuplesort_mk.c
@@ -2878,7 +2878,6 @@ writetup_heap(Tuplesortstate_mk *state, LogicalTape *lt, MKEntry *e)
 	uint32		tuplen = memtuple_get_size(e->ptr);
 	long		ret = tuplen;
 
-	Assert(tuplen > sizeof(uint32));
 	if (tuplen <= sizeof(uint32))
 	{
 		const char *sortMethod = "";


src/backend/utils/sort/tuplesort_mk.c
```diff
@@ -2878,6 +2878,18 @@ writetup_heap(Tuplesortstate_mk *state, LogicalTape *lt, MKEntry *e)
 	uint32		tuplen = memtuple_get_size(e->ptr);
 	long		ret = tuplen;
 
+	Assert(tuplen > sizeof(uint32));
```
RekGRpth, line 2881: this assert is redundant

src/backend/utils/sort/tuplesort_mk.c
```diff
@@ -2878,6 +2878,18 @@ writetup_heap(Tuplesortstate_mk *state, LogicalTape *lt, MKEntry *e)
...
+	if (tuplen <= sizeof(uint32))
+	{
+		const char *sortMethod = "";
+		const char *sortSpaceType = "";
+		long sortSpaceUsed = 0;
+		
+		tuplesort_get_stats_mk(state, &sortMethod, &sortSpaceType, &sortSpaceUsed);
+		elog(ERROR, "invalid tuple len %u. Sort method: %s, space type: %s, space used: %ld	",
```
RekGRpth, line 2889: use `ereport` instead

src/backend/utils/sort/tuplesort_mk.c
```diff
@@ -2899,13 +2911,25 @@ freetup_heap(MKEntry *e)
...
+
 static void
 readtup_heap(Tuplesortstate_mk *state, TuplesortPos_mk *pos, MKEntry *e, LogicalTape *lt, uint32 len)
 {
 	uint32		tuplen;
 	size_t		readSize;
 
 	Assert(is_under_sort_or_exec_ctxt(state));
+	
+	if (!is_len_memtuplen(len) && (memtuple_size_from_uint32(len) < sizeof(uint32)))
```
RekGRpth, line 2923: why not move these checks to `getlen`?

src/backend/utils/sort/tuplesort_mk.c
dkovalev1: Okay, removed

src/backend/utils/sort/tuplesort_mk.c
```diff
@@ -2878,6 +2878,18 @@ writetup_heap(Tuplesortstate_mk *state, LogicalTape *lt, MKEntry *e)
 	uint32		tuplen = memtuple_get_size(e->ptr);
 	long		ret = tuplen;
 
+	Assert(tuplen > sizeof(uint32));
+	if (tuplen <= sizeof(uint32))
```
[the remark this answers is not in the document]
dkovalev1: This is the only critical place in mk sort

PR #44, opened 2025-08-11
ADBDEV-7999 Make the gp_orphaned_files test stable

andr-sokolov:
Make the gp_orphaned_files test stable

In the case of a segfault on all segments the MPP operation can be cancelled on
some segments, because the cancel request can come faster than the segfault
happens on these segments. So the qe_exec_finished inject fault is reset
explicitly to avoid unexpected segfaults later. The output of the inject fault
resetting is ignored, because it's not a problem if we sometimes get an error.
The reason of the error is recovering after segfault. The segfault resets all
inject faults.
Suppress FTS during the test case 3 execution to avoid failovers.
Make a checkpoint to remove orphaned files from segments where segfault did
not happen.

Ticket: ADBDEV-7999

### DIFF
--- a/src/backend/utils/sort/test/Makefile
+++ b/src/backend/utils/sort/test/Makefile
@@ -2,6 +2,6 @@ subdir=src/backend/utils/sort
 top_builddir=../../../../..
 include $(top_builddir)/src/Makefile.global
 
-TARGETS=string_wrapper
+TARGETS=string_wrapper tuplesort_mk
 
 include $(top_builddir)/src/backend/mock.mk
--- a/src/backend/utils/sort/test/tuplesort_mk_test.c
+++ b/src/backend/utils/sort/test/tuplesort_mk_test.c
@@ -1,0 +1,219 @@
+#include <stdarg.h>
+#include <stddef.h>
+#include <setjmp.h>
+
+#include "cmockery.h"
+
+#include "postgres.h"
+#include "utils/elog.h"
+#include "utils/tuplesort.h"
+#include "utils/tuplesort_mk.h"
+
+#undef USE_ASSERT_CHECKING
+#include "../tuplesort_mk.c"
+
+
+struct LogicalTapeSet
+{
+	BufFile    *pfile;			/* underlying file for whole tape set */
+	long		nFileBlocks;	/* # of blocks used in underlying file */
+	bool		forgetFreeSpace;		/* if we need to keep track of free
+										 * space */
+	bool		blocksSorted;	/* is freeBlocks[] currently in order? */
+	long	   *freeBlocks;		/* resizable array */
+	long		nFreeBlocks;	/* # of currently free blocks */
+	long		freeBlocksLen;	/* current allocated length of freeBlocks[] */
+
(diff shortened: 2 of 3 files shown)


src/test/regress/sql/gp_orphaned_files.sql
```diff
@@ -280,6 +290,25 @@ select 1 from gp_dist_random('gp_id');
...
+-- request can come faster than the segfault happens on these segments. So we
+-- should reset the qe_exec_finished inject fault explicitly to avoid segfaults.
+select resetInjectFaults(0);
+select resetInjectFaults(1);
+select resetInjectFaults(2);
+-- Segments, where the segfault happended, are restarting at this moment. Wait
+-- for the segments to be ready to accept connections. Sometimes we can get the
+-- "FTS detected connection lost" error, so ignore the output.
+select 1 from gp_dist_random('gp_id');
+-- Recover if failover happened
```
whitehawk, line 304: Is it the failover due to FTS? Maybe in this case, just suppress FTS during the test execution with
```
SELECT gp_inject_fault_infinite('fts_probe', 'skip', dbid) ...
```
? And do not involve `gprecoverseg`, which might be time consuming and which may also alter the disc contents (as it performs the stuff like `pg_basebackup` and etc during segment recovery) and intervene the normal workflow of the orphaned files removal feature...

src/test/regress/sql/gp_orphaned_files.sql
```diff
@@ -330,7 +359,14 @@ select 1 from gp_dist_random('gp_id');
 -- Rollback the transaction to make it possible to run queries after the error
 rollback;
 
-select force_mirrors_to_catch_up();
+-- start_ignore
```
whitehawk, line 362: why do we need any change for the case when only single segment has a segfault? In this case there shouldn't be any race condition between segv and cancel signal from the Coordinator...

src/test/regress/sql/gp_orphaned_files.sql
```diff
@@ -280,6 +290,25 @@ select 1 from gp_dist_random('gp_id');
 -- Rollback the transaction to make it possible to run queries after the error
 rollback;
 
+-- start_ignore
+-- The MPP operation can be cancelled on some segments, because the cancel
+-- request can come faster than the segfault happens on these segments. So we
+-- should reset the qe_exec_finished inject fault explicitly to avoid segfaults.
+select resetInjectFaults(0);
```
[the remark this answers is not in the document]
andr-sokolov: It is necessary to reset inject faults with separate queries. When a segment is in recovery mode after segfault, we get an error, so execution of the query is terminated and inject faults are not reset on some segment

src/test/regress/sql/gp_orphaned_files.sql
andr-sokolov: Fixed

vbwagner (approved):
We can remove unneccessary begin/end lines in the function resetInjectFaults, by making its language SQL rather than plpgpsql. But it is pure cosmetic.

PR #49, opened 2025-08-26
Automatic modernization (Python 2.x to Python 3.x) and formatting of legacy python code.

SaelKimberly:
Hello everyone!

Perhaps, as part of a real modernization of the code base, it is worth introducing the use of Ruff as a linter and formatter for python code.
And move once and for all from python2 (dead for many years) to python3 (ideally 3.9+, since older ones are no longer supported by the community).

This PR is an attempt to automatically (by 99.9%) migrate to more or less modern code.

If it takes off, you can add checks for known errors and formatting problems to CI before building:

- `ruff check --silent --target-version py37 --select <error codes>`
- `ruff format --check --silent --target-version py37`

## Here are some reminders before you submit the pull request
- [ ] Add tests for the change
- [ ] Document changes
- [ ] Communicate in the mailing list if needed
- [ ] Pass `make installcheck`
- [ ] Review a PR in return to support the community


Stolb27:
Hello, @SaelKimberly! We are glad to see your contribution! Python 2 reached EOL 5 years ago.

As far I understand, you've checked the possibility to make cluster and proceed basic cluster operations. Let's run it on our CI. For this purpose, I ask you to separate code linting to including in the further patch. It helps us to concentrate on functional changes first. Can you make it?

The next step is expanding the test suite by running more tests. At least \@concourse_cluster tests should be brought up. We'll plan this work, but the similar help are welcome.

Suggested by you changes require modification not only in the core part of Greengage but in the user third-party utilities. E.g.: utilities that use built-in gppylib for cluster connection. So we need some time to collect feedback from other users about these changes.

In case of success, we'll be able to integrate this patch past November, taking in account internal working plan.

SaelKimberly:
Hello, @Stolb27! I have not built yet GreengageDB with these changes.
Perhaps `Ruff` should be tuned more granular to keep some pieces of legacy code intact, but this utility applies only *safe* fixes to codebase, so I don't know, if this needed or not.
The `modernize` utility also should not break the source code, but if some processes uses Python 2.x, it definitely may be broken since.

This PR is, first and foremost, proof of the fundamental possibility of automatic code migration (99.9% of code changes took less than 1 minute).
When I was still working with Greenplum, the approach to writing Python code in this project was quite sloppy, and maintaining such a codebase seemed very challenging to me even back then - precisely due to the significant lag behind standards adopted by the community. What really got to me was the use of Python2 in some parts of the project, despite its support having ended quite some time ago. It either shouldn't be there at all if it's not being used, or we should move away from using it.

SaelKimberly:
I see. As in PyPi docs of `pygresql`:
```
The following Python versions are supported:

PyGreSQL 4.x and earlier: Python 2 only

PyGreSQL 5.x: Python 2 and Python 3

PyGreSQL 6.x and newer: Python 3 only
```

And also, in [docs](https://pygresql.readthedocs.io/en/latest/contents/install.html)

```
The current version of PyGreSQL (*6.1.0*) has been tested with Python versions 3.7 to 3.13, and PostgreSQL versions 10 to 17.
```

Maybe, I don't know, version 4 of pygresql is a little bit outdated...

I think there are definitely no live packages for Python that require Python 2 to run, except for the really rare, battle-tested ones. PyGreSQL is certainly battle-tested, but the Python 2 requirement is also probably a security issue, and proof that the mammoths are alive.

Unless you're using Postgres 9 here, then I really agree it's may be necessary. In this case, of course, it is worth bypassing this package when migrating code.

KnightMurloc:
@SaelKimberly Hello again. We eventually came to our own solution for Python 3 support in the 6X branch while maintaining Python 2 support. See #221. In addition to updating PyGreSQL and applying futurize, it also took a lot of manual changes to bring the code to a working state, mainly related to the string types.
I think this PR can be closed.

PR #50, opened 2025-08-27
Enable isolation2 test: parallel_retrieve_cursor/status_check

red1452:
Enable isolation2 test: parallel_retrieve_cursor/status_check

It was disabled due to unstable output. Sometimes test was failed, because error
after creating parallel retrieve cursor was printed immediately, but should be
printed, when "rollback" is executed.

The flakiness was caused by the fact that the session hadn't been disconnected
in the previous test. But the previous test generates "query finish request to
process <PID>", which is also sent to session of the next test. It leads to
immediate processing all errors, which are generated by cursors. But at the
normal case errors should be processed at the end of transaction.

This patch adds command to quit all connection after previous test. With this
change one test wouldn't affect another.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,55 +1,93 @@
 # Greengage CI Workflow
 
-This repository contains the main CI pipeline for the Greengage project, orchestrating the build, test, and upload stages for containerized environments. The pipeline is designed to be flexible, with parameterized inputs for version and target operating systems, allowing it to adapt to different branches and configurations.
+This directory contains the CI pipelines for the Greengage project,
+orchestrating the build, test, and upload stages for containerized
+environments. The pipeline is designed to be flexible, with parameterized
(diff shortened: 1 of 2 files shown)


vbwagner (commented):
> In the PR description first sentence doesn't look grammatically correct.

I'd recommend to reword it following way:

The flakiness was caused by the fact that the session hadn't been disconnected in the previous test.

In the same  paragraph I suggest also following changes:

It lead to immediately processing -> It leads to immediate processing.

should be processing -> should be processed


> Patch adds quitting for all connections to the previous test. After that, the
previous test does not affect the next test.

I'd reword it as:  'This patch adds command to quit all connection after previous test
With this change one test wouldn't affect another"

to the previous test. After that, the
previous test does not affect the next test.

vbwagner (approved):
Now, everything seems OK

PR #56, opened 2025-09-10
Support foreign table column selection as a distribution key

whitehawk:
Support foreign table column selection as a distribution key

This patch provides means to mark a subset of columns of a foreign table as the 
distribution keys of the table. It allows to eliminate the Redistribute motion 
when, for example, the user wants to insert data from a Greengage 
hash-distributed table into a foreign table. Before this patch, the insert plan 
would contain Redistribute motion, as the foreign table was always assumed as 
randomly distributed. With this patch, the foreign table is handled as a 
hash-distributed table during planning, allowing the planner to remove 
Redistribute motion. 

This patch introduces new options for a column definition:
1. 'insert_dist_by_key', which expects a boolean value. It marks the column as
a distribution key.
2. 'insert_dist_by_key_weight', which expects a non-negative integer value. It
sets the weight of the column when the distribution key set contains more than
one column. The overall order of distribution keys is following:
a) all columns that have an explicit weight, sorted by their weights
ascendingly;
b) then all other columns in the order they were set during the table creation.

This feature works only for foreign tables that have 'mpp_execute' set to
'all segments' (either explicitly for the table or derived from SERVER or
FOREIGN DATA WRAPPER options).

Tests are provided in postgres_fdw. A small change in postgres_fdw is introduced
to support the new column options.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,55 +1,93 @@
 # Greengage CI Workflow
 
-This repository contains the main CI pipeline for the Greengage project, orchestrating the build, test, and upload stages for containerized environments. The pipeline is designed to be flexible, with parameterized inputs for version and target operating systems, allowing it to adapt to different branches and configurations.
+This directory contains the CI pipelines for the Greengage project,
+orchestrating the build, test, and upload stages for containerized
+environments. The pipeline is designed to be flexible, with parameterized
+inputs for version and target operating systems, allowing it to adapt to
+different branches and configurations.
+
+## ⚠️ Important Notice
+
+Whenever the list of **NAMES of required jobs** in the workflow (including any
+**reusable workflows**) is **added, removed, or renamed**, you must contact a
+repository administrator to update the **Branch Protection Rules** accordingly.
+Without this, new, deleted, or renamed jobs will not be recognized as required
+when checking Pull Requests.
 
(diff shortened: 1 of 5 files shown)


whitehawk:
> The specification says that if it is impossible to create a distribution for a column type, then we need to throw a warning and use a random distribution. But in the current implementation, we will get an error:
> 
> ```
> postgres=# CREATE FOREIGN TABLE ft9 (
>         a int,
>         c point OPTIONS(insert_dist_by_key 'true'),
>         b int                                   
> )
>         SERVER foreign_server
>         OPTIONS (schema_name 'public', table_name 't1', mpp_execute 'all segments');
> CREATE FOREIGN TABLE
> postgres=# insert into ft9 values (1, null, 1);
> ERROR:  data type point has no default operator class for access method "hash"
> HINT:  You must specify an operator class or define a default operator class for the data type.
> ```

Agree, changed to WARNING.

KnightMurloc:
I think it's also worth adding a test for the minimum and maximum values of `insert_dist_by_key_weight`. Just in case.

whitehawk:
> I think it's also worth adding a test for the minimum and maximum values of `insert_dist_by_key_weight`. Just in case.

added

dkovalev1 (approved):

PR #60, opened 2025-09-18
ADBDEV-8076: Make the gp_orphaned_files isolation2 test stable

andr-sokolov:
Make the gp_orphaned_files isolation2 test stable

In the case of a segfault on all segments the MPP operation can be cancelled on
some segments, because the cancel request can come faster than the segfault
happens on these segments. So the qe_exec_finished inject fault is reset
explicitly to avoid unexpected segfaults later. The output of the inject fault
resetting is ignored, because it's not a problem if we sometimes get an error.
The reason of the error is recovering after segfault. The segfault resets all
inject faults.
Suppress FTS during the test cases 2 and 3 execution to avoid failovers.
Make a checkpoint to remove orphaned files from segments where segfault did
not happen.
The idea of fixes is taken from 780483134615792d628581bf8626c418e69ca246.

Ticket:ADBDEV-8076

### DIFF
--- a/src/test/isolation2/sql/gp_orphaned_files.sql
+++ b/src/test/isolation2/sql/gp_orphaned_files.sql
@@ -171,6 +171,19 @@ begin
 end
 $$ language plpgsql;
 
+1: create or replace function resetInjectFaults(p_contentid int) returns void as
+$$
+begin
+  perform gp_inject_fault('qe_exec_finished', 'reset', dbid),
+          gp_inject_fault('checkpoint',       'reset', dbid)
+  from gp_segment_configuration
+  where role = 'p' and content = p_contentid; /**/
+end
+$$ language plpgsql;
+
+-- Skip FTS probes
+1: select gp_inject_fault_infinite('fts_probe', 'skip', 1);
+
 -- Test case 2.1
 -- Segfault on all segments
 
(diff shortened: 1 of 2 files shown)


vbwagner (approved):
I think that use of PL/pgSQL function is a bit of overkill. SQL languate function of one select statement would do. But it is pure stylistic note.
Also I found misleading that function definition and drop do appear  in output file, but function invocatiions are not. But it seems that it is neccessary because we cannot say beforehand which segments are already restarted and which are not.

whitehawk (approved):

src/test/isolation2/sql/gp_orphaned_files.sql
```diff
@@ -200,8 +213,21 @@ $$ language plpgsql;
 1: rollback;
 2: rollback;
 
+-- start_ignore
+-- The MPP operation can be cancelled on some segments, because the cancel
+-- request can come faster than the segfault happens on these segments. So we
+-- should reset the qe_exec_finished inject fault explicitly to avoid segfaults.
+1: select resetInjectFaults(0);
+1: select resetInjectFaults(1);
+1: select resetInjectFaults(2);
```
RekGRpth, line 222: Why couldn't you just execute the SQL request?

PR #61, opened 2025-09-19
Fix SIGSEGV when running the 'execute on initplan' function

whitehawk:
Fix SIGSEGV when running the 'execute on initplan' function

Problem description:
A query with an 'execute on initplan' function having a parameter that refers
the other relation in the query caused SIGSEGV during execution.

Root cause:
As the function was executed in the initplan, during the execution the value
of the parameter was not yet defined, as the initplan is processed before the
main plan.

Fix:
When appending an initplan node for the function scan, set the subplan's
'parent_root' to NULL, isolating it from plan params of outer querie's.
Now all such params don't get into the valid parameter list in
SS_finalize_plan(), and the planner emits an error for such a query.

### DIFF
--- a/src/backend/optimizer/plan/createplan.c
+++ b/src/backend/optimizer/plan/createplan.c
@@ -7241,7 +7241,14 @@ append_initplan_for_function_scan(PlannerInfo *root, Path *best_path, Plan *plan
 	subroot = (PlannerInfo *) palloc(sizeof(PlannerInfo));
 	memcpy(subroot, root, sizeof(PlannerInfo));
 	subroot->query_level++;
-	subroot->parent_root = root;
+	/*
+	 * We set parent_root to NULL here in order to isolate the initplan
+	 * from all params ('plan_params') of outer queries. Otherwise, we may
+	 * recognize the parameter of the initplan function, referring to the
+	 * outer query, as an eligible param. We will set it to 'root'
+	 * after SS_make_initplan_from_plan().
+	 */
+	subroot->parent_root = NULL;
 	/* reset subplan-related stuff */
 	subroot->plan_params = NIL;
 	subroot->init_plans = NIL;
@@ -7265,6 +7272,8 @@ append_initplan_for_function_scan(PlannerInfo *root, Path *best_path, Plan *plan
 	 * result rows in tuplestore instead of a scalar param
 	 */
 	prm = SS_make_initplan_from_plan(subroot, (Plan *)initplan, InvalidOid, -1, InvalidOid, true);
+
+	subroot->parent_root = root;
 
 	fsplan->param = prm;
 	fsplan->resultInTupleStore = true;
--- a/src/backend/optimizer/plan/createplan.c
+++ b/src/backend/optimizer/plan/createplan.c
@@ -7273,8 +7273,6 @@ append_initplan_for_function_scan(PlannerInfo *root, Path *best_path, Plan *plan
 	 */
 	prm = SS_make_initplan_from_plan(subroot, (Plan *)initplan, InvalidOid, -1, InvalidOid, true);
 
-	subroot->parent_root = root;
-
 	fsplan->param = prm;
 	fsplan->resultInTupleStore = true;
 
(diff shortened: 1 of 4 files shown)


src/test/regress/sql/function_extensions.sql
```diff
@@ -417,3 +417,41 @@ insert into test_ao3 values(1,2),(2,3),(3,4);
...
+return query
+with
+-- though cte1 is not used, we need it here to reproduce the SIGSEGV,
+-- without it the query below failed with 'invalid Datum pointer' error.
+cte1 as (select hostname from gp_segment_configuration),
+cte2 as (select a from test_table_cte where b = param_text_in)
+select param_text_in::text, param_int_in::int4 from cte2;
+end;
+$function$
+language plpgsql execute on initplan;
```
RekGRpth, line 450: You can simplify it a little like this
```suggestion
with
-- though cte1 is not used, we need it here to reproduce the SIGSEGV,
-- without it the query below failed with 'invalid Datum pointer' error.
cte1 as (select hostname from gp_segment_configuration),
cte2 as (select a from test_table_cte where b = param_text_in)
select param_text_in::text, param_int_in::int4 from cte2;
$function$
language sql execute on initplan;
```

whitehawk: Ok

src/test/regress/sql/function_extensions.sql
```diff
@@ -417,3 +417,41 @@ insert into test_ao3 values(1,2),(2,3),(3,4);
...
+-- start_ignore
+drop table if exists test_table_cte;
+drop table if exists test_table;
+-- end_ignore
+
+create table test_table_cte(a text, b text);
+insert into test_table_cte values('-', '-');
+create table test_table(a int, b text);
+
+create or replace function test_function(param_text_in text, param_int_in int4) returns
```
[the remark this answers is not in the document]
whitehawk: Ok

src/backend/optimizer/plan/createplan.c
```diff
@@ -7266,6 +7273,8 @@ append_initplan_for_function_scan(PlannerInfo *root, Path *best_path, Plan *plan
 	 */
 	prm = SS_make_initplan_from_plan(subroot, (Plan *)initplan, InvalidOid, -1, InvalidOid, true);
 
+	subroot->parent_root = root;
+
```
bandetto, line 7277: Is this assignment even needed? I don't see where `subroot->parent_root` could be used, `subroot` seems like a dummy for this one initplan-related call?

src/backend/optimizer/plan/createplan.c
```diff
@@ -7241,7 +7241,14 @@ append_initplan_for_function_scan(PlannerInfo *root, Path *best_path, Plan *plan
 	subroot = (PlannerInfo *) palloc(sizeof(PlannerInfo));
 	memcpy(subroot, root, sizeof(PlannerInfo));
 	subroot->query_level++;
-	subroot->parent_root = root;
+	/*
+	 * We set parent_root to NULL here in order to isolate the initplan
+	 * from all params ('plan_params') of outer queries. Otherwise, we may
+	 * recognize the parameter of the initplan function, referring to the
+	 * outer query, as an eligible param. We will set it to 'root'
+	 * after SS_make_initplan_from_plan().
```
RekGRpth, line 7249: Is this comment still relevant?

src/test/regress/sql/function_extensions.sql
```diff
@@ -417,3 +417,39 @@ insert into test_ao3 values(1,2),(2,3),(3,4);
...
+insert into test_table_cte values('-', '-');
+create table test_table(a int, b text);
+
+create function test_function(param_text_in text, param_int_in int4) returns
+table (param_text_out text, param_int_out int4) as
+$function$
+with
+-- though cte1 is not used, we need it here to reproduce the SIGSEGV,
+-- without it the query below failed with 'invalid Datum pointer' error.
+cte1 as (select hostname from gp_segment_configuration),
```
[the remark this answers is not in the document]
RekGRpth: I got `SIGSEGV` even with this simplification:
```sql
create function test_function(a text) returns table (t text) as $$
    select a;
$$ language sql execute on initplan;

--explain (costs off, verbose)
create table tnew as
select * from test_table l join test_function(l.a) r on l.a = r.t;
```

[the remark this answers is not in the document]
whitehawk: As with the `sql`-lang function the SIGSEGV is reproduced even without cte1, I think we can remove it.

bandetto (approved):

PR #64, opened 2025-09-23
Cleanup idle reader gang after utility statements

bimboterminator1:
Cleanup idle reader gang after utility statements

Reader gangs use local snapshot to access catalog, as a result, it will
not synchronize with the sharedSnapshot from write gang which will
lead to inconsistent visibility of catalog table on idle reader gang.
Considering the case:

select * from t, t t1; -- create a reader gang.
begin;
create role r1;
set role r1;  -- set command will also dispatched to idle reader gang

When set role command dispatched to idle reader gang, reader gang
cannot see the new tuple t1 in catalog table pg_auth.
To fix this issue, we should drop the idle reader gangs after each
utility statement which may modify the catalog table.

Reviewed-by: Zhenghua Lyu <zlv@pivotal.io>
(cherry picked from commit https://github.com/GreengageDB/greengage/commit/d1ba4da5ae27ffb7af80a464c0e3d62a5c540bca)


red1452 (approved):
Clear cherry-pick, that fixes the problem.

silent-observer (approved):

PR #65, opened 2025-09-23
ADBDEV-8371: Fix hash join spill file issue when rescanning

silent-observer:
Fix hash join spill file issue when rescanning

Previously in some cases when rescanning a hash join was required, and the
hash join was executed in multiple batches that were spilled to disk, several
issues occurred when rescan happened in the middle of a batch:
1. Outer batch files (which contain outer tuples for all the next batches)
were never cleared, and after rescan they were filled again. When
gp_workfile_compression is enabled, this caused a BufFile to be in an
unexpected state (reading instead of writing), causing an error.
2. When inner batch file is read into the in-memory hash table, it is deleted.
After processing the batch, the hash table is spilled again into the batch
file in case rescan is possible. However, if rescan happens in the middle of a
batch, the hash table was never spilled and its contents were lost, resulting
in missing data in the output.

To fix these issues, when rescan is executed without rebuilding the whole hash
table, clear all the outer batch files and also spill the current batch to an
inner batch file.

### DIFF
--- a/src/backend/executor/nodeHashjoin.c
+++ b/src/backend/executor/nodeHashjoin.c
@@ -1231,11 +1231,38 @@ ExecReScanHashJoin(HashJoinState *node)
 			 */
 			node->hj_OuterNotEmpty = false;
 
+			/*
+			 * Outer batch files have to be cleared before restarting the hash join,
+			 * because they will be written again when batch 0 is processed.
+			 *
+			 * TODO: it might be possible to optimize by saving batch 0 outer file too,
+			 * in which case we might avoid rescanning the outer plan again.
+			 */
+			if (node->hj_HashTable->outerBatchFile)
+			{
+				for (int i = 0; i < node->hj_HashTable->nbatch; i++)
+				{
+					if (node->hj_HashTable->outerBatchFile[i])
+						BufFileClose(node->hj_HashTable->outerBatchFile[i]);
+					node->hj_HashTable->outerBatchFile[i] = NULL;
+				}
+			}
+
 			/* ExecHashJoin can skip the BUILD_HASHTABLE step */
 			node->hj_JoinState = HJ_NEED_NEW_OUTER;
 
 			if (node->hj_HashTable->nbatch > 1)
 			{
+				/*
+				 * If we rescan in the middle of a batch, inner batch file for the current
+				 * batch is actually deleted, its contents only present in the hash table.
+				 * We must spill it to disk to not lose it.
+				 */
+				if (node->reuse_hashtable &&
+					node->hj_HashTable->innerBatchFile[node->hj_HashTable->curbatch] == NULL)
+				{
+					SpillCurrentBatch(node);
+				}
 				/* Force reloading batch 0 upon next ExecHashJoin */
 				node->hj_HashTable->curbatch = -1;
 			}
--- a/src/backend/executor/nodeHashjoin.c
+++ b/src/backend/executor/nodeHashjoin.c
@@ -1247,8 +1247,10 @@ ExecReScanHashJoin(HashJoinState *node)
 				for (int i = 0; i < hashtable->nbatch; i++)
 				{
 					if (hashtable->outerBatchFile[i])
+					{
 						BufFileClose(hashtable->outerBatchFile[i]);
-					hashtable->outerBatchFile[i] = NULL;
+						hashtable->outerBatchFile[i] = NULL;
+					}
 				}
 			}
 
--- a/src/backend/executor/nodeHashjoin.c
+++ b/src/backend/executor/nodeHashjoin.c
@@ -1197,6 +1197,8 @@ ExecHashJoinGetSavedTuple(HashJoinState *hjstate,
 void
 ExecReScanHashJoin(HashJoinState *node)
 {
+	HashJoinTable hashtable = node->hj_HashTable;  
+
 	/*
 	 * In a multi-batch join, we currently have to do rescans the hard way,
 	 * primarily because batch temp files may have already been released. But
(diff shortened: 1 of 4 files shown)


src/backend/executor/nodeHashjoin.c
```diff
@@ -1231,11 +1231,38 @@ ExecReScanHashJoin(HashJoinState *node)
...
+				/*
+				 * If we rescan in the middle of a batch, inner batch file for the current
+				 * batch is actually deleted, its contents only present in the hash table.
+				 * We must spill it to disk to not lose it.
+				 */
+				if (node->reuse_hashtable &&
+					node->hj_HashTable->innerBatchFile[node->hj_HashTable->curbatch] == NULL)
+				{
+					SpillCurrentBatch(node);
+				}
```
RekGRpth, line 1265:
```suggestion
				HashJoinTable hashtable = node->hj_HashTable;

				if (node->reuse_hashtable &&
					hashtable->innerBatchFile[hashtable->curbatch] == NULL)
				{
					SpillCurrentBatch(node);
				}
```

src/backend/executor/nodeHashjoin.c
```diff
@@ -1231,11 +1231,38 @@ ExecReScanHashJoin(HashJoinState *node)
...
+			 * in which case we might avoid rescanning the outer plan again.
+			 */
+			if (node->hj_HashTable->outerBatchFile)
+			{
+				for (int i = 0; i < node->hj_HashTable->nbatch; i++)
+				{
+					if (node->hj_HashTable->outerBatchFile[i])
+						BufFileClose(node->hj_HashTable->outerBatchFile[i]);
+					node->hj_HashTable->outerBatchFile[i] = NULL;
+				}
```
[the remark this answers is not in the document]
silent-observer: How about introducing `hashtable` variable at the start of the function, the way it's done in `ExecHashJoinNewBatch`? This whole function has a lot of mentions of `node->hj_HashTable`

src/test/regress/sql/bfv_joins.sql
```diff
@@ -575,10 +575,28 @@ set enable_seqscan=off;
...
+truncate r_table1;
+insert into l_table select i, i%100 from generate_series(1, 20000) i;
+insert into r_table1 select i%100, i from generate_series(1, 10000) i;
+insert into r_table1 select i%100, i from generate_series(1, 10000) i;
+analyze l_table;
+analyze r_table1;
+
+set statement_mem="1MB";
+explain analyze select * from r_table2 where ra2 in ( select ra1 from l_table join r_table1 on b = rb1);
+select * from r_table2 where ra2 in ( select ra1 from l_table join r_table1 on b = rb1);
```
KnightMurloc, line 589: What's the point of doing an explain analyze and then doing a select? Maybe just do an explain and then select?

src/test/regress/sql/bfv_joins.sql
```diff
@@ -575,10 +575,28 @@ set enable_seqscan=off;
...
+
+set gp_workfile_compression=on;
+select * from r_table2 where ra2 in ( select ra1 from l_table join r_table1 on b = rb1);
+reset gp_workfile_compression;
+
 reset optimizer;
 reset enable_nestloop;
 reset enable_bitmapscan;
 reset enable_seqscan;
+reset statement_mem;
```
KnightMurloc, line 599: Why not do it immediately after the test, just like with `gp_workfile_compression`?

src/test/regress/sql/bfv_joins.sql
silent-observer: Explain analyze output contains info about spill files created in the hash join, which are a necessary condition for this bug. If regular explain is used, we wouldn't know if the spill files are not created anymore, which invalidates the test.

RekGRpth: `gp_workfile_entries` or `gp_workfile_usage_per_segment` or `gp_workfile_usage_per_query` won't help in this case?

silent-observer: I think explain analyze better illustrates that it's the hash join node creating spill files, and that spill files are created for outer child too (which might not happen sometimes). Also, wouldn't using `gp_workfile_entries` require making this an isolation test, since spill files are temporary?

src/test/regress/sql/bfv_joins.sql
silent-observer: Done

src/backend/executor/nodeHashjoin.c
```diff
@@ -1231,30 +1233,57 @@ ExecReScanHashJoin(HashJoinState *node)
...
 			/* ExecHashJoin can skip the BUILD_HASHTABLE step */
 			node->hj_JoinState = HJ_NEED_NEW_OUTER;
 
-			if (node->hj_HashTable->nbatch > 1)
+			if (hashtable->nbatch > 1)
 			{
+				/*
+				 * If we rescan in the middle of a batch, inner batch file for the current
+				 * batch is actually deleted, its contents only present in the hash table.
+				 * We must spill it to disk to not lose it.
```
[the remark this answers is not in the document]
silent-observer: Done

src/backend/executor/nodeHashjoin.c
```diff
@@ -1231,30 +1233,57 @@ ExecReScanHashJoin(HashJoinState *node)
 			 */
 			node->hj_OuterNotEmpty = false;
 
+			/*
+			 * Outer batch files have to be cleared before restarting the hash join,
+			 * because they will be written again when batch 0 is processed.
+			 *
+			 * TODO: it might be possible to optimize by saving batch 0 outer file too,
+			 * in which case we might avoid rescanning the outer plan again.
```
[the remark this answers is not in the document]
silent-observer: Done

src/test/regress/sql/bfv_joins.sql
silent-observer: These functions seem to be used on top of `explain analyze` anyway. Here we need both a plan with spill info and query outputs, so we still end up with `explain analyze`. Also these queries still won't show whether spill files belong to outer child or inner child

KnightMurloc (dismissed):

src/backend/executor/nodeHashjoin.c
```diff
@@ -1231,30 +1233,60 @@ ExecReScanHashJoin(HashJoinState *node)
...
+			 * file too, in which case we might avoid rescanning the outer
+			 * plan again.
+			 */
+			if (hashtable->outerBatchFile)
+			{
+				for (int i = 0; i < hashtable->nbatch; i++)
+				{
+					if (hashtable->outerBatchFile[i])
+						BufFileClose(hashtable->outerBatchFile[i]);
+					hashtable->outerBatchFile[i] = NULL;
```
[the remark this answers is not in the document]
silent-observer: This was copied from `ExecHashJoinNewBatch` but alright

PR #76, opened 2025-09-30
ADBDEV-8456: Hide fault-injector macros

RekGRpth:
Hide fault-injector macros

Currently, not all fault-injector macros are hidden behind the define. For
example, in a recent patch 82b02d7. These macros should be hidden.

Ticket: ADBDEV-8456


vbwagner (approved):
Looks right. There were three functions where fault injections were added instead of dtm_debug checks, and two injection points in each functions. So all six points are protected by ifdeds.

PR #77, opened 2025-09-30
ADBDEV-8355: Convince the compiler that the variable was initialized

RekGRpth:
Convince the compiler that the variable was initialized

GPOS_ASSERT disappears on the build without asserts, and an error will be given:
Ctranslatordxltoplstmt.cpp: 625: 9: Error: ‘Rel_distr_policy_Expected’ May Uninitialized in this Function

It is necessary to add __builtin_unreachable, as is done in other places of the ORCA.

Ticket: ADBDEV-8355


silent-observer (approved):

vbwagner (approved):
Patch is exactly as we discussed and agreeed in mattermost.

PR #78, opened 2025-10-01
Fix the absence of auxiliary functions due to some isolation2 tests

whitehawk:
Fix the absence of auxiliary functions due to some isolation2 tests

Problem description:
Script 'setup.sql' is executed before the isolation2 test suite. It contains
lots of auxiliary functions that are used throughout the tests. But after
running the 'export_distributed_snapshot' test, the auxiliary functions from
'setup.sql' written in Python, like 'pg_ctl()', were no longer available.
That broke any test that attempted to use these functions if it was executed
after the 'export_distributed_snapshot' test.

Root cause:
'export_distributed_snapshot' test dropped cascade the 'plpython3u' language at
its start, and then recreated the language again. It caused all Python
functions created before that moment to be removed.
Also, the issue is relevant for tests 'resgroup/resgroup_dumpinfo' and
'dependency', that also drop the 'plpython3u' language. 

Fix:
Remove the 'drop language cascade' and 'create language' from the
'export_distributed_snapshot' test. The language is created in 'setup.sql' once
for all isolation2 tests. There is no need to re-create it.
Also, do the same for 'resgroup/resgroup_dumpinfo' test. But in the
'dependency' test the drop of the language is done by design, so it can't be
removed. For now we'll keep this test the last one in the schedule, so it
doesn't spoil any other test. A proper disclaimer is added to the schedule file.


hilltracer:
Should we also do something with this removal (maybe in a separate ticket)?
https://github.com/GreengageDB/greengage/blob/7a4c80d7e3dbab86cebda28365cb62d11eb10fe9/src/test/isolation2/sql/resgroup/resgroup_dumpinfo.sql#L101-L102

Where is [setup.sql](https://github.com/GreengageDB/greengage/blob/7.x/src/test/isolation2/sql/setup.sql) run?
Does it run for src/test/isolation2/isolation2_resgroup_v2_schedule also?

whitehawk:
> Should we also do something with this removal (maybe in a separate ticket)?
> 
> https://github.com/GreengageDB/greengage/blob/7a4c80d7e3dbab86cebda28365cb62d11eb10fe9/src/test/isolation2/sql/resgroup/resgroup_dumpinfo.sql#L101-L102
> 
> Where is [setup.sql](https://github.com/GreengageDB/greengage/blob/7.x/src/test/isolation2/sql/setup.sql) run? Does it run for src/test/isolation2/isolation2_resgroup_v2_schedule also?

'setup' is launched at the beginning of every isolation2 suite [from the launcher](https://github.com/GreengageDB/greengage/blob/d809c722087277bf14abcb7af209de5378e923aa/src/test/isolation2/isolation2_main.c#L146).
`resgroup_dumpinfo` is almost the last test in the respective resgroup schedule, but I think we'd better to remove the `drop` there as well. Just in case lsomebody tries to add new tests that use the aux functions to the end of resgroup schedule.

I've found another issue - [isolation2/sql/dependency.sql](https://github.com/GreengageDB/greengage/blob/d809c722087277bf14abcb7af209de5378e923aa/src/test/isolation2/sql/dependency.sql#L113) drops the `plpython3u` language, and it is done by design, so we can't just remove it from the test. So I guess will need to keep the `dependency` the last one in the schedule. And add there a proper comment describing why we need to keep it as the last test.

src/test/isolation2/sql/resgroup/resgroup_dumpinfo.sql
```diff
@@ -99,4 +99,3 @@ SELECT count(*) FROM pg_resgroup_get_status_kv(NULL);
 DROP ROLE role_dumpinfo_test;
 DROP ROLE role_permission;
 DROP RESOURCE GROUP rg_dumpinfo_test;
-DROP LANGUAGE plpython3u CASCADE;
```
hilltracer, line 102: Let's remove `CREATE LANGUAGE plpython3u;` in the top of this file. Because it confuse when we create the language but don't drop it.

PR #87, opened 2025-10-09
ADBDEV-7948: Bring back hot standby TAP tests

bandetto:
Bring back hot standby TAP tests

Previously, commit https://github.com/GreengageDB/greengage/commit/8359bfa07335a262339b31344288333fbb9224ce reverted changes related to a TAP tests that
required hot standby functionality since it is not available in 6.X.

Standby errored out before it could fully start due to several functions that
threw an error. Skip the error if connection is in utility mode.

---

The PR should be rebased to preserve the revert.

### DIFF
--- a/src/test/recovery/t/020_archive_status.pl
+++ b/src/test/recovery/t/020_archive_status.pl
@@ -5,7 +5,7 @@ use strict;
 use warnings;
 use PostgresNode;
 use TestLib;
-use Test::More tests => 7;
+use Test::More tests => 16;
 use Config;
 
 my $primary = get_new_node('master');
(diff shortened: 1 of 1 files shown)


bandetto:
> Could you provide a more detailed description of how these functions are implemented? Why was it done this way, and not like in vanilla Postgres of that version?

This is done in comments to the modified functions via the explanation of the difference between 7.X (as in Postgres), and 6.X, which does not use the "two-phase state file", and relies on the custom implementation.

RekGRpth:
> > Could you provide a more detailed description of how these functions are implemented? Why was it done this way, and not like in vanilla Postgres of that version?
> 
> This is done in comments to the modified functions via the explanation of the difference between 7.X (as in Postgres), and 6.X, which does not use the "two-phase state file", and relies on the custom implementation.

Well, I would like it to be in one place first - in the commit description.

src/backend/access/transam/twophase.c
```diff
@@ -1883,7 +1917,94 @@ GetOldestPreparedTransaction()
...
+				 "could not read WAL record for prepared transaction %u: %s",
+					 xid, errormsg);
+			}
+			continue;			/* Skip this transaction */
+		}
+
+		buf = XLogRecGetData(record);
+		hdr = (TwoPhaseFileHeader *) buf;
+
+		/* Skip if this transaction has been committed/aborted already */
```
hilltracer: Should we check it here? Is it possible move it before XLogReadRecord? To avoid unnecessary reading.

src/backend/access/transam/twophase.c
```diff
@@ -1883,7 +1917,94 @@ GetOldestPreparedTransaction()
...
+		char	   *bufptr;
+		int			i;
+
+		xlogrecptr = entry->xlogrecptr;
+
+		if (xlogrecptr == InvalidXLogRecPtr)
+			continue;
+
+		record = XLogReadRecord(xlogreader, xlogrecptr, &errormsg);
+		if (record == NULL)
```
hilltracer: Should we check that the record corresponds to a prepare transaction?

RekGRpth (approved):

hilltracer (approved):

PR #94, opened 2025-10-16
fix ADBDEV-8480: planner issue that leads to rescan of the Motion

ShapatinAA:
Fix issue with rescan of Motion

In some cases on path creation stage, the planner didn't account, that
the outer relation of a join in case of parametrization by outer
parameter from join above, might be rescanned.

Example of approximate plans that planner was able to produce, which lead to error.

<details>
  <summary>HashJoin plan</summary>
  
```sql
Gather Motion
    ->  Nested Loop
            ->  Broadcast Motion
                ->  Seq Scan (we get param here)
            ->  Materialize
                ->  Hash Join
                        ->  Broadcast Motion
                            ->  Foreign Scan
                        ->  Hash
                            ->  Function Scan (we use param here)
```
  
</details>

<details>
  <summary>MergeJoin plan</summary>
  
```sql
Nested Loop Semi Join
   ->  Gather Motion
         ->  Seq Scan
   ->  Materialize
         ->  Merge Left Join
               ->  Gather Motion
                     ->  Sort
                           ->  Seq Scan
               ->  Sort
                     ->  Result
                           Filter (we use param here)
                           ->  Materialize
                                 ->  Gather Motion
                                       ->  Seq Scan
```
  
</details>

In HashJoin we are getting error when we try to rescan Foreign Scan. We do it
because Seq Scan can produce >1 rows so, because of parametrization
we will be forced to rebuild HashTable and rescan Foreign Scan too.
And in MergeJoin we are getting error when we try to rescan Gather Motion of 
outer side of Merge Join, for the same reason. 

This patch applies same fix, previously used to address same problem
for NestLoop Join.
After creating a motions in join, we check if we have outer
parameters, which can lead to rescanning of outer path, and
if we do have them and we're not rescanable, we add Materialization
to outer path. We do not apply this logic for inner paths because they
are already rescannable my themselves in some form.

### DIFF
--- a/src/test/regress/sql/bfv_joins.sql
+++ b/src/test/regress/sql/bfv_joins.sql
@@ -750,6 +750,17 @@ reset from_collapse_limit;
 reset enable_mergejoin;
 reset enable_hashjoin;
 reset allow_system_table_mods;
+-- test that empty table gets hashed instead of a bigger table
+create table big_table(s smallint) distributed replicated;
+create table empty_table(i int) distributed replicated;
+insert into big_table select generate_series(1, 10000);
+analyze big_table;
+analyze empty_table;
+
+explain select * from big_table, empty_table where s = i;
+
+drop table big_table;
+drop table empty_table;
 
 -- Clean up. None of the objects we create are very interesting to keep around.
 reset search_path;
--- a/contrib/postgres_fdw/expected/gp2pg_postgres_fdw_optimizer.out
+++ b/contrib/postgres_fdw/expected/gp2pg_postgres_fdw_optimizer.out
@@ -1660,19 +1660,23 @@ SELECT t1.c1, t2.c1 FROM (SELECT c1 FROM ft4 WHERE c1 between 50 and 60) t1 FULL
  Sort
    Output: ft4.c1, ft5.c1
    Sort Key: ft4.c1, ft5.c1
-   ->  Hash Full Join
+   ->  Merge Full Join
          Output: ft4.c1, ft5.c1
-         Hash Cond: (ft4.c1 = ft5.c1)
-         ->  Foreign Scan on public.ft4
-               Output: ft4.c1
-               Remote SQL: SELECT c1 FROM "S 1"."T 3" WHERE (((c1 >= 50) AND (c1 <= 60)))
-         ->  Hash
+         Merge Cond: (ft5.c1 = ft4.c1)
+         ->  Sort
                Output: ft5.c1
+               Sort Key: ft5.c1
                ->  Foreign Scan on public.ft5
                      Output: ft5.c1
                      Remote SQL: SELECT c1 FROM "S 1"."T 4" WHERE (((c1 >= 50) AND (c1 <= 60)))
+         ->  Sort
+               Output: ft4.c1
+               Sort Key: ft4.c1
+               ->  Foreign Scan on public.ft4
+                     Output: ft4.c1
+                     Remote SQL: SELECT c1 FROM "S 1"."T 3" WHERE (((c1 >= 50) AND (c1 <= 60)))
  Optimizer: GPORCA
-(15 rows)
+(19 rows)
 
 SELECT t1.c1, t2.c1 FROM (SELECT c1 FROM ft4 WHERE c1 between 50 and 60) t1 FULL JOIN (SELECT c1 FROM ft5 WHERE c1 between 50 and 60) t2 ON (t1.c1 = t2.c1) ORDER BY t1.c1, t2.c1;
  c1 | c1 
@@ -1762,8 +1766,8 @@ SELECT t1.c1, ss.a, ss.b FROM (SELECT c1 FROM ft4 WHERE c1 between 50 and 60) t1
 -- c. test deparsing the remote query as nested subqueries
 EXPLAIN (VERBOSE, COSTS OFF)
 SELECT t1.c1, ss.a, ss.b FROM (SELECT c1 FROM ft4 WHERE c1 between 50 and 60) t1 FULL JOIN (SELECT t2.c1, t3.c1 FROM (SELECT c1 FROM ft4 WHERE c1 between 50 and 60) t2 FULL JOIN (SELECT c1 FROM ft5 WHERE c1 between 50 and 60) t3 ON (t2.c1 = t3.c1) WHERE t2.c1 IS NULL OR t2.c1 IS NOT NULL) ss(a, b) ON (t1.c1 = ss.a) ORDER BY t1.c1, ss.a, ss.b;
-                                                    QUERY PLAN                                                    
-------------------------------------------------------------------------------------------------------------------
+                                                 QUERY PLAN                                                 
+------------------------------------------------------------------------------------------------------------
  Sort
    Output: ft4_1.c1, ft4.c1, ft5.c1
    Sort Key: ft4_1.c1, ft4.c1, ft5.c1
@@ -1773,24 +1777,28 @@ SELECT t1.c1, ss.a, ss.b FROM (SELECT c1 FROM ft4 WHERE c1 between 50 and 60) t1
          ->  Result
                Output: ft4.c1, ft5.c1
                Filter: ((ft4.c1 IS NULL) OR (NOT (ft4.c1 IS NULL)))
-               ->  Hash Full Join
+               ->  Merge Full Join
                      Output: ft4.c1, ft5.c1
-                     Hash Cond: (ft4.c1 = ft5.c1)
-                     ->  Foreign Scan on public.ft4
-                           Output: ft4.c1
-                           Remote SQL: SELECT c1 FROM "S 1"."T 3" WHERE (((c1 >= 50) AND (c1 <= 60)))
-                     ->  Hash
+                     Merge Cond: (ft5.c1 = ft4.c1)
+                     ->  Sort
                            Output: ft5.c1
+                           Sort Key: ft5.c1
                            ->  Foreign Scan on public.ft5
                                  Output: ft5.c1
                                  Remote SQL: SELECT c1 FROM "S 1"."T 4" WHERE (((c1 >= 50) AND (c1 <= 60)))
+                     ->  Sort
+                           Output: ft4.c1
+                           Sort Key: ft4.c1
+                           ->  Foreign Scan on public.ft4
+                                 Output: ft4.c1
+                                 Remote SQL: SELECT c1 FROM "S 1"."T 3" WHERE (((c1 >= 50) AND (c1 <= 60)))
          ->  Hash
                Output: ft4_1.c1
                ->  Foreign Scan on public.ft4 ft4_1
                      Output: ft4_1.c1
                      Remote SQL: SELECT c1 FROM "S 1"."T 3" WHERE (((c1 >= 50) AND (c1 <= 60)))
  Optimizer: GPORCA
-(26 rows)
+(30 rows)
 
 SELECT t1.c1, ss.a, ss.b FROM (SELECT c1 FROM ft4 WHERE c1 between 50 and 60) t1 FULL JOIN (SELECT t2.c1, t3.c1 FROM (SELECT c1 FROM ft4 WHERE c1 between 50 and 60) t2 FULL JOIN (SELECT c1 FROM ft5 WHERE c1 between 50 and 60) t3 ON (t2.c1 = t3.c1) WHERE t2.c1 IS NULL OR t2.c1 IS NOT NULL) ss(a, b) ON (t1.c1 = ss.a) ORDER BY t1.c1, ss.a, ss.b;
  c1 | a  | b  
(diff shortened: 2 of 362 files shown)


src/test/regress/sql/bfv_joins.sql
```diff
@@ -605,6 +605,44 @@ ANALYZE ext_stats_tbl;
 
 explain SELECT 1 FROM ext_stats_tbl t11 FULL JOIN ext_stats_tbl t12 ON t12.c2;
 
+-- This test is introduced to verify that outer side of HashJoin path
+-- is rescannable, if this HashJoin is parametrized by the inner side of
+-- another join.
+create table ot (oid oid);
+insert into ot values (10::oid), (10::oid);
+
+create external web table w (oid int) execute 'echo 1' on all format 'text';
```
RekGRpth:
```suggestion
create external web table w (i int) execute 'echo 1' on all format 'text';
```

src/test/regress/sql/bfv_joins.sql
```diff
@@ -605,6 +605,44 @@ ANALYZE ext_stats_tbl;
...
+set enable_mergejoin = 'off';
+set enable_hashjoin = 'on';
+set from_collapse_limit = '1';
+set join_collapse_limit = '1';
+set random_page_cost = '1';
+
+select 1
+from ot,
+     foo(ot.oid) as f
+join w on w.oid = f.oid;
```
RekGRpth: Could you also fix the plan with verbose and cost off?

src/backend/optimizer/util/pathnode.c
```diff
@@ -4056,6 +4056,36 @@ create_hashjoin_path(PlannerInfo *root,
 	if (CdbPathLocus_IsNull(join_locus))
 		return NULL;
 
+	/*
+	 * We check if our path will be rescanned due to parametrization
+	 * from outer path after this node.
+	 * For all information see the comments in create_nestloop_path().
```
RekGRpth: Maybe it would be better to copy all the comments from there? Or even move the shared section of code into a separate function?

RekGRpth:
It looks like the test only checks one part of the changes, could you come up with a test for the rest of the changes as well?

RekGRpth:
fix failed tests

src/backend/optimizer/util/pathnode.c
```diff
@@ -4056,6 +4056,36 @@ create_hashjoin_path(PlannerInfo *root,
...
+	if (!outer_path->rescannable && !bms_is_empty(required_outer))
+	{
+		MaterialPath *matouter = create_material_path(root, outer_path->parent, outer_path);
+
+		matouter->cdb_shield_child_from_rescans = true;
+
+		outer_path = (Path *) matouter;
+	}
+
+	if (!inner_path->rescannable && !bms_is_empty(required_outer))
```
Stolb27: Is there any chance to be non rescannable and depends on param changing for inner part of hash join? Please describe the meaning of this code path.

src/test/regress/sql/bfv_joins.sql
```diff
@@ -605,6 +605,44 @@ ANALYZE ext_stats_tbl;
 
 explain SELECT 1 FROM ext_stats_tbl t11 FULL JOIN ext_stats_tbl t12 ON t12.c2;
 
+-- This test is introduced to verify that outer side of HashJoin path
+-- is rescannable, if this HashJoin is parametrized by the inner side of
+-- another join.
+create table ot (oid oid);
+insert into ot values (10::oid), (10::oid);
```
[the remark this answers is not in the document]
ShapatinAA: fixed

ShapatinAA:
Managed to get same error with rescan of Merge Join with this query
<details>
  <summary>Query</summary>
  
  ```sql
create table tcorr1(a int, b int);
create table tcorr2(a int, b int);

insert into tcorr1 values (1,99);
insert into tcorr2 values (1,1);
insert into tcorr1 values (1,99);
insert into tcorr2 values (1,1);

analyze tcorr1;
analyze tcorr2;                    
                                                                                          
create table a (i int4);
create table b (i int4);
insert into a select g from generate_series(1,1) g;
insert into b select g from generate_series(1,1) g;

analyze a;
analyze b;

set allow_system_table_mods = on;
update pg_class set reltuples=10000000 where oid ='a'::regclass;
update pg_class set relpages=10000 where oid ='a'::regclass;
                                                            
update pg_class set reltuples=1000000 where oid ='b'::regclass;
update pg_class set relpages=1000 where oid ='b'::regclass; 

set join_collapse_limit=1;
set from_collapse_limit=1;

set enable_mergejoin=on;
set enable_hashjoin=off;
set optimizer=off;
set jit=on;
set jit_above_cost=0;

select *
from tcorr1 out
where out.b in (select coalesce(tcorr2_d.c, 99)
                from (select a.i as a from a join b on a.i=b.i) tcorr1 left outer join (select a, count(*) as c
                                             from tcorr2
                                             where tcorr2.b = out.b
                                             group by a order by a) tcorr2_d on tcorr1.a=tcorr2_d.a);
  ```
  
</details>

So now will apply same logic for it and write test.

src/backend/optimizer/util/pathnode.c
```diff
@@ -2071,6 +2075,69 @@ create_material_path(PlannerInfo *root, RelOptInfo *rel, Path *subpath)
...
+
+		outer_path = (Path *) matouter;
+	}
+
+	/*
+	 * We need to try to materialize inner path only in case of
+	 * Nested loop join. This is because all other types of join
+	 * already support rescan of inner part in some form.
+	 */
+	if ((tag == T_NestLoop) && !inner_path->rescannable &&
```
RekGRpth: use `IsA(node, NestLoop)` instead

RekGRpth: Or separate only the part for outer into a separate function.

ShapatinAA: I would rather use IsA in this function.
Which solution is more preferable?

ShapatinAA: On second thought, wouldn’t splitting up (as it appears to me) a single functionality make the code harder to read and maintain?
(Btw I remembered, that at the moment we're using this function, we do not have any memory allocated for any "node" to `IsA()` it)

RekGRpth:
actuallize pr descr

src/test/regress/expected/gporca.out
```diff
@@ -12424,8 +12424,9 @@ where out.b in (select coalesce(tcorr2_d.c, 99)
    ->  Materialize  (cost=1.08..2.17 rows=3 width=8)
          ->  Hash Left Join  (cost=1.08..2.15 rows=3 width=8)
                Hash Cond: (tcorr1.a = tcorr2.a)
-               ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1.03 rows=1 width=4)
-                     ->  Seq Scan on tcorr1  (cost=0.00..1.01 rows=1 width=4)
+               ->  Materialize  (cost=0.00..1.03 rows=1 width=4)
+                     ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1.03 rows=1 width=4)
+                           ->  Seq Scan on tcorr1  (cost=0.00..1.01 rows=1 width=4)
```
RekGRpth: explain this change in pr descr

KnightMurloc: Why did this plan work before the patch?

RekGRpth: > I will add more rows inserted in tables in this test

Perhaps in this case no further mention will be necessary.

ShapatinAA: Do you mean, if we add a few rows to this test in `gporca.sql`, then we can remove one from `bvf_join.sql`?

ShapatinAA: Okay, so then there's only left PR description: what should I add? I think general case about adding materialization on outer in any join is already described.

RekGRpth: > what should I add?

For PR? Probably nothing so far.

src/test/regress/sql/bfv_joins.sql
```diff
@@ -605,6 +605,105 @@ ANALYZE ext_stats_tbl;
...
+update pg_class set relpages=1000 where oid ='b'::regclass;
+
+set join_collapse_limit=1;
+set from_collapse_limit=1;
+
+set enable_mergejoin=on;
+set enable_hashjoin=off;
+set optimizer=off;
+set jit=on;
+set jit_above_cost=0;
```
KnightMurloc: Why is this necessary?

src/test/regress/sql/bfv_joins.sql
```diff
@@ -605,6 +605,105 @@ ANALYZE ext_stats_tbl;
...
+
+reset join_collapse_limit;
+reset from_collapse_limit;
+reset enable_mergejoin;
+reset enable_hashjoin;
+reset optimizer;
+reset jit;
+reset jit_above_cost;
+reset allow_system_table_mods;
+
```
KnightMurloc: Drop the tables created in the test.

src/test/regress/sql/bfv_joins.sql
```diff
@@ -605,6 +605,105 @@ ANALYZE ext_stats_tbl;
...
+update pg_class set reltuples=1000000 where oid ='b'::regclass;
+update pg_class set relpages=1000 where oid ='b'::regclass;
+
+set join_collapse_limit=1;
+set from_collapse_limit=1;
+
+set enable_mergejoin=on;
+set enable_hashjoin=off;
+set optimizer=off;
```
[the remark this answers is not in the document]
ShapatinAA: Removed this.

src/test/regress/expected/gporca.out
ShapatinAA: Added value to test.

src/test/regress/sql/gporca.sql
```diff
@@ -2652,7 +2652,7 @@ drop table if exists tcorr1, tcorr2;
 create table tcorr1(a int, b int);
 create table tcorr2(a int, b int);
 
-insert into tcorr1 values (1,99);
+insert into tcorr1 values (1,99) (1,99);
```
RekGRpth:
```suggestion
insert into tcorr1 values (1,99), (1,99);
```

src/test/regress/expected/gporca.out
ShapatinAA: fixed

src/test/regress/sql/gporca.sql
```diff
@@ -2683,6 +2683,12 @@ where out.b in (select max(tcorr2.b + out.b - 1)
                 from tcorr2
                 where tcorr2.a=out.a);
 
+-- We insert one additional row in this case for testing that materialization
+-- of outer side of parametrized from above join works as intended.
+-- For some time this test passed fine on wrong plan because
+-- there were only one row in table.
+insert into tcorr1 values (1,88);
+analyze tcorr1;
```
RekGRpth: I think this is already excessive.

ShapatinAA: This get's printed out. Only 1 row for tcorr1 showed in plan.
```sql
postgres=# drop table if exists tcorr1, tcorr2;

create table tcorr1(a int, b int);
create table tcorr2(a int, b int);

insert into tcorr1 values (1,99);
insert into tcorr2 values (1,1);
analyze tcorr1;
analyze tcorr2;
DROP TABLE
CREATE TABLE
CREATE TABLE
INSERT 0 1
INSERT 0 1
ANALYZE
ANALYZE
postgres=# insert into tcorr1 values (1,88);
explain
select *
from tcorr1 out
where out.b in (select coalesce(tcorr2_d.c, 99)
                from tcorr1 left outer join (select a, count(*) as c
                                             from tcorr2
                                             where tcorr2.b = out.b
                                             group by a) tcorr2_d on tcorr1.a=tcorr2_d.a);
delete from tcorr1 where b = 88;
INSERT 0 1
                                                      QUERY PLAN                                                      
----------------------------------------------------------------------------------------------------------------------
 Nested Loop Semi Join  (cost=10000000001.08..10000000003.20 rows=3 width=8)
   Join Filter: ("out".b = COALESCE((count(*)), '99'::bigint))
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..1.03 rows=1 width=8)
         ->  Seq Scan on tcorr1 "out"  (cost=0.00..1.01 rows=1 width=8)
   ->  Materialize  (cost=1.08..2.17 rows=3 width=8)
         ->  Hash Left Join  (cost=1.08..2.15 rows=3 width=8)
               Hash Cond: (tcorr1.a = tcorr2.a)
               ->  Materialize  (cost=0.00..1.03 rows=1 width=4)
                     ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1.03 rows=1 width=4)
                           ->  Seq Scan on tcorr1  (cost=0.00..1.01 rows=1 width=4)
               ->  Hash  (cost=1.07..1.07 rows=1 width=12)
                     ->  HashAggregate  (cost=1.05..1.06 rows=1 width=12)
                           Group Key: tcorr2.a
                           ->  Result  (cost=0.00..1.04 rows=1 width=4)
                                 Filter: (tcorr2.b = "out".b)
                                 ->  Materialize  (cost=0.00..1.03 rows=1 width=8)
                                       ->  Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1.03 rows=1 width=8)
                                             ->  Seq Scan on tcorr2  (cost=0.00..1.01 rows=1 width=8)
(19 rows)

```

ShapatinAA: Node material will be added even when there were no insertion. 
And the plan will have exactly the same view as it was in one of the first variants of this patch

RekGRpth:
new flaky?
```diff
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/regress/expected/ic.out /home/gpadmin/gpdb_src/src/test/regress/results/ic.out
--- /home/gpadmin/gpdb_src/src/test/regress/expected/ic.out	2025-10-23 13:49:17.402496634 +0000
+++ /home/gpadmin/gpdb_src/src/test/regress/results/ic.out	2025-10-23 13:49:17.422496606 +0000
@@ -613,7 +612,11 @@
 --  * udpifc, "ERROR:  fault triggered"
 --  * tcp,    "ERROR:  canceling statement due to statement timeout"
 begin; select count(*) from qe_errors_ic; rollback;
-ERROR:  fault triggered, fault name:'exec_mpp_query_start' fault type:'error'
+ count 
+-------
+    10
+(1 row)
+
 reset statement_timeout;
 select gp_inject_fault('exec_mpp_query_start', 'reset', 2);
  gp_inject_fault 
```

RekGRpth (approved):

PR #95, opened 2025-10-16
ADBDEV-8454: Fix hash join spill file issue when rescanning

silent-observer:
Keep skew hashtable in memory for hash joins that require reuse (#95)

When the most common values occuring in the children relations are known, hash
join might enable the skew optimization, placing the most common values into a
separate "skew hashtable", effectively forcing them to be processed in the 0th
batch. This avoids spilling them to disk, which is an important performance
optimization for skewed tables. It is normally okay to free this skew
hashtable after the 0th batch is processed. However when the hash join might
be rescanned, we have to keep it in memory to not lose its contents. This
means that some memory will be freed at the end of the hash join, and not at
the end of 0th batch, but it is acceptable since skew hashtable uses at most
2% of the total hashtable memory.

Ticket: ADBDEV-8454

---

Fix hash join spill file issue when rescanning (#95)

Previously in some cases when rescanning a hash join was required, and the
hash join was executed in multiple batches that were spilled to disk, several
issues occurred when rescan happened in the middle of a batch:
1. Outer batch files (which contain outer tuples for all the next batches)
were never cleared, and after rescan they were filled again. When
gp_workfile_compression is enabled, this caused a BufFile to be in an
unexpected state (reading instead of writing), causing an error.
2. When inner batch file is read into the in-memory hash table, it is deleted.
After processing the batch, the hash table is spilled again into the batch
file in case rescan is possible. However, if rescan happens in the middle of a
batch, the hash table was never spilled and its contents were lost, resulting
in missing data in the output.

To fix these issues, when rescan is executed without rebuilding the whole hash
table, clear all the outer batch files and also spill the current batch to an
inner batch file.

Changes to original commit:
- Test refactored completely to cover more cases

Ticket: ADBDEV-8371
(cherry picked from commit d016d74884f420e227ada85437e09c04970e9d10)

---

### DIFF
--- a/src/backend/executor/nodeHashjoin.c
+++ b/src/backend/executor/nodeHashjoin.c
@@ -1608,6 +1608,8 @@ ExecHashJoinGetSavedTuple(HashJoinState *hjstate,
 void
 ExecReScanHashJoin(HashJoinState *node)
 {
+	HashJoinTable hashtable = node->hj_HashTable;
+
 	/*
 	 * In a multi-batch join, we currently have to do rescans the hard way,
 	 * primarily because batch temp files may have already been released. But
@@ -1615,12 +1617,12 @@ ExecReScanHashJoin(HashJoinState *node)
 	 * inner subnode, then we can just re-use the existing hash table without
 	 * rebuilding it.
 	 */
-	if (node->hj_HashTable != NULL)
+	if (hashtable != NULL)
 	{
-		node->hj_HashTable->first_pass = false;
+		hashtable->first_pass = false;
 
 		if (node->js.ps.righttree->chgParam == NULL &&
-			!node->hj_HashTable->eagerlyReleased)
+			!hashtable->eagerlyReleased)
 		{
 			/*
 			 * Okay to reuse the hash table; needn't rescan inner, either.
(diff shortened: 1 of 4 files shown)


hilltracer:
```
+ERROR:  Error on receive from seg0 slice1 172.17.0.2:7002 pid=73182: server closed the connection unexpectedly
+	This probably means the server terminated abnormally
+	before or while processing the request.
```
SigSegv during added test?

src/backend/executor/nodeHashjoin.c
```diff
@@ -1642,18 +1644,46 @@ ExecReScanHashJoin(HashJoinState *node)
 			 */
 			node->hj_OuterNotEmpty = false;
 
+			/*
+			 * Outer batch files have to be cleared before restarting the hash
+			 * join, because they will be written again when batch 0 is
+			 * processed.
+			 */
+			if (hashtable->outerBatchFile)
```
dkovalev1: it seems this path is not covered by the test, is it?

silent-observer: Updated the test data again. The uncertainty came from the skew hashtable - since the most common values are decided randomly by ANALYZE, the test data might not be present in the main hashtable. After this fix (adding both common and uncommon values in a big enough amount to fill the skew hashtable) the test fails consistently when reverting each of 3 separate fixes.

hilltracer (approved):

PR #96, opened 2025-10-16
ADBDEV-8582: Fix issue with sort stat output during EXPLAIN ANALYZE

hilltracer:
EXPLAIN always shows the planned execution tree, while EXPLAIN ANALYZE overlays 
what actually ran.  If a plan node exists in the plan tree but the executor
never invoked it, it is still printed — marked as "never executed" in text
format, or with zero-valued statistics in JSON format.

In 6.x, the code loops over all possible sort methods when collecting QEs' sort
statistics and prints sort information only for methods that were actually
used. In 7.x, commits 923d9a5 and e3339bb reworked this logic. A new
shared_info structure was introduced to aggregate sort statistics from all
workers, which is populated during execution. The shared_info is now used
directly in show_sort_info() to obtain tuplesort metrics. 

However, when all workers' sinstrument->sortMethod is
SORT_TYPE_STILL_IN_PROGRESS (e.g., when the sort node was never executed), the
pointers sortMethod and spaceType remained uninitialized. This caused a crash
during JSON output in escape_json() or produced wrong output such as (null)
during TEXT output.

Do not print sort information during EXPLAIN ANALYZE when no completed
statistics exist. This aligns the 7.x behavior with 6.x.

Ticket: ADBDEV-8582

### DIFF
--- a/contrib/postgres_fdw/expected/gp2pg_postgres_fdw.out
+++ b/contrib/postgres_fdw/expected/gp2pg_postgres_fdw.out
@@ -9549,3 +9549,378 @@ SELECT 1 FROM ft1 LIMIT 1;
 ALTER SERVER pgserver OPTIONS (ADD use_remote_estimate 'off');
 -- The invalid connection gets closed in pgfdw_xact_callback during commit.
 COMMIT;
+-- ===================================================================
+-- test distribution keys for foreign tables
+-- ===================================================================
+--
+-- Test simple case for foreign table with 1 distribution key
+--
+\! env PGOPTIONS='' psql -p ${PG_PORT} contrib_regression -c 'CREATE TABLE "S 1"."test_distr" (i int);'
+CREATE TABLE
+SET gp_force_random_redistribution TO on;
+DROP TABLE IF EXISTS test_internal;
+NOTICE:  table "test_internal" does not exist, skipping
+CREATE TABLE test_internal(a int) DISTRIBUTED BY (a);
+-- Check foreign table without distribution keys
+DROP FOREIGN TABLE IF EXISTS ftest_distr_randomly;
+NOTICE:  foreign table "ftest_distr_randomly" does not exist, skipping
+CREATE FOREIGN TABLE ftest_distr_randomly(
+	i int
+) SERVER pgserver
+OPTIONS(schema_name 'S 1', table_name 'test_distr', mpp_execute 'all segments');
+-- Plan should contain Redistribute motion
+EXPLAIN (COSTS OFF) INSERT INTO ftest_distr_randomly SELECT a FROM test_internal;
+                      QUERY PLAN                      
+------------------------------------------------------
+ Insert on ftest_distr_randomly
+   ->  Redistribute Motion 3:3  (slice1; segments: 3)
+         ->  Seq Scan on test_internal
+ Optimizer: Postgres query optimizer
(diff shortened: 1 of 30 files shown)


hilltracer:
> flaky?

I see this in log:
```
2025-10-16T23:55:42.9455509Z Performing test 'mix_distribution_policy'
2025-10-16T23:55:42.9526446Z [ERROR]: Found tables created using both legacy and non legacy hashops in distribution policy.Please run the gpcheckcat.distpolicy.sql file to list the tables.
2025-10-16T23:55:42.9529058Z Total runtime for test 'mix_distribution_policy': 0:00:00.00
2025-10-16T23:55:42.9559523Z 
2025-10-16T23:55:42.9559963Z SUMMARY REPORT: FAILED
2025-10-16T23:55:42.9560628Z ===================================================================
2025-10-16T23:55:42.9562834Z Completed 14 test(s) on database 'regression' at 2025-10-16 23:55:42 with elapsed time 0:00:18
2025-10-16T23:55:42.9564164Z Failed test(s) that are not reported here: mix_distribution_policy
2025-10-16T23:55:42.9565140Z See /home/gpadmin/gpAdminLogs/gpcheckcat_20251016.log for detail
```
I think it's [that](https://tracker.yandex.ru/ADBDEV-8382) flucky

src/backend/commands/explain.c
```diff
@@ -3479,28 +3479,28 @@ show_sort_info(SortState *sortstate, ExplainState *es)
...
-			peakSpaceUsed = Max(peakSpaceUsed, sinstrument->spaceUsed);
-			totalSpaceUsed += sinstrument->spaceUsed;
+	bool have_stats = false;
+
+	TuplesortInstrumentation *sinstrument;
+	for (int n = 0; n < sortstate->shared_info->num_workers; n++) {
+		sinstrument = &sortstate->shared_info->sinstrument[n];
+		if (sinstrument->sortMethod == SORT_TYPE_STILL_IN_PROGRESS)
+			continue;	/* ignore any unfilled slots */
+		if (!have_stats) {
```
dkovalev1: Now we assume that we take only first value for sortMethod and spaceType, and if they are different for different sinstrument, the rest of them will be lost. 
Is it possible situation when sortMethod and spaceType are different? Would it be better to take this into account?

hilltracer: Thanks, it looks like bug in 7.x. 6.x [prints](https://github.com/GreengageDB/greengage/blob/9e2704e8f0399072f0ce6384922782af613a5cbd/src/backend/commands/explain.c#L2709) all used methods. I created separated task for it.

src/backend/commands/explain.c
```diff
@@ -3479,28 +3481,28 @@ show_sort_info(SortState *sortstate, ExplainState *es)
...
```
RekGRpth:
```suggestion
	for (int n = 0; n < sortstate->shared_info->num_workers; n++)
	{
		TuplesortInstrumentation *sinstrument = &sortstate->shared_info->sinstrument[n];

		if (sinstrument->sortMethod == SORT_TYPE_STILL_IN_PROGRESS)
			continue;	/* ignore any unfilled slots */

		if (!have_stats)
		{
```

hilltracer: done

silent-observer (approved):

PR #103, opened 2025-10-24
Fix functions for working with orphaned files to handle custom tablespaces correctly

hilltracer:
Adopt patch 7bdf316b6 from our 6.x branch to 7.x.

Functions from extension worked very badly with custom tablespaces.
This patch fixes all known related issues and adds tests to cover them.

1. gp_check_orphaned_files did not handle absent directories

gp_check_orphaned_files failed when it encountered tablespaces whose physical 
directories were not yet created for a database (e.g.
pg_tblspc/<tablespace_oid>/<GPDB_version>/<db_oid> or base/<db_oid>).

Set missing_ok=true parameter for pg_ls_dir() inside __get_exist_files, so it
returns an empty set instead of raising an error when the directory is missing.

2. gp_check_orphaned_files returned duplicate `global//…` paths

Normalise the literal `global/` to `global` inside __get_exist_files.

3. False orphan/missing-file reports for databases in non-default tablespaces

The extension reported phantom orphan and missing files whenever a database
lived in a user-defined tablespace rather than pg_default.  Helper views
hard-coded tablespace OID `0`, ignoring the database’s real default tablespace
(pg_database.dattablespace).

__get_exist_files now uses the real pg_default OID.

__get_expect_files uses pg_database.dattablespace when `reltablespace = 0`, so
expected locations match the actual default tablespace.

__get_expect_files_ext introduces a CTE class_info that pre-computes the
correct tablespace value and joins AO/CO branches to it.

4. __gp_check_orphaned_files_func used NOWAIT ineffectively

In GPDB, NOWAIT inside PL/pgSQL did not throw an exception when attempting to
lock pg_class; it waited silently. The problem is reproduced even in PostgreSQL
17.

It should be fixed in the future. For now, remove NOWAIT and the associated
exception handling from __gp_check_orphaned_files_func and
gp_move_orphaned_files to prevent misunderstanding.

5. gp_move_orphaned_files failed when orphaned files were located across
different filesystems

The function relied on the rename syscall, which cannot cross filesystems,
causing an “Invalid cross-device link”.

Add gp_move_orphaned_files_by_tablespace_location, allowing users to specify a
target directory per tablespace; plus an internal helper called by wrappers.
Catch errors during renaming and emit a hint explaining with instructions for
user.

6.  __get_exist_files used hardcoded tablespace OIDs and condition
tablespace.oid > 1664 to identify user-defined tablespaces.

Replace OIDs with subqueries using pg_tablespace.spcname to dynamically resolve
pg_default and pg_global OIDs.  Updated the user-defined tablespace filter to
spcname NOT IN ('pg_default', 'pg_global'). According to PostgreSQL
documentation, only pg_default and pg_global are built-in; all others are
user-defined.

Changes caused by the new 7.x code:

- Use relkind != 'v' instead of relstorage; add AM via pg_am. Foreign and partitioned
relations have no storage, so we don’t explicitly exclude them.

- Simplify the activity probe to gp_stat_activity with a backend_type filter; remove
UNIONs with gp_dist_random('pg_stat_activity').

- Join gp_settings by gp_segment_id (instead of pg_settings hacks) in move functions.

- Set search_path = pg_catalog, pg_temp for safety.

- Remove the test workaround; the issue does not reproduce on 7.x.

Ticket: ADBDEV-7719

### DIFF
--- a/gpcontrib/gp_toolkit/Makefile
+++ b/gpcontrib/gp_toolkit/Makefile
@@ -2,7 +2,7 @@ EXTENSION = gp_toolkit
 DATA = gp_toolkit--1.1--1.2.sql gp_toolkit--1.0--1.1.sql gp_toolkit--1.0.sql \
 		gp_toolkit--1.2--1.3.sql gp_toolkit--1.3.sql gp_toolkit--1.3--1.4.sql \
 		gp_toolkit--1.4--1.5.sql gp_toolkit--1.5--1.6.sql \
-		gp_toolkit--1.6--1.7.sql
+		gp_toolkit--1.6--1.7.sql gp_toolkit--1.7--1.8.sql
 MODULE_big = gp_toolkit
 ifeq ($(shell uname -s), Linux)
 OBJS = resgroup.o gp_partition_maint.o gp_segment_files_maint.o
--- a/gpcontrib/gp_toolkit/gp_toolkit--1.7--1.8.sql
+++ b/gpcontrib/gp_toolkit/gp_toolkit--1.7--1.8.sql
@@ -1,0 +1,466 @@
+/* gpcontrib/gp_toolkit/gp_toolkit--1.7--1.8.sql */
+
+-- complain if script is sourced in psql, rather than via ALTER EXTENSION
+\echo Use "ALTER EXTENSION gp_toolkit UPDATE TO '1.8" to load this file. \quit
+
+--------------------------------------------------------------------------------
+-- @view:
+--        gp_toolkit.__get_exist_files
+--
+-- @doc:
+--        Retrieve a list of all existing data files in the default
+--        and user tablespaces.
+--
+--------------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


hilltracer:
For simplify review see these diffs:
[new_code_6x_vs_7x.diff.txt](https://github.com/user-attachments/files/23128673/new_code_6x_vs_7x.diff.txt)
[new_test_6x_vs_7x.diff.txt](https://github.com/user-attachments/files/23128676/new_test_6x_vs_7x.diff.txt)
[old_code_6x_vs_7x.diff.txt](https://github.com/user-attachments/files/23128677/old_code_6x_vs_7x.diff.txt)
[old_test_6x_vs_7x.diff.txt](https://github.com/user-attachments/files/23128678/old_test_6x_vs_7x.diff.txt)

red1452:
Is the patch cherry-pick? If yes, it should contain one commit: cherry-pick from original patch with addiotal description about differences between original patch and current patch.

red1452 (approved):

PR #114, opened 2025-11-06
ADBDEV-8758: Add support for Python 3 in bundled PyGreSql

KnightMurloc:
This patch adds python 3 support for the supplied PyGreSQL 4 library.
Changes:
1. Add a basestring declaration that does not exist in python 3
2. Make print compatible with python 3
3. Use items instead of iteritems to iterate through the dictionary.
4. Add a header file for partial compatibility with python 3 in the C module
(taken from PyGreSQL 5).
5. Add a function to form a string representation of the query result instead of
output to stdout (taken from PyGreSQL 5).
6. Use latin-1 encoded for strings in the escaping and unescaping functions for
uniformity between python 2 and 3

Updating PyGreSQL leads to a large number of changes in the behavior of the
library. Such as closing connections, type conversions, etc. Which in turn can
lead to unexpected problems.

Ticket: ADBDEV-8758

### DIFF
--- a/gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
+++ b/gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
@@ -60,6 +60,10 @@
 #define PyInt_FromLong PyLong_FromLong
 #define PyInt_AsLong PyLong_AsLong
 
+/* Floats */
+
+#define PyFloat_FromString(str, pend) PyFloat_FromString(str)
+
 /* Module init */
 
 #define MODULE_INIT_FUNC(name) \
@@ -79,10 +83,6 @@
 
 #define PyBytes_AS_STRING PyString_AS_STRING
 #define _PyBytes_Resize _PyString_Resize
-
-/* Floats */
-
-#define PyFloat_FromString(str) PyFloat_FromString(str, NULL)
 
 /* Module init */
 
@@ -379,7 +379,6 @@ format_result(const PGresult *res)
 			if (buffer)
 			{
 				char *p = buffer;
-				PyObject *result;
 
 				/* create the header */
 				for (j = 0; j < n; ++j)
@@ -1251,7 +1250,7 @@ static PyTypeObject PgSourceType = {
 	.tp_basicsize = sizeof(pgsourceobject),
 	/* methods */
 	.tp_dealloc = (destructor) pgsource_dealloc,
-#if PY_MAJOR_VERSION >= 3
+#if PY_MAJOR_VERSION < 3
 	.tp_print = (printfunc) pgsource_print,
 #endif
 	.tp_getattro = (getattrofunc) pgsource_getattr,
@@ -1742,7 +1741,7 @@ pgconnect(pgobject * self, PyObject * args, PyObject * dict)
 	static const char *kwlist[] = {"dbname", "host", "port", "opt",
 	"tty", "user", "passwd", NULL};
 
-	char	   *pghost,
+	const char *pghost,
 			   *pgopt,
 			   *pgtty,
 			   *pgdbname,
@@ -2139,7 +2138,7 @@ pgquery_getresult(pgqueryobject * self, PyObject * args)
 
 					case 3:
 						tmp_obj = PyStr_FromString(s);
-						val = PyFloat_FromString(tmp_obj);
+						val = PyFloat_FromString(tmp_obj, NULL);
 						Py_DECREF(tmp_obj);
 						break;
 
@@ -2169,7 +2168,7 @@ pgquery_getresult(pgqueryobject * self, PyObject * args)
 						else
 						{
 							tmp_obj = PyStr_FromString(s);
-							val = PyFloat_FromString(tmp_obj);
+							val = PyFloat_FromString(tmp_obj, NULL);
 						}
 						Py_DECREF(tmp_obj);
 						break;
@@ -2267,7 +2266,7 @@ pgquery_dictresult(pgqueryobject * self, PyObject * args)
 
 					case 3:
 						tmp_obj = PyStr_FromString(s);
-						val = PyFloat_FromString(tmp_obj);
+						val = PyFloat_FromString(tmp_obj, NULL);
 						Py_DECREF(tmp_obj);
 						break;
 
@@ -2297,7 +2296,7 @@ pgquery_dictresult(pgqueryobject * self, PyObject * args)
 						else
 						{
 							tmp_obj = PyStr_FromString(s);
-							val = PyFloat_FromString(tmp_obj);
+							val = PyFloat_FromString(tmp_obj, NULL);
 						}
 						Py_DECREF(tmp_obj);
 						break;
@@ -2474,7 +2473,7 @@ pg_query(pgobject * self, PyObject * args)
 
 						if (ret[0])		/* return number of rows affected */
 						{
-							PyObject* obj = PyString_FromString(ret);
+							PyObject* obj = PyStr_FromString(ret);
 							PQclear(result);
 							return obj;
 						}
@@ -3116,11 +3115,7 @@ pg_notices(pgobject *self, PyObject *args)
 	else
 		retval = PyList_New(0);
 
-	if (self->notices)
-	{
-		Py_DECREF(self->notices);
-		self->notices = NULL;
-	}
+	Py_CLEAR(self->notices);
 	return retval;
 }
  


dkovalev1:
Are we going to support both python 2 and python 3? If so, may state in in the description?
Which exact versions of python 3 we support now?

> Updating PyGreSQL leads to a large number of changes in the behavior of the
> library. Such as closing connections, type conversions, etc. Which in turn can
> lead to unexpected problems.

is it related to python3 support?

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/py3c.h
```diff
@@ -0,0 +1,136 @@
...
+#define PyInt_AsUnsignedLongLongMask PyLong_AsUnsignedLongLongMask
+#define PyInt_AsSsize_t PyLong_AsSsize_t
+
+/* Module init */
+
+#define MODULE_INIT_FUNC(name) \
+    PyMODINIT_FUNC PyInit_ ## name(void); \
+    PyMODINIT_FUNC PyInit_ ## name(void)
+
+#else
```
RekGRpth:
```suggestion
#else /* if PY_MAJOR_VERSION >= 3 */
```

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -702,7 +860,7 @@ pgsource_fetch(pgsourceobject * self, PyObject * args)
 				str = Py_None;
 			}
 			else
-				str = PyString_FromString(PQgetvalue(self->last_result, self->current_row, j));
+				str = PyStr_FromString(PQgetvalue(self->last_result, self->current_row, j));
```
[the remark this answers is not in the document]
KnightMurloc: Wouldn't it be better to use function names from python 3? So that after upgrading to python 3, it would not be strange to use function names from python 2.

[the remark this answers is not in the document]
KnightMurloc: > Furthermore, you can move the entire contents of `py3c.h` to the top of `pgmodule.c` file.

the `pgmodule.c` already a big file, is it worth adding this to it?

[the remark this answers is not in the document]
RekGRpth: Why have a separate header if it's only used in one file?

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -2003,12 +2224,16 @@ pgquery_getresult(pgqueryobject * self, PyObject * args)
 						if (decimal)
 						{
 							tmp_obj = Py_BuildValue("(s)", s);
-							val = PyEval_CallObject(decimal, tmp_obj);
+#if PY_MAJOR_VERSION >= 3
+								val = PyObject_Call(decimal, tmp_obj, NULL);
+#else
+								val = PyEval_CallObject(decimal, tmp_obj);
+#endif
```
RekGRpth: why not
```c
#if PY_MAJOR_VERSION >= 3
#define PyEval_CallObject(callable, arg) PyObject_Call(callable, arg, NULL)
#endif

File: gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c [hunk 4]
```diff
@@ -42,6 +42,71 @@
...
+
+/* Strings */
+
+#define PyStr_AsUTF8String(str) (Py_INCREF(str), (str))
+#define PyStr_AsUTF8AndSize(pystr, sizeptr) \
+	((*sizeptr=PyString_Size(pystr)), PyString_AsString(pystr))
+
+/* Floats */
+
+#define PyFloat_FromString(str) PyFloat_FromString(str, NULL)
```
RekGRpth: How does this work if `PyFloat_FromString` is already defined in `Python.h`? Should we do the opposite to reduce the diff?
```c
#if PY_MAJOR_VERSION >= 3
#define PyFloat_FromString(str, pend) PyFloat_FromString(str)
#endif
```

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -133,9 +195,8 @@ static PyObject *pg_default_passwd;	/* default password */
 #ifdef HANDLE_NOTICES
 #define MAX_BUFFERED_NOTICES 100              /* max notices to keep for each connection */
 static void notice_processor(void *arg, const char *message);
-#endif /* HANDLE_NOTICES */ 
+#endif /* HANDLE_NOTICES */
```
RekGRpth: remove this change

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -54,9 +119,6 @@ typedef int Py_ssize_t;
 #define PY_SSIZE_T_MIN INT_MIN
 #endif
 
-/* taken from fileobject.c */
-#define BUF(v) PyString_AS_STRING((PyStringObject *)(v))
```
RekGRpth: why not
```suggestion
#define BUF(v) PyBytes_AS_STRING((PyStringObject *)(v))
```

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -156,7 +217,7 @@ typedef struct
 #endif /* HANDLE_NOTICES */
 }	pgobject;
 
-staticforward PyTypeObject PgType;
+static PyTypeObject PgType;
```
RekGRpth: why not
```c
#if PY_MAJOR_VERSION >= 3
#define staticforward static
#endif
```

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -1228,15 +1453,15 @@ pglarge_read(pglargeobject * self, PyObject * args)
...
+	if ((size = lo_read(self->pgcnx->cnx, self->lo_fd, PyBytes_AS_STRING(buffer), size)) < 0)
 	{
 		PyErr_SetString(PyExc_IOError, "error while reading.");
 		Py_XDECREF(buffer);
 		return NULL;
 	}
 
 	/* resize buffer and returns it */
-	_PyString_Resize(&buffer, size);
+	_PyBytes_Resize(&buffer, size);
```
RekGRpth: why not
```c
#if PY_MAJOR_VERSION >= 3
#define _PyString_Resize _PyBytes_Resize
#endif
```

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -325,6 +386,166 @@ print_result(FILE *fout, const PGresult *res)
...
 }
 
+/* format result (mostly useful for debugging) */
+/* Note: This is similar to the Postgres function PQprint().
+ * PQprint() is not used because handing over a stream from Python to
+ * Postgres can be problematic if they use different libs for streams
+ * and because using PQprint() and tp_print is not recommended any more.
+ */
+static PyObject *
+format_result(const PGresult *res)
```
RekGRpth: Why not modify the existing `print_result` function so that it can be used with both `reprfunc` and `printfunc`?

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -2609,7 +2838,7 @@ pg_inserttable(pgobject * self, PyObject * args)
 			}
 			else if (PyString_Check(item))
 			{
-				const char* t = PyString_AS_STRING(item);
+				const char* t = PyString_AsString(item);
```
RekGRpth: why not
```c
#if PY_MAJOR_VERSION >= 3
#define PyString_AS_STRING PyString_AsString
#endif
```

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -2907,12 +3136,12 @@ pg_loimport(pgobject * self, PyObject * args)
 #endif /* LARGE_OBJECTS */
 
 #ifdef HANDLE_NOTICES
- 
+
```
RekGRpth: remove this change

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -2907,12 +3136,12 @@ pg_loimport(pgobject * self, PyObject * args)
...
 /* fetch accumulated backend notices */
 static char pg_notices__doc__[] =
   "notices() -- returns and clears the list of currently accumulated backend notices for the connection.";
- 
-static void 
+
+static void
```
RekGRpth: remove this change

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -2936,7 +3165,7 @@ notice_processor(void *arg, const char *message)
 		PySequence_DelItem(self->notices, 0);
 	PyGILState_Release(gstate);
 }
- 
+
```
RekGRpth: remove this change

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -2951,10 +3180,14 @@ pg_notices(pgobject *self, PyObject *args)
 	else
 		retval = PyList_New(0);
 
-	Py_CLEAR(self->notices);
+	if (self->notices)
+	{
+		Py_DECREF(self->notices);
+		self->notices = NULL;
+	}
```
RekGRpth: Why is this change necessary?

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
dkovalev1: > why not
> 
> ```c
> #if PY_MAJOR_VERSION >= 3
> #define PyEval_CallObject(callable, arg) PyObject_Call(callable, arg, NULL)
> #endif
> ```

I agree with current code. If we use this one, we will semantically introduce a python2 simulation level for python3, which seems to be a bit strange considering that python2 is deprecated.

RekGRpth: This is a temporary solution, and the diff will be smaller.

dkovalev1: there is nothing more permanent than a temporary solution

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -1076,25 +1299,27 @@ pgsource_print(pgsourceobject * self, FILE *fp, int flags)
...
+static PyTypeObject PgSourceType = {
+	PyVarObject_HEAD_INIT(NULL, 0)
 
-	0,							/* ob_size */
-	"pgsourceobject",			/* tp_name */
-	sizeof(pgsourceobject),		/* tp_basicsize */
-	0,							/* tp_itemsize */
+	.tp_name = "pgsourceobject",
+	.tp_basicsize = sizeof(pgsourceobject),
+	.tp_itemsize = 0,
```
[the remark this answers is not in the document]
RekGRpth: fix in other places too

gpMgmt/bin/pythonSrc/PyGreSQL-4.0/pgmodule.c
```diff
@@ -55,7 +122,7 @@ typedef int Py_ssize_t;
 #endif
 
 /* taken from fileobject.c */
-#define BUF(v) PyString_AS_STRING((PyStringObject *)(v))
+#define BUF(v) PyBytes_AS_STRING((PyStringObject *)(v))
```
dkovalev1:
```suggestion
#define BUF(v) PyBytes_AS_STRING((PyObject *)(v))
```
because PyBytes_AS_STRING is declared as
```
static inline char* PyBytes_AS_STRING(PyObject *op)
```

PR #117, opened 2025-11-12
ADBDEV-8698: Make reindextable_while_reindex_idx_*_part_btree tests stable

hilltracer:
Problem description:

In 6.x, tests work in the following way:

- s1: reindex whole partitioned table in a separate transaction
- s2: try to reindex the default-partition index only -> wait for s1 to finish
- s1: commit -> s2 completes

In 7.x, 67a7c0f changed this logic, because the REINDEX TABLE command can't
be inside a transaction block.

- s1: lock whole partitioned table
- s2: try to reindex the whole partitioned table (don't use transaction) -> wait 
  for s1 to release lock
- s3: try to reindex default-partition index only, in a separate transaction ->
  wait for s1 to release lock
- s1: release lock -> race condition between s2 and s3 -> s3 wins the race

But sometimes s2 won the race, which broke the test.

Patch effects:

Remove lock logic and race condition by suspending REINDEX TABLE with help of a
fault injector. 

- main session: set fault injector to only one execution inside
  reindex_relation()
- s1: try to reindex whole partitioned table -> suspend.
- s2: try to reindex default-partition index only, in a separated transaction ->
  wait for s1 to finish
- main session: reset fault injector -> s1 completes -> s2 completes

During reindex of the whole partitioned table, the traversal order of 
partitions is not deterministic. But we want to always suspend during
default-partition processing. Remove p_one partition from the test table data
to make deterministic fault injector behavior.

Change isolation2_schedule to force sequential execution of tests because
a fault is injected.

Ticket: ADBDEV-8698

### DIFF
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -6,7 +6,7 @@ on:
     branches: ['7.x']  # Trigger on push to main (after merged PR)
     tags: ['7.*']   # Trigger on tags for versioned releases
   pull_request:
-    branches: ['*'] # Trigger on pull requests for all branches
+    branches: ['**'] # Trigger on pull requests for all branches
 
 # Concurrency control to cancel previous runs on new push to same PR/branch
 concurrency:


KnightMurloc:
What are these tests testing? Is it possible to do this without a race at all?

PR #118, opened 2025-11-13
ADBDEV-7477: Ensure that creation of an empty relfile is fsync'd at checkpoint.

KnightMurloc:
If you create a table and don't insert any data into it, the relation file is
never fsync'd. You don't lose data, because an empty table doesn't have any data
to begin with, but if you crash and lose the file, subsequent operations on the
table will fail with "could not open file" error.

To fix, register an fsync request in mdcreate(), like we do for mdwrite().

c request in mdcreate(), like we do for mdwrite().

Per discussion, we probably should also fsync the containing directory after
creating a new file. But that's a separate and much wider issue.

Backpatch to all supported versions.

Reviewed-by: Andres Freund, Thomas Munro
Discussion: https://www.postgresql.org/message-id/d47d8122-415e-425c-d0a2-e0160829702d%40iki.fi

6X changes:
In 6X, the smgr_which field is not used and not set. This patch adds setting
this field. To maintain binary compatibility, this field is set outside the
smgropen function.

(cherry picked from commit 1b4f1c6f8a6c0b764636f25e17d37d8cab7e82be)

### DIFF
--- a/src/backend/access/transam/xlogutils.c
+++ b/src/backend/access/transam/xlogutils.c
@@ -313,6 +313,7 @@ XLogReadBufferExtended(RelFileNode rnode, ForkNumber forknum,
 
 	/* Open the relation at smgr level */
 	smgr = smgropen(rnode, InvalidBackendId);
+	smgr->smgr_which = RELSTORAGE_HEAP;
 
 	/*
 	 * Create the target file if it doesn't already exist.  This lets us cope
--- a/src/backend/catalog/storage.c
+++ b/src/backend/catalog/storage.c
@@ -114,6 +114,7 @@ RelationCreateStorage(RelFileNode rnode, char relpersistence, char relstorage)
 	}
 
 	srel = smgropen(rnode, backend);
+	srel->smgr_which = relstorage;
 	smgrcreate(srel, MAIN_FORKNUM, false);
 
 	if (needs_wal)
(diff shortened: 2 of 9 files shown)


KnightMurloc:
> Is link for cherry-picked commit correct? Your commit contains changes for 8 files, but origin commit - just for 1 file. Fix it or add info to the description.

I updated the description.

KnightMurloc:
> Is link for cherry-picked commit correct? Your commit contains changes for 8 files, but origin commit - just for 1 file.

we need to modify more files since the field we need is not set in 6X.

src/backend/access/transam/xlogutils.c
```diff
@@ -313,6 +313,7 @@ XLogReadBufferExtended(RelFileNode rnode, ForkNumber forknum,
 
 	/* Open the relation at smgr level */
 	smgr = smgropen(rnode, InvalidBackendId);
+	smgr->smgr_which = 'h';
```
whitehawk: Is it possible to use RELSTORAGE_HEAP and similar instead of bare 'h'?
It is defined in pg_class.h:
```
#define		  RELSTORAGE_HEAP	'h'
#define		  RELSTORAGE_AOROWS	'a'
#define 	  RELSTORAGE_AOCOLS	'c'
```

src/backend/catalog/storage.c
```diff
@@ -541,6 +543,7 @@ smgr_redo(XLogRecPtr beginLoc, XLogRecPtr lsn, XLogRecord *record)
 		SMgrRelation reln;
 
 		reln = smgropen(xlrec->rnode, InvalidBackendId);
+		reln->smgr_which = 0;
```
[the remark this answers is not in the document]
KnightMurloc: done

src/backend/catalog/storage.c
whitehawk: I suppose we need to do smth with `reln->smgr_which` here for `XLOG_SMGR_TRUNCATE` as well

PR #119, opened 2025-11-14
ADBDEV-7269: Fix problem with handling outer references in join clauses

GirgerS:
Fix problem with handling outer references in join clauses

When the Postgres planner creates a motion for a join inside 
cdbpath_motion_for_join(), it looks up the 
RestrictInfo.contain_outer_query_references flag for all present
RestrictInfos to determine if a join contains outer references.

The error occurred when original clause was treated as a part of 
EquivalenceClass inside distribute_qual_to_rels(), and never 
went through the distribute_restrictinfo_to_rels(), 
responsible for setting the flag.

To fix the issue, it is enough to set the flag when RestrictInfo 
is being created from EquivalenceClass, inside build_implied_join_equality.

Ticket: ADBDEV-7269

### DIFF
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -23,7 +23,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: write # Required for GHCR access
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-build.yml@v4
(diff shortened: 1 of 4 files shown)


GirgerS:
Regarding the last test inside join_hash.sql: we were not able to run the original test because we ran into a problem with rescanning a motion, which is currently not allowed. It turns out that the planner wasn't aware of the issue for the same reason: the RestrictInfo.contain_outer_query_references flag was not set. As the error is now fixed, and the planner creates materialize nodes to avoid rescanning, we can restore the test to its initial state.

hilltracer (approved):

PR #121, opened 2025-11-17
ADBDEV-7272: Fix escaping for perfmon metrics export

VoidZeroNull0:
# Fix escaping for perfmon metrics export

## What occurs?
Whenever new log rows, containing database names with some special characters,
were tried to be uploaded from `_queries_history.dat` file to the
`queries_history` external table an error considering the size of a `db` column
would occur. It would prevent any new logs to be loaded whatsoever.

## Why it occurs?
The `db` column has a type of `VARCHAR(64)`, thus it means that a larger string
was tried to be put. Obvious reason for this - incorrect escaping or lack of
such whatsoever. Only two symbols were observed to lead to errors: `"` and `|`.
In case of a pipe - it leads to another error about incorrect row structure,
which is logical, as pipes are used as delimiters inside the file.

## How do we fix it?
Before the patch (1f67d39) `db` field used to be written to
`_queries_history.dat`  without double quotes, so no escaping was present.  Now
it has it, as we enclose every database name in double quotes. Also, we double
the already present in database name double quotes.

## Does it fix the problem?
Yes, as now we escape the whole database name string with all of its special
(or not) characters - the very same method is used for the SQL query command
and it works. Doubling double quotes is needed as we need to escape the quotes
to not to end the string to early.

## What was changed?
Minor code additions for escaping inside the `src/backend/gpmon/gpmon.c`

## Tests?
New BDD auto test to check that before mentioned logs are added to
`queries_history` correctly (as they appear there). But, not all symbols can be
auto tested using present testing functions: `'`, `|` and UTF-8 symbols. Errors
occur at different steps of a test for different type of symbol and are
connected to the way commands are issued to the DBSM. Nevertheless, during hand
testing these symbols passed.

## Observations?
`dbuser` column also is not escaped, but I have not managed to recreate the
same issue with it. Yet, it may be worth to add escaping to it in future, but
now it seems like an extremely edge case scenario.

### DIFF
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -41,7 +41,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: read  # Explicit for GHCR access clarity
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-behave.yml@v8
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-behave.yml@v9
     with:
       version: 6
       target_os: ${{ matrix.target_os }}
@@ -59,7 +59,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: read  # Explicit for GHCR access clarity
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-regression.yml@v8
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-regression.yml@v9
     with:
       version: 6
       target_os: ${{ matrix.target_os }}
@@ -77,7 +77,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: read  # Explicit for GHCR access clarity
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-orca.yml@v8
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-orca.yml@v9
     with:
       version: 6
       target_os: ${{ matrix.target_os }}
@@ -113,7 +113,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: write # Required for GHCR access
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-upload.yml@v8
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-upload.yml@v9
     with:
       version: 6
       target_os: ${{ matrix.target_os }}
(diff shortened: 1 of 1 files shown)


src/backend/utils/gpmon/gpmon.c
```diff
@@ -104,6 +105,45 @@ void gpmon_init(void)
 	gpmon.pid = pid;
 }
 
+/* Puts a database name inside double quotes, doubles the quotes for correct escape */
+static char *escape_database_name(char *dbname)
```
bandetto: Is this function relevant to this implementation? Can it be used instead?
https://github.com/greengagedb/greengage/blob/1b63b1ba5480c5a1af4c6499c2ce7771c1ed1dbd/contrib/pg_upgrade/util.c#L257-L282

src/backend/utils/gpmon/gpmon.c
```diff
@@ -104,6 +105,45 @@ void gpmon_init(void)
...
+		{
+			escaped_dbname[j++] = '"';
+			escaped_dbname[j++] = '"';
+		}
+		else
+			escaped_dbname[j++] = c;
+	}
+	escaped_dbname[j++] = '"';
+	escaped_dbname[j] = '\0';
+	pfree(dbname);
```
bandetto: Assuming ownership of `dbname` seems like a bad idea to me. Better use `const ...` in parameters and manage the value in it's initial scope.

src/backend/utils/gpmon/gpmon.c
```diff
@@ -104,6 +105,45 @@ void gpmon_init(void)
...
+
+	/* Count '"'s as they will count as two */
+	for (i = 0; i < len; i++)
+		if (dbname[i] == '"')
+			extra_quotes++;
+
+	size_t	new_len = len + extra_quotes + 3;
+	char *escaped_dbname = (char *) palloc(new_len);
+
+	Assert(escaped_dbname == NULL);
```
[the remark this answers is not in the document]
VoidZeroNull0: Got it.

VoidZeroNull0:
> Could you rephrase it in simpler words?

When gpperfmon tries to to log database with `"`s in name, overflow of `db` column happens, as `"` was not escaped.

>Are you refer db column in the queries_history table? Why have you called it external? It looks like it's not.

Yes, i am referring to it.  I certainly remember that it was external, as it did not allow me to alter types of columns. I am going to put output of `DESCRIBE TABLE` in the next comment.

> Are you referring to the commit you just made to this PR? 

Yup.

> This commit opened with a warming "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

Usage of `--amend` and `--force` could have caused this.  Looks like I should not use them.

> Is the pipe symbol also should be covered with the patch?

It is covered indirectly as all special symbols are escaped when we wrap a string inside double quotes.

> Do we really need to always enclose database name with double quotes?

No, we don't. [This](https://github.com/GreengageDB/greengage/blob/1d42b1f9a119310c60da8eb22cbf68808a149602/src/backend/utils/adt/ruleutils.c#L9808-L9892) recently discovered function checks for cases, when we don't need to wrap in quotes.

bandetto (dismissed):

bandetto (approved):

PR #124, opened 2025-11-17
ADBDEV-8836: Delete shared tuplestores on abort

silent-observer:
Delete shared tuplestores on abort

Previously, the creator of the shared tuplestore was responsible for manually
deleting it after it's no longer used. As a result, tuplestores created by
ShareInputScans weren't deleted if an error occured in the middle of a query.
This patch completely reworks the way shared tuplestores are cleaned up.

This patch adds a shared hashtable that stores the sharing state of all shared
tuplestores. The creator of the tuplestore must specify in advance how many
readers/writers will the tuplestore have. Then the tuplestore is automatically
deleted once the last reader/writer closes it, as signified by num_total_left
counter reaching zero.

For error handling, there's an additional global list on each backend,
containing the tuplestores that are currently open. If abort happens in the
middle of a query, all those tuplestores are closed, and deleted if possible.

It's possible that a backend that was supposed to use a tuplestore aborts
before it even opens it. In that case, the num_total_left counter never
reaches zero, and we need to delete the tuplestore prematurely. In this case,
the aborting process is allowed to delete all tuplestores (of the aborting
session) that aren't currently in use, even if future use was expected. For
this, a separate atomic counter num_current is used.

In case when abort happens in a different backend than the one using the
tuplestore, it can't be deleted immediately since it is currently. Instead,
the aborting backend sets an atomic flag in shared memory, and once all the
tuplestore users are done (either while also aborting, or exiting normally),
the tuplestore will be deleted. This way it is not possible for the tuplestore
to not be deleted, since we can guarantee num_current will reach zero once all
backends exit, and that aborting flag will be set before that point.

Since ShareInputScans might be executing inside InitPlans, which are
dispatched as several queries, the EState's lifetime is not enough for these
tuplestores. We instead store them in current transaction's memory context and
resource owner.

The only other place that uses shared tuplestores are functions executing on
initplans. They aren't directly affected by this patch, since the tuplestores
are present on coordinator and would be deleted anyway, but they still get the
new error handling. Since in that case the number of tuplestore users is 2 (1
writer + 1 reader), 2 is chosen as the default value. This case is checked by
function_extensions regress test.

### DIFF
--- a/src/backend/utils/sort/tuplestore.c
+++ b/src/backend/utils/sort/tuplestore.c
@@ -154,6 +154,9 @@ typedef struct
 	 * Please only set under exclusive lock to prevent races.
 	 */
 	pg_atomic_uint32 aborting;
+	/* Data to identify the SharedFileSet that was used */
+	pid_t		sfs_creator_pid;
+	uint32		sfs_number;
 } TuplestoreSharingState;
 
 /*
@@ -1923,6 +1926,8 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
 	pg_atomic_init_u32(&sstate->num_current, 1);
 	pg_atomic_init_u32(&sstate->num_total_left, ntotal);
 	pg_atomic_init_u32(&sstate->aborting, 0);
+	sstate->sfs_creator_pid = fileset->creator_pid;
+	sstate->sfs_number = fileset->number;
 
 	LWLockRelease(ShareInputScanLock);
 
@@ -2092,9 +2097,14 @@ AtAbort_SharedTuplestores()
 		/*
 		 * No one is using it, and we're aborting so no one will use it
 		 * in the future either. It's safe to delete the files now.
-		 * Also delete the shared memory entry.
+		 * Also delete the shared memory entry. We do not need to delete
+		 * files if the SharedFileSet was already cleaned up.
 		 */
-		BufFileDeleteShared(sisc_fileset, sstate->tag);
+		if (sstate->sfs_creator_pid == sisc_fileset->creator_pid &&
+			sstate->sfs_number == sisc_fileset->number)
+		{
+			BufFileDeleteShared(sisc_fileset, sstate->tag);
+		}
 
 		if (hash_search(shared_tuplestores,
 						sstate->tag,


src/backend/utils/sort/tuplestore.c
```diff
@@ -1709,6 +1730,16 @@ tuplestore_make_shared(Tuplestorestate *state, SharedFileSet *fileset, const cha
...
+	/*
+	 * QE must create shared tuplestores only in transaction lifetime, since InitPlans
+	 * are executed as separate queries and so Executor lifetime is not enough.
+	 * 
+	 * For QD and Utility mode it's fine to use Executor lifetime since tuplestore must
+	 * live only for the duration of one query.
+	 */
+	AssertImply(Gp_role == GP_ROLE_EXECUTE, state->resowner == CurTransactionResourceOwner);
+	AssertImply(Gp_role == GP_ROLE_EXECUTE, CurrentMemoryContext == CurTransactionContext);
+
```
hilltracer: Please explain, how `transaction lifetime` is related with this asserts?

src/test/regress/sql/shared_scan.sql
```diff
@@ -123,3 +123,37 @@ reset gp_cte_sharing;
...
+--
+-- Check error handling in shared scans
+--
+-- Helper function to count the number of temporary files in
+-- pgsql_tmp.
+create or replace function get_temp_file_num() returns int as
+$$
+import os
+fileNum = 0
+for root, directories, filenames in os.walk('base/pgsql_tmp'):
```
hilltracer: Maybe use PG_TEMP_FILES_DIR constant like here:
https://github.com/GreengageDB/greengage/blob/06d1802e5826cb39c6c0e79aa6dbdc7c071c2e19/src/test/regress/regress_gp.c#L2085

src/test/regress/sql/shared_scan.sql
```diff
@@ -123,3 +123,37 @@ reset gp_cte_sharing;
...
+  for filename in filenames:
+    fileNum += 1
+return fileNum
+$$ language plpython3u;
+
+create table sisc(i int) distributed by (i);
+insert into sisc select generate_series(1, 100);
+
+-- Temp file number before running Shared Scan queries
+select get_temp_file_num() as num_temp_files_before
```
hilltracer: Where this call is performed? on QD? Do I undestand right that we have to check QEs?
I've checked the test with comment xaxt.c changes, and it's still passed.

src/backend/utils/sort/tuplestore.c
```diff
@@ -565,6 +573,19 @@ tuplestore_end(Tuplestorestate *state)
...
+ *	Release resources and clean up. Also deletes the tuplestore from the list.
+ */
+void
+tuplestore_end(Tuplestorestate *state)
+{
+	if (state->share_status == TSHARE_WRITER)
+		shared_tuplestores = list_delete_ptr(shared_tuplestores, state);
+	tuplestore_cleanup(state);
+}
+
```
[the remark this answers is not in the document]
silent-observer: It shouldn't in this case, since aborting transaction happens in the error handling path, so ExecutorEnd won't be called anymore after that. However, this is a real issue when support for InitPlans is added, so I think it's better to rework this patch instead of putting it off for later.

src/test/regress/sql/shared_scan.sql
silent-observer: Fixed, now the calls are executed on QEs.

src/backend/utils/sort/tuplestore.c
```diff
@@ -1694,21 +1797,83 @@ tuplestore_set_instrument(Tuplestorestate *state,
...
+	bool		found;
 
 	state->work_set = workfile_mgr_create_set("SharedTupleStore", filename, true /* hold pin */);
 
 	Assert(state->status == TSS_INMEM);
 	Assert(state->tuples == 0);
 	Assert(state->share_status == TSHARE_NOT_SHARED);
+	/* We only support one shared fileset currently */
+	Assert(fileset == get_shareinput_fileset());
+	/*
```
hilltracer: Can we avoid hard-coding the fileset assumption in this low-level module?
To make tuplestore usable in more general contexts.

Maybe store the SharedFileSet pointer inside TuplestoreSharingState
(at the top level).

hilltracer (approved):

bimboterminator1 (approved):

PR #129, opened 2025-11-20
Fix vacuum failure to clean up some segment files of AOCS tables

whitehawk:
Fix vacuum failure to clean up some segment files of AOCS tables

Problem description:
Vacuum failed to clean up segment files for AOCS tables, that were created in a
transaction, which was rolled back.

Root cause:
For AOCS tables, a transaction that creates segment files for the very first
time, inserts a corresponding record into 'pg_aoseg.pg_aocsseg_<oid>' in 
'InsertInitialAOCSFileSegInfo()'. But if the transaction was rolled back, this
record in 'pg_aoseg.pg_aocsseg_<oid>' was no more available. But the physical
segment files still existed.
Vacuum process in AOCSTruncateToEOF() relies on the information from
'pg_aoseg.pg_aocsseg_<oid>' to get all segments vacuum needs to scan. Obviously,
the segment files from the aborted transaction were not visible to it.

Fix:
Add 'heap_freeze_tuple_wal_logged(segrel, segtup)' into
'InsertInitialAOCSFileSegInfo()', so vacuum now can see the new segment files.
It is already done so in 7x (refer to 1306d47183375e3acef0bc23735b450187165d05).

But this change interferes with commit 9e106f51d295fcfb37fc97383286ebeb22426402,
as freezing the tuple partially reverts its logic.
Part of this interference is resolved by preceding cherry-pick of
0a2d3cb7cfe35010afc769cdb970df7d5bb0e0a1, which handles the same problem as 
9e106f51d295fcfb37fc97383286ebeb22426402, but in a different way.
But for 6x additional changes are required in 'UpdateAOCSFileSegInfo()', so this
patch adds usage of 'deformAOCSVPInfo()' into it in the manner intended by
0a2d3cb7cfe35010afc769cdb970df7d5bb0e0a1.

Plus this change requires update of the output of the test
'uao_crash_compaction_column', because the output of 'gp_toolkit.__gp_aocsseg()'
now contains records about segment files created by the interrupted vacuum
command. 

----------------

This PR has to be rebased, as it contains cherry-pick of 0a2d3cb7cfe35010afc769cdb970df7d5bb0e0a1

### DIFF
--- a/src/backend/access/aocs/aocssegfiles.c
+++ b/src/backend/access/aocs/aocssegfiles.c
@@ -115,6 +115,7 @@ InsertInitialAOCSFileSegInfo(Relation prel, int32 segno, int32 nvp)
 	segtup = heap_form_tuple(RelationGetDescr(segrel), values, nulls);
 
 	CatalogTupleInsert(segrel, segtup);
+	heap_freeze_tuple_wal_logged(segrel, segtup);
 
 	heap_freetuple(segtup);
 	heap_close(segrel, RowExclusiveLock);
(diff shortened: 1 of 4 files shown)


RekGRpth (dismissed):

src/include/access/aocssegfiles.h
```diff
@@ -137,7 +137,6 @@ extern AOCSFileSegInfo *GetAOCSFileSegInfo(Relation prel,
 				   Snapshot appendOnlyMetaDataSnapshot,
 				   int32 segno);
 
-
```
KnightMurloc: remove this change

whitehawk:
BTW, I've found another scenario that leads to orphaned files of AOCS tables, but it has completely different root cause, so it should be handled separately. Details are in ADBDEV-8873 - FYI.

src/backend/access/aocs/aocssegfiles.c
```diff
@@ -947,11 +948,15 @@ UpdateAOCSFileSegInfo(AOCSInsertDesc idesc)
...
-	Assert(VARSIZE(dv) == aocs_vpinfo_size(nvp));
-	AOCSVPInfo *oldvpinfo = (AOCSVPInfo *) dv;
+	AOCSFileSegInfo *seginfo;
+	seginfo = (AOCSFileSegInfo *) palloc0(aocsfileseginfo_size(nvp));
+
+	deformAOCSVPInfo(prel,
+					 (struct varlena *) DatumGetPointer(d_tmp),
+					 seginfo);
+
+	AOCSVPInfo *oldvpinfo = (AOCSVPInfo *) &seginfo->vpinfo;
```
KnightMurloc: does this change more related to the previous commit?
We also don't need to cast seginfo->vpinfo to AOCSVPInfo

RekGRpth (approved):

PR #133, opened 2025-11-22
Add the temp_tables_stat extension

andr-sokolov:
Add the temp_tables_stat extension

This extension is useful to find out how much disk space is occupied by
temporary tables and which user created them in which session.
The extension tracks the creation and deletion of temporary table files on
segments. The list of files is stored in the shared memory of each segment.
If memory allocated using ShmemInitStruct is not enough, the extension uses DSM.
The extension adds the tts_get_seg_files function to get the tracked files,
their sizes, user and session where they was created from segments.

### DIFF
--- a/.abi-check/6.29.1/postgres.symbols.ignore
+++ b/.abi-check/6.29.1/postgres.symbols.ignore
@@ -1,0 +1,1 @@
+ConfigureNamesBool_gp
--- a/.abi-check/6.29.2/postgres.symbols.ignore
+++ b/.abi-check/6.29.2/postgres.symbols.ignore
@@ -1,0 +1,1 @@
+ConfigureNamesBool_gp
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -37,6 +37,41 @@ operating systems:
   DockerHub. Runs for push to `main` (retags to `latest`) and tags (uses tag
   like `6.28.2`) after build.
 
+## Release Workflow
+
+A separate workflow, `Greengage release`, handles the uploading of Debian packages
+to GitHub releases. It is triggered when a release is published and uses a
+composite action to manage package deployment.
+
+### Key Features
+
+- **Triggers:** `release: [published]` - Runs when a release is published,
+including re-publishing.
+- **Concurrency:** Uses the same concurrency group as the CI workflow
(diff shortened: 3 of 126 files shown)


gpcontrib/temp_tables_stat/Makefile
```diff
@@ -0,0 +1,20 @@
+# gpcontrib/temp_tables_stat/Makefile
+
+MODULE_big = temp_tables_stat
+OBJS = $(MODULE_big).o $(WIN32RES)
```
whitehawk: What is the purpose of adding $(WIN32RES)?

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,384 @@
...
+	bool		found;
+
+	if (prev_shmem_startup_hook)
+		(*prev_shmem_startup_hook)();
+
+	head = ShmemInitStruct("temp_tables_stat", sizeof(TTSHeadNode), &found);
+	if (found)
+		return;
+
+	LWLockInitialize(&head->lock, 0);
```
whitehawk: If we've used 
```
RequestAddinLWLocks(1);
```
we should use `*LWLockAssign()` to obtain a lw lock.

`LWLockInitialize()` should be used with `int	LWLockNewTrancheId(void)`.

Please refer [here](https://github.com/GreengageDB/greengage/blob/b5f63c6571e4300187a1d88e266520d0187fdd82/src/include/storage/lwlock.h#L229)

whitehawk:
@andr-sokolov , please add a README file with some general info about the extension and description of main use cases.

src/test/isolation2/expected/temp_tables_stat.out
whitehawk: I suggest adding test cases with partitioned tables.

src/test/isolation2/sql/temp_tables_stat.sql
whitehawk: Can the test be moved to `gpcontrib/temp_tables_stat`? So it will be also a part of the extension

whitehawk (approved):

silent-observer (approved):

PR #139, opened 2025-11-28
Fix orphaned segment files for added and rolled back columns for AOCS tables

whitehawk:
Fix orphaned segment files for added and rolled back columns for AOCS tables

Problem description:
In case we added a new column to a AOCS table in a transaction that was later
rolled back, a non-empty segment file could be left, and vacuum didn't truncate
it.

Root cause:
Vacuum does truncation of AOCS segment files in 2 places:
1. In AOCSSegmentFileTruncateToEOF(). It iterated over 'vpinfo.nEntry', which
doesn't count the added and rolled back attribute, so its segment file was not
processed here.  
2. In AOCSCompaction_DropSegmentFile() function (if compaction is required for
the segment). It iterates over attributes retrieved by
RelationGetNumberOfAttributes(). But, the added and rolled back attribute is not
counted by RelationGetNumberOfAttributes(), so its segment file was not
processed here as well.

Fix:
In AOCSSegmentFileTruncateToEOF() iterate and truncate all segment files for
a given 'segno' until they exist. We stop once we didn't find a file and column
number is above 'vpinfo.nEntry'. We assume that the segment files exist or stop
existing for all columns thereafter - similar logic we already use in
ao_foreach_extent_file().

We do it not in AOCSCompaction_DropSegmentFile(), as it will not be called if
compaction is not required, while the orphaned segment files may still exist in
this case.

### DIFF
--- a/src/backend/access/aocs/aocs_compaction.c
+++ b/src/backend/access/aocs/aocs_compaction.c
@@ -118,7 +118,6 @@ AOCSSegmentFileTruncateToEOF(Relation aorel,
 	{
 		int64		segeof = 0;
 		char		filenamepath[MAXPGPATH];
-
 		File		fd;
 		int32		fileSegNo;
 


src/test/regress/sql/vacuum_full_ao.sql
```diff
@@ -49,5 +62,45 @@ vacuum table_ao_col;
...
-drop function getTableSegFiles(t regclass, out gp_contentid smallint, out filepath text);
 drop table table_ao_col;
+
+-- Test that vacuum can process segment files created for a new column in an aborted transaction (case 1)
+create table table_ao_col_1 (i int, j int, k int) with (appendonly='true', orientation="column");
+insert into table_ao_col_1 select i, i + 1, i + 2 from generate_series(1, 20) i;
+
+begin;
+alter table table_ao_col_1 add column a int;
+update table_ao_col_1 set a = 1 where true;
```
bandetto:
```suggestion
update table_ao_col_1 set a = 1;
```

src/test/regress/sql/vacuum_full_ao.sql
```diff
@@ -49,5 +62,45 @@ vacuum table_ao_col;
...
+rollback;
+
+select cmdCheckSegmentFileSizes('table_ao_col_1') check_segfiles_size_1
+\gset
+
+delete from table_ao_col_1 where true;
```
bandetto:
```suggestion
delete from table_ao_col_1;
```

src/backend/access/aocs/aocs_compaction.c
```diff
@@ -108,16 +108,24 @@ AOCSSegmentFileTruncateToEOF(Relation aorel,
 
 	segno = fsinfo->segno;
 
-	for (j = 0; j < fsinfo->vpinfo.nEntry; ++j)
+	/*
+	 * We try to truncate all segment files beyond
+	 * `vpinfo.nEntry`, as we may have non-empty segment files
+	 * left by ADD COLUMN, which was rolled back. It is similar to logic in
+	 * ao_foreach_extent_file().
```
bandetto: `ao_foreach_extent_file()` always checks all segment numbers from 0 to 128. The logic below stops checking on the least segment number that doesn't have a corresponding file.
Can we miss a segment file if the numbering is not consequential, e.g. 1, 2, 4?

PR #140, opened 2025-11-30
[ADBDEV-8852] Refactor gpcheckcat query

bimboterminator1:
Refactor leaked schema retrieving query

During GGDB functioning lots of various temp schema might be created.
In case of any mailfunction, when backend could not release resources
and drop temp objects, some orphaned temp schemas may be left.
Essentially, autovacuum processes objects of such schema and drops
relations, but by its design it can't drop whole schemas.

Orphaned temp tables could be also created during test suits run,
leaving such schemas either on segments or coordinator only. Sometimes
orphaned schemas may have a really specific id in its 'pg_temp_%'
pattern, which does not allow manual clean up queries to recognize
them. Known examples are: 1) after isolation2 test (autovacuum-temp-table) there
is orphaned temp schema is left at seg0. Since the schema was created in
utility session it is assigned very specific id (pg_temp_06 at the time of this
patch). And during leaked schemas cleanup by gpcheckcat it couldn't be
found because query in leak_schema_dropper.py succeded to join the
schema id with active coordinator background worker (logical replication
launcher), thus leaving the schema untouched. 2) gpcheckcat session id
may get into collision with leaked schema id, what leads to the same
consequences.

First of all, this patch rewrites the gpcheckcat's query by uniting the temp schema
retrieval into the cte clause. Then the  the WHERE clause conditions are
complemented, checking the database correspondence of temp schema and
active session (gpcheckcat connects to a database but orphaned schema from
current database has joined with session connected to another database or which
does not have database) and the session id collision absence.

The behave test is added, testing case 1).

### DIFF
--- a/contrib/auto_explain/auto_explain.c
+++ b/contrib/auto_explain/auto_explain.c
@@ -387,10 +387,6 @@ explain_ExecutorEnd(QueryDesc *queryDesc)
 		MemoryContext oldcxt;
 		double		msec;
 
-		/* Wait for completion of all qExec processes. */
-		if (queryDesc->estate->dispatcherState && queryDesc->estate->dispatcherState->primaryResults)
-			cdbdisp_checkDispatchResult(queryDesc->estate->dispatcherState, DISPATCH_WAIT_NONE);
-
 		/*
 		 * Make sure we operate in the per-query context, so any cruft will be
 		 * discarded later during ExecutorEnd.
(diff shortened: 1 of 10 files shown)


gpMgmt/bin/gpcheckcat_modules/leaked_schema_dropper.py
```diff
@@ -3,35 +3,36 @@
...
+          WHERE nspname ~ '^pg_t(emp|oast_temp)_[0-9]+'
+        )
+        SELECT DISTINCT nspname as schema
+        FROM temp_schemas n 
+        LEFT JOIN pg_stat_activity x USING (sess_id)
+        WHERE x.sess_id IS NULL 
+           OR x.backend_type LIKE 'autovacuum%'
+           OR x.datname IS NULL 
+           OR x.datname <> current_database()
+           OR x.pid = pg_backend_pid()
```
RekGRpth: Could you please elaborate on the meaning of these three new additional conditions?

RekGRpth: > 1. x.datname <> current_database()

We ran `gpcheckcat` in one database, but we want it to delete schemas in another? How will it do this if it's not connected to it?

PR #141, opened 2025-12-01
ADBDEV-9025: Prohibit skip-level correlated queries in Postgres planner

VoidZeroNull0:
Prohibit skip-level correlated queries in Postgres planner

What occurs?
Postgres planner creates an invalid plan with skip-level correlation involving
master-only and replicated tables. The plan fails to be executed during
execution.

Why it occurs?
Skip-level correlation (at least for distributed tables) seems to be prohibited
in Postgres planner due to rescanning issues that arise - they appear, because
the lowest level plan is made an initplan without correct motion node applied.
ORCA planner produces valid plans for such queries, but not with master-only
tables - it seems to lack a feature to work with them whatsoever.

How do we fix this?
Let's prohibit any skip-level correlation for Postgres Legacy planner as it
leads to rescan issues because of absent motions. 

Alternative fixes?
Changing some checks for initplans creation and enabling skip-level correlation
make Postgres produces correct plan - initplan is changed to subplan and
correct motions are applied. Yet, one test from rangefuncs fails. So, some
research is needed.

Tests?
Newly added tests check that ORCA is able to plan and execute simple skip-level
correlated plans, whereas Postgres planner outputs an error about lack of
skip-level correlation support.

### DIFF
--- a/src/backend/optimizer/plan/subselect.c
+++ b/src/backend/optimizer/plan/subselect.c
@@ -420,7 +420,8 @@ get_first_col_type(Plan *plan, Oid *coltype, int32 *coltypmod,
 }
 
 /**
- * Returns true if query refers to any distributed table.
+ * Returns true if query refers to any table on this level or on some other
+ * subquery level, if recursive is true.
  */
 bool QueryHasDistributedRelation(Query *q, bool recursive)
 {
--- a/src/test/regress/sql/qp_correlated_query.sql
+++ b/src/test/regress/sql/qp_correlated_query.sql
@@ -846,24 +846,19 @@ DROP TABLE IF EXISTS skip_correlated_replicated;
 CREATE TABLE skip_correlated_partitioned (
     a INT
 ) DISTRIBUTED BY (a);
-INSERT INTO skip_correlated_partitioned VALUES(1);
-INSERT INTO skip_correlated_partitioned VALUES(2);
-INSERT INTO skip_correlated_partitioned VALUES(3);
+INSERT INTO skip_correlated_partitioned VALUES (1), (2), (3);
 
 CREATE TABLE skip_correlated_random (
     b INT
 ) DISTRIBUTED RANDOMLY;
-INSERT INTO skip_correlated_random VALUES(1);
-INSERT INTO skip_correlated_random VALUES(2);
-INSERT INTO skip_correlated_random VALUES(3);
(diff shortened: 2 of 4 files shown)


VoidZeroNull0:
Just to note, eventual PR description will be much smaller, wrapped around 80 symbols and without fancy formatting - it looks as it is now to provide more context to the problem and be more readable than the plain text.

src/test/regress/sql/qp_correlated_query.sql
```diff
@@ -833,7 +833,78 @@ DROP TABLE skip_correlated_t3;
...
+EXPLAIN (COSTS OFF)
+SELECT (
+    SELECT (
+        SELECT dbid FROM gp_segment_configuration WHERE dbid = a
+    )
+) FROM skip_correlated_partitioned;
 
+DROP TABLE IF EXISTS skip_correlated_partitioned;
+DROP TABLE IF EXISTS skip_correlated_random;
+DROP TABLE IF EXISTS skip_correlated_replicated;
```
RekGRpth: put `drop tables if exists` BEFORE new test in skip block, and here - WITHOUT `if exists`

src/backend/optimizer/plan/subselect.c
```diff
@@ -435,17 +436,10 @@ bool QueryHasDistributedRelation(Query *q, bool recursive)
 				&& QueryHasDistributedRelation(rte->subquery, true))
 			return true;
 
+		/* Really, any kind of distribution policy causes rescan issues */
 		if (rte->relid != InvalidOid
 				&& rte->rtekind == RTE_RELATION)
-		{
-			GpPolicy *policy = GpPolicyFetch(rte->relid);
-			if (GpPolicyIsPartitioned(policy))
```
ShapatinAA: Why can we return true not only for partitioned, but for replicated tables also? Why should we prohibit skip-level for replicated tables?

VoidZeroNull0:
> It would be nice to also have the final commit text so that it can be reviewed too.

PR comment updated and ready for review.

silent-observer (approved):

ShapatinAA:
Add name of PR (without ticket) as a first line of description.

VoidZeroNull0:
> Add name of PR (without ticket) as a first line of description.

Done.

PR #142, opened 2025-12-01
ADBDEV-8899: Prevent ORCA from generating redundant redistributions on joins

GirgerS:
Prevent ORCA from generating redundant redistributions on joins

When performing a join on types within the same distribution opfamily,
there is no need to create a redistribution motion because 
the hash functions of these types are compatible with each other.

ORCA, on the other hand, is not aware of this and creates a redistribution.

The easy part of the fix is to detect such cases. 
The hard part is to integrate this logic into the existing one. 
It can be done in at least two ways:

Inside the distribution matching logic, the current distribution 
can be safely changed to the desired one, as they are 
practically equivalent.

Alternatively, the actual distribution of the scan can be returned. 
In this case, every piece of code that expects the derived distribution 
to directly match the required one should be changed as well.

The full-fledged version of the first approach would require adding 
a new enforcing type to the EPropEnforcingType enum, as well as changing 
CDistributionSpec::Matches and CDistributionSpec: FSatisfies
virtual functions overloads. The enforcing logic inside 
CDistributionSpecHashed::AppendEnforcers should also be changed. 
To minimize the amount of changes, only 
CDistributionSpecHashed::AppendEnforcers function can be modified
to decide how to enforce a distribution on its own, 
but it doesn’t seem to fit into the current logic. 

The second approach requires thorough investigation to determine every 
piece of code that should perform distribution matching instead of direct 
expression matching.

Each of these approaches is not trivial. The second one was chosen, 
because it makes the logic explicit. It allows the expression’s parent to 
know about the derived distribution, enabling specialized logic if necessary.

In this patch, an overload of CUtils::Equals supporting distribution 
matching is added.
Other matching functions are modified to support it as well. 
Where required, expression-matching versions of these functions are 
changed to distribution-matching ones.

Also, to make the casting logic of the planners easier to observe, 
a new test file was added.

Ticket: ADBDEV-8899

### DIFF
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -41,7 +41,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: read  # Explicit for GHCR access clarity
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-behave.yml@v15
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-behave.yml@v18
     with:
       version: 7
       target_os: ${{ matrix.target_os }}
--- a/ci/docker-compose.yaml
+++ b/ci/docker-compose.yaml
@@ -1,55 +1,42 @@
+x-common-service: &common-service
+  image: "${IMAGE}"
+  privileged: true
+  sysctls:
+    kernel.sem: 500 1024000 200 4096
+  init: true
+  ulimits:
+    nofile: 65535
+  entrypoint: >
+    sleep infinity
+
 services:
   cdw:
-    image: "${IMAGE}"
+    <<: *common-service
     working_dir: /home/gpadmin
     hostname: cdw
     volumes:
-      - "$PWD/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa"
-      - "$PWD/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub"
-      - "$PWD/allure-results:/tmp/allure-results"
-      - "$PWD/logs_cdw:/logs"
-    privileged: true
-    sysctls:
-      kernel.sem: 500 1024000 200 4096
-    init: true
-    entrypoint: >
-      sleep infinity
+      - ${PWD}/allure-results:/tmp/allure-results
+      - ${PWD}/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub
+      - ${PWD}/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa
+      - ${PWD}/logs_cdw:/logs
   sdw1:
-    image: "${IMAGE}"
-    privileged: true
+    <<: *common-service
     hostname: sdw1
     volumes:
-      - "$PWD/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa"
-      - "$PWD/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub"
-      - "$PWD/logs_sdw1:/logs"
-    sysctls:
-      kernel.sem: 500 1024000 200 4096
-    init: true
-    entrypoint: >
-      sleep infinity
+      - ${PWD}/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub
+      - ${PWD}/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa
+      - ${PWD}/logs_sdw1:/logs
   sdw2:
-    image: "${IMAGE}"
-    privileged: true
+    <<: *common-service
     hostname: sdw2
     volumes:
-      - "$PWD/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa"
-      - "$PWD/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub"
-      - "$PWD/logs_sdw2:/logs"
-    sysctls:
-      kernel.sem: 500 1024000 200 4096
-    init: true
-    entrypoint: >
-      sleep infinity
+      - ${PWD}/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub
+      - ${PWD}/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa
+      - ${PWD}/logs_sdw2:/logs
   sdw3:
-    image: "${IMAGE}"
-    privileged: true
+    <<: *common-service
     hostname: sdw3
     volumes:
-      - "$PWD/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa"
-      - "$PWD/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub"
-      - "$PWD/logs_sdw3:/logs"
-    sysctls:
-      kernel.sem: 500 1024000 200 4096
-    init: true
-    entrypoint: >
-      sleep infinity
+      - ${PWD}/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub
+      - ${PWD}/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa
+      - ${PWD}/logs_sdw3:/logs
--- a/ci/scripts/behave_collect_logs.bash
+++ b/ci/scripts/behave_collect_logs.bash
@@ -1,0 +1,13 @@
+feature=$1
+
+params=(
+  "./ d gpAdminLogs"
+  "gpdb_src/gpAux/gpdemo/datadirs/ d log"
+  "gpdb_src/gpAux/gpdemo/datadirs/ d pg_log"
+)
+for param in "${params[@]}"; do
+  read -r path type name <<< "$param"
+  [ -d "$path" ] && find "$path" -name "$name" -type "$type" \
+    -exec tar -rf "/logs/behave_${feature}_${name}.tar" "{}" \;
+done
+chmod -R a+rwX /logs
--- a/ci/scripts/run_behave_tests.bash
+++ b/ci/scripts/run_behave_tests.bash
@@ -8,6 +8,8 @@ behave_tests_dir="gpMgmt/test/behave/mgmt_utils"
 
 clusters="~concourse_cluster"
 
+docker_compose_path="ci/docker-compose.yaml"
+
 if [ $# -eq 0 ]
 then
   # TODO cross_subnet and gpssh tests are excluded
@@ -44,12 +46,12 @@ run_feature() {
     local project="${feature}_demo"
   fi
   echo "Started $feature behave tests on cluster $cluster and project $project"
-  docker-compose -p $project -f ci/docker-compose.yaml --env-file ci/.env up -d
+  docker compose -p $project -f "$docker_compose_path" --env-file ci/.env up -d
   # Prepare ALL containers first
-  local services=$(docker-compose -p $project -f ci/docker-compose.yaml config --services | tr '\n' ' ')
+  local services=$(docker compose -p $project -f "$docker_compose_path" config --services | tr '\n' ' ')
   for service in $services
   do
-    docker-compose -p $project -f ci/docker-compose.yaml exec -T \
+    docker compose -p $project -f "$docker_compose_path" exec -T \
       $service bash -c "mkdir -p /data/gpdata && chmod -R 777 /data &&
         source gpdb_src/concourse/scripts/common.bash && install_gpdb &&
        ./gpdb_src/concourse/scripts/setup_gpadmin_user.bash" &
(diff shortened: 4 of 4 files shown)


src/test/regress/sql/gp_cast.sql
```diff
@@ -0,0 +1,207 @@
...
+    join cst_int4 using(a);
+
+reset optimizer_join_order;
+
+
+-- Confirm that the same logic is correct for other join types
+explain (verbose) select * from cst_int2 full join cst_int4 using(a);
+select * from cst_int2 full join cst_int4 using(a);
+
+-- BUG: this test is failing for ORCA
```
whitehawk: Was this bug reproduced before the patch? If yes, do we have a ticket for this problem?

src/test/regress/sql/gp_cast.sql
```diff
@@ -0,0 +1,207 @@
...
+where cst_int4.a not in (select * from cst_int2);
+
+explain (verbose)
+select * from cst_int2 natural join cst_int4;
+select * from cst_int2 natural join cst_int4;
+
+
+-- Here, insead of an implicit cast, an explicit one is present.
+--    The postgres-based planner should require a redistribution, because
+--    distribution of the cst_int2 table is not directly equal to the left-hand side of the expression
```
whitehawk:
```suggestion
--    distribution of the cst_int2 table is not directly equal to the left-hand side of the expression.
```

src/test/regress/sql/gp_cast.sql
```diff
@@ -0,0 +1,207 @@
...
+create table cst_int2_int4 as (select gen::int2 as a, gen::int4 as b from generate_series(1, 10) as gen) distributed by (a, b);
+create table cst_int4_int8 as (select gen::int4 as a, gen::int8 as b from generate_series(1, 10) as gen) distributed by (a, b);
+
+
+-- Perform an inner join on all hash opfamilies containing more than one type.
+-- We shouldn't see any redistributions, because:
+--    The postgres-based planner has operators that work on different types
+--    ORCA doesn't support such operations, but it knows when a cast doesn't change a distribution
+
+set optimizer_join_order = query;
```
whitehawk: why do we need setting of this GUC?

src/backend/gporca/libgpopt/src/base/CDistributionSpecHashed.cpp
```diff
@@ -267,7 +267,7 @@ CDistributionSpecHashed::FMatchSubset(
 				opfamily_other = (*pdsHashed->m_opfamilies)[ulInner];
 			}
 
-			if (CUtils::Equals(pexprOwn, pexprOther) &&
+			if (CUtils::Equals(pexprOwn, pexprOther, true) &&
```
whitehawk: why not

```suggestion
			if (CUtils::EqualDistributions(pexprOwn, pexprOther) &&
```
?

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -1513,6 +1516,78 @@ CUtils::Equals(const CExpression *pexprLeft, const CExpression *pexprRight)
...
+				if (CUtils::Equals(mdidSourceOpfamily, mdidTargetOpfamily))
+				{
+					pexprLeft = pexprChild;
+				}
+			}
+		}
+
+		if (pexprRight->Pop()->Eopid() == COperator::EopScalarCast &&
+			pexprLeft->Pop()->Eopid() == COperator::EopScalarIdent)
+		{
```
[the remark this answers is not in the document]
GirgerS: We can, but I don't think it is necessary, as long as it is obvious that these blocks are duplicated. And it should be, since they are they are next to each other. 

The reason I think it won't cause any problems is that if someone is going to change one of the blocks, it should be obvious that the other block needs to be changed as well.

But yes, it should be trivial to add, lets say, a 'fAllowedToSkipCast' function or something, and then call it twice.

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -1513,6 +1516,78 @@ CUtils::Equals(const CExpression *pexprLeft, const CExpression *pexprRight)
 		return true;
 	}
 
+	// Allow a column reference (int2_column) and the same column reference casted to
+	// a different type within the same distribution opfamily (int2_column::int4) be considered equal
+	// In this case, presence if the cast doesn't change how
+	// the value is distributed, meaning that redistribution motion is not required
+	//
+	// Note that this check is not quite permissive, because distribution might not be changed even if there are multiple casts in a row,
```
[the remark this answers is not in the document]
whitehawk: > Do we obey this rule in ORCA?

Not strictly. But this comment outstands too much all surrounding code, so it would be nice to fix it.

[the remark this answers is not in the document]
GirgerS: Makes sense. Should've been fixed [here](https://github.com/GreengageDB/greengage/pull/142/changes/c65a60df1c423ff7ddae34ca5f4dce7c736fe50b).

src/backend/gporca/libgpopt/src/base/CUtils.cpp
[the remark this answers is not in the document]
ShapatinAA: Maybe not even a function but a swap of `pexprLeft` and `pexprRight` values?

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -1513,6 +1552,40 @@ CUtils::Equals(const CExpression *pexprLeft, const CExpression *pexprRight)
 		return true;
 	}
 
+	// Allow a column reference (int2_column) and the same column reference
```
whitehawk: What is `int2_column`? If the column with type `int2` is meant here, please use more precise wording like `(for example a column having 'int2' type)`. I'm saying it because `int2_column` looks more like an identifier in the code, and the reader might start looking for it.

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -1513,6 +1552,40 @@ CUtils::Equals(const CExpression *pexprLeft, const CExpression *pexprRight)
...
+	// casted to a different type within the same distribution opfamily
+	// (int2_column::int4) be considered equal
```
whitehawk:
```suggestion
	// (int2_column::int4) be considered equal.
```

Please use dots here and in other similar places, where we have more than one sentence. It improves readability.

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -1495,9 +1496,47 @@ CUtils::Equals(const CExpressionArray *pdrgpexprLeft,
...
+
+	CMDAccessor *mda = COptCtxt::PoctxtFromTLS()->Pmda();
+	CScalarCast *popCast = CScalarCast::PopConvert(pexprFirst->Pop());
+	CExpression *pexprChild = (*pexprFirst)[0];
+	if (COperator::EopScalarIdent != pexprChild->Pop()->Eopid())
+	{
+		return nullptr;
+	}
+
+	CScalarIdent *popChild = CScalarIdent::PopConvert(pexprChild->Pop());
```
whitehawk: can we guarantee that `pexprChild`, `popChild`, `popCase`, `mda`... (everything we dereference here in any possible way) are not null?

GirgerS: *Pop()* explicitly checks that *m_pop* is not null.

*PopConvert()* and *RetrieveType()* may return nullptr from *dynamic_cast*. But they also assert that we can perform the cast using available information about a subtype. Meaning that if these function fail, there is a severe bug in the logic, so none of the places that call them double-checks the returned pointer. 

*MdidType()* can fail only if the required private members are not set properly, again, implying a bug in the logic.

*pexprChild* should be present, because the functions used to create a *CScalarCast* expression (*CUtils::PexprCast*, *CCastUtils::PexprCast*) set it. Also, the same thing is already done [here](https://github.com/GreengageDB/greengage/blob/10e350d628d9399179ff82c34e43f1cfc2b9716b/src/backend/gporca/libgpopt/src/operators/CPhysicalJoin.cpp#L640-L652).

*CMDAccessor* is a little bit more difficult to trace because of the abstractions, but it is it is essentially a global variable that is assigned [here](https://github.com/GreengageDB/greengage/blob/10e350d628d9399179ff82c34e43f1cfc2b9716b/src/backend/gporca/libgpopt/src/optimizer/COptimizer.cpp#L287). It has no business being null.

whitehawk (dismissed):

src/test/regress/expected/qp_join_union_all_optimizer.out
```diff
@@ -343,26 +343,28 @@ explain analyze select c1 from dist_view_large_subquery join inner_1 on c1 = cc;
          ->  Append  (cost=0.00..879.51 rows=66667 width=4) (never executed)
                ->  Nested Loop  (cost=0.00..447.37 rows=33334 width=4) (never executed)
                      Join Filter: true
-                     ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..431.01 rows=1 width=8) (never executed)
-                           ->  Finalize Aggregate  (cost=0.00..431.01 rows=1 width=8) (actual time=11.339..11.339 rows=1 loops=1)
-                                 ->  Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..431.01 rows=1 width=8) (actual time=8.341..11.310 rows=3 loops=1)
-                                       ->  Partial Aggregate  (cost=0.00..431.01 rows=1 width=8) (actual time=1.028..1.029 rows=1 loops=1)
-                                             ->  Seq Scan on dist_small_1  (cost=0.00..431.01 rows=334 width=1) (actual time=3.334..3.385 rows=340 loops=1)
-                     ->  Index Scan using dist_large_1_index on dist_large_1  (cost=0.00..15.65 rows=33334 width=4) (never executed)
+                     ->  Redistribute Motion 1:3  (slice2)  (cost=0.00..431.01 rows=1 width=8) (never executed)
```
ShapatinAA: Looks a little bit strange to me. Especially in context of making query slower. Why did that happen? Previously we broadcasted one segment to all others, so we got full information everywhere, but here we redistribute data from one node to other nodes (by hash). 
So after that we forced to have broadcast motion in other part of nestloop. 
There's no joins below redistribute motion, so i don't quite get why this plan got changes.

src/test/regress/expected/gporca_optimizer.out
```diff
@@ -13166,21 +13166,22 @@ from foo l1 where b in (select ab
...
+         ->  Seq Scan on foo
          ->  Materialize
-               ->  HashAggregate
-                     Group Key: tbtree.b
-                     ->  Redistribute Motion 3:3  (slice2; segments: 3)
-                           Hash Key: tbtree.b
-                           ->  Seq Scan on tbtree
- Optimizer: Pivotal Optimizer (GPORCA)
-(12 rows)
+               ->  Redistribute Motion 3:3  (slice2; segments: 3)
```
ShapatinAA: This redistribute motion looks alike one I commented in `qp_join_union_all`. Like planner doesn't do well with aggregate functions.

ShapatinAA (approved):

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -4834,35 +4916,57 @@ CUtils::PexprMatchEqualityOrINDF(
...
+		IMDId *mdidOpfamilyOuter =
+			mdAccessor->RetrieveType(pmdidTypeOuter)->GetDistrOpfamilyMdid();
+
+		// Callers of this function expect that they can distribute by the
+		// expression returned, and there were no guarantees
+		// about it before this call.
+		//
+		// For the time being, just don't return such expressions,
+		// but it doesn't seem that this logic belongs here.
+		// TODO: check if refactoring it is possible.
```
whitehawk: If there are no determined plans for the refactoring, I suggest not to pollute the code with such TODOs.

GirgerS: To clarify, is the problem the presence of the TODO, or the last sentence itself? Because I think it is worth clarifying that the logic belongs here for legacy reasons, not because it is the best place for it. But the TODO may indeed be useless given how development is organized.

whitehawk:
> And considering the fact that such cast chains shouldn't occur without a user manually adding them, I thought that this case is not worth the effort.

Ok, I would agree with it. This scenario seems to have not much practical usage, I think we can leave it as is for now.

But I have another scenario.
Given 3 tables:
```
create table cst_int2 as (select a::int2 from generate_series(0, 10) as a) distributed by (a);
create table cst_int4 as (select a::int4 from generate_series(5, 15) as a) distributed by (a);
create table cst_int8 as (select a::int8 from generate_series(5, 15) as a) distributed by (a);
```

Their join without casts is planned without redistribute motion:
```
 explain (costs off) select * from cst_int2 l join cst_int4 r on l.a = r.a join cst_int8 p on l.a = p.a;
                   QUERY PLAN
------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)
   ->  Hash Join
         Hash Cond: ((l.a)::integer = r.a)
         ->  Hash Join
               Hash Cond: (p.a = (l.a)::bigint)
               ->  Seq Scan on cst_int8 p
               ->  Hash
                     ->  Seq Scan on cst_int2 l
         ->  Hash
               ->  Seq Scan on cst_int4 r
 Optimizer: GPORCA
(11 rows)

```
But if I add a cast to the second join condition, there is a redistribute motion:
```
explain (costs off) select * from cst_int2 l join cst_int4 r on l.a = r.a join cst_int8 p on l.a::int4 = p.a::int4;
                            QUERY PLAN
------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)
   ->  Hash Join
         Hash Cond: ((l.a)::integer = r.a)
         ->  Hash Join
               Hash Cond: (int4(p.a) = (l.a)::integer)
               ->  Redistribute Motion 3:3  (slice2; segments: 3)
                     Hash Key: int4(p.a)
                     ->  Seq Scan on cst_int8 p
               ->  Hash
                     ->  Seq Scan on cst_int2 l
         ->  Hash
               ->  Seq Scan on cst_int4 r
 Optimizer: GPORCA
(13 rows)
```

Can you please check it?

src/backend/gporca/libgpopt/src/base/CUtils.cpp
GirgerS: Agreed.

ShapatinAA (approved):

PR #144, opened 2025-12-02
ADBDEV-8673: Add support for ShareInputScans inside InitPlans

silent-observer:
Add support for ShareInputScans inside InitPlans

Previously ShareInputScans could not be used inside InitPlans due to
synchronization issues. This happened because InitPlans are executed
sequentially: they are separate queries dispatched to QEs, and each query has
to complete before the next one is started.

This conflicts with ShareInputScan execution (in cross-slice case), which
assumes that producer and consumer slices are both available at the same time,
so producer must notify consumers through shared memory when it's done
preparing the tuplestore, and the consumers must notify the producer when
they're done reading it, so it's okay to producer to delete the tuplestore.

After the tuplestore cleanup logic was moved to tuplestore.c by
80b6ac8c6058ebdb68a61a1409b39cd3c3565d6e, "done" notifications are no longer
necessary and can be removed. We also can move "ready" notifications to
tuplestore's shared state, which now can handle InitPlans properly, including
cleanup during abort.

This patch also removes some unnecessary atomics from shared tuplestore state,
since these values are only accessed under lock anyway due to synchronization
requirements.

### DIFF
--- a/gpMgmt/bin/gpcheckcat_modules/leaked_schema_dropper.py
+++ b/gpMgmt/bin/gpcheckcat_modules/leaked_schema_dropper.py
@@ -3,35 +3,30 @@ from gppylib.utils import escapeDoubleQuoteInSQLString
 class LeakedSchemaDropper:
 
     # This query does a union of all the leaked temp schemas on the coordinator as well as all the segments.
-    # The first part of the query uses gp_dist_random which gets the leaked schemas from only the segments
-    # The second part of the query gets the leaked temp schemas from just the coordinator
-    # The simpler form of this query that pushed the union into the
-    # inner select does not run correctly on 3.2.x
     # Also note autovacuum lancher and worker will not generate temp namespace
+    # Additionally, we should take into account orphaned schemas, whose session_id may
+    # conflict with coordinator bgworkers, which shouldn't create orphaned schemas.
+    # Also try to avoid connection id collision with truly orphaned schema id.
     leaked_schema_query = """
-        SELECT distinct nspname as schema
-        FROM (
-          SELECT nspname, replace(nspname, 'pg_temp_','')::int as sess_id
-          FROM   gp_dist_random('pg_namespace')
-          WHERE  nspname ~ '^pg_temp_[0-9]+'
-          UNION ALL
-          SELECT nspname, replace(nspname, 'pg_toast_temp_','')::int as sess_id
-          FROM   gp_dist_random('pg_namespace')
-          WHERE  nspname ~ '^pg_toast_temp_[0-9]+'
-        ) n LEFT OUTER JOIN pg_stat_activity x using (sess_id)
-        WHERE x.sess_id is null OR x.backend_type like 'autovacuum%'
-        UNION
-        SELECT nspname as schema
-        FROM (
-          SELECT nspname, replace(nspname, 'pg_temp_','')::int as sess_id
-          FROM   pg_namespace
-          WHERE  nspname ~ '^pg_temp_[0-9]+'
-          UNION ALL
-          SELECT nspname, replace(nspname, 'pg_toast_temp_','')::int as sess_id
-          FROM   pg_namespace
-          WHERE  nspname ~ '^pg_toast_temp_[0-9]+'
-        ) n LEFT OUTER JOIN pg_stat_activity x using (sess_id)
-        WHERE x.sess_id is null OR x.backend_type like 'autovacuum%'
+        WITH temp_schemas AS (
+          SELECT nspname, 
+                 regexp_replace(nspname, '\D', '', 'g')::int as sess_id
+          FROM gp_dist_random('pg_catalog.pg_namespace')
+          WHERE nspname ~ '^pg_t(emp|oast_temp)_\d+'
+          UNION
+          SELECT nspname,
+                regexp_replace(nspname, '\D', '', 'g')::int as sess_id
+          FROM pg_catalog.pg_namespace
+          WHERE nspname ~ '^pg_t(emp|oast_temp)_\d+'
+        )
+        SELECT DISTINCT nspname as schema
+        FROM temp_schemas n
+        LEFT JOIN pg_stat_activity x USING (sess_id)
+        WHERE x.sess_id IS NULL 
+           OR x.backend_type LIKE 'autovacuum%'
+           OR x.datname IS NULL 
+           OR x.datname <> current_database()
+           OR x.pid = pg_backend_pid()
     """
 
     def __get_leaked_schemas(self, db_connection):
--- a/gpMgmt/test/behave/mgmt_utils/gpcheckcat.feature
+++ b/gpMgmt/test/behave/mgmt_utils/gpcheckcat.feature
@@ -14,9 +14,9 @@ Feature: gpcheckcat tests
         And gpcheckcat should not print "Execution error:" to stdout
         And the user runs "dropdb all_good"
 
-    Scenario: gpcheckcat should drop leaked schemas
+    Scenario Outline: gpcheckcat should drop leaked schemas
         Given database "leak_db" is dropped and recreated
-        And the user runs the command "psql leak_db -f 'test/behave/mgmt_utils/steps/data/gpcheckcat/create_temp_schema_leak.sql'" in the background without sleep
+        And the user runs the command "<command>" in the background without sleep
         And waiting "1" seconds
         Then read pid from file "test/behave/mgmt_utils/steps/data/gpcheckcat/pid_leak" and kill the process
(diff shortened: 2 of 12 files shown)


whitehawk:
I've faced flaky test result:
```diff
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /usr/re/share/review/ws137-ADBDEV-8673-2/greengage/src/test/regress/expected/bfv_cte_optimizer.out /usr/re/share/review/ws137-ADBDEV-8673-2/greengage/src/test/regress/results/bfv_cte.out
--- /usr/re/share/review/ws137-ADBDEV-8673-2/greengage/src/test/regress/expected/bfv_cte_optimizer.out  2025-12-05 13:19:00.396569399 +1000
+++ /usr/re/share/review/ws137-ADBDEV-8673-2/greengage/src/test/regress/results/bfv_cte.out     2025-12-05 13:19:00.440569953 +1000
@@ -512,8 +512,8 @@
 -- Filter out irrelevant LOG messages from segments other than seg2.
 \! cat /tmp/bfv_cte.out | grep -P '^(?!LOG)|^(LOG.*seg2)' | grep -vP 'LOG.*fault|decreased xslice state refcount'
 LOG:  SISC READER (shareid=0, slice={2|4}): initializing because squelched
-LOG:  SISC (file=SIRW_X_X_0, slice={2|4}): initialized shared state
 LOG:  SISC READER (shareid=0, slice={2|4}): initializing because squelched
+LOG:  SISC (file=SIRW_X_X_0, slice={2|4}): initialized shared state
 LOG:  SISC WRITER (shareid=0, slice=1): No tuplestore yet, creating tuplestore
 LOG:  SISC WRITER (file=SIRW_X_X_0, slice=1): wrote notify_ready
 LOG:  SISC (file=SIRW_X_X_0, slice=1): file deleted

```

whitehawk:
In the description:
> producer must notify consumers through shared when it's done

'Through shared' memory? It looks like a word is missing...

src/backend/utils/sort/tuplestore.c
```diff
@@ -607,37 +644,40 @@ tuplestore_cleanup(Tuplestorestate *state, bool should_abort)
...
+		if (sstate->num_current == 0) /* safe to delete, no one is using it */
 		{
 			/*
 			 * Either our backend is aborting, or some other backend was
 			 * and set the flag for us. It's this backend's job to delete
 			 * the files now.
 			 */
-			uint32 aborting = pg_atomic_read_u32(&sstate->aborting);
-			if (aborting || num_total_left == 0) /* should delete */
+			if (sstate->aborting || sstate->num_done >= sstate->num_total) /* should delete */
```
whitehawk: just above there is an assert
```
Assert(sstate->num_done <= sstate->num_total);
```
so  situation `sstate->num_done > sstate->num_total` is not possible. Therefore

```suggestion
			if (sstate->aborting || sstate->num_done == sstate->num_total) /* should delete */
```

src/backend/utils/sort/tuplestore.c
```diff
@@ -1836,6 +1876,59 @@ SharedTuplestoreShmemInit(void)
...
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_OUT_OF_MEMORY),
+					errmsg("out of shared tuplestore slots")));
+		}
+		sstate->session_id = gp_session_id;
+		sstate->num_current = 0;
+		sstate->num_done = 0;
+		sstate->aborting = 0;
+		sstate->created = 0;
```
whitehawk:
```suggestion
		sstate->aborting = false;
		sstate->created = false;
```

src/backend/utils/sort/tuplestore.c
silent-observer: Done

src/backend/executor/nodeShareInputScan.c
```diff
@@ -801,290 +678,11 @@ get_shareinput_fileset(void)
...
-	/* If we were the last consumer, wake up the producer. */
-	if (ndone >= nconsumers)
-		ConditionVariableBroadcast(&state->ready_done_cv);
-
-	elog((Debug_shareinput_xslice ? LOG : DEBUG1), "SISC READER (shareid=%d, slice=%d): wrote notify_done",
-		 ref->share_id, currentSliceId);
-}
-
 void
 ShareInputReaderNotifyDone(ShareInputScanState *node)
```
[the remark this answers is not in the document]
silent-observer: ABI compatibility. Are we allowed to add this to the ignore list?

src/backend/utils/sort/tuplestore.c
```diff
@@ -1896,6 +1987,20 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
...
+	PG_TRY();
+	{
+		state->shared_state = get_shared_state(fileset, filename);
+		state->shared_state->num_total = ntotal;
+		state->shared_state->created = true;
+	}
+	PG_CATCH();
+	{
+		BufFileDeleteShared(fileset, filename);
+		PG_RE_THROW();
```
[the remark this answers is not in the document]
silent-observer: The locks are automatically released on abort anyway, but yes, I think it's better to release it explicitly

src/backend/utils/sort/tuplestore.c
```diff
@@ -607,37 +644,40 @@ tuplestore_cleanup(Tuplestorestate *state, bool should_abort)
...
-				BufFileDeleteShared(state->fileset, state->shared_filename);
+				if (sstate->created)
+					BufFileDeleteShared(state->fileset, state->shared_filename);
 				if (hash_search(shared_tuplestores,
 								state->shared_filename,
 								HASH_REMOVE, NULL) == NULL)
 					elog(ERROR, "hash table corrupted");
+				
+				elog((Debug_shareinput_xslice ? LOG : DEBUG1), "SISC (file=%s, slice=%d): file deleted",
+					 state->shared_filename, currentSliceId);
```
KnightMurloc: This code is not only used by Shared scans. Can we give it more general names?

src/backend/utils/sort/tuplestore.c
```diff
@@ -2023,24 +2167,47 @@ tuplestore_open_shared(SharedFileSet *fileset, const char *filename)
...
 	state->share_status = TSHARE_READER;
 	state->frozen = false;
 	state->fileset = fileset;
 	state->shared_filename = pstrdup(filename);
-	state->shared_state = sstate;
+
+	/* Early exit if requested */
+	if (skip_open)
+		return state;
+
```
hilltracer: In comment above, you says:
```
 * Waiting is determined by skip_open parameter. If it is set to true,
 * this call will block until the tuplestore is filled by the writer.
```

src/backend/executor/nodeShareInputScan.c
```diff
@@ -498,14 +400,11 @@ ExecInitShareInputScan(ShareInputScan *node, EState *estate, int eflags)
...
-	 * in current slice, to wake the corresponding CTE producer up for
-	 * cleaning the materialized tuplestore, during squelching.
+	 * To accumulate the number of CTE participant executed in this slice.
+	 * This variable will be used by the last finishing CTE participant
+	 * in current slice, to delete the tuplestore during squelching.
 	 */
-	if (currentSliceId == node->this_slice_id &&
-		currentSliceId != node->producer_slice_id)
-		local_state->nsharers++;
+	local_state->nsharers++;
```
hilltracer: Is it ok that we miss filters for currentSliceId after this patch?

src/backend/utils/sort/tuplestore.c
```diff
@@ -2023,24 +2167,47 @@ tuplestore_open_shared(SharedFileSet *fileset, const char *filename)
...
+	/* Wait until writer is ready */
+	tuplestore_reader_waitready(sstate);
+
+	state->myfile = BufFileOpenShared(fileset, filename);
+	state->readptrs[0].file = 0;
+	state->readptrs[0].offset = 0L;
+	state->status = TSS_READFILE;
 
 	/* Remember this tuplestore in case we have to close it while aborting */
 	local_shared_tuplestores = lappend(local_shared_tuplestores, state);
```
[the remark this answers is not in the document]
silent-observer: I guess theoretically it could lead to issues if an error occurs exactly between opening the tuplestore with skip_open and `tuplestore_end` during squelch. Not sure how that can happen, but yes, I'll fix it.

src/backend/executor/nodeShareInputScan.c
silent-observer: Yes, the meaning of `nsharers` changes after this patch to mean "number of all participants in the local slice", not just "number of all readers in the local slice". This is compensated by the fact that `local_state->ndone++` now happens for both readers and writers, while previously it only happened for readers.

whitehawk (approved):

PR #150, opened 2025-12-05
ADBDEV-8960: Update storage options setting for partitions

ShapatinAA:
Update storage options setting for partitions

Previously storage options declared for partitions via `WITH` keyword were not
handled correctly: storage options from parent dominated over partitions.
This was not a case in 6-th version and was not explained by
any articles in docs.
For example:

```
CREATE TABLE part_table (id int) WITH (appendoptimized=true, 
        compresstype=zstd, compresslevel=1, orientation='column')
PARTITION BY RANGE(id)
(PARTITION p1 start(1) end(10) WITH (compresstype=zstd, compresslevel=7));
```

Will create partition with attributes having `compresslevel = 1`.
While expected behavior (according to 6th version and documentation)
is `compresslevel = 7`.

After this patch this issue is fixed and with clauses are accounted correctly.

The root problem was that the parental encodings, columns specified encodings
and default encodings were being interpreted on early stage and collided in
single attribute encoding field, without accounting encodings, which were
in `WITH` options.

So to fix this issue this patch adds separate function
`merge_inherited_and_default_encodings()` in which storage options from `WITH`
are parsed and collided with parental options and default ones, with next priority:
default options > with options > parental options.
"_Default options_" mean options which are declared on a whole table by using
constructions `DEFAULT ENCODING` or `COLUMN i ENCODING`.

To create "_with options_" function iterates over whole options of partition
table and selects from them only storage options (using
`form_default_storage_directive()`) and then to make them
`ColumnReferenceStorageDirective` it adds to them names of columns, which
can be got from parent table (skipping dropped ones). Final result of this
parsing is options in list `with_cls`. 

Ticket ADBDEV-8960.

### DIFF
--- a/src/backend/executor/execUtils.c
+++ b/src/backend/executor/execUtils.c
@@ -60,7 +60,6 @@
 #include "executor/execUtils.h"
 #include "executor/executor.h"
 #include "executor/execPartition.h"
-#include "executor/nodeShareInputScan.h"
 #include "jit/jit.h"
 #include "mb/pg_wchar.h"
 #include "miscadmin.h"
@@ -213,8 +212,6 @@ CreateExecutorState(void)
 	estate->eliminateAliens = false;
 
 	estate->gp_bypass_unique_check = false;
-
-	estate->sharedScanConsumers = NIL;
 
 	/*
 	 * Return the executor state structure
@@ -1537,22 +1534,6 @@ void mppExecutorFinishup(QueryDesc *queryDesc)
 		!(estate->es_top_eflags & EXEC_FLAG_EXPLAIN_ONLY))
 	{
 		ExecSquelchNode(queryDesc->planstate);
-	}
-
-	/*
-	 * Notify corresponding producer that the reading is completed. We need to
-	 * do notification before processing the results in order to avoid a
-	 * deadlock. Deadlock could occur when the consumer of a cross slice Shared
-	 * Scan was in slice 0 and the producer was in another slice. QE with
-	 * producer couldn't complete until the consumer sent the notification,
-	 * however that didn't happen because consumer at slice 0 send a
-	 * notification after all QEs are completed.
-	 */
-	ListCell *cell;
-	foreach (cell, estate->sharedScanConsumers)
-	{
-		ShareInputScanState *node = lfirst(cell);
-		ShareInputReaderNotifyDone(node);
 	}
 
 	/*
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -35,14 +35,6 @@
  * the whole tuplestore, and advertises that it's ready in shared memory.
  * Consumer slices wait for that before trying to read the store.
  *
- * The producer and the consumers communicate the status of the scan using
- * shared memory. There's a hash table in shared memory, containing a
- * 'shareinput_Xslice_state' struct for each shared scan. The producer uses
- * a condition variable to wake up consumers, when the tuplestore is fully
- * materialized, and the consumers use the same condition variable to inform
- * the producer when they're done reading it. The producer slice keeps the
- * underlying tuplestore open, until all the consumers have finished.
- *
  *
  * Portions Copyright (c) 2007-2008, Greenplum inc
  * Portions Copyright (c) 2012-Present VMware, Inc. or its affiliates.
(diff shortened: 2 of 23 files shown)


src/backend/parser/parse_partition_gp.c
```diff
@@ -1819,6 +1807,49 @@ generatePartitions(Oid parentrelid, GpPartitionDefinition *gpPartSpec,
...
+		 * transformColumnEncoding() will dominate over this WITH options.
+		 */
+		if (elem->options != NIL) 
+		{
+			List	   *tmpenc = NIL;
+			/* 
+			 * We need to analyze only storage options, so if there is none of
+			 * them, then we can skip this step.
+			 */
+			if ((tmpenc = form_default_storage_directive(elem->options)) != NIL)
```
RekGRpth: why not simply
```suggestion
			List	   *tmpenc = form_default_storage_directive(elem->options);
			/* 
			 * We need to analyze only storage options, so if there is none of
			 * them, then we can skip this step.
			 */
			if (tmpenc != NIL)
```

src/backend/parser/parse_partition_gp.c
```diff
@@ -1819,6 +1807,49 @@ generatePartitions(Oid parentrelid, GpPartitionDefinition *gpPartSpec,
...
+			{
+				ColumnReferenceStorageDirective *deflt = NULL;
+				TupleDesc tupdesc = parentrel->rd_att;
+				int natts = tupdesc->natts;
+				for (int i = 0; i < natts; i++)
```
RekGRpth:
```suggestion
				for (int i = 0; i < tupdesc->natts; i++)
```

src/backend/parser/parse_partition_gp.c
ShapatinAA: Updated

src/backend/parser/parse_partition_gp.c
```diff
@@ -1819,6 +1807,49 @@ generatePartitions(Oid parentrelid, GpPartitionDefinition *gpPartSpec,
...
+		{
+			List	   *tmpenc = NIL;
+			/* 
+			 * We need to analyze only storage options, so if there is none of
+			 * them, then we can skip this step.
+			 */
+			if ((tmpenc = form_default_storage_directive(elem->options)) != NIL)
+			{
+				ColumnReferenceStorageDirective *deflt = NULL;
+				TupleDesc tupdesc = parentrel->rd_att;
```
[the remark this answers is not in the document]
ShapatinAA: Updated.

src/backend/parser/parse_partition_gp.c
ShapatinAA: Updated.

src/backend/parser/parse_partition_gp.c
```diff
@@ -1120,6 +1127,61 @@ merge_partition_encoding(ParseState *pstate, List *elem_colencs, List *penc)
...
+			}
+		}
+	}
+	
+	/*
+	 * Merge encodings specified for parent table level and partition
+	 * configuration level. (Each partition element level encoding will be
+	 * merged later to this). 
+	 */
+	out_cls = merge_partition_encoding(pstate, with_cls, parent_tblenc);
```
whitehawk: it looks that `pstate` is used here only to pass it to `merge_partition_encoding()`. And in `merge_partition_encoding()` it is also not used. I suggest to remove it completely from the both function's parameters.

src/backend/parser/parse_partition_gp.c
```diff
@@ -1120,6 +1127,61 @@ merge_partition_encoding(ParseState *pstate, List *elem_colencs, List *penc)
 	return elem_colencs;
 }
 
+static List *
+mergeInheritedAndDefaultEncodings(GpPartDefElem *elem,
```
whitehawk: In the file there is a mix of snake and camel cases. But the functions right above and below the new one are in snake case. So I suggest to use it here as well:
```suggestion
merge_inherited_and_default_encodings(GpPartDefElem *elem,
```

ShapatinAA:
> In the description there is very few information about the way the problem was solved - only one general sentence "After this patch this issue is fixed and with clauses are accounted correctly.". Please provide more details about the solution in the description.

Updated description.

src/test/regress/sql/partition_storage.sql
```diff
@@ -498,3 +498,76 @@ drop table if exists co_can cascade;
...
+SELECT p.relname, a.attname, e.attoptions
+FROM pg_attribute_encoding AS e
+	JOIN pg_class AS p ON e.attrelid=p.oid
+	JOIN pg_attribute AS a ON a.attrelid=e.attrelid AND a.attnum=e.attnum
+WHERE p.relname LIKE 'part_table%'
+ORDER BY e.attrelid, e.attnum;
+
+SELECT reloptions FROM pg_class WHERE relname = 'part_table_1_sss_2_prt_sp1';
+
+--start_ignore
```
whitehawk: I think we can do `drop` without `ignore` section here - tables should always exist.

src/backend/parser/parse_partition_gp.c
ShapatinAA: Updated.

ShapatinAA:
> In the description:
> 
> > without accounting encodings, which was
> > in WITH options.
> 
> if 'which was' relates to 'encodings', we should change to 'which were'.

Updated.

ShapatinAA:
> In the description:
> 
> > So to fix this issue this patch adds separate function mergeInheritedAndDefaultEncodings()
> 
> function name was changed in the code, the description should be aligned as well.

Updated.

src/test/regress/expected/AOCO_Compression.out
```diff
@@ -1834,49 +1834,49 @@ alter table co_wt_sub_partrle_type8192_1_2 add partition new_p values('C')  WITH
...
+ a38    | text                        |           |          |                                                            | extended |              | rle_type         | 1                 | 8192       | 
+ a39    | bit(1)                      |           |          |                                                            | extended |              | rle_type         | 1                 | 8192       | 
+ a40    | bit varying(5)              |           |          |                                                            | extended |              | rle_type         | 1                 | 8192       | 
+ a41    | smallint                    |           |          |                                                            | plain    |              | rle_type         | 1                 | 8192       | 
+ a42    | integer                     |           |          |                                                            | plain    |              | rle_type         | 1                 | 8192       |
```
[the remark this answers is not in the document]
ShapatinAA: `compresstype` was changed, because we added new partition, but previously we defined subpartitions template also, so the storage options in this subpartition are set accordingly to this definition from now on.
```
subpartition template (default subpartition df_sp, start(1)  end(5000) every(1000) 
 WITH (appendonly=true, orientation=column,compresstype=rle_type,compresslevel=1,blocksize=8192))
```

red1452 (approved):

PR #151, opened 2025-12-09
Fix vacuum doesn't compact AO RO table after rolled back transaction

whitehawk:
Fix vacuum doesn't compact AO RO table after rolled back transaction

Problem description:
In case a column was added to a AO row-oriented table in a transaction (and
some values were set to this new column), and this transaction was rolled back,
vacuum didn't compact the table.

Root cause:
In case a new column is added to a AO row-oriented table, the table has to be
rewritten, and a new 'pg_aoseg_...' table is created. Segment data from the new
segment table is also stored to 'AppendOnlyHash' hash table entry.  If we
rollback the transaction, 'pg_aoseg_...' table is rolled back to the old one. 
But the hash entry continued to exist with the values from the transient
pg_aoseg table. So, during vacuum we tried to get the list of segments we had
to compact and called SetSegnoForCompaction(), which requested data from the
hash (via AORelGetOrCreateHashEntry()) and got the hash entry with 0 tupcount.
Therefore, compaction didn't happen.

Fix:
Remove the hash entry from 'AppendOnlyHash' in 'AtEOXact_AppendOnly_Relation()'
if the transaction is aborted. It is done for both row and column oriented
tables, as both are susceptible to the non-valid hash entry.

'AtEOXact_AppendOnly_Relation()' previously contained the code to remove the
hash entry, but it was under condition of GUC
'test_AppendOnlyHash_eviction_vs_just_marking_not_inuse'. This GUC did nothing
more, and was disabled by default. So, as this patch adds proper handling for
the hash entry eviction, it also makes this GUC deprecated and removes its
usage from 'ao-insert-eof' isolation test.

### DIFF
--- a/src/backend/gporca/libnaucrates/src/parser/CParseHandlerDynamicTableScan.cpp
+++ b/src/backend/gporca/libnaucrates/src/parser/CParseHandlerDynamicTableScan.cpp
@@ -162,7 +162,7 @@ CParseHandlerDynamicTableScan::EndElement(const XMLCh *const,  // element_uri,
 			m_mp, table_descr, m_part_index_id, m_part_index_id_printable);
 
 	m_dxl_node = GPOS_NEW(m_mp) CDXLNode(m_mp, dxl_op);
-	// set statictics and physical properties
+	// set statistics and physical properties
 	CParseHandlerUtils::SetProperties(m_dxl_node, prop_parse_handler);
 
 	// add constructed children
--- a/src/backend/gporca/libnaucrates/src/parser/CParseHandlerGatherMotion.cpp
+++ b/src/backend/gporca/libnaucrates/src/parser/CParseHandlerGatherMotion.cpp
@@ -160,7 +160,7 @@ CParseHandlerGatherMotion::EndElement(const XMLCh *const,  // element_uri,
 	GPOS_ASSERT(NULL != child_parse_handler);
 
 	m_dxl_node = GPOS_NEW(m_mp) CDXLNode(m_mp, m_dxl_op);
-	// set statictics and physical properties
+	// set statistics and physical properties
 	CParseHandlerUtils::SetProperties(m_dxl_node, prop_parse_handler);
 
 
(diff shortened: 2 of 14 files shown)


src/backend/access/appendonly/appendonlywriter.c
```diff
@@ -2180,7 +2176,7 @@ AtEOXact_AppendOnly_Relation(AORelHashEntry aoentry, TransactionId currentXid)
  * AtAbort_AppendOnly as well.
  */
 void
-AtEOXact_AppendOnly(void)
+AtEOXact_AppendOnly()
```
RekGRpth: remove this change

src/backend/utils/misc/guc_gp.c
```diff
@@ -1500,17 +1499,6 @@ struct config_bool ConfigureNamesBool_gp[] =
 		NULL, NULL, NULL
 	},
 
-	{
-		{"test_AppendOnlyHash_eviction_vs_just_marking_not_inuse", PGC_SUSET, DEVELOPER_OPTIONS,
```
RekGRpth: Shouldn't we somehow warn users who have this GUC enabled in the configuration file that their cluster won't start after the patch?

whitehawk: The GUC does look like something created for testing purposes. I guess there is low probability that someone uses it in production. Though never zero.
We can highlight it in the release notes. Or simply revert the GUC removal and keep this junk code. What do you think?

src/backend/access/appendonly/appendonlywriter.c
```diff
@@ -2152,16 +2158,8 @@ AtEOXact_AppendOnly_Relation(AORelHashEntry aoentry, TransactionId currentXid)
...
 						  aoentry->txns_using_rel)));
-	}
 
-	if (test_AppendOnlyHash_eviction_vs_just_marking_not_inuse)
-	{
-		/*
-		 * If no transaction is using this entry, it can be removed if
-		 * hash-table gets full. So perform the same here if the above GUC is
-		 * set.
-		 */
```
RekGRpth:
```suggestion
		/*
		 * If no transaction is using this entry, it can be removed if
		 * hash-table gets full.
		 */
```

RekGRpth: Perhaps we could add a relevant comment then?

src/backend/utils/misc/guc_gp.c
whitehawk: made the guc deprecated

PR #154, opened 2025-12-09
ADBDEV-8975: Report resource groups to pg_stat_activity for segment backends

bandetto:
Report resource groups to pg_stat_activity for segment backends

Current state of affairs, from upstream docs:

    When resource groups are enabled. Only query dispatcher (QD) processes will
    have a rsgid and rsgname. Other server processes such as a query executer
    (QE) process or session connection processes will have a rsgid value of 0
    and a rsgname value of unknown. QE processes are managed by the same
    resource group as the dispatching QD process.

As a developer, I encountered a need to see backend's resource group on the
segment directly via utility mode for PARALLEL RETRIEVE CURSOR. I found no
adequate reason why it was done this way.

1. Only the DISPATCHER and EXECUTOR roles change resource groups. UTILITY does
   not interact with resource groups.
2. The patch forces segments (EXECUTOR) to report their resource group status
   to `pg_stat_activity`.
3. Due to point above, processes with the EXECUTOR role on the coordinator
   (`IS_QUERY_DISPATCHER()`) now also receive a resource group.
4. `pg_resgroup_move_query()` sends a request from the coordinator to segment
   processes with the same session ID, which in turn is looked up by the PID in
   shared memory.
5. The resource group is set only for the duration of the request and detaches
   at the end of the request, resetting the value in `pg_stat_activity`. The
   queries need to be suspended/long enough to catch their resource group in
   `pg_stat_activity`.
6. The resource group attaches/detaches from the process with a slight delay.
   This leads to the backend seeming to see its resource group only for itself
   between requests when querying `pg_stat_activity`.

This patch makes it so resource group stats to `pg_stat_activity` on segment
backends are reported, making `rsgid` and `rsgname` columns filled for the
time of slice execution.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -37,6 +37,41 @@ operating systems:
   DockerHub. Runs for push to `main` (retags to `latest`) and tags (uses tag
   like `6.28.2`) after build.
 
+## Release Workflow
+
+A separate workflow, `Greengage release`, handles the uploading of Debian packages
+to GitHub releases. It is triggered when a release is published and uses a
+composite action to manage package deployment.
+
+### Key Features
+
+- **Triggers:** `release: [published]` - Runs when a release is published,
+including re-publishing.
+- **Concurrency:** Uses the same concurrency group as the CI workflow
+(`Greengage CI-${{ github.ref }}`) to ensure proper sequencing and prevent race
+conditions.
+- **Cache-based Artifacts:** Restores built packages from cache using the
+commit SHA as the key, rather than downloading artifacts from previous jobs.
+- **Manual Recovery:** If the cache is missing, the workflow checks the status
+of the last build for the tag and provides clear instructions for manual
+intervention. It does not automatically trigger builds to avoid infinite loops.
+- **Safe Uploads:** Uploads packages with fixed naming patterns and optional
+overwrite (`clobber` flag).
+
+### Behavior
+
+1. **Normal Flow (Cache Available):** Restores packages from cache, renames
+them to the pattern `${PACKAGE_NAME}${VERSION}.${EXT}`, and uploads to the
+release.
+2. **Cache Miss Scenarios:**
+   - **No previous build or previous build successful:** Provides instructions
+   to manually trigger the CI build, then restart the release workflow.
+   - **Previous build failed:** Reports the failure with a link to the failed
+   run and requires manual fixing before retrying.
+
+The release workflow is designed to be robust and provide clear feedback when
+issues occur, ensuring that releases are always consistent and reliable.
+
 ## Configuration
 
 The workflow is parameterized to support flexibility:
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -133,7 +133,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: write # Required for GHCR access
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-package.yml@v9
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-package.yml@v10
     with:
       version: 6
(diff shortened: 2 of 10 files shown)


src/test/isolation2/input/resgroup/resgroup_move_query.source
```diff
@@ -306,12 +306,21 @@ SELECT num_running FROM gp_toolkit.gp_resgroup_status WHERE rsgname='rg_move_que
 --spawn all backends at first short call to guarantee correct pg_resgroup_move_query() execution
 1: SELECT * FROM gp_dist_random('gp_id'), pg_sleep(1) LIMIT 1;
 1&: SELECT * FROM gp_dist_random('gp_id'), pg_sleep(3) LIMIT 1;
+-- segments should display the resource group name
+1U: SELECT rsgname, query FROM pg_stat_activity WHERE rsgname LIKE 'rg_move_query%';
+-- there are two backends on coordinator, the second one is for gather motion
 2: SELECT gp_toolkit.pg_resgroup_move_query(pid, 'rg_move_query') FROM pg_stat_activity WHERE query LIKE '%pg_sleep%' AND rsgname='rg_move_query_mem_small';
+1U: SELECT pg_sleep(1);
```
RekGRpth: Is it really a good idea to use a sleep test? Why not wait for an inject-fault?

src/backend/postmaster/pgstat.c
```diff
@@ -2953,6 +2953,8 @@ pgstat_report_resgroup(TimestampTz queueStart, Oid groupid)
 		beentry->st_waiting = PGBE_WAITING_RESGROUP;
 	}
 
+	elog(DEBUG1, "Changed resource groupId to %d", groupid);
```
RekGRpth: This function seems to be about updating statistics, not about changing the resource group.

src/test/isolation2/input/resgroup/resgroup_move_query.source
bandetto: I tested fault inject locally, and I think it's too complex for this case. The actual delay is negligible and a second should be well enough.

src/backend/utils/resgroup/resgroup.c
```diff
@@ -4930,6 +4930,8 @@ HandleMoveResourceGroup(void)
 		/* Add into cgroup */
 		ResGroupOps_AssignGroup(self->groupId, &(self->caps), MyProcPid);
 	}
+
+	pgstat_report_resgroup(GetCurrentTimestamp(), groupId);
```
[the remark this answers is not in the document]
bandetto: In `SwitchResGroupOnSegment()`, it is not precisely known at which time the role which initiated the query obtained it's resource group. It would make no sense to timestamp it, and it is not timestamped (0 as timestamp means don't touch the previosly assigned timestamp).

Piece of code you're referring to actually changes a resource group to a different one.

src/test/isolation2/output/resgroup/resgroup_bypass.source
```diff
@@ -357,6 +357,31 @@ ERROR:  canceling statement due to user request
...
+----------
+          
+          
+          
+(3 rows)
+
+61q: ... <quitting>
+0Uq: ... <quitting>
+
+
```
silent-observer: The test fails because of this extra newline

src/backend/utils/resgroup/resgroup.c
```diff
@@ -2840,6 +2840,9 @@ SwitchResGroupOnSegment(const char *buf, int len)
 		
 		/* Record the bypass memory limit of current query */
 		self->bypassMemoryLimit = self->memUsage + RESGROUP_BYPASS_MODE_MEMORY_LIMIT_ON_QE;
+
+		pgstat_report_resgroup(0, bypassedSlot.groupId);
+
 		return;
```
RekGRpth: remove redundant return

silent-observer (approved):

PR #156, opened 2025-12-14
Fix vacuum failure to clean up some segment files of AOCS tables

whitehawk:
Fix vacuum failure to clean up some segment files of AOCS tables

Problem description:
Vacuum failed to clean up segment files for AOCS tables, that were created in a
transaction, which was rolled back.

Root cause:
For AOCS tables, a transaction that creates segment files for the very first
time, inserts a corresponding record into 'pg_aoseg.pg_aocsseg_<oid>' in
'InsertInitialAOCSFileSegInfo()'. But if the transaction was rolled back, this
record in 'pg_aoseg.pg_aocsseg_<oid>' was no more available. But the physical
segment files still existed.
Vacuum process in AOCSTruncateToEOF() relies on the information from
'pg_aoseg.pg_aocsseg_<oid>' to get all segments vacuum needs to scan. Obviously,
the segment files from the aborted transaction were not visible to it.

Fix:
Add 'heap_freeze_tuple_wal_logged(segrel, segtup)' into
'InsertInitialAOCSFileSegInfo()', so vacuum now can see the new segment files.
It is already done so in 7x (refer to https://github.com/GreengageDB/greengage/commit/1306d47183375e3acef0bc23735b450187165d05).

But this change interferes with commit https://github.com/GreengageDB/greengage/commit/9e106f51d295fcfb37fc97383286ebeb22426402,
as freezing the tuple partially reverts its logic.
Part of this interference is resolved by preceding cherry-pick of
https://github.com/GreengageDB/greengage/commit/0a2d3cb7cfe35010afc769cdb970df7d5bb0e0a1, which handles the same problem as
https://github.com/GreengageDB/greengage/commit/9e106f51d295fcfb37fc97383286ebeb22426402, but in a different way.
But for 6x additional changes are required in 'UpdateAOCSFileSegInfo()', so this
patch adds usage of 'deformAOCSVPInfo()' into it in the manner intended by
https://github.com/GreengageDB/greengage/commit/0a2d3cb7cfe35010afc769cdb970df7d5bb0e0a1.

Plus this change requires update of the output of the test
'uao_crash_compaction_column', because the output of 'gp_toolkit.__gp_aocsseg()'
now contains records about segment files created by the interrupted vacuum
command.

(cherry picked from commit https://github.com/GreengageDB/greengage/commit/20c446b3393953c35c2d8a0d9dd75a0709ad08f7)

Changes comparing to the original commit:
 - function 'getTableSegFiles()' in the test was updated to SRF as 'EXECUTE ON
ALL SEGMENTS is only supported for set-returning functions'.

Note: part of changes from the original commit (call of
'heap_freeze_tuple_wal_logged(segrel, segtup)' inside
'InsertInitialAOCSFileSegInfo()') is already presented in 7x. The other part
of changes in 'UpdateAOCSFileSegInfo()' doesn't affect the test execution on 7x,
but we still cherry-pick these changes as usage of `deformAOCSVPInfo()` should
be a more proper and safe way of extracting vpinfo.

Also, changes in 'uao_crash_compaction_column' test are already presented in 7x.

-------------
SHOULD BE REBASED

### DIFF
--- a/src/backend/parser/parse_partition_gp.c
+++ b/src/backend/parser/parse_partition_gp.c
@@ -15,6 +15,7 @@
 
 #include "access/table.h"
 #include "access/tableam.h"
+#include "access/reloptions.h"
 #include "catalog/partition.h"
 #include "catalog/pg_collation.h"
 #include "catalog/gp_partition_template.h"
@@ -81,6 +82,11 @@ static List *generateDefaultPartition(ParseState *pstate,
 									  CreateStmtOrigin origin);
 
 static char *extract_tablename_from_options(List **options);
+
+static List *merge_inherited_and_default_encodings(GpPartDefElem *elem,
+											   Relation parentrel,
+											   List *penc_cls,
+											   List *parent_tblenc);
 
 /*
  * qsort_stmt_cmp
@@ -1047,7 +1053,7 @@ split_encoding_clauses(List *encs, List **non_def,
 }
 
 static List *
-merge_partition_encoding(ParseState *pstate, List *elem_colencs, List *penc)
+merge_partition_encoding(List *elem_colencs, List *penc)
 {
 	List	   *elem_nondefs = NIL;
 	List	   *part_nondefs = NIL;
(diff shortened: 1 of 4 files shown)


src/test/regress/sql/vacuum_full_ao.sql
```diff
@@ -22,3 +22,36 @@ select pg_relation_size((select segrelid from pg_appendonly where relid = 'vfao'
 
 vacuum full vfao;
 select pg_relation_size((select segrelid from pg_appendonly where relid = 'vfao'::regclass)) from gp_dist_random('gp_id') where gp_segment_id = 1;
+
+-- Test that vacuum can process segment files created in an aborted transaction
```
RekGRpth: Is it expected that this test passes even without this patch?

PR #157, opened 2025-12-17
ADBDEV-9011: Prevent ORCA from losing distribution requests

GirgerS:
Prevent ORCA from losing distribution requests

In the case when the join condition contains AND clauses, hash join performs
redistribution request by every subexpression in the condition.

The problem occurred when several of these subexpressions had the same
right-hand side, making the matching logic inside 
CPhysicalHashJoin::PdshashedMatching select the same subexpression 
for every request, therefore losing possible redistributions.

This patch adds a parameter to the function that specifies the index 
of the initial distribution request index. Subexpressions matching 
this index are now prioritized in the matching logic.

Also, as a side-effect of the research went into the patch, several 
explanatory comments are added to make the code easier to understand.

Ticket: ADBDEV-9011

### DIFF
--- a/src/test/regress/expected/qp_correlated_query.out
+++ b/src/test/regress/expected/qp_correlated_query.out
@@ -3965,7 +3965,7 @@ DROP TABLE skip_correlated_t4;
 -- For more info, please refer to the comments inside
 -- CPhysicalJoin::PdshashedMatching
 --------------------------------------------------------------------------------
-EXPLAIN (VERBOSE)
+EXPLAIN
 SELECT *
 FROM a
 WHERE NOT EXISTS (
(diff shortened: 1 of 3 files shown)


src/backend/gporca/libgpopt/src/operators/CPhysicalHashJoin.cpp
```diff
@@ -592,8 +595,87 @@ CPhysicalHashJoin::PdshashedMatching(
...
+		//
+		// here, the join condition has two expressions, and all of them have the same
+		// right-hand side. It means that for every snggle-key redistribution request we get the same
+		// pexprDlvrd value (t1.a)
+		// and, when we go through the source expressions in order,
+		// the first expression will always be selected as matching for every redistribution request,
+		// meaning that the second expression has no chance to be matched, and therefore the distribution
+		// by t2.a is not considered
+		//
+		// also, I haven't checked if the same problem might arise when matching a distribution containing
```
ShapatinAA: I think we should avoid first-person narrative here. Using impersonal language keeps the focus on what changed and why.

src/backend/gporca/libgpopt/src/operators/CPhysicalHashJoin.cpp
```diff
@@ -592,8 +595,87 @@ CPhysicalHashJoin::PdshashedMatching(
...
+		//                           Sort Key: t2.a
+		//                           ->  Seq Scan on public.t2  (cost=0.00..431.00 rows=1 width=8)
+		//                                 Output: t2.a
+		//         ->  Hash  (cost=431.00..431.00 rows=1 width=8)
+		//               Output: t1.a
+		//               ->  Seq Scan on public.t1  (cost=0.00..431.00 rows=1 width=8)
+		//                     Output: t1.a
+		//
+		// here, the join condition has two expressions, and all of them have the same
+		// right-hand side. It means that for every snggle-key redistribution request we get the same
```
ShapatinAA:
```suggestion
		// right-hand side. It means that for every single-key redistribution request we get the same
```

src/backend/gporca/libgpopt/include/gpopt/base/CReqdPropPlan.h
```diff
@@ -16,6 +16,8 @@
 #include "gpopt/base/CColRef.h"
 #include "gpopt/base/CReqdProp.h"
 
+#define GPOPT_INVALID_OPT_REQUEST gpos::ulong_max
```
ShapatinAA: Should we use this macro instead `GPOPT_INVALID_OPTCTXT_ID`, if it fits here?

GirgerS: Optimization context is a different thing, it is used to store a request inside a group. And they can be created by completely different jobs generated from different expression from different groups. While optimization request indices are local to a single optimization job.

src/backend/gporca/libgpopt/src/operators/CPhysicalHashJoin.cpp
GirgerS: Thanks. Should be fixed in the last commit.

src/backend/gporca/libgpopt/src/operators/CPhysicalHashJoin.cpp
GirgerS: Agreed. Please check the last commit.

ShapatinAA (approved):

PR #158, opened 2025-12-17
ADBDEV-9015: Fix gpmon const warning

VoidZeroNull0:
What happened?
Usage of `quote_identifier()` caused a warning about discarding `const`
qualifier when assigning return of the function to the `escaped_dbname`.

How do we fix this?
`escaped_dbname` should be made `const char *` and casted to `void *`
inside `pfree()` to avoid another warning.

### DIFF
--- a/src/backend/utils/gpmon/gpmon.c
+++ b/src/backend/utils/gpmon/gpmon.c
@@ -223,7 +223,7 @@ void gpmon_qlog_packet_init(gpmon_packet_t *gpmonPacket)
 {
 	const char *username = NULL;
 	char *dbname = NULL;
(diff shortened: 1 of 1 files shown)


bandetto:
> escaped_dbname could be made const char*, but it is changed during its lifetime

Are you sure you're not confusing `const char *` and `char *const`?

VoidZeroNull0:
> Are you sure you're not confusing `const char *` and `char *const`?

Yup.
https://github.com/GreengageDB/greengage/blob/390d0636ff5485adea3f28b8b55aba45ca229ed3/src/backend/utils/adt/ruleutils.c#L9808-L9816

PR #165, opened 2025-12-29
ADBDEV-9065: Prohibit skip-level correlated queries in Postgres planner for 7.x

VoidZeroNull0:
Prohibit skip-level correlated queries in Postgres planner

What occurs?
Postgres planner creates an invalid plan with skip-level correlation involving
master-only and replicated tables. The plan fails to be executed during
execution. The error occurs when on bottom level (nested subplan) replicated or
master-only table is used and on upper - distributed or replicated.

Why it occurs?
Skip-level correlation (at least for distributed tables) seems to be prohibited
in Postgres planner due to rescanning issues that arise - they appear, because
the lowest level plan is made an initplan without correct motion node applied.
ORCA planner produces valid plans for such queries, but not with master-only
tables - it seems to lack a feature to work with them whatsoever.

How do we fix this?
Let's prohibit replicated tables or master-distributed (distributed-master)
skip-level correlation for Postgres Legacy planner as it
leads to rescan issues because of absent motions.

Alternative fixes?
Changing some checks for initplans creation and enabling skip-level correlation
make Postgres produces correct plan - initplan is changed to subplan and
correct motions are applied. Yet, one test from rangefuncs fails. So, some
research is needed.

Tests?
Newly added tests check that ORCA is able to plan and execute simple skip-level
correlated plans, whereas Postgres planner outputs an error about lack of
skip-level correlation support.

Patch was cherry-picked from e3bae2b669a2b819ec0391ff2d4dc15f8a6536a0 with
additions to tests and reworked skip-level logic.

### DIFF
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -41,7 +41,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: read  # Explicit for GHCR access clarity
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-behave.yml@v14
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-behave.yml@v15
     with:
       version: 7
       target_os: ${{ matrix.target_os }}
--- a/gpMgmt/test/behave/mgmt_utils/cross_subnet.feature
+++ b/gpMgmt/test/behave/mgmt_utils/cross_subnet.feature
@@ -1,4 +1,4 @@
-@cross_subnet
+@cross_subnet @skip
 Feature: Tests for a cross_subnet cluster
 
     Scenario: gpinitsystem works across subnets
--- a/gpMgmt/test/behave/mgmt_utils/gpexpand.feature
+++ b/gpMgmt/test/behave/mgmt_utils/gpexpand.feature
@@ -1,4 +1,4 @@
-@gpexpand
+@gpexpand @skip
 Feature: expand the cluster by adding more segments
 
     @gpexpand_no_mirrors
--- a/gpMgmt/test/behave/mgmt_utils/gpssh.feature
+++ b/gpMgmt/test/behave/mgmt_utils/gpssh.feature
@@ -1,4 +1,4 @@
-@gpssh
+@gpssh @skip
 Feature: gpssh behave tests
 
     Scenario: gpssh -d and -t options


src/test/regress/expected/subselect_gp_optimizer.out
```diff
@@ -3778,17 +3778,15 @@ drop table table_left;
 drop table table_right;
 -- test cross params of initplan
 -- https://github.com/GreengageDB/greengage/issues/16268
+-- ADBDEV-9065, https://github.com/GreengageDB/greengage/pull/165
```
red1452: Link to PR is enough. Remove number of ticket because it useless for community.

red1452 (commented):
Please, use "git cherry-pick -x" to cherry-pick comiit from another branch and save description. Add differences from original patch to the end of description.

src/backend/optimizer/plan/subselect.c
```diff
@@ -158,7 +158,8 @@ get_first_col_type(Plan *plan, Oid *coltype, int32 *coltypmod,
 }
 
 /**
- * Returns true if query refers to a distributed table.
+ * Returns true if query refers to any table on this level or on some other
+ * subquery level, if recursive is true.
  */
 bool QueryHasDistributedRelation(Query *q, bool recursive)
```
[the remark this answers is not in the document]
VoidZeroNull0: True, we no longer fetch the policy, so it is no longer related. Better name for function could be chosen, but it was decided in original [PR](https://github.com/GreengageDB/greengage/pull/141) to keep it as it is for ABI checks.

dkovalev1:
Before the patch the following scenario worked:
```
postgres=# create external web table wt(a varchar) execute 'ls -1' on master format 'text';
CREATE EXTERNAL TABLE
postgres=# create external web table wtb(b varchar) execute 'ls -1' on master format 'text';
CREATE EXTERNAL TABLE
postgres=# explain select (select (select b from wtb where b = a)) from wt;
                                    QUERY PLAN                                     
-----------------------------------------------------------------------------------
 Foreign Scan on wt  (cost=0.00..451.90 rows=1000000 width=8)
   SubPlan 1
     ->  Result  (cost=0.00..65800474.78 rows=1 width=8)
           Filter: ((wtb.b)::text = (wt.a)::text)
           ->  Materialize  (cost=0.00..474.78 rows=1000000 width=8)
                 ->  Foreign Scan on wtb  (cost=0.00..451.90 rows=1000000 width=8)
 Optimizer: GPORCA
(7 rows)
```
Now it does not:
```
postgres=# create external web table wt(a varchar) execute 'ls -1' on master format 'text';
CREATE EXTERNAL TABLE
postgres=# create external web table wtb(b varchar) execute 'ls -1' on master format 'text';
CREATE EXTERNAL TABLE
postgres=# explain select (select (select b from wtb where b = a)) from wt;
ERROR:  correlated subquery with skip-level correlations is not supported
```
Is it expected?

VoidZeroNull0:
> Are the tests with master only table needed here?

Originally, problem arose from skip-level correlated query with master-only table. Also it seems logical, that if patch prohibits skip-level queries using any tables, it should include tests for all types of these tables.

VoidZeroNull0:
> Pardon, double checked with Postgres optimizer:

Thanks! When preparing initial research I concluded that skip-level correlation with 2 master-only tables lead to exact problem - turns out no. New test and logic to cover skip-level correlation using only master-only tables was added.

VoidZeroNull0:
> I see too much differences between original and your patch. I think, this sentence should be updated.

Done.

>> Patch was cherry-picked from e3bae2b669a2b819ec0391ff2d4dc15f8a6536a0 with additions to tests and reworked skip-level logic.

src/test/regress/expected/subselect_gp.out
```diff
@@ -3752,17 +3752,15 @@ drop table table_left;
 drop table table_right;
 -- test cross params of initplan
 -- https://github.com/GreengageDB/greengage/issues/16268
```
red1452: This link does not work. Maybe, remove it?

red1452 (approved):

dkovalev1 (approved):

PR #173, opened 2026-01-13
ADBDEV-8988: Fix orphaned segment files for added and rolled back columns for AOCS tables

silent-observer:
Fix orphaned segment files for added and rolled back columns for AOCS tables

Problem description:
In case we added a new column to a AOCS table in a transaction that was later
rolled back, a non-empty segment file could be left, and vacuum didn't truncate
it.

Root cause:
Vacuum does truncation of AOCS segment files in 2 places:
1. In AOCSSegmentFileTruncateToEOF(). It iterated over 'vpinfo.nEntry', which
doesn't count the added and rolled back attribute, so its segment file was not
processed here.
2. In AOCSCompaction_DropSegmentFile() function (if compaction is required for
the segment). It iterates over attributes retrieved by
RelationGetNumberOfAttributes(). But, the added and rolled back attribute is not
counted by RelationGetNumberOfAttributes(), so its segment file was not
processed here as well.

Fix:
In AOCSSegmentFileTruncateToEOF() iterate and truncate all segment files for
a given 'segno' until they exist. We stop once we didn't find a file and column
number is above 'vpinfo.nEntry'. We assume that the segment files exist or stop
existing for all columns thereafter - similar logic we already use in
ao_foreach_extent_file().

We do it not in AOCSCompaction_DropSegmentFile(), as it will not be called if
compaction is not required, while the orphaned segment files may still exist in
this case.

Changes from original commit:
We cannot use 'attnum' to iterate through segment files, since in GPDB 7 the
filenumber is stored explicitly in 'pg_attribute_encoding' system table.
Changes to this table are also rolled back, so we won't be able to find an
entry for the 'attnum' in there. Instead, iterate through possible filenums
and find the corresponding 'attnum'. If the file exists but has no entry in
'pg_attribute_encoding', it's safe to drop.
Notably, we can only stop when both filenum and filenum+1600 are missing,
since table rewrites can replace filenum with filenum+1600. This means we
must check filenums in order 1, 1601, 2, 1602, 3, 1603 etc. to ensure we
don't miss any.
Since filenums are allocated in order, for N-column table it's enough to
check filenums up to N and N+1600, and then more only if they exist.
(cherry picked from commit b5f63c6571e4300187a1d88e266520d0187fdd82)

---
Note: Do not squash to preserve authorship

### DIFF
--- a/src/backend/access/aocs/aocs_compaction.c
+++ b/src/backend/access/aocs/aocs_compaction.c
@@ -108,67 +108,91 @@ void
 AOCSSegmentFileTruncateToEOF(Relation aorel, int segno, AOCSVPInfo *vpinfo, AOVacuumRelStats *vacrelstats)
 {
 	const char *relname = RelationGetRelationName(aorel);
-	int			j;
 
 	Assert(RelationStorageIsAoCols(aorel));
 
-	for (j = 0; j < vpinfo->nEntry; ++j)
+	/*
+	 * We try to truncate all segment files beyond
+	 * `vpinfo.nEntry`, as we may have non-empty segment files
+	 * left by ADD COLUMN, which was rolled back. It is similar to logic in
+	 * ao_foreach_extent_file().
+	 */
+	for (int i = 1;; ++i)
 	{
-		int64		segeof;
-		char		filenamepath[MAXPGPATH];
-		AOCSVPInfoEntry *entry;
-		File		fd;
-		int32		fileSegNo;
-		/* Filenum for the column */
-		FileNumber  filenum = GetFilenumForAttribute(RelationGetRelid(aorel), j + 1);
+		bool		found = false;
 
-		entry = &vpinfo->entry[j];
-		segeof = entry->eof;
-
-		/* Open and truncate the relation segfile to its eof */
-		MakeAOSegmentFileName(aorel, segno, filenum, &fileSegNo, filenamepath);
-
-		elogif(Debug_appendonly_print_compaction, LOG,
-			   "Opening AO COL relation \"%s.%s\", relation id %u, relfilenode %u column #%d, logical segment #%d (physical segment file #%d, logical EOF " INT64_FORMAT ")",
-			   get_namespace_name(RelationGetNamespace(aorel)),
-			   relname,
-			   aorel->rd_id,
-			   aorel->rd_node.relNode,
-			   j,
-			   segno,
-			   fileSegNo,
-			   segeof);
-
-		fd = OpenAOSegmentFile(filenamepath, segeof);
-		if (fd >= 0)
+		for (int j = 0; j < 2; ++j)
 		{
-			TruncateAOSegmentFile(fd, aorel, fileSegNo, segeof, vacrelstats);
-			CloseAOSegmentFile(fd);
+			int64		segeof = 0;
+			char		filenamepath[MAXPGPATH];
+			File		fd;
+			int32		fileSegNo;
+			/*
+			 * Filenum for the column.
+			 * We cannot directly use GetFilenumForAttribute() here since it
+			 * accesses pg_attribute_encoding table, which might not have
+			 * entries for some of the files, for example if a column was added
+			 * and then rolled back. Instead we have to iterate through
+			 * filenums directly. Note the nested loop to ensure we iterate in
+			 * order 1, 1601, 2, 1602, 3, 1603 etc.
+			 */
+			FileNumber	filenum = i + j * MaxHeapAttributeNumber;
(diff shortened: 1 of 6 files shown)


src/backend/catalog/pg_attribute_encoding.c
```diff
@@ -763,6 +763,47 @@ GetFilenumForAttribute(Oid relid, AttrNumber attnum)
...
+ */
+AttrNumber
+GetAttnumForFilenum(Oid relid, FileNumber filenum)
+{
+	Relation    rel;
+	SysScanDesc scan;
+	ScanKeyData skey[2];
+	HeapTuple	tup;
+	AttrNumber  attnum = InvalidAttrNumber;
+	bool        isnull;
```
RekGRpth: Reduce the visibility of this variable (put it in the conditional block below where it is only used).

RekGRpth: pgindent your changes

src/backend/access/aocs/aocs_compaction.c
```diff
@@ -108,67 +108,84 @@ void
...
-			   aorel->rd_node.relNode,
-			   j,
-			   segno,
-			   fileSegNo,
-			   segeof);
-
-		fd = OpenAOSegmentFile(filenamepath, segeof);
-		if (fd >= 0)
+		bool found = false;
+		for (j = 0; j < 2; ++j)
```
RekGRpth: why not
```suggestion
	for (int i = 1; ; i++)
	{
		bool found = false;
		for (int j = 0; j < 2; j++)
```

src/backend/access/aocs/aocs_compaction.c
```diff
@@ -108,67 +108,84 @@ void
...
 		{
-			TruncateAOSegmentFile(fd, aorel, fileSegNo, segeof, vacrelstats);
-			CloseAOSegmentFile(fd);
+			int64		segeof = 0;
+			char		filenamepath[MAXPGPATH];
+			File		fd;
+			int32		fileSegNo;
+			/* Filenum for the column */
+			FileNumber  filenum = i + j * MaxHeapAttributeNumber;
+			AttrNumber  attnum = GetAttnumForFilenum(RelationGetRelid(aorel), filenum);
```
RekGRpth: pgindent your changes

src/backend/catalog/pg_attribute_encoding.c
silent-observer: Done

silent-observer: Fixed

src/backend/access/appendonly/aomd.c
```diff
@@ -667,10 +668,19 @@ ao_rel_get_physical_size(Relation aorel)
 							pg_aoseg_dsc, &isNull);
 			vpinfo = (AOCSVPInfo *) PG_DETOAST_DATUM(d);
 
-			for (col = 0; col < vpinfo->nEntry; ++col)
+			for (i = 1; ; ++i)
 			{
-				FileNumber filenum = GetFilenumForAttribute(RelationGetRelid(aorel), col + 1);
-				total_physical_size += ao_segfile_get_physical_size(aorel, segno, filenum);
+				bool found = false;
+				for (j = 0; j < 2; j++)
```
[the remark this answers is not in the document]
silent-observer: Fixed

RekGRpth (approved):

src/backend/access/aocs/aocs_compaction.c
```diff
@@ -108,67 +108,83 @@ void
...
+		for (int j = 0; j < 2; ++j)
 		{
-			TruncateAOSegmentFile(fd, aorel, fileSegNo, segeof, vacrelstats);
-			CloseAOSegmentFile(fd);
+			int64		segeof = 0;
+			char		filenamepath[MAXPGPATH];
+			File		fd;
+			int32		fileSegNo;
+			/* Filenum for the column */
+			FileNumber	filenum = i + j * MaxHeapAttributeNumber;
```
whitehawk: Is it possible to add a test case to cover the situation when we have `filenum` >` MaxHeapAttributeNumber`?

whitehawk: I suggest adding here a comment why we can't use `GetFilenumForAttribute()` directly.

silent-observer: Yes, modified your test to contain filenums above 1600 (both in the original columns and in the newly added ones)

whitehawk (approved):

RekGRpth (approved):

PR #176, opened 2026-01-16
[GG-89] Fix gpexpand behave tests

KnightMurloc:
The previous commit (c0bc5c2e78b912bccfd709f6683cb025dd2b5d65) added a few
regressions. The regression was related to replacing the comparison condition of
the comparison with 't' with a truth check. This change is due to the fact that
in PyGreSQL 5, unlike the 4th version, it converts the bool values. But it was
not taken into account that such values can be set in Python code.
The error of calling verify in TestDML has also been fixed. The verify method
was called without passing a connection, and although the verify implementation
in the class itself does not require a connection, this function may be
overloaded in a child class.

### DIFF
--- a/gpMgmt/bin/gppylib/gparray.py
+++ b/gpMgmt/bin/gppylib/gparray.py
@@ -524,7 +524,7 @@ def createSegmentRows( hostlist
     content = 0
 
     for host in hostlist:
-        isprimary='t'
+        isprimary=True
         port=primary_portbase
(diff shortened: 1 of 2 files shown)


red1452 (commented):
LGTM, just add link to commit, which broke tests.

red1452 (approved):

silent-observer (approved):

PR #200, opened 2026-01-27
ADBDEV-9055 Optimize Greengage 7 development image size

uaqq:
### Optimize Greengage 7 development image size (#200)

- Move `WORKDIR` to the top to fix hadolint warning about relative COPY
destination.
- Sort multi-line arguments in apt-get and pip for easier maintenance. 
- Clean apt cache metadata after `apt-get update`.
- Disable pip cache during package installation.
- Move the list of apt packages to `README.Ubuntu.bash`
- Sort packages in `README.Ubuntu.bash` alphabetically.
- Use hardlinks in `compile_gpdb.bash` to eliminate duplicate binary
archives.
- Remove openjdk-17-jdk package, since it is already installed during
PXF build.
- Combine ENV into a single instruction.

Image size reduction: from 940 MB to 580 MB.

Before: `ggdb7_ubuntu_510157a53649339b0ec3da8ac1289eaf1014755b 940 MB`
After: `ggdb7_ubuntu_5958b17abf0be403ee90aed09b635f01c5feae6e 580 MB`

Task: ADBDEV-9055

### DIFF
--- a/src/backend/postmaster/pgstat.c
+++ b/src/backend/postmaster/pgstat.c
@@ -3530,6 +3530,8 @@ pgstat_report_resgroup(Oid groupid)
 	beentry->st_rsgid = groupid;
 	beentry->st_changecount++;
 	Assert((beentry->st_changecount & 1) == 0);
+
+	elog(DEBUG1, "Reported resource group ID of %d", groupid);
 }
 
 /* ----------
--- a/src/backend/utils/resgroup/resgroup.c
+++ b/src/backend/utils/resgroup/resgroup.c
@@ -1709,6 +1709,8 @@ SwitchResGroupOnSegment(const char *buf, int len)
 		cgroupOpsRoutine->attachcgroup(bypassedGroup->groupId, MyProcPid,
 									   caps.cpuMaxPercent == CPU_MAX_PERCENT_DISABLED);
 
+		pgstat_report_resgroup(bypassedSlot.groupId);
+
 		return;
 	}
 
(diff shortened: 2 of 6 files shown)


dkovalev1:
I am okay with the patch, but got some questions.

> Before: ggdb7_ubuntu_510157a53649339b0ec3da8ac1289eaf1014755b 940 MB
> After: ggdb7_ubuntu_5958b17abf0be403ee90aed09b635f01c5feae6e 580 MB

I've run image build according to the readme.md on my local machine and got before the patch:
```
$ docker image ls                                              
                                                                            i Info →   U  In Use
IMAGE                                  ID             DISK USAGE   CONTENT SIZE   EXTRA
gpdb7_u22:latest                       2a559105f812       2.68GB             0B 
```
and after the patch:
```
$ docker image ls
                                                                            i Info →   U  In Use
IMAGE                                  ID             DISK USAGE   CONTENT SIZE   EXTRA
gpdb7_u22:latest                       b82767f5e112       1.96GB             0B        
```
The patch is definitely useful and reduces size, but why the images could be so bigger on my local host?

magf:
> When using the image and running tests within a container there is a step in scripts performing `chown -R gpadmin:gpadmin /home/gpadmin`. It takes significant time (minutes) and probably affects layers size. Is it possible to eliminate this step and move it to the image.
> It seemd there is `COPY --chown=<XXX>` in the dockerfile available, of course, the `gpadmin` must be created before.

Thank You, that's right but:
1. this change is beyond the scope of this patch;
2. this change cannot be implemented selectively because the Concourse scripts are an integrated system and require an epic task for review and refactoring.

dkovalev1 (approved):

PR #205, opened 2026-01-28
GG-143 Add skip tags for features, bump CI (6.x)

uaqq:
### Add skip tags for features, bump CI (6.x) (#205)

The grep-based filter used to exclude certain features from the CI matrix has been removed from the CI workflow.

Instead of filtering features in CI, the `@skip` tag is now used directly in feature files for tests that should be excluded from CI.

The behave tests workflow version was bumped to use workflow without the grep-based filter.

Tests: https://github.com/uaqq/greengage/actions/runs/21430675390

Task: [GG-125](https://tracker.yandex.ru/GG-125)

### DIFF
--- a/gpMgmt/bin/gppylib/commands/unix.py
+++ b/gpMgmt/bin/gppylib/commands/unix.py
@@ -580,7 +580,7 @@ class Rsync(Command):
         # of each line and redirects it to progress_file
         if progress_file:
             cmd_tokens.append(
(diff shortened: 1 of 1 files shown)


RekGRpth:
> in the upstream repository

The current repository is the upstream, so it seems strange to link to some other upstream from it.

silent-observer (approved):

RekGRpth (approved):

PR #208, opened 2026-01-29
GG-130: Fix infinite recursion in ProcessUtility hook chain

silent-observer:
Fix infinite recursion in ProcessUtility hook chain

Loading gg_tables_tracking together with gpadcc caused stack overflow
due to infinite recursion in adcc_ProcessUtility_hook() when executing
simple queries.

Both gpadcc and gg_tables_tracking declared next_ProcessUtility_hook as a global
variable with the same name. When extensions are loaded in order 'gpadcc,
gg_tables_tracking', the toolkit's initialization overwrites the global
next_ProcessUtility_hook symbol, making it point back to
adcc_ProcessUtility_hook() instead of the standart hook. This creates a circular
reference. If they are loaded in different order, the next next_ProcessUtility_hook
points to toolkit's one leading to the same issue.

This patch solves the issue by declaring next_ProcessUtility_hook as static in
toolkit extension.

(cherry picked from commit arenadata/gpdb@eb827128c232b7cf8dba5c40e502617382d1c39b)

---

### DIFF
--- a/src/backend/optimizer/plan/subselect.c
+++ b/src/backend/optimizer/plan/subselect.c
@@ -158,7 +158,7 @@ get_first_col_type(Plan *plan, Oid *coltype, int32 *coltypmod,
 }
 
 /**
- * Returns true if query refers to a distributed table.
+ * Returns true if query refers to a distributed or replicated table.
  */
 bool QueryHasDistributedRelation(Query *q, bool recursive)
 {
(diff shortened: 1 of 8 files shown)


ShapatinAA:
Don't we need header in description?

ShapatinAA (dismissed):

ShapatinAA:
Test `resource_queue_terminate` failed, but I don't see diff.

silent-observer:
> ??? There is no arenadata_toolkit in 7X!!!

This is a port from 6X, the corresponding code was moved verbatim to gg_tables_tracking extension in 7X. I will update commit message.

bimboterminator1 (approved):

PR #211, opened 2026-01-30
GG-121: Fix gpssh-exkeys behave tests

RekGRpth:
Fix gpssh-exkeys behave tests

The gpssh_exkeys behave tests passed. However, they did't check anything
because all scenarios were skipped. This is due to a mismatch between the file
name and the tag name. In other words, we're running tests with a tag based on
the file name, but there's no such tag.

1) Rename the gpssh_exkeys.feature file to gpssh-exkeys.feature to ensure an
exact match between the file name and the tag name.

2) In the Docker container, the /dev/stdout and /dev/stderr files are not
directly writable by an unprivileged user. Capture stdout/stderr via run_cmd.

3) Some tests require each host to have its own files in the /home/gpadmin/.ssh
directory because the tests delete them, but not on all hosts. Mount the
ssh_keys directory as a whole, not individual files, and mount it to
/home/gpadmin/.ssh.src and before running the tests, copy .ssh.src to .ssh so
that each host has its own copy of the (initially identical) files.

4) Add -x flag when calling the set command to make it clear in the logs which
commands were run.

Ticket: GG-121

### DIFF
--- a/src/backend/gporca/data/dxl/minidump/3WayJoinUsingOperatorsOfNonDefaultOpfamily.mdp
+++ b/src/backend/gporca/data/dxl/minidump/3WayJoinUsingOperatorsOfNonDefaultOpfamily.mdp
@@ -396,7 +396,7 @@
         </dxl:And>
       </dxl:LogicalJoin>
     </dxl:Query>
-    <dxl:Plan Id="0" SpaceSize="18">
+    <dxl:Plan Id="0" SpaceSize="24">
       <dxl:GatherMotion InputSegments="0,1,2" OutputSegments="-1">
         <dxl:Properties>
           <dxl:Cost StartupCost="0" TotalCost="1324463.445795" Rows="3.413333" Width="8"/>
--- a/src/backend/gporca/data/dxl/minidump/CTE15HAReplicated.mdp
+++ b/src/backend/gporca/data/dxl/minidump/CTE15HAReplicated.mdp
@@ -416,7 +416,7 @@
         </dxl:LogicalCTEAnchor>
       </dxl:LogicalCTEAnchor>
     </dxl:Query>
-    <dxl:Plan Id="0" SpaceSize="57585">
+    <dxl:Plan Id="0" SpaceSize="61743">
       <dxl:GatherMotion InputSegments="0,1,2" OutputSegments="-1">
         <dxl:Properties>
           <dxl:Cost StartupCost="0" TotalCost="2156.065806" Rows="400.000000" Width="267"/>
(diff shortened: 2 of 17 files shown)


RekGRpth:
> > Rename the gpssh_exkeys.feature file to gpssh-exkeys.feature
> 
> have you forgot to add gpssh-exkeys.feature?

indeed, thanx!

ci/scripts/init_containers.sh
```diff
@@ -13,28 +13,20 @@ else
...
-
-# Add ip and host to /etc/hosts
-for service in $services
-do
-  docker compose -p $project -f ci/docker-compose.yaml exec -T \
-    $service bash -c "for HOST in \"cdw sdw1 sdw2 sdw3 sdw4 sdw5 sdw6\"; do echo \"\$(host \"\$HOST\" | grep 'has address' | head -n 1 | cut -d ' ' -f 4) \$HOST\" >>/etc/hosts; done" &
+  docker compose -p "$project" -f ci/docker-compose.yaml exec -T "$service" bash -exc "
+    # Copy /home/gpadmin/.ssh
+    cp -rf /home/gpadmin/.ssh.src /home/gpadmin/.ssh;
+    # Prepare ALL containers first
```
dkovalev1: Perhaps this comment related to the whole for loop and may be moved upper?

ci/scripts/init_containers.sh
```diff
@@ -13,28 +13,20 @@ else
...
-done
-wait
-
-# Add ip and host to /etc/hosts
-for service in $services
-do
-  docker compose -p $project -f ci/docker-compose.yaml exec -T \
-    $service bash -c "for HOST in \"cdw sdw1 sdw2 sdw3 sdw4 sdw5 sdw6\"; do echo \"\$(host \"\$HOST\" | grep 'has address' | head -n 1 | cut -d ' ' -f 4) \$HOST\" >>/etc/hosts; done" &
+  docker compose -p "$project" -f ci/docker-compose.yaml exec -T "$service" bash -exc "
+    # Copy /home/gpadmin/.ssh
```
dkovalev1: Perhaps this comment is redundant and just repeats the line below. May be something like "make a fresh unmodified copy of  /home/gpadmin/.ssh?"

dkovalev1 (dismissed):

hilltracer (dismissed):
LGTM.
Please move `Co-authored-by: ...` line in the end of patch description. In this case github will  display it properly.

hilltracer (approved):

PR #217, opened 2026-02-02
ADBEDV-9169: Backport prohibit skip-level correlated queries in Postgres planner for 6.x

VoidZeroNull0:
Backport prohibit skip-level correlated queries in Postgres

What occurs?
Postgres planner puts an error in case of master-only to master-only skip-level
correlation, but this was executed correctly before.

Why it occurs?
QueryHasDistributedRelation() bans any table, without regard of its
distribution.

How do we fix this?
Make a more soft condition: prohibit only replicated and distributed tables
inside QueryHasDistributedRelation(), but additionally check master-only with
replicated/distributed scenarios using QueryHasMasterOnlyRelation() inside
check_multi_subquery_correlated().

Tests?
Test to check the planning of master-only to master-only skip-level correlation
without error was added.

---------

Co-authored-by: Denis Kovalev <d.kovalev@arenadata.io>
(cherry picked from commit 623167954a2f25109c8420b4a7f659c5596fe3a8)

### DIFF
--- a/src/backend/optimizer/plan/subselect.c
+++ b/src/backend/optimizer/plan/subselect.c
@@ -420,8 +420,7 @@ get_first_col_type(Plan *plan, Oid *coltype, int32 *coltypmod,
 }
 
 /**
- * Returns true if query refers to any table on this level or on some other
- * subquery level, if recursive is true.
+ * Returns true if query refers to a distributed or replicated table.
  */
 bool QueryHasDistributedRelation(Query *q, bool recursive)
 {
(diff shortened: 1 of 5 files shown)


VoidZeroNull0:
> Since this patch contains only updates from review of 7.x port, maybe the description should only describe them? We already have a commit in 6.x with basically the same description

I agree.

VoidZeroNull0:
> LGTM, but what about other tests mentioned in the ticket, like master-distributed case?

They were in previous PRs, so I decided not to write about them in description.

dkovalev1 (approved):

PR #218, opened 2026-02-02
GG-160: Report resource groups to pg_stat_activity for segment backends

silent-observer:
Report resource groups to pg_stat_activity for segment backends (#218)

Current state of affairs, from upstream docs:

    When resource groups are enabled. Only query dispatcher (QD) processes will
    have a rsgid and rsgname. Other server processes such as a query executer
    (QE) process or session connection processes will have a rsgid value of 0
    and a rsgname value of unknown. QE processes are managed by the same
    resource group as the dispatching QD process.

As a developer, I encountered a need to see backend's resource group on the segment directly via utility mode for PARALLEL RETRIEVE CURSOR. I found no adequate reason why it was done this way.

1. Only the DISPATCHER and EXECUTOR roles change resource groups. UTILITY does not interact with resource groups.
2. The patch forces segments (EXECUTOR) to report their resource group status to `pg_stat_activity`.
3. Due to point above, processes with the EXECUTOR role on the coordinator (`IS_QUERY_DISPATCHER()`) now also receive a resource group.
4. `pg_resgroup_move_query()` sends a request from the coordinator to segment processes with the same session ID, which in turn is looked up by the PID in shared memory.
5. The resource group is set only for the duration of the request and detaches at the end of the request, resetting the value in `pg_stat_activity`. The queries need to be suspended/long enough to catch their resource group in `pg_stat_activity`.
6. The resource group attaches/detaches from the process with a slight delay. This leads to the backend seeming to see its resource group only for itself between requests when querying `pg_stat_activity`.

This patch makes it so resource group stats to `pg_stat_activity` on segment backends are reported, making `rsgid` and `rsgname` columns filled for the time of slice execution.

Changes from original commit:
- pgstat_report_resgroup doesn't have a timestamp argument in 7X
- Updated tests to match 7X
- Added second row to output in two more places in resgroup_move_query, in the
same way as in the original commit.
- Merge conflicts resolved

Ticket: GG-160
(cherry picked from commit a0457d258acec64a1140b350f3f03d795c7a0aeb)

---

### DIFF
--- a/src/backend/utils/resgroup/resgroup.c
+++ b/src/backend/utils/resgroup/resgroup.c
@@ -1709,6 +1709,8 @@ SwitchResGroupOnSegment(const char *buf, int len)
 		cgroupOpsRoutine->attachcgroup(bypassedGroup->groupId, MyProcPid,
 									   caps.cpuMaxPercent == CPU_MAX_PERCENT_DISABLED);
 
+		pgstat_report_resgroup(bypassedSlot.groupId);
+
 		return;
 	}
 
@@ -1775,6 +1777,8 @@ SwitchResGroupOnSegment(const char *buf, int len)
 	/* Add into cgroup */
 	cgroupOpsRoutine->attachcgroup(self->groupId, MyProcPid,
 								   self->caps.cpuMaxPercent == CPU_MAX_PERCENT_DISABLED);
+
+	pgstat_report_resgroup(group->groupId);
 }
 
 /*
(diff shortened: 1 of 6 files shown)


src/backend/utils/resgroup/resgroup.c
```diff
@@ -3427,6 +3431,8 @@ HandleMoveResourceGroup(void)
 		cgroupOpsRoutine->attachcgroup(self->groupId, MyProcPid,
 									   self->caps.cpuMaxPercent == CPU_MAX_PERCENT_DISABLED);
 	}
+
+	pgstat_report_resgroup(groupId);
```
RekGRpth: Where is the removal of the line `pgstat_report_resgroup(self->groupId);` in this function (as in the original patch)?

bandetto (approved):

PR #221, opened 2026-02-03
ADBDEV-8369: Add Python 3 support

KnightMurloc:
This patch set implements Python 3 support for the Greenplum 6 cluster
maintenance utilities to support current distributions. Python 2.7 support is
retained, and building is still possible with both versions. In addition to
adapting the code, the PyGreSQL module used for connecting to cluster segments
was updated and replaced with a vanilla version.

To maintain the functionality of all catalog objects, one external table was
modified. This was deemed acceptable, as transferring data catalogs between
hosts with different distributions is not intended as a migration method. A
clean installation is required. Otherwise, manual modification of the external
table definition will be required for all databases.

---------

Must be rebased as is.

### DIFF
--- a/README.linux.md
+++ b/README.linux.md
@@ -29,18 +29,21 @@
   sudo ln -s python2 /usr/bin/python
   ```
 
-## For Ubuntu (22.04):
+## For Ubuntu (versions 22.04 or 24.04):
 
 - Install dependencies using README.ubuntu.bash script:
   ```bash
   sudo ./README.ubuntu.bash
   ```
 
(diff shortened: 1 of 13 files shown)


ShapatinAA:
> The `minirepro` tests failed, so I think it was necessary to do [this](https://github.com/GreengageDB/greengage/pull/271#discussion_r2832227069).

I think they have been applied from mentioned commit. Maybe there's more places where we need to use encode.

Stolb27 (approved):

PR #235, opened 2026-02-06
ADBDEV-9165: Collect SQL dump after new image built on main branches

magf:
## Сreates a new workflow `Greengage SQL Dump` in the `main` branch

A separate workflow `Greengage SQL Dump` is responsible for generating SQL dump
artifacts after the main CI process completes successfully. It is triggered
automatically upon the completion of the `Greengage CI` workflow.

### Key Features

- **Triggers:** `workflow_run: workflows: ["Greengage CI"], types: [completed]`
- **Branch Targeting:** Runs only for the `main` and `7.x` branches.
- **Version Detection:** Automatically determines the database version (6 or 7)
based on the triggering branch.
- **Artifact Creation:** Executes regression tests with the `dump_db: "true"`
parameter to generate a SQL dump archive, which is then uploaded as a workflow
artifact.
- **Controlled Execution:** Since the main CI workflow runs on `main` and `7.x`
branches only for push events (which occur after final merge of a PR), SQL dump
are generated exclusively for verified, approved patches after they are merged
into the main branches.
- **Artifact Retention:** The generated SQL dump artifact is retained 90 days
after the last download. Each new run of the `behave tests gpexpand` workflow
(which consumes this artifact as a consumer) resets this retention period to
90 days when it downloads the artifact.

Task: [ADBDEV-9165](https://tracker.yandex.ru/ADBDEV-9165)

### DIFF
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -41,7 +41,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: read  # Explicit for GHCR access clarity
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-behave.yml@v15
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-behave.yml@v18
     with:
       version: 6
       target_os: ${{ matrix.target_os }}
--- a/ci/docker-compose.yaml
+++ b/ci/docker-compose.yaml
@@ -1,64 +1,42 @@
----
+x-common-service: &common-service
+  image: "${IMAGE}"
+  privileged: true
+  sysctls:
+    kernel.sem: 500 1024000 200 4096
+  init: true
+  ulimits:
+    nofile: 65535
+  entrypoint: >
+    sleep infinity
 
-version: "3"
 services:
   cdw:
-    image: "${IMAGE}"
+    <<: *common-service
     working_dir: /home/gpadmin
     hostname: cdw
     volumes:
-      - "$PWD/ssh_keys/id_rsa:/home/gpadmin/.ssh/id_rsa"
-      - "$PWD/ssh_keys/id_rsa.pub:/home/gpadmin/.ssh/id_rsa.pub"
-      - "$PWD/allure-results:/tmp/allure-results"
-      - "$PWD/logs_cdw:/logs"
-    privileged: true
-    sysctls:
-      kernel.sem: 500 1024000 200 4096
-    init: true
-    ulimits:
-      nofile: 65535
-    entrypoint: >
-      sleep infinity
+      - ${PWD}/allure-results:/tmp/allure-results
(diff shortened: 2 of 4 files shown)


.github/workflows/greengage-sql-dump.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          echo ""
+          echo "================== Workflow Trigger Information =================="
+          echo "Triggered by:   ${{ github.event.workflow_run.name }} (#${{ github.event.workflow_run.id }})"
+          echo "Target branch:  ${{ github.event.workflow_run.head_branch }}"
+          echo "Source commit:  ${{ github.event.workflow_run.head_sha }}"
+          echo "Commit message: ${{ github.event.workflow_run.head_commit.message }}"
+          echo ""
+          echo "Author: ${{ github.event.workflow_run.head_commit.author.name }}"
+          echo "=================================================================="
```
bandetto: Why do we need trigger information?

.github/workflows/greengage-sql-dump.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Move Docker storage to /mnt/docker
+        run: |
+          sudo bash -c '
+            systemctl stop docker
+            mkdir -p /mnt/docker
+            rsync -a /var/lib/docker/ /mnt/docker
+            chown -R docker:docker /mnt/docker
+            echo "{ \"data-root\": \"/mnt/docker\" }" > /etc/docker/daemon.json
+            systemctl start docker
+          '
```
bandetto: I see that this step is to use a device with more available memory. Can you add a comment here to help the next person avoid reading README.md?

magf: This universal method has already been chosen and already used in other jobs.
Using different methods for the same tasks is a bad idea.
I can think of a dozen ways to solve this problem, and we'll never choose the best one.

.github/workflows/greengage-sql-dump.yml
```diff
@@ -0,0 +1,81 @@
...
+          echo "=================================================================="
+
+      - name: Move Docker storage to /mnt/docker
+        run: |
+          sudo bash -c '
+            systemctl stop docker
+            mv /var/lib/docker /mnt/docker
+            echo "{ \"data-root\": \"/mnt/docker\" }" > /etc/docker/daemon.json
+            systemctl start docker
+          '
```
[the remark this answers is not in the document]
magf: already. this [branch](https://github.com/GreengageDB/greengage-ci/tree/feat/mv-docker).
new [action](https://github.com/GreengageDB/greengage-ci/blob/feat/mv-docker/.github/actions/move-docker/action.yml).
a separate task is needed for implementation.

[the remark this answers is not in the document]
magf: I'm using it for example in my repo
```yaml
- name: Move Docker storage to /mnt/docker
  uses: greengagedb/greengage-ci/.github/actions/move-docker@feat/mv-docker
```
If you insist, we'll open a new task and do it right away.

RekGRpth (approved):

PR #239, opened 2026-02-06
GG-166: Add Ubuntu 24 support

silent-observer:
Add Ubuntu 24 support

Parameterize the Dockerfile and README.ubuntu.bash to support building both on
Ubuntu 22 and Ubuntu 24. The following was changed:
1. README.ubuntu.bash now installs only Python 3 on Ubuntu 24.
2. Different base image is used depending on the build argument.
3. Python 2 is set as default only on Ubuntu 22.
4. --break-system-packages is used on Ubuntu 24 to force overriding externally
managed environment with pip.
5. PYTHON3 variable is only set on Ubuntu 22 to avoid compiling plpython3
twice on Ubuntu 24.
6. Debian package build files were updated to support both versions.

---

P.S. To build with Ubuntu 24 + Python 3 merge GG-124 branch here.

### DIFF
--- a/ci/Dockerfile.ubuntu
+++ b/ci/Dockerfile.ubuntu
@@ -25,9 +25,9 @@ RUN set -eux; \
     echo 'precedence ::ffff:0:0/96  100' >> /etc/gai.conf; \
 # Install allure-behave for behave tests
     if [ "$UBUNTU_VERSION" = "22.04" ]; then \
-        pip2 install --no-cache-dir allure-behave==2.4.0 future; \
+        pip2 install --no-cache-dir allure-behave==2.4.0; \
     else \
-        pip3 install --no-cache-dir --break-system-packages allure-behave==2.4.0 future; \
+        pip3 install --no-cache-dir --break-system-packages allure-behave==2.4.0; \
     fi; \
 # Cleanup to reduce image size
     apt-get clean && \


ci/Dockerfile.ubuntu
RekGRpth: Do we really need this if we only compile with Python 3?

ci/Dockerfile.ubuntu
```diff
@@ -1,12 +1,16 @@
-FROM ubuntu:22.04 AS base
+ARG UBUNTU_VERSION=22.04
+
+FROM ubuntu:${UBUNTU_VERSION} AS base
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 WORKDIR /home/gpadmin
 
+ENV SET_PYTHON2_DEFAULT=1
+ENV INSTALL_ALLURE_BEHAVE=1
```
RekGRpth: Why do we need these environment variables if they can't be set externally when building an image?

gpAux/debian/control
```diff
@@ -14,13 +14,12 @@ Depends: ${shlibs:Depends},
...
 	 openssl,
-	 python2,
+	 python3,
 	 rsync,
 	 zip,
 	 net-tools,
 	 ${pythonRequires}
 Conflicts: greengage-loaders,
-	 python-is-python3
 Description: Greengage MPP database engine
```
RekGRpth: The dependency change doesn't seem universal. Will the release definitely work?

ci/Dockerfile.ubuntu
silent-observer: No, but do we really want complicated bash in Dockerfiles? I would rather put it into a separate script instead

silent-observer: I'll make a version with bash conditionals in Dockerfile to compare complexity

ci/Dockerfile.ubuntu
```diff
@@ -1,12 +1,18 @@
-FROM ubuntu:22.04 AS base
+ARG UBUNTU_VERSION=22.04
+
+FROM ubuntu:${UBUNTU_VERSION} AS base
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 WORKDIR /home/gpadmin
+ARG UBUNTU_VERSION
```
RekGRpth: ???

README.ubuntu.bash
```diff
@@ -43,10 +43,22 @@ apt-get install -y \
...
+
+if [[ -n "$USE_PYTHON3_ONLY" ]]; then
+  apt-get install -y \
+    python3-pip \
+    python-is-python3;
+else
+  apt-get install -y \
+    python-pip \
+    python2 \
+    python2-dev
```
RekGRpth: How does this work outside of a dockerfile without a symlink?

gpAux/debian/rules
```diff
@@ -10,8 +10,13 @@ LSB_SR := $(shell lsb_release -sr)
 
 ifeq ($(LSB_SI),Ubuntu)
   ifeq ($(LSB_SR),22.04)
-    DEPS = python2.7
+    DEPS = python2,python2.7
+    CONFLICTS = python-is-python3
+  else
+    DEPS = python3
```
RekGRpth: So we won't be shipping a build with Python 3 on Ubuntu 22?

README.ubuntu.bash
```diff
@@ -43,10 +43,18 @@ apt-get install -y \
...
   protobuf-compiler \
-  python-pip \
-  python2 \
-  python2-dev \
   python3-dev \
   rsync \
   sudo \
   zlib1g-dev
+
+if [ "$(lsb_release -si)" == "Ubuntu" ] && [ "$(lsb_release -sr)" == "24.04" ]; then
```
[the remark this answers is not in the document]
silent-observer: What do you mean? The condition remains the same, it just applies directly to the apt-get commands instead of setting a variable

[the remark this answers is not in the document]
RekGRpth: to simplify support for future versions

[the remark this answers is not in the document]
RekGRpth: like in `gpAux/debian/rules`

RekGRpth:
how about to change base branch?

PR #241, opened 2026-02-09
GG-147: Fix gpssh behave tests on Python 3

RekGRpth:
Fix gpssh behave tests on Python 3

gpssh behave tests didn't pass on Python 3.

1) Fix the error "TypeError: '<=' not supported between instances of 'int' and
'NoneType'" when comparing retry_attempt <= num_retries when num_retries is
None.

2) Fix the error "TypeError: write() argument must be str, not bytes" when
DEBUG_VERBOSE_PRINTING is True.

3) In Python 3, file operations return bytes, not strings, so the CRNL for
comparison must be in bytes.

Co-authored-by: Viktor Kurilko <v.kurilko@arenadata.io>

(cherry picked from commit 1b76372)

Changes from original commit:
- CRNL comparison adapted for both Python 2 and Python 3

Ticket: GG-147

---
!!!REBASE!!!

### DIFF
--- a/gpMgmt/bin/gpssh_modules/gppxssh_wrapper.py
+++ b/gpMgmt/bin/gpssh_modules/gppxssh_wrapper.py
@@ -36,7 +36,7 @@ class PxsshWrapper(pxssh.pxssh):
         num_retries = self.sync_retries
         retry_attempt = 0
         success = False
-        while (not success) and retry_attempt <= num_retries:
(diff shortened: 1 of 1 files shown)


silent-observer (dismissed):

KnightMurloc:
> I think that for this PR this is not very important, because it works on both Python 2 and 3 without additional changes.

Yes, but it doesn't make much sense without python 3. It seems logical to me since we have a branch for things related to python 3.

PR #246, opened 2026-02-10
ADBDEV-9228: Test expression refers to not projected the nested subplan attribute for 7.x

VoidZeroNull0:
What happens?
ORCA's ready DXL plan fails to be converted further with an error:
"DXL-to-PlStmt Translation: Attribute number *some_number* not found in project
list". This happens on queries with WHERE filter. In the filter, there is a NOT
IN condition between two subplans. For example of such query, please, look at
the tests.

Why it happens?
Logical chain could be traced like this:
1. During cost-based optimization, ORCA decides to use GroupExpression of
PhysicalCorrelatedInnerNLJoin type for inner subplan, due to it being more cost
efficient.
2. Later, during translation of ready plan to DXL, columns needed for TestExpr
of outer subplan (NOT IN condition) are determined (PdxlnQuantifiedSubplan()) -
some of them must be coming from inner subplan.
3. Then, inner subplan is converted to BooleanScalarSubplan and not the usual
one inside PdxlnCorrelatedNLJoin(). This happens due to it meeting needed
conditions (as it is PhysicalCorrelatedInnerNLJoin and has const true as a
scalar, join condition).
4. As BooleanScalarSubplan is placed inside the filter with its projection
replaced with const true and outer subplan expecting to get column of inner
subplan from projection or directly from subplan - the error occurs.

Overall, logic on step 3 seems to have never expected, that output of subplan
may be needed somewhere else, so conditions for converting it to
BooleanScalarSubplan are too soft.

How do we fix this?
Let's additionally check, that no columns, which can potentially come from
subplan, are needed on some upper level. If some of them are needed, we should
not make a BooleanScalarSubplan and make usual instead. Usual subplan is then
placed in place of needed col_id, so plan is converted and executed properly.

Tests?
Tests to check correct execution and ORCA planning of case-specific NOT IN
query was added. In these tests, generated DXL plan is also checked.

---------

Co-authored-by: Vasiliy Ivanov <ivi@arenadata.io>
(cherry picked from commit f7a679d24d9167877cd01c71f00527324fbdd3a2)


GirgerS:
I've changed the query a bit, and it doesn't seem to work for me. Can you double-check it? 

```sql
CREATE TABLE t1 (
    text_t1 text,
    int_t1 integer
  ) DISTRIBUTED BY (int_t1);
  CREATE TABLE t2 (
    text_t2 text,
    int_t2 integer
  ) DISTRIBUTED BY (int_t2);

  INSERT INTO t1 SELECT '0', generate_series (0,1);
  INSERT INTO t2 SELECT '0', generate_series (0,1);
  INSERT INTO t1 SELECT '1', generate_series (0,1);

  SELECT disable_xform('CXformImplementInnerJoin');
  explain SELECT * FROM t1 WHERE
  (SELECT int_t1 FROM t1 LIMIT 1)
    !=
  (SELECT int_t2 FROM t2 WHERE int_t2 = int_t1 OR t1.text_t1 = '0');

2026-02-13 17:21:46:180662 MSK,THD000,ERROR,"DXL-to-PlStmt Translation: Attribute number 10 not found in project list",
LOG:  GPORCA failed to produce plan (unexpected)
                                                  QUERY PLAN
---------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.02..901.36 rows=49550 width=36)
   InitPlan 1 (returns $0)  (slice3)
     ->  Limit  (cost=0.00..0.02 rows=1 width=4)
           ->  Gather Motion 3:1  (slice4; segments: 3)  (cost=0.00..0.05 rows=3 width=4)
                 ->  Limit  (cost=0.00..0.01 rows=1 width=4)
                       ->  Seq Scan on t1 t1_1  (cost=0.00..199.33 rows=16533 width=4)
   ->  Seq Scan on t1  (cost=0.00..240.67 rows=16517 width=36)
         Filter: ($0 <> (SubPlan 2))
         SubPlan 2
           ->  Result  (cost=0.00..1604.67 rows=49600 width=4)
                 Filter: ((t2.int_t2 = t1.int_t1) OR (t1.text_t1 = '0'::text))
                 ->  Materialize  (cost=0.00..1108.67 rows=49600 width=4)
                       ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=0.00..860.67 rows=49600 width=4)
                             ->  Seq Scan on t2  (cost=0.00..199.33 rows=16533 width=4)
(15 rows)
```

VoidZeroNull0:
> I've changed the query a bit, and it doesn't seem to work for me.

Thanks for the notice! I was testing it during research, but it worked at the time (maybe because I was not disabling any transformations). Dump looks similar to the original problem with `NOT IN`, but the query seems to use different code for translation to DXL. Overall, looks like subplan was not "pulled" in place of needed column. Gonna research this further.

GirgerS:
> I think newly found problem is worth its own PR, not to clog port for 7.x and make subsequent backport.

I agree, it does seem to be related to the patch (unlike I originally thought, it is a mistake on my part). Thanks for investigating it.

whitehawk:
In the description:
'needed somwhere else' -> 'needed somewhere else'

VoidZeroNull0:
> If yes, please add a reference to this PR into the commit's summary.

Done in fe470ea.

whitehawk (approved):

PR #248, opened 2026-02-11
GG-178: Fix the use of errno as an error indicator

silent-observer:
Fix the use of errno as an error indicator

Most library functions have the following convention: if an error happens,
errno is set and some return code is returned, and if the operation completes
successfully, errno is preserved. This means the correct way to handle errors
is to check the return code, and if an error happens, errno is guaranteed to
be set.

However, some functions have a different convention. These functions are fixed
in this patch:

1. readdir() returns NULL both on an error condition, and on successfully
reaching the end of the directory. errno must be checked to determine if an
error happened, so you have to set errno to 0 before using readdir() in all
cases.
2. read() and write() return -1 on error condition and guarantee that errno is
set in this case. However, they might also do a partial read/write and errno
isn't guaranteed to be set. So in all cases when we check partial read/write
to report an error, we must set errno to 0 before read()/write().
3. BIO_write() from OpenSSL might return a negative value without setting
errno. We should set it to 0 for it to have a defined value before the call.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -39,7 +39,7 @@ operating systems:
 
 ## Release Workflow
 
-A separate workflow, `Greengage release`, handles the uploading of Debian packages
+A separate workflow `Greengage release` handles the uploading of Debian package
 to GitHub releases. It is triggered when a release is published and uses a
 composite action to manage package deployment.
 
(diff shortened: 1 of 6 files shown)


dkovalev1:
Interesting problem. Have you checked all the library functions where resetting errno might be useful? It seems that at least `strtol` and similar may benefit too (even `man strtol` has such an example)?  Do you have a list of such a functions? Have you checked all call occurrences of them?

silent-observer:
About `strtol`: just checked it and it seems like all cases are already covered. Another important condition for this issue is that the `errno` has to actually be used for something, and in a lot of cases for `strtol` we check an error by some other criteria and never touch `errno` at all. Hence, what I used is just checking all uses of `errno`.

Also, most of these places aren't actually critical, for example I don't even know how to achieve a partial `read()` without setting `errno`, even though documentation allows it. The only actually problematic cases are `readdir()` in `initdb.c` and `write()` in `nodeShareInputScan.c`, even `write()` in `fe-lobj.c` only affects error message.

P.S. Postgres doesn't seem to have an equivalent patch, since the actually problematic cases occur in GPDB-specific code.

src/bin/initdb/initdb.c
KnightMurloc: Maybe do it as is done in walkdir: 
https://github.com/GreengageDB/greengage/blob/f2e95a4aaf9fdc7fd8779fb437b0e43fd5d5a4f7/src/bin/initdb/initdb.c#L680-L714

KnightMurloc:
> if the operation completes successfully, errno is preserved

This is not true. from errno man:
The value in errno is significant only when the return value of
the call indicated an error (i.e., -1 from most system calls; -1
or NULL from most library functions); **a function that succeeds is
allowed to change errno.**  The value of errno is never set to zero
by any system call or library function.

PR #305, opened 2026-03-06
Implement cluster rebalance rollback and cleanup

whitehawk:
Implement cluster rebalance rollback and cleanup

This patch adds:
1. Full rollback (via '-r' flag) capability for the rebalance flow.
2. Error checking for the steps marked as errored due to the interruption of the
tool execution (during both normal and rollback flows). Errored steps can be
retried, rolled back individually, or cancelled.
3. Cleanup capability when the execution was interrupted during rebalance.

In order to facilitate the capabilities above:
1. Added SegmentStatus command class to check the status of a running segment
via 'pg_ctl'.
2. Added fault injections inside the 'gprecoverseg' tool for the test purposes.
3. Renamed 'wrap_state_func_with_faults' into 'wrap_func_with_faults' in the
fault injector, as now it is used not only with the state machine functions.
4. In RebalanceStep class removed rollback-related statuses, as they duplicate
the statuses of the normal flow. Instead, 'rollback' flag is added to
distinguish between the steps of the normal and rollback flows.
5. Introduced updates in the main and shrink state machines in order to
coordinate cleanup and rollback between shrink and rebalance.
6. Removed 'gprecoverseg' invocation from 'check_down_segments()', as now the
proper error handling is done by the rebalance execution engine itself.
7. Moved 'check_down_segments()' to an earlier point to make FTS probe before
'check_running_gputils()' (otherwise, in some test cases
'check_running_gputils()' fails when accessing the cluster).
8. Added '--skip-resource-estimation' option to 'gpmovemirrors'. We use this
option when moving mirrors, as, when retrying a failed operation, in some cases
there will be no old datadir yet, and 'gpmovemirrors' will fail at resource
estimation. So we skip resource estimation at 'gpmovemirrors' side. It is not a
problem, as we have our own resource estimation.
9. Added new behave tests for rollback and cleanup.
10. Added new steps for behave tests.
11. Removed cases, related to failures during switchovers, from rebalance test
'3', as these cases are not checked in the newly added test cases differently.
12. Removed cases with delay from rebalance test '3', as now such scenarios are
tested with the help of fault injections in the 'gprecoverseg' at particular
points of interest.


gpMgmt/bin/ggrebalance_modules/ggrebalance_sm.py
```diff
@@ -315,9 +361,199 @@ def get_state_after_interrupt(self, prev_state) -> str:
...
+                        if catalog_segment_info.isSegmentUp() and catalog_segment_info.isSegmentModeSynchronized():
+                            self.logger.info('The step is complete, mark it as done')
+                            step.setStatus(RebalanceStep.Status.DONE, step.isRollback())
+                            self.rebalance_schema.updateExecutionStep(step)
+                            break
+
+                    time.sleep(SLEEP_PERIOD_SEC)
+                    time_waited = time_waited + SLEEP_PERIOD_SEC
+                    if time_waited >= TIMEOUT_SEC and self.interactive_check('Timeout waiting for segment start, wait again?'):
+                        time_waited = 0
```
KnightMurloc: Is there a way to check the replication progress?

gpMgmt/bin/ggrebalance_modules/ggrebalance_sm.py
```diff
@@ -315,9 +361,199 @@ def get_state_after_interrupt(self, prev_state) -> str:
...
+
+            # Continue with the next step, if we already marked this one
+            if step.getStatus() == RebalanceStep.Status.DONE:
+                continue
+
+            if not gp_segment_configuration_updated and self.interactive_check(f'Retry step?'):
```
KnightMurloc:
```suggestion
            if not gp_segment_configuration_updated and self.interactive_check('Retry step?'):
```

gpMgmt/bin/ggrebalance_modules/ggrebalance_sm.py
```diff
@@ -315,9 +361,199 @@ def get_state_after_interrupt(self, prev_state) -> str:
...
+
+        if not cmd.was_successful():
+            self.logger.info(f"Failed to get port from postgresql.conf on {hostname}: {cmd.get_stderr()}")
+            return -1
+
+        output = cmd.get_stdout().strip()
+        if not output or not output.isdigit():
+            return -1
+
+        return int(output)
```
KnightMurloc: Can we use something like that:
https://github.com/GreengageDB/greengage/blob/a39c24ff358c8755f4be38786e8c0111b44f109f/gpMgmt/bin/gpinitstandby#L728-L745

gpMgmt/bin/ggrebalance_modules/ggrebalance_sm.py
```diff
@@ -303,9 +344,14 @@ def state_is_final(self, state: str) -> bool:
         return state == self.states_main_rebalance_flow[-1]
 
     def get_state_after_interrupt(self, prev_state) -> str:
-        if (prev_state == 'STATE_REBALANCE_EXECUTION_STARTED' or
-            prev_state == 'STATE_REBALANCE_MOVES_SUCCEEDED' or
-            prev_state == 'STATE_REBALANCE_EXECUTION_AWAITING_SWITCHOVER_APPROVE_DONE'):
+        if prev_state in self.states_rollback_rebalance_flow[:-1]:
+            prev_idx = self.states_rollback_rebalance_flow.index(prev_state)
+            return self.states_rollback_rebalance_flow[prev_idx + 1]
```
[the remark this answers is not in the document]
whitehawk: The state is saved into the ggrebalance schema at the state completion. So, if we got `STATE_REBALANCE_ROLLBACK_PREPARE_MOVES_STARTED` from the schema, it means it is already complete, and we can proceed to `STATE_REBALANCE_ROLLBACK_PREPARE_MOVES_DONE`.

gpMgmt/bin/ggrebalance_modules/ggrebalance_sm.py
whitehawk: Ok, will update

gpMgmt/bin/ggrebalance_modules/ggrebalance_sm.py
whitehawk: Currently there is no replication progress monitoring

bimboterminator1 (commented):
There is a [backup state check](https://github.com/GreengageDB/greengage/blob/4995f00e554ec771711b1e5b00e6e23a698741a3/gpMgmt/bin/gppylib/programs/clsRecoverSegment_triples.py#L331-L344) for mirrors we are going to move in recoverseg. Should we cover that by setting appropriate move status, or we it's better to assume that nothing runs in parallel?

whitehawk:
> There is a [backup state check](https://github.com/GreengageDB/greengage/blob/4995f00e554ec771711b1e5b00e6e23a698741a3/gpMgmt/bin/gppylib/programs/clsRecoverSegment_triples.py#L331-L344) for mirrors we are going to move in recoverseg. Should we cover that by setting appropriate move status, or we it's better to assume that nothing runs in parallel?

We have a check at the start of ggrebalance in `check_running_gputils()`. So I think we can assume that nothing runs in parallel.

whitehawk:
@KnightMurloc @bimboterminator1 

Latest [commit](https://github.com/GreengageDB/greengage/pull/305/changes/e8ddb985d235d8df0f44ae596608502ddaeb9f9a) was done because of feedback from our architect - we should allow rollback of fully rebalanced cluster. Therefore, logic of an existing test was altered.

whitehawk:
@bimboterminator1 ,

I'd like to highlight that during merging with the latest changes on the feature branch, I've reworked the part in the rebalance executor, that you've recently updated for handling of 3rd phase when swapping via 3rd host. Please take a look.

gpMgmt/bin/ggrebalance_modules/ggrebalance_sm.py
```diff
@@ -1,521 +1,832 @@
...
-
-    @wrap_state_func_with_faults
-    def on_enter_STATE_ERROR(self) -> None:
-        raise Exception('Rebalance execution entered STATE_ERROR')
-
-    # state callbacks end here
+#!/usr/bin/env python3
+
+from transitions import Machine
+from enum import Enum
```
KnightMurloc, line 4: Why there is diff?

whitehawk: It appears that when merging the line ending were updated from CR LF to LF in some files. I suppose we'd better leave LF endings. You can hide this delta with `hide whitespace`

gpMgmt/bin/ggrebalance_modules/ggrebalance_sm.py
```diff
@@ -1,521 +1,832 @@
...
+                segment_current_info = move.seg
+                if self.lookup_seg(gparray, segment_current_info):
+                    cfg_line = f'{segment_current_info.getSegmentHostName()}|{segment_current_info.getSegmentPort()}|{segment_current_info.getSegmentDataDirectory()} '
+                else:
+                    cfg_line = f'{move.dstHost.hostname}|{move.target_port}|{move.target_datadir} '
+                if step.isRollback():
+                    cfg_line += f'{segment_current_info.getSegmentHostName()}|{segment_current_info.getSegmentPort()}|{segment_current_info.getSegmentDataDirectory()}\n'
+                else:
+                    cfg_line += f'{move.dstHost.hostname}|{move.target_port}|{move.target_datadir}\n'
+                fp.write(cfg_line)
```
KnightMurloc, line 336: Add comment about this logic.

whitehawk: added

bimboterminator1 (approved):

KnightMurloc (approved):

PR #306, opened 2026-03-08
GG-267: [7X] Place pg_regress into test docker image

GirgerS:
Install pg_isolation2_regress for use in extension tests (#274)

Previously, it was impossible to run isolation tests without directly building
pg_isolation2_regress from source. This complicates testing infrastructure in
extensions, requiring to specify ISOLATION2_ROOT to run isolation tests. This
patch instead builds and installs pg_isolation2_regress during the compilation
step.

Regular pg_regress was already built and installed in the pgxs directory, but
for some reason it wasn't actually being used. This patch instead uses the
installed version of pg_regress, and adds pg_isolation2_regress to the same
place. Other necessary files are added as symlinks to regress installation.

test_python step is moved from `all` target to `installcheck` target, since it
requires installed pygresql to work.

Ticket: GG-206
(cherry picked from commit https://github.com/GreengageDB/greengage/commit/44e96ec948f9caa7c8c667716a6d6c2d71e29716)

Changes compared to the original commit:
- Update the gg_tables_tracking extension to use the new pg_isolation2_regress
  path
- Make the installcheck target depend on the extended_protocol_test
  execulable, since it is built in the source directory and therefore
  not included in the docker image

---
Restore backwards compatibility for isolation2 tests (#302)

After commit https://github.com/GreengageDB/greengage/commit/44e96ec948f9caa7c8c667716a6d6c2d71e29716, even after make install
in isolation2 directory there were missing symlinks in the source directory.
These symlinks were moved to installation directory, but to preserve backwards
compatibility we should leave them in old place as well, so that isolation2
tests can still be run the old way, from source directory.

Ticket: ADBDEV-9345
(cherry picked from commit https://github.com/GreengageDB/greengage/commit/8b18119c950f18f6f96edea19101e6e214c83ea3)

---

Note: merge with rebase.

### DIFF
--- a/gpcontrib/gg_tables_tracking/isolation2/Makefile
+++ b/gpcontrib/gg_tables_tracking/isolation2/Makefile
@@ -1,8 +1,7 @@
 top_builddir = ../../../
-
-ISOLATION2_ROOT = $(top_builddir)src/test/isolation2
+include $(top_builddir)/src/Makefile.global
 
 installcheck:
-	cd $(ISOLATION2_ROOT) && ./pg_isolation2_regress $(EXTRA_REGRESS_OPTS) \
-	--dbname=trackingisolation --schedule=${CURDIR}/tracking_schedule --inputdir=${CURDIR} --outputdir=${CURDIR} \
+	$(pg_isolation2_regress_installcheck) \
+	--dbname=trackingisolation --schedule=tracking_schedule \
 	--load-extension=gp_inject_fault --load-extension=plpython3u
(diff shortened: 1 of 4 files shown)


GirgerS:
Regarding arendata_toolkit, `docker/tests.sh` seems to work locally for me. I'd suggest double-checking it, just to be sure.

GirgerS:
> Update link to PR in the commits' descriptions to the current PR. The previous PRs may be found throw the link to cherry-picked commits.

Done

red1452 (dismissed):

silent-observer (dismissed):

silent-observer (approved):

red1452 (approved):

PR #309, opened 2026-03-10
Fix calculating "total_queue_duration" in gp_toolkit.gp_resgroup_status

red1452:
Fix calculating "total_queue_duration" in gp_toolkit.gp_resgroup_status

View gp_toolkit.gp_resgroup_status uses "select" from the function
pg_resgroup_get_status. It gets statistics of resource groups by executing
function ResGroupGetStat. For the total_queue_duration column, value from field
totalQueuedTimeMs is used. But at the function addTotalQueueDuration value is
calculated incorrectly. Variables groupWaitStart and groupWaitEnd contain data
in microseconds and such value is written to totalQueuedTimeMs without
converting. As a result, we get total_queue_duration which is greater 1000
times, than correct value.

Patch adds converting microseconds to milliseconds to the function
addTotalQueueDuration, and View gp_toolkit.gp_resgroup_status returns correct
value in the "total_queue_duration" column.

### DIFF
--- a/src/backend/utils/resgroup/resgroup.c
+++ b/src/backend/utils/resgroup/resgroup.c
@@ -930,7 +930,7 @@ ResGroupGetStat(Oid groupId, ResGroupStatType type)
 			 * justify_interval() if she wants.
 			 */
 			interval = (Interval *) palloc(sizeof(Interval));
-			interval->time = group->totalQueuedTimeMs;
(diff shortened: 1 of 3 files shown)


silent-observer (approved):

VoidZeroNull0 (approved):
LGTM.

PR #316, opened 2026-03-12
[GG-192] Adapt pg_wait_sampling to Greengage

bimboterminator1:
Adapt pg_wait_sampling to Greengage

This patch adapts the pg_wait_sampling extension from
[postgrespro/pg_wait_sampling](https://github.com/postgrespro/pg_wait_sampling) project.
The commit history is preserved. As initial point the https://github.com/postgrespro/pg_wait_sampling/commit/89dda55 is taken

The exported part of extension - SQL interface and extension name - is renamed.
Now SQL functions and view have gg_* prefix instead of pg_*. The C symbols
are not renamed in order to simplify future patch porting from upstream.
The SQL interface was reworked for distributed queries support. Old
views now retrieve wait events from all segments and coordinator.

Main GGDB specific changes are related to queryId handling. Instead of
vanila queryId entity the QueryItem type is introduced. It contains several
fields which help to identify running query more precisely. Original queryId is
now stored there and is evaluated in ggws_post_parse_analyze() hook. Since pg
does not have unsigned types, uint64 values may be represented with overflow in
sql output. The main architecture is unchanged.
The SQL interface is extended with distributed versions of original views.
The profile reset is now made through view, not a function as it was assumed
in upstream interface.

Some irrelevant for GG files are deleted. Regression and isolation2 tests are
added covering base functionality. Resgroup wait events tests are
separated from main isolation2 installcheck. Corresponding license is
left as it is, it's also referenced in NOTICE file in repo root.
README.md is mostly rewritten for GGDB specific adaptation.


RekGRpth:
> This patch adapts the pg_wait_sampling extension from
> [postgrespro/pg_wait_sampling](https://github.com/postgrespro/pg_wait_sampling) project. The commit history is preserved with
> references to original repo commits.

why not fork and sub-module?

RekGRpth:
> SQL functions and view have gg_* prefix instead of pg_*

why not add new `gg_*` and save old `pg_*` prefixes in sql?

gpcontrib/gg_wait_sampling/pg_wait_sampling.h
```diff
@@ -21,19 +21,27 @@
 #define PGWS_QUEUE_LOCK				0
 #define PGWS_COLLECTOR_LOCK			1
 
+typedef struct
+{
+	int		tmid;			/* pg start time */
```
RekGRpth: 1) why `pg start time`? `InstrumentationSlot` and `gpmon_qlogkey_t` comment `tmid` as `transaction time`.

3) why `int`? `InstrumentationSlot` and `gpmon_qlogkey_t` contain `tmid`, `ssid` and `ccnt` as `int32`.

RekGRpth:
may be symlinking instead renaming files?

RekGRpth:
where is readme?

gpcontrib/gg_wait_sampling/gg_wait_sampling.control
```diff
@@ -0,0 +1,5 @@
+# gg_wait_sampling extension
+comment = 'sampling based statistics of wait events'
+default_version = '1.0'
```
RekGRpth: why not `1.1` to simplify future syncing?

gpcontrib/gg_wait_sampling/pg_wait_sampling.h
```diff
@@ -21,19 +21,27 @@
...
+	int		ccnt;			/* command count */
+	uint64	sql_id;			/* queryId from Query */
+} WSQueryId;
+
 typedef struct
 {
 	int			pid;
 	uint32		wait_event_info;
-	uint64		queryId;
+	WSQueryId	queryId;
```
RekGRpth: why not four different fields as in `pgws_should_sample_proc`?

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -790,12 +820,12 @@ pg_wait_sampling_get_profile(PG_FUNCTION_ARGS)
 		else
 			nulls[2] = true;
 
-		if (pgws_profileQueries)
```
RekGRpth: why removed?

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.0.sql
```diff
@@ -0,0 +1,119 @@
+/* gpcontrib/gg_wait_sampling/gg_wait_sampling--1.0.sql */
+
```
RekGRpth, line 5:
```suggestion
-- complain if script is sourced in psql, rather than via CREATE EXTENSION
\echo Use "CREATE EXTENSION gg_wait_sampling" to load this file. \quit
```

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.0.sql
```diff
@@ -0,0 +1,119 @@
...
+RETURNS VOID
+LANGUAGE plpgsql VOLATILE
+AS $$
+BEGIN
+PERFORM gg_wait_sampling_reset_profile();
+PERFORM gg_wait_sampling_reset_profile_segments();
+END;
+$$;
+
+REVOKE ALL ON FUNCTION gg_wait_sampling_cluster_reset_profile() FROM PUBLIC;
```
RekGRpth: why not simple view and cluster view?

gpcontrib/gg_wait_sampling/isolation2/sql/resgroup_wait_events.sql
```diff
@@ -0,0 +1,33 @@
+--
+-- Test statistics collection with GGDB specific wait events. Resgroup
+--
+!\retcode gpconfig -c gp_resource_manager -v group;
```
RekGRpth, line 4: this test for cgroup v1 ot v2?

gpcontrib/gg_wait_sampling/expected/gg_queries.out
```diff
@@ -0,0 +1,136 @@
...
+SELECT count(*) = 1 AS ok
+FROM gg_wait_sampling_get_current(pg_backend_pid());
+ ok 
+----
+ t
+(1 row)
+
+-- A non-existent pid must raise an error.
+SELECT gg_wait_sampling_get_current(-1);
+ERROR:  backend with pid=-1 not found (pg_wait_sampling.c:467)
```
RekGRpth: maybe ignore so write
```suggestion
ERROR:  backend with pid=-1 not found
```

RekGRpth:
new flaky?
```diff
[51](https://github.com/GreengageDB/greengage/actions/runs/23028465733/job/66882967761?pr=316#step:5:1852)
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/regress/expected/select_parallel.out /home/gpadmin/gpdb_src/src/test/regress/results/select_parallel.out
--- /home/gpadmin/gpdb_src/src/test/regress/expected/select_parallel.out	2026-03-12 23:41:01.599140311 +0000
+++ /home/gpadmin/gpdb_src/src/test/regress/results/select_parallel.out	2026-03-12 23:41:01.651140779 +0000
@@ -599,10 +599,10 @@
                ->  Sort (actual rows=3386 loops=1)
                      Sort Key: tenk1.ten
                      Sort Method:  quicksort  Memory: xxx
-                     Executor Memory: xxx  Segments: 3  Max: 203kB (segment 1)
+                     Executor Memory: xxx  Segments: 3  Max: 204kB (segment 1)
                      ->  Seq Scan on tenk1 (actual rows=3386 loops=1)
                            Filter: (ten < 100)
-GP_IGNORE:(11 rows)
+GP_IGNORE:(12 rows)
 
 reset enable_indexscan;
 reset enable_hashjoin;
```

LICENSE-THIRD-PARTY
```diff
@@ -0,0 +1,25 @@
+This project includes code from pg_wait_sampling extension https://github.com/postgrespro/pg_wait_sampling
```
Stolb27: What difference between this and existing NOTICE file?

bimboterminator1: I'll put the license into the NOTICE file, thanks

gpcontrib/gg_wait_sampling/pg_wait_sampling.h
bimboterminator1: > why pg start time

https://github.com/GreengageDB/greengage/blob/7.x/src/backend/executor/instrument.c#L528-L541

What comment would be more correct to use?

> why int?

Changed to int32.

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.0.sql
bimboterminator1: Done

gpcontrib/gg_wait_sampling/expected/gg_queries.out
bimboterminator1: Fixed matchsubs

gpcontrib/gg_wait_sampling/isolation2/sql/resgroup_wait_events.sql
bimboterminator1: Tested with v1. It does not work with v2?

gpcontrib/gg_wait_sampling/pg_wait_sampling.h
bimboterminator1: It seemed to me more readable and easier to initialize. Do you think we should simply expand it in exisiting stucture?

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: I guess it's zeroed anyway. Or you mean we should reduce diff?

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.0.sql
bimboterminator1: I've decided to preserve functional semantics since it does not return any rows

gpcontrib/gg_wait_sampling/gg_wait_sampling.control
bimboterminator1: Done

LICENSE-THIRD-PARTY
bimboterminator1: Put into the NOTICE

bimboterminator1 (commented):
> why not add new gg_* and save old pg_* prefixes in sql?
I believe that if we 've started to rename the whole extension to `gg_wait_sampling`, the sql should be also coherent with that. Otherwise it's a little bit confusing. However, if we used the extension as submodule, we would leave the `pg_*` namings. 

> may be symlinking instead renaming files?

gg_wait_sampling -> pg_wait_sampling?

>new flaky?

Seems like that. I'll create a ticker for that.
Failed job https://github.com/GreengageDB/greengage/actions/runs/23048426085/job/66945048017?pr=316

gpcontrib/gg_wait_sampling/pg_wait_sampling.h
RekGRpth: > What comment would be more correct to use?

I think, `transaction time`

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
RekGRpth: > you mean we should reduce diff?

yes

RekGRpth:
Shouldn't we add `gg_wait_sampling` to `gpcontrib/Makefile`?

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.0.sql
RekGRpth:
```sql
CREATE FUNCTION gg_wait_sampling_reset_profile()
RETURNS VOID
AS '$libdir/gg_wait_sampling', 'pg_wait_sampling_reset_profile'
LANGUAGE C VOLATILE STRICT;

CREATE VIEW gg_wait_sampling_reset_profile AS
    SELECT true FROM gg_wait_sampling_reset_profile();

CREATE VIEW gg_wait_sampling_cluster_reset_profile AS
    SELECT -1 AS gp_segment_id, * FROM gg_wait_sampling_reset_profile
    UNION ALL
    SELECT gp_execution_segment(), * FROM gp_dist_random('gg_wait_sampling_reset_profile');

explain (verbose, costs off) select * from gg_wait_sampling_cluster_reset_profile;
                                                QUERY PLAN                                                 
-----------------------------------------------------------------------------------------------------------
 Append
   ->  Function Scan on public.gg_wait_sampling_reset_profile
         Output: '-1'::integer, true
         Function Call: gg_wait_sampling_reset_profile()
   ->  Gather Motion 3:1  (slice1; segments: 3)
         Output: (gp_execution_segment()), gg_wait_sampling_reset_profile_1.bool
         ->  Subquery Scan on gg_wait_sampling_reset_profile_1
               Output: gp_execution_segment(), gg_wait_sampling_reset_profile_1.bool
               ->  Function Scan on public.gg_wait_sampling_reset_profile gg_wait_sampling_reset_profile_2
                     Output: true
                     Function Call: gg_wait_sampling_reset_profile()
(12 rows)
```

gpcontrib/gg_wait_sampling/isolation2/sql/resgroup_wait_events.sql
RekGRpth: And on CI, which groups are regression tests run on? I suspect it's the v2.

RekGRpth:
> gg_wait_sampling -> pg_wait_sampling?

yes

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
```diff
@@ -0,0 +1,122 @@
...
+
+CREATE VIEW gg_wait_sampling_current AS
+	SELECT * FROM gg_wait_sampling_get_current(NULL::integer);
+
+GRANT SELECT ON gg_wait_sampling_current TO PUBLIC;
+
+CREATE VIEW gg_wait_sampling_cluster_current AS
+	SELECT * FROM gg_wait_sampling_current
+	UNION ALL
+	SELECT * FROM gp_dist_random('gg_wait_sampling_current');
```
RekGRpth: Is there a potential vulnerability here? In a similar location, the `gp_toolkit` extension uses a [qualified name](https://github.com/GreengageDB/greengage/blob/8e42739e4d06ce4c721973477470f0037368a1d8/gpcontrib/gp_toolkit/gp_toolkit--1.7--1.8.sql#L156) and a [constant schema](https://github.com/GreengageDB/greengage/blob/8e42739e4d06ce4c721973477470f0037368a1d8/gpcontrib/gp_toolkit/gp_toolkit.control#L5).

gpcontrib/gg_wait_sampling/pg_wait_sampling.h
RekGRpth: may be
```suggestion
	QueryItem	query_item;
```
where
```c
typedef struct
{
	uint64		queryId;		/* queryId from Query */
	int32		tmid;			/* transaction time */
	int32		ssid;			/* session id */
	int32		ccnt;			/* command count */
} QueryItem;
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.h
```diff
@@ -70,7 +78,7 @@ extern bool pgws_sampleCpu;
 /* pg_wait_sampling.c */
 extern CollectorShmqHeader *pgws_collector_hdr;
 extern shm_mq *pgws_collector_mq;
-extern uint64 *pgws_proc_queryids;
+extern WSQueryId *pgws_proc_queryids;
```
RekGRpth:
```suggestion
extern QueryItem *pgws_proc_query_items;
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -1060,14 +1103,17 @@ pgws_ExecutorFinish(QueryDesc *queryDesc)
...
 		else
 			pgws_proc_queryids[i] = save_queryId;
 	}
 	PG_CATCH();
 	{
 		nesting_level--;
-		pgws_proc_queryids[i] = save_queryId;
+		if (nesting_level == 0)
+			pgws_proc_queryids[i] = (WSQueryId) {0};
+		else
```
RekGRpth: Can you clarify this point?

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: Yep, added schema creation. We could use  @extschema@, but the extension won't be relocatable

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: restored

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: Yep, that's redundant since executor end does it anyway

bimboterminator1:
> gg_wait_sampling -> pg_wait_sampling?

Since changes in sql are drastic, the syncing process will cause conflics anyway. I suggest remove unused files, since they confuse users and packagers, create false maintenance burden, bloat extension directory.

gpcontrib/gg_wait_sampling/gg_wait_sampling.control
```diff
@@ -1,5 +1,5 @@
-# pg_wait_sampling extension
+# gg_wait_sampling extension
 comment = 'sampling based statistics of wait events'
 default_version = '1.1'
-module_pathname = '$libdir/pg_wait_sampling'
+module_pathname = '$libdir/gg_wait_sampling'
 relocatable = true
```
RekGRpth, line 5:
```suggestion
relocatable = false
```

RekGRpth:
where is readme?

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -31,6 +33,7 @@
 #include "utils/builtins.h"
 #include "utils/guc.h"
 #include "utils/memutils.h"
+#include "utils/queryjumble.h"
```
RekGRpth: Place GPDB-specific includes separately below, this will make it easier to sync with upstream.

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -140,6 +145,9 @@ bool		pgws_profilePid = true;
 int			pgws_profileQueries = PGWS_PROFILE_QUERIES_TOP;
 bool		pgws_sampleCpu = true;
 
+/*---- GGDB funcs ----*/
+static QueryItem ggws_calculate_query_item(uint64 queryId, bool with_commandId);
```
RekGRpth:
```suggestion
static QueryItem ggws_calculate_query_item(uint64 queryId, bool with_commandId);
static void ggws_post_parse_analyze(ParseState *pstate, Query *query);
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -791,11 +821,15 @@ pg_wait_sampling_get_profile(PG_FUNCTION_ARGS)
...
 		if (pgws_profileQueries)
-			values[3] = UInt64GetDatum(item->queryId);
-		else
-			values[3] = (Datum) 0;
+		{
+			values[3] = UInt64GetDatum(item->query_item.queryId);
+			values[5] = Int32GetDatum(item->query_item.ssid);
+			values[6] = Int32GetDatum(item->query_item.ccnt);
+			values[7] = Int32GetDatum(item->query_item.tmid);
+		}
```
RekGRpth, line 847:
```suggestion
		}
		else
		{
			values[3] = (Datum) 0;
			values[5] = (Datum) 0;
			values[6] = (Datum) 0;
			values[7] = (Datum) 0;
		}
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -935,12 +981,12 @@ pgws_planner_hook(Query *parse,
 {
 	PlannedStmt *result;
 	int			i = MyProc - ProcGlobal->allProcs;
-	uint64		save_queryId = 0;
+	QueryItem	save_queryId = {0};
```
RekGRpth:
```suggestion
	QueryItem	save_query_item = {0};
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -959,19 +1005,20 @@ pgws_planner_hook(Query *parse,
 									  query_string,
 #endif
 									  cursorOptions, boundParams);
+		result->queryId = parse->queryId;
```
RekGRpth, line 1054: Could you explain where this line came from?

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -1010,7 +1057,7 @@ pgws_ExecutorRun(QueryDesc *queryDesc,
 )
 {
 	int			i = MyProc - ProcGlobal->allProcs;
-	uint64		save_queryId = pgws_proc_queryids[i];
+	QueryItem	save_queryId = pgws_proc_query_items[i];
```
RekGRpth:
```suggestion
	QueryItem	save_query_item = pgws_proc_query_items[i];
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -1049,7 +1096,7 @@ static void
 pgws_ExecutorFinish(QueryDesc *queryDesc)
 {
 	int			i = MyProc - ProcGlobal->allProcs;
-	uint64		save_queryId = pgws_proc_queryids[i];
+	QueryItem	save_queryId = pgws_proc_query_items[i];
```
RekGRpth:
```suggestion
	QueryItem	save_query_item = pgws_proc_query_items[i];
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -1108,12 +1155,12 @@ pgws_ProcessUtility(PlannedStmt *pstmt,
 )
 {
 	int			i = MyProc - ProcGlobal->allProcs;
-	uint64		save_queryId = 0;
+	QueryItem	save_queryId = (QueryItem) {0};
```
RekGRpth:
```suggestion
	QueryItem	save_query_item = (QueryItem) {0};
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -1147,18 +1194,59 @@ pgws_ProcessUtility(PlannedStmt *pstmt,
...
+			pgws_proc_query_items[i] = save_queryId;
 		PG_RE_THROW();
 	}
 	PG_END_TRY();
 }
+
+static QueryItem
+ggws_calculate_query_item(uint64 queryId, bool with_commandId)
+{
+	QueryItem res_id = {0};
```
RekGRpth:
```suggestion
	QueryItem query_item = {0};
```

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
```diff
@@ -0,0 +1,130 @@
...
+CREATE FUNCTION gg_wait_sampling.gg_wait_sampling_reset_profile_segments()
+RETURNS SETOF bool
+EXECUTE ON ALL SEGMENTS
+LANGUAGE SQL VOLATILE
+AS $$
+    SELECT gg_wait_sampling.gg_wait_sampling_reset_profile();
+    SELECT true;
+$$;
+
+REVOKE ALL ON FUNCTION gg_wait_sampling.gg_wait_sampling_reset_profile_segments() FROM PUBLIC;
```
RekGRpth: Is this function still really needed?

gpcontrib/gg_wait_sampling/collector.c
```diff
@@ -122,6 +126,17 @@ handle_sigterm(SIGNAL_ARGS)
...
+handle_sighup(SIGNAL_ARGS)
+{
+	int			save_errno = errno;
+
+	got_sighup = true;
+	SetLatch(MyLatch);
+
+	errno = save_errno;
+}
+
```
RekGRpth, line 140: why not something like
```suggestion
#if PG_VERSION_NUM < 130000
void SignalHandlerForConfigReload(SIGNAL_ARGS);
...
void
SignalHandlerForConfigReload(SIGNAL_ARGS)
{
	int			save_errno = errno;

	ConfigReloadPending = true;
	SetLatch(MyLatch);

	errno = save_errno;
}
#endif
```

gpcontrib/gg_wait_sampling/sql/setup.sql
```diff
@@ -0,0 +1,9 @@
+--Setup shared_preload_libraries and create schema for tests
+--start_ignore
+\! gpconfig -c shared_preload_libraries -v "$(psql -At -c "SELECT array_to_string(array_append(string_to_array(current_setting('shared_preload_libraries'), ','), 'gg_wait_sampling'), ',')" postgres)"
+\! gpconfig -c gg_wait_sampling.history_period -v 10000
+\! gpconfig -c gg_wait_sampling.profile_period -v 10000
+\! gpstop -raq -M fast
+--end_ignore
+\c contrib_regression;
+create schema gg_wait_sampling;
```
RekGRpth: I don't quite understand why we need to create a schema manually?

gpcontrib/gg_wait_sampling/sql/teardown.sql
```diff
@@ -0,0 +1,7 @@
+--Teardown tests
+--start_ignore
+\! gpconfig -c shared_preload_libraries -v "$(psql -At -c "SELECT array_to_string(array_remove(string_to_array(current_setting('shared_preload_libraries'), ','), 'gg_wait_sampling'), ',')" postgres)"
+\! gpstop -raq -M fast
+--end_ignore
+\c contrib_regression;
+drop schema gg_wait_sampling cascade;
```
RekGRpth: I don't quite understand why we need to drop a schema manually?

gpcontrib/gg_wait_sampling/sql/gg_queries.sql
```diff
@@ -0,0 +1,72 @@
+-- start_matchsubs
+-- m/ERROR:  backend with pid=-?\d+ not found( \([[:alnum:]_]+\.[ch]:\d+\))?/
+-- s/ \([[:alnum:]_]+\.[ch]:\d+\)//
+-- end_matchsubs
+
+CREATE EXTENSION gg_wait_sampling;
+SET search_path=gg_wait_sampling;
+
+-- Restore normal collection periods so the sampler gathers data
```
RekGRpth: Is that really the word "Restore"?

gpcontrib/gg_wait_sampling/sql/gg_queries.sql
```diff
@@ -0,0 +1,72 @@
...
+SELECT * from gg_wait_sampling_cluster_reset_profile;
+SELECT gp_segment_id, sleep_on_segment(6) FROM gp_dist_random('gp_id');
+-- Validate
+SELECT event_type, event
+FROM gg_wait_sampling_cluster_profile
+WHERE event_type = 'Timeout'
+AND event = 'PgSleep';
+--start_ignore
+\! gpconfig -c gg_wait_sampling.history_period -v 10000
+\! gpconfig -c gg_wait_sampling.profile_period -v 10000
```
RekGRpth:
```suggestion
\! gpconfig -r gg_wait_sampling.history_period
\! gpconfig -r gg_wait_sampling.profile_period
```

gpcontrib/gg_wait_sampling/sql/gg_queries.sql
```diff
@@ -0,0 +1,72 @@
...
+CREATE FUNCTION sleep_on_segment(seconds INT)
+RETURNS VOID
+LANGUAGE plpgsql
+AS $$
+BEGIN
+    PERFORM pg_sleep(seconds);
+END;
+$$;
+SELECT * from gg_wait_sampling_cluster_reset_profile;
+SELECT gp_segment_id, sleep_on_segment(6) FROM gp_dist_random('gp_id');
```
RekGRpth:
```suggestion
SELECT gp_segment_id, pg_sleep(6) FROM gp_dist_random('gp_id');
```
or even
```suggestion
SELECT -1 AS gp_segment_id, pg_sleep(6)
UNION ALL
SELECT gp_segment_id, pg_sleep(6) FROM gp_dist_random('gp_id');
```

gpcontrib/gg_wait_sampling/sql/gg_queries.sql
```diff
@@ -0,0 +1,72 @@
...
+
+-- Restore normal collection periods so the sampler gathers data
+--start_ignore
+\! gpconfig -c gg_wait_sampling.history_period -v 5
+\! gpconfig -c gg_wait_sampling.profile_period -v 5
+\! gpstop -u
+--end_ignore
+-- BASIC QUERYABILITY
+-- Row counts are timing-dependent; we only assert they are non-negative
+SELECT count(*) >= 0 AS ok FROM gg_wait_sampling_current;
```
RekGRpth: Why make a separate request on the coordinator if the next request already includes a request on the coordinator?

gpcontrib/gg_wait_sampling/sql/gg_queries.sql
```diff
@@ -0,0 +1,72 @@
...
+LANGUAGE plpgsql
+AS $$
+BEGIN
+    PERFORM pg_sleep(seconds);
+END;
+$$;
+SELECT * from gg_wait_sampling_cluster_reset_profile;
+SELECT gp_segment_id, sleep_on_segment(6) FROM gp_dist_random('gp_id');
+-- Validate
+SELECT event_type, event
```
RekGRpth:
```suggestion
SELECT segid, event_type, event
```

gpcontrib/gg_wait_sampling/isolation2/sampling_schedule
```diff
@@ -0,0 +1,3 @@
+test: lock_wait_events
+test: resgroup_wait_events
```
RekGRpth: Is it really a good idea to run resource group tests in regression, rather than separately as is done in the core?

RekGRpth: The regression environment is not prepared for running resource group tests.

gpcontrib/gg_wait_sampling/gg_wait_sampling.control
bimboterminator1: Why? It will block ALTER EXTENSION  SET SCHEMA

RekGRpth: And then all cluster views will stop working!

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: Done

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: Done

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: Done

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: Done. Thanks for 300 identical comments

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: Added the comment. Result->queryId is not always set since query is not usually jumbled.

bimboterminator1 (commented):
@RekGRpth , I came across an issue, when running multi slice query:
```sql
SELECT ABS(cnt_hist - cnt_prof) < 10 FROM
(SELECT pid, event, count(event) as cnt_hist
FROM gg_wait_sampling_cluster_history WHERE event='ResourceGroup'
GROUP BY pid, event) s1 JOIN (SELECT pid, event, count as cnt_prof
FROM gg_wait_sampling_cluster_profile WHERE event='ResourceGroup') s2 USING(pid);

ERROR:  Unexpected internal error (assert.c:44)  (seg0 slice1 127.0.0.1:7002 pid=1048972) (assert.c:44)

DETAIL:  FailedAssertion("!(mq->mq_receiver == MyProc)", File: "shm_mq.c", Line: 864)

```
The `pg_wait_sampling_get_profile` is executed at slice1, `pg_wait_sampling_get_history` is executed at slice2. However, the locking mechanism of GGDB allows processes under the same mpp session to acquire the same lock, even if the locktype is conflicting:
https://github.com/GreengageDB/greengage/blob/d64b1fcba042fb0dac819c83c92ac4283dc2478e/src/backend/storage/lmgr/lock.c#L1682-L1687

https://github.com/GreengageDB/greengage/blob/d64b1fcba042fb0dac819c83c92ac4283dc2478e/src/backend/storage/lmgr/lock.c#L1695-L1699

Thus, when both slices acquire `queueTag`  **shm_mq is not accessed exclusively** .

Here are 3 possible variants to handle it:
1. Prohibit execution of extension functions at different slices.
2. Use LWLock mechanism.
3. Patch core to cause blocking behaviour for LOCKTAG with `USER_LOCKMETHOD`  - advisory lock will be affected.

I need your opinion. We could discuss it in person as well.

gpcontrib/gg_wait_sampling/sql/gg_queries.sql
```diff
@@ -0,0 +1,61 @@
...
+SELECT * from gg_wait_sampling.gg_wait_sampling_cluster_reset_profile;
+
+-- Test Timeout event at coordinator
+SELECT pg_sleep(6);
+-- The pg_sleep above must have produced a Timeout/pg_sleep sample.
+SELECT count(*) > 0 AS sleep_event_captured
+FROM gg_wait_sampling.gg_wait_sampling_profile
+WHERE pid = pg_backend_pid()
+AND event_type = 'Timeout'
+AND event = 'PgSleep';
```
RekGRpth: Do we really need to check the slip on the coordinator, when it is already checked below?

RekGRpth:
> The `pg_wait_sampling_get_profile` is executed at slice1, `pg_wait_sampling_get_history` is executed at slice2.

Could You provide plan?

RekGRpth:
> Resgroup wait events tests are
> separated from main isolation2 installcheck.

Resource group tests not running on CI?

bimboterminator1:
>Resource group tests not running on CI?

Made gpcontrib perform resgroup tests.

RekGRpth:
fix resgroup tests

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -646,8 +686,8 @@ receive_array(SHMRequest request, Size item_size, Size *count)
 	MemoryContext oldctx;
 
 	/* Ensure nobody else trying to send request to queue */
-	pgws_init_lock_tag(&queueTag, PGWS_QUEUE_LOCK);
-	LockAcquire(&queueTag, ExclusiveLock, false, false);
+	queueLock = &(GetNamedLWLockTranche("gg_wait_sampling"))->lock;
```
RekGRpth: I think this can be done once in `pgws_shmem_startup` function when `ShmemInitStruct` is not found.

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
```diff
@@ -0,0 +1,179 @@
...
+LANGUAGE SQL VOLATILE EXECUTE ON COORDINATOR;
+
+GRANT EXECUTE ON FUNCTION gg_wait_sampling.gg_wait_sampling_get_current_coordinator TO PUBLIC;
+
+CREATE VIEW gg_wait_sampling.gg_wait_sampling_current AS
+	SELECT * FROM gg_wait_sampling.gg_wait_sampling_get_current_coordinator()
+	UNION ALL
+	SELECT * FROM gg_wait_sampling.gg_wait_sampling_get_current_segments();
+
+GRANT SELECT ON gg_wait_sampling.gg_wait_sampling_current TO PUBLIC;
```
RekGRpth: why not simply two functions?
```suggestion
CREATE FUNCTION gg_wait_sampling.gg_wait_sampling_get_current_segments (
	pid int4,
	OUT pid int4,
	OUT event_type text,
	OUT event text,
	OUT queryid int8,
	OUT mppsessionid int4,
	OUT command_id int4,
	OUT tmid int4,
	OUT segid int4
)
RETURNS SETOF record
AS '$libdir/gg_wait_sampling', 'pg_wait_sampling_get_current'
LANGUAGE C VOLATILE CALLED ON NULL INPUT EXECUTE ON ALL SEGMENTS;

GRANT EXECUTE ON FUNCTION gg_wait_sampling.gg_wait_sampling_get_current_segments TO PUBLIC;

CREATE FUNCTION gg_wait_sampling.gg_wait_sampling_get_current_coordinator (
	pid int4,
	OUT pid int4,
	OUT event_type text,
	OUT event text,
	OUT queryid int8,
	OUT mppsessionid int4,
	OUT command_id int4,
	OUT tmid int4,
	OUT segid int4
)
RETURNS SETOF record
AS '$libdir/gg_wait_sampling', 'pg_wait_sampling_get_current'
LANGUAGE C VOLATILE CALLED ON NULL INPUT EXECUTE ON COORDINATOR;

GRANT EXECUTE ON FUNCTION gg_wait_sampling.gg_wait_sampling_get_current_coordinator TO PUBLIC;

CREATE VIEW gg_wait_sampling.gg_wait_sampling_current AS
	SELECT * FROM gg_wait_sampling.gg_wait_sampling_get_current_coordinator(NULL::integer)
	UNION ALL
	SELECT * FROM gg_wait_sampling.gg_wait_sampling_get_current_segments(NULL::integer);

GRANT SELECT ON gg_wait_sampling.gg_wait_sampling_current TO PUBLIC;
```

gpcontrib/gg_wait_sampling/sql/gg_queries.sql
```diff
@@ -0,0 +1,55 @@
...
+FROM gg_wait_sampling.gg_wait_sampling_get_current(pg_backend_pid());
+-- A non-existent pid must raise an error.
+SELECT gg_wait_sampling.gg_wait_sampling_get_current(-1);
+
+-- Start with a clean profile on every segment and coordinator.
+SELECT * from gg_wait_sampling.gg_wait_sampling_reset_profile;
+
+-- Test Timeout event at coordinator
+SELECT pg_sleep(6);
+-- The pg_sleep above must have produced a Timeout/pg_sleep sample.
```
RekGRpth: I still don't understand why we need a separate test for a coordinator sleep when the next test below also does a coordinator sleep?

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: I think passing the pid argument into clusterwide functions is redundant. I see no issues in current approach.

RekGRpth: But there is no extra call and this is how it is done in the original extension.

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -816,9 +879,8 @@ pg_wait_sampling_reset_profile(PG_FUNCTION_ARGS)
 
 	check_shmem();
 
-	pgws_init_lock_tag(&queueTag, PGWS_QUEUE_LOCK);
-
-	LockAcquire(&queueTag, ExclusiveLock, false, false);
+	queueLock = &(GetNamedLWLockTranche("gg_wait_sampling"))->lock;
```
RekGRpth: Shouldn't the lock be in shared memory? Look at how it's done in the stat-statement.

bimboterminator1: I am a little bit confused, isn't the pointer to shared lock always the same ?
Secondly, I recall this comment https://github.com/arenadata/gpdb/pull/1079#discussion_r1808503075 
Thirdly, I can revert last changes and call GetNamedLWLockTranche each time the lock is needed

In fact we could put initialization out of !found branch. Initialize it unconditionally.
>Initialize it unconditionally.

Done this way. Nothing criminal here

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: I'll do that only with pid argument removal from signature.

RekGRpth: compare such plans
your
```sql
explain (verbose, costs off)
select * from gg_wait_sampling_current;
                                                                                                                                                                                                           QUERY PLAN                                                                                                                                      
                                                                     
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------
 Append
   ->  Function Scan on gpadmin.gg_wait_sampling_get_current_coordinator
         Output: gg_wait_sampling_get_current_coordinator.pid, gg_wait_sampling_get_current_coordinator.event_type, gg_wait_sampling_get_current_coordinator.event, gg_wait_sampling_get_current_coordinator.queryid, gg_wait_sampling_get_current_coordinator.mppsessionid, gg_wait_sampling_get_current_coordinator.command_id, gg_wait_sampling_get_curr
ent_coordinator.tmid, gg_wait_sampling_get_current_coordinator.segid
         Function Call: gg_wait_sampling_get_current_coordinator()
   ->  Gather Motion 3:1  (slice1; segments: 3)
         Output: gg_wait_sampling_get_current_segments.pid, gg_wait_sampling_get_current_segments.event_type, gg_wait_sampling_get_current_segments.event, gg_wait_sampling_get_current_segments.queryid, gg_wait_sampling_get_current_segments.mppsessionid, gg_wait_sampling_get_current_segments.command_id, gg_wait_sampling_get_current_segments.tmid,
 gg_wait_sampling_get_current_segments.segid
         ->  Function Scan on gpadmin.gg_wait_sampling_get_current_segments
               Output: gg_wait_sampling_get_current_segments.pid, gg_wait_sampling_get_current_segments.event_type, gg_wait_sampling_get_current_segments.event, gg_wait_sampling_get_current_segments.queryid, gg_wait_sampling_get_current_segments.mppsessionid, gg_wait_sampling_get_current_segments.command_id, gg_wait_sampling_get_current_segments
.tmid, gg_wait_sampling_get_current_segments.segid
               Function Call: gg_wait_sampling_get_current_segments()
(10 rows)
```
vs mine
```sql
explain (verbose, costs off)
select * from gg_wait_sampling_current;
                                                                                                                                                                                                           QUERY PLAN                                                                                                                                      
                                                                     
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------
 Append
   ->  Function Scan on gpadmin.gg_wait_sampling_get_current_coordinator
         Output: gg_wait_sampling_get_current_coordinator.pid, gg_wait_sampling_get_current_coordinator.event_type, gg_wait_sampling_get_current_coordinator.event, gg_wait_sampling_get_current_coordinator.queryid, gg_wait_sampling_get_current_coordinator.mppsessionid, gg_wait_sampling_get_current_coordinator.command_id, gg_wait_sampling_get_curr
ent_coordinator.tmid, gg_wait_sampling_get_current_coordinator.segid
         Function Call: gg_wait_sampling_get_current_coordinator(NULL::integer)
   ->  Gather Motion 3:1  (slice1; segments: 3)
         Output: gg_wait_sampling_get_current_segments.pid, gg_wait_sampling_get_current_segments.event_type, gg_wait_sampling_get_current_segments.event, gg_wait_sampling_get_current_segments.queryid, gg_wait_sampling_get_current_segments.mppsessionid, gg_wait_sampling_get_current_segments.command_id, gg_wait_sampling_get_current_segments.tmid,
 gg_wait_sampling_get_current_segments.segid
         ->  Function Scan on gpadmin.gg_wait_sampling_get_current_segments
               Output: gg_wait_sampling_get_current_segments.pid, gg_wait_sampling_get_current_segments.event_type, gg_wait_sampling_get_current_segments.event, gg_wait_sampling_get_current_segments.queryid, gg_wait_sampling_get_current_segments.mppsessionid, gg_wait_sampling_get_current_segments.command_id, gg_wait_sampling_get_current_segments
.tmid, gg_wait_sampling_get_current_segments.segid
               Function Call: gg_wait_sampling_get_current_segments(NULL::integer)
(10 rows)
```

RekGRpth: At first glance, the plans are identical. The only difference is that yours calls SQL functions, while mine calls C functions directly.

RekGRpth: That is, you make an extra call to SQL functions.

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -266,20 +285,23 @@ pgws_shmem_startup(void)
...
 	{
 		/* Attach to existing shared objects */
 		toc = shm_toc_attach(PG_WAIT_SAMPLING_MAGIC, pgws);
 		pgws_collector_hdr = shm_toc_lookup(toc, 0, false);
 		pgws_collector_mq = shm_toc_lookup(toc, 1, false);
-		pgws_proc_queryids = shm_toc_lookup(toc, 2, false);
+		pgws_proc_query_items = shm_toc_lookup(toc, 2, false);
 	}
 
+	queueLock = &(GetNamedLWLockTranche("gg_wait_sampling"))->lock;
```
RekGRpth: In `pg_stat_statements`, a similar lock is located in shared memory
https://github.com/GreengageDB/greengage/blob/ff4e72934cc9924e482b6194eedfb0567de18b9c/contrib/pg_stat_statements/pg_stat_statements.c#L456
but in yours it is in a global variable.

RekGRpth:
> LWLock is used instead

Could you please explain in more detail why the original extension had the heavy `LockAcquire` lock, and why it can be replaced with the light `LWLockAcquire` lock?

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: >That is, you make an extra call to SQL functions.

And what? What's wrong with that? I guess an impact is minimal.

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: It works with global variable variant. It's a legit way. What's the point of storing it there?

bimboterminator1:
>had the heavy LockAcquire lock

Advisory lock probably utilizes less cpu.

RekGRpth:
> > had the heavy LockAcquire lock
> 
> Advisory lock probably utilizes less cpu.

Does a heavy lock utilize less cpu than a light one?

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
RekGRpth: You have to perform an extra search each time, whereas with the other option the search is done only once.

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
RekGRpth: Extra SQL function calls do extra work, and will obviously make the query run slower.

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
RekGRpth: yours
```sql
explain (verbose, costs on, analyze on)
select * from gg_wait_sampling_current;
                                                                                                                                                                                                           QUERY PLAN                                                                                                                                                                                                           
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Append  (cost=0.25..80.50 rows=4000 width=92) (actual time=13.023..13.146 rows=34 loops=1)
   ->  Function Scan on gpadmin.gg_wait_sampling_get_current_coordinator  (cost=0.25..10.25 rows=1000 width=92) (actual time=13.020..13.033 rows=10 loops=1)
         Output: gg_wait_sampling_get_current_coordinator.pid, gg_wait_sampling_get_current_coordinator.event_type, gg_wait_sampling_get_current_coordinator.event, gg_wait_sampling_get_current_coordinator.queryid, gg_wait_sampling_get_current_coordinator.mppsessionid, gg_wait_sampling_get_current_coordinator.command_id, gg_wait_sampling_get_current_coordinator.tmid, gg_wait_sampling_get_current_coordinator.segid
         Function Call: gg_wait_sampling_get_current_coordinator()
         work_mem: 18kB  Segments: 1  Max: 18kB (segment -1)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.25..50.25 rows=3000 width=92) (actual time=0.012..0.043 rows=24 loops=1)
         Output: gg_wait_sampling_get_current_segments.pid, gg_wait_sampling_get_current_segments.event_type, gg_wait_sampling_get_current_segments.event, gg_wait_sampling_get_current_segments.queryid, gg_wait_sampling_get_current_segments.mppsessionid, gg_wait_sampling_get_current_segments.command_id, gg_wait_sampling_get_current_segments.tmid, gg_wait_sampling_get_current_segments.segid
         ->  Function Scan on gpadmin.gg_wait_sampling_get_current_segments  (cost=0.25..10.25 rows=1000 width=92) (actual time=0.356..0.365 rows=8 loops=1)
               Output: gg_wait_sampling_get_current_segments.pid, gg_wait_sampling_get_current_segments.event_type, gg_wait_sampling_get_current_segments.event, gg_wait_sampling_get_current_segments.queryid, gg_wait_sampling_get_current_segments.mppsessionid, gg_wait_sampling_get_current_segments.command_id, gg_wait_sampling_get_current_segments.tmid, gg_wait_sampling_get_current_segments.segid
               Function Call: gg_wait_sampling_get_current_segments()
               work_mem: 52kB  Segments: 3  Max: 18kB (segment 0)
 Planning Time: 10.773 ms
   (slice0)    Executor memory: 10373K bytes.  Work_mem: 18K bytes max.
   (slice1)    Executor memory: 58K bytes avg x 3 workers, 58K bytes max (seg0).  Work_mem: 18K bytes max.
 Memory used:  128000kB
 Execution Time: 13.552 ms
(17 rows)
```
vs mine
```sql
explain (verbose, costs on, analyze on)
select * from gg_wait_sampling_current;
                                                                                                                                                                                                           QUERY PLAN                                                                                                                                                                                                           
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Append  (cost=0.00..80.00 rows=4000 width=92) (actual time=0.038..0.418 rows=34 loops=1)
   ->  Function Scan on gpadmin.gg_wait_sampling_get_current_coordinator  (cost=0.00..10.00 rows=1000 width=92) (actual time=0.036..0.047 rows=10 loops=1)
         Output: gg_wait_sampling_get_current_coordinator.pid, gg_wait_sampling_get_current_coordinator.event_type, gg_wait_sampling_get_current_coordinator.event, gg_wait_sampling_get_current_coordinator.queryid, gg_wait_sampling_get_current_coordinator.mppsessionid, gg_wait_sampling_get_current_coordinator.command_id, gg_wait_sampling_get_current_coordinator.tmid, gg_wait_sampling_get_current_coordinator.segid
         Function Call: gg_wait_sampling_get_current_coordinator(NULL::integer)
         work_mem: 18kB  Segments: 1  Max: 18kB (segment -1)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..50.00 rows=3000 width=92) (actual time=0.270..0.299 rows=24 loops=1)
         Output: gg_wait_sampling_get_current_segments.pid, gg_wait_sampling_get_current_segments.event_type, gg_wait_sampling_get_current_segments.event, gg_wait_sampling_get_current_segments.queryid, gg_wait_sampling_get_current_segments.mppsessionid, gg_wait_sampling_get_current_segments.command_id, gg_wait_sampling_get_current_segments.tmid, gg_wait_sampling_get_current_segments.segid
         ->  Function Scan on gpadmin.gg_wait_sampling_get_current_segments  (cost=0.00..10.00 rows=1000 width=92) (actual time=0.067..0.076 rows=8 loops=1)
               Output: gg_wait_sampling_get_current_segments.pid, gg_wait_sampling_get_current_segments.event_type, gg_wait_sampling_get_current_segments.event, gg_wait_sampling_get_current_segments.queryid, gg_wait_sampling_get_current_segments.mppsessionid, gg_wait_sampling_get_current_segments.command_id, gg_wait_sampling_get_current_segments.tmid, gg_wait_sampling_get_current_segments.segid
               Function Call: gg_wait_sampling_get_current_segments(NULL::integer)
               work_mem: 52kB  Segments: 3  Max: 18kB (segment 0)
 Planning Time: 12.582 ms
   (slice0)    Executor memory: 60K bytes.  Work_mem: 18K bytes max.
   (slice1)    Executor memory: 63K bytes avg x 3 workers, 63K bytes max (seg0).  Work_mem: 18K bytes max.
 Memory used:  128000kB
 Execution Time: 0.912 ms
(17 rows)
```

RekGRpth: `Execution Time: 13.552 ms` vs `Execution Time: 0.912 ms`

bimboterminator1: >  extra search each time

When exactly? If common backend is forked the `shmem_startup_hook` is not called

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: Ok, I'll change it

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
RekGRpth: https://github.com/GreengageDB/greengage/blob/ff4e72934cc9924e482b6194eedfb0567de18b9c/src/backend/storage/ipc/ipci.c#L109-L110

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: Done

bimboterminator1:
Returned advisory lock. Review is not blokcked on core chages. Only merging

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
RekGRpth: I meant all functions and views, not just these.

bimboterminator1: Done

RekGRpth:
> Instead of
> vanila queryId entity the WSQueryId type is introduced.

actuallize commit/PR descr

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
```diff
@@ -0,0 +1,160 @@
...
+
+DO $$
+BEGIN
+	IF NOT EXISTS (SELECT 1 FROM pg_catalog.pg_namespace WHERE nspname = 'gg_wait_sampling')
+	THEN
+		CREATE SCHEMA gg_wait_sampling;
+	END IF;
+END$$;
+
+CREATE FUNCTION gg_wait_sampling.gg_wait_sampling_get_current_segments (
```
RekGRpth: Do we still need fqn "gg_wait_sampling."?

gpcontrib/gg_wait_sampling/gg_wait_sampling.control
```diff
@@ -0,0 +1,5 @@
+# gg_wait_sampling extension
+comment = 'sampling based statistics of wait events'
+default_version = '1.1'
+module_pathname = '$libdir/gg_wait_sampling'
+relocatable = false
```
RekGRpth: Do we still need no-relocatable?

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: I think it'd be better if we have it. We have fully rewritten the sql file anyway.

gpcontrib/gg_wait_sampling/README.md
```diff
@@ -40,144 +39,153 @@ When `pg_wait_sampling` is enabled, it collects two kinds of statistics.
...
-[PostgreSQL](http://www.postgresql.org/about/licence/)
-and supports PostgreSQL 13+.
-
-Installation
-------------
-
-Pre-built `pg_wait_sampling` packages are provided in official PostgreSQL
-repository: https://download.postgresql.org/pub/repos/
+In Greengage, the extension also exposes cluster-wide views built with
+gp_dist_random(...) so you can inspect both coordinator-local and
```
red1452: There is no difference between using `gp_dist_rnadom` and without it. All segments and coordinator are present at the output. Is there function or view at the extension, which may change outpup with `gp_dist_rnadom`?

red1452 (commented):
Typo:
`uint64 values my be represented` -> `uint64 values may be represented`

RekGRpth:
I think we need to use `LWLockAcquire(AddinShmemInitLock`, `PGWS_QUEUE_LOCK_NAME` and `PGWS_COLLECTOR_LOCK_NAME` from PR https://github.com/postgrespro/pg_wait_sampling/pull/111.

gpcontrib/gg_wait_sampling/README.md
bimboterminator1: Updated the readme.

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
RekGRpth:
```suggestion
 *	  gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
```diff
@@ -990,19 +1043,24 @@ pgws_planner_hook(Query *parse,
 									  , es
 #endif
 									  );
+		/*
+		 * Set queryId int PlannedStmt. GPORCA currently does not copy
```
RekGRpth: may be
```suggestion
		 * Set queryId in PlannedStmt. GPORCA currently does not copy
```

RekGRpth: Why
> GPORCA currently does not copy queryId to PlannedStmt

? Maybe it's a bug and needs to be fixed?

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
RekGRpth: Still, I suggest removing "pg_wait_sampling."

RekGRpth:
Why did resource group tests fail?

gpcontrib/gg_wait_sampling/pg_wait_sampling.c
bimboterminator1: queryId is not used/generated during normal operartion. Extensions now set it themselves.

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: What? Where?  "gg_wait_sampling."?

RekGRpth: yes, remove "gg_wait_sampling."

RekGRpth (dismissed):

red1452 (dismissed):

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
```diff
@@ -0,0 +1,152 @@
+/* gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql */
+
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "CREATE EXTENSION gg_wait_sampling" to load this file. \quit
+
+CREATE FUNCTION gg_wait_sampling_get_current_segments (
```
RekGRpth, line 6: Please add the following functions for ease of calling from the ADCC (on the coordinator and on the segments).
```sql
CREATE FUNCTION gg_wait_sampling_get_current (
	pid int4,
	OUT pid int4,
	OUT event_type text,
	OUT event text,
	OUT queryid int8,
	OUT mppsessionid int4,
	OUT command_id int4,
	OUT tmid int4,
	OUT segid int4
)
RETURNS SETOF record
AS '$libdir/gg_wait_sampling', 'pg_wait_sampling_get_current'
LANGUAGE C VOLATILE CALLED ON NULL INPUT;

CREATE FUNCTION gg_wait_sampling_get_history (
	OUT pid int4,
	OUT ts timestamptz,
	OUT event_type text,
	OUT event text,
	OUT queryid int8,
	OUT mppsessionid int4,
	OUT command_id int4,
	OUT tmid int4,
	OUT segid int4
)
RETURNS SETOF record
AS '$libdir/gg_wait_sampling', 'pg_wait_sampling_get_history'
LANGUAGE C VOLATILE STRICT;

CREATE FUNCTION gg_wait_sampling_get_profile (
	OUT pid int4,
	OUT event_type text,
	OUT event text,
	OUT queryid int8,
	OUT count int8,
	OUT mppsessionid int4,
	OUT command_id int4,
	OUT tmid int4,
	OUT segid int4
)
RETURNS SETOF record
AS '$libdir/gg_wait_sampling', 'pg_wait_sampling_get_profile'
LANGUAGE C VOLATILE STRICT;

CREATE FUNCTION gg_wait_sampling_reset_profile()
RETURNS SETOF BOOL
AS '$libdir/gg_wait_sampling', 'pg_wait_sampling_reset_profile'
LANGUAGE C VOLATILE STRICT;
```

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
```diff
@@ -0,0 +1,152 @@
...
+	OUT segid int4
+)
+RETURNS SETOF record
+AS '$libdir/gg_wait_sampling', 'pg_wait_sampling_get_current'
+LANGUAGE C VOLATILE CALLED ON NULL INPUT EXECUTE ON ALL SEGMENTS;
+
+GRANT EXECUTE ON FUNCTION gg_wait_sampling_get_current_segments TO PUBLIC;
+
+CREATE FUNCTION gg_wait_sampling_get_current_coordinator (
+	pid int4,
```
RekGRpth: why not
```suggestion
	pid int4 DEFAULT NULL,
```

RekGRpth:
Upstream has properly [reworked](https://github.com/postgrespro/pg_wait_sampling/commit/89dda551bebb6b326d1d8adbb7c0a0b1438ffc2c) the light blocks, I think we should grab this for ourselves, especially since we haven't merged the patch yet.

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: Added everything besides the gg_wait_sampling_reset_profile() since user may confuse it with a view. I guess it's ok

gpcontrib/gg_wait_sampling/gg_wait_sampling--1.1.sql
bimboterminator1: added

gpcontrib/gg_wait_sampling/sql/queries.sql
```diff
@@ -1,18 +1,22 @@
...
+WITH t as (SELECT sum(0) FROM gg_wait_sampling_profile)
 	SELECT sum(0) FROM generate_series(1, 2), t;
 
 -- Some dummy checks just to be sure that all our functions work and return something.
-SELECT count(*) = 1 as test FROM pg_wait_sampling_get_current(pg_backend_pid());
-SELECT count(*) >= 0 as test FROM pg_wait_sampling_get_profile();
-SELECT count(*) >= 0 as test FROM pg_wait_sampling_get_history();
-SELECT pg_wait_sampling_reset_profile();
+SELECT count(*) = 1 as test FROM gg_wait_sampling_get_current_coordinator(NULL::INTEGER);
+SELECT count(*) = 1 as test FROM gg_wait_sampling_get_current_segments(NULL::INTEGER);
```
RekGRpth:
```suggestion
SELECT count(*) = 1 as test FROM gg_wait_sampling_get_current_coordinator();
SELECT count(*) = 1 as test FROM gg_wait_sampling_get_current_segments();
```

gpcontrib/gg_wait_sampling/isolation2/sampling_schedule
```diff
@@ -0,0 +1,2 @@
+test: lock_wait_events
+test: teardown
```
RekGRpth: add empty line

red1452 (approved):

RekGRpth:
new flaky?
```diff
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/isolation2/expected/vacuum_progress_column.out /home/gpadmin/gpdb_src/src/test/isolation2/results/vacuum_progress_column.out
--- /home/gpadmin/gpdb_src/src/test/isolation2/expected/vacuum_progress_column.out	2026-05-19 09:46:17.323430334 +0000
+++ /home/gpadmin/gpdb_src/src/test/isolation2/results/vacuum_progress_column.out	2026-05-19 09:46:17.335430618 +0000
@@ -417,14 +417,14 @@
 select gp_segment_id, relid::regclass as relname, phase, heap_blks_total, heap_blks_scanned, heap_blks_vacuumed, index_vacuum_count, max_dead_tuples, num_dead_tuples from gp_stat_progress_vacuum where gp_segment_id > -1;
  gp_segment_id | relname                   | phase                         | heap_blks_total | heap_blks_scanned | heap_blks_vacuumed | index_vacuum_count | max_dead_tuples | num_dead_tuples 
 ---------------+---------------------------+-------------------------------+-----------------+-------------------+--------------------+--------------------+-----------------+-----------------
- 0             | vacuum_progress_ao_column | append-optimized post-cleanup | 0               | 0                 | 9                  | 2                  | 0               | 0               
- 1             | vacuum_progress_ao_column | append-optimized post-cleanup | 0               | 0                 | 9                  | 2                  | 0               | 0               
- 2             | vacuum_progress_ao_column | append-optimized post-cleanup | 0               | 0                 | 9                  | 2                  | 0               | 0               
+ 0             | vacuum_progress_ao_column | append-optimized post-cleanup | 0               | 0                 | 0                  | 0                  | 0               | 0               
+ 1             | vacuum_progress_ao_column | append-optimized post-cleanup | 0               | 0                 | 0                  | 0                  | 0               | 0               
+ 2             | vacuum_progress_ao_column | append-optimized post-cleanup | 0               | 0                 | 0                  | 0                  | 0               | 0               

real	139m27.730s
user	8m8.679s
sys	4m5.075s
+ exit_code=1
+ echo 'Script gpdb_src/concourse/scripts/ic_gpdb.bash finished with 1'
 (3 rows)
 select relid::regclass as relname, phase, heap_blks_total, heap_blks_scanned, heap_blks_vacuumed, index_vacuum_count, max_dead_tuples, num_dead_tuples from gp_stat_progress_vacuum_summary;
  relname                   | phase                         | heap_blks_total | heap_blks_scanned | heap_blks_vacuumed | index_vacuum_count | max_dead_tuples | num_dead_tuples 
 ---------------------------+-------------------------------+-----------------+-------------------+--------------------+--------------------+-----------------+-----------------
- vacuum_progress_ao_column | append-optimized post-cleanup | 0               | 0                 | 27                 | 6                  | 0               | 0               
+ vacuum_progress_ao_column | append-optimized post-cleanup | 0               | 0                 | 0                  | 0                  | 0               | 0               
 (1 row)
 
 2: SELECT gp_inject_fault('vacuum_ao_post_cleanup_end', 'reset', dbid) FROM gp_segment_configuration WHERE content > -1 AND role = 'p';
```

bimboterminator1:
Seems so, updated the wiki list and created a corresponding task

RekGRpth (approved):

PR #318, opened 2026-03-13
GG-283: Fix distribution by resjunk column (6.x)

GirgerS:
In the PostgreSQL planner, some queries require auxiliary resjunk columns in
the targetlist.

These columns are necessary for query execution but do not appear in the
resulting relation.

Our logic for CTAS expressions, however, did not account this case.
If a query didn't contain an explicit DISTRIBUTED BY clause,
and other columns in the targetlist were not hashable,
the planner could use resjunk column in the distribution policy,
resulting in a segmentation fault.

To fix this issue, simply filter out resjunk columns.

Also, make attribute selection more efficient, as the targetlist
was re-examined for each entry.

Ticket: GG-283


GirgerS:
Also, the message we get when there is no column to distribute by is pretty cryptic:
```
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column(s) named '' as the Greengage Database data distribution key for this table.
```

The logic in orca is different, but the message is much better:
```
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause. Creating a NULL policy entry.
```

Maybe in such case we want so say something like:
```
NOTICE:  Could not find a suitable column for distribution. Creating a NULL policy entry.
```

src/backend/cdb/cdbmutate.c
```diff
@@ -493,9 +493,14 @@ apply_motion(PlannerInfo *root, Plan *plan, Query *query)
 
 							if (target)
```
silent-observer: Why not just

```suggestion
							if (target && !target->resjunk)
```

GirgerS: Well, I personally prefer to separate such filtering conditions, so it is immediately clear which one of them is triggered during debugging. But the way how existing code is organized, yeah, your variant is better, it just didn't occur to me.

bandetto:
Please wrap the description by 80 characters :^)

src/backend/cdb/cdbmutate.c
```diff
@@ -491,7 +491,7 @@ apply_motion(PlannerInfo *root, Plan *plan, Query *query)
 							TargetEntry *target =
 							get_tle_by_resno(plan->targetlist, i + 1);
 
-							if (target)
```
bandetto, line 494: Could there be possibly similar errors for other commands?

bandetto:
Looking at the ticket, the customer already has tables distributed by a junk column. Does this patch account for the mitigation of this problem for tables already affected by the issue? How would we fix this on the customer cluser after the patch?

src/backend/cdb/cdbmutate.c
GirgerS: I don't think it is possible to distribute by resjunk in any other case because we need to create a table, and then fill it with values from a query that needs resjunk columns. And I think that only CTAS can do these two things at the same time.

Other similar cases that I can think of work fine (though I didn't dive deep into the source code):
```
postgres=# CREATE TABLE alter_test AS (SELECT 10::int4 AS a, 11::int4 AS b) DISTRIBUTED BY (a);
SELECT 1
postgres=# ALTER TABLE alter_test SET DISTRIBUTED BY (b);
ALTER TABLE
postgres=# ALTER TABLE alter_test SET DISTRIBUTED BY (tableoid);
ERROR:  cannot distribute by system column "tableoid"
postgres=# ALTER TABLE alter_test ALTER b SET DATA TYPE point USING (point(b, b));
ERROR:  cannot alter type of a column used in a distribution policy
postgres=# CREATE TABLE create_test (a int4) DISTRIBUTED BY (tableoid);
ERROR:  column "tableoid" named in 'DISTRIBUTED BY' clause does not exist
```

If you have any other ideas, let me know, it is always good to test more things.

GirgerS:
> Please wrap the description by 80 characters :^)

The classic. Done.

GirgerS:
> Looking at the ticket, the customer already has tables distributed by a junk column. Does this patch account for the mitigation of this problem for tables already affected by the issue? How would we fix this on the customer cluser after the patch?

Great question, I didn't think about it.

I'm not sure if it is necessary, because interacting with such tables would immediately result in a segmentation fault. So unless it was a one-of-a-kind table, it shouldn't be hard to just recreate it.

But maybe we still want to try to migrate such tables, just to be completely sure. I would guess that it is enough to 
compare distribution policy from `gp_distribution_policy` with the actual columns of the table, and then reset distribution keys if necessary.

GirgerS (commented):
Just a note to myself.

What does the line `based on this, we were able to get a reproduction - the problem of DXL translation when working with the json data type.` in the parent ticket mean?

I didn't pay attention to it initially, and now I'm confused.

Edit: they thought that the problem is in the `SELECT` query, based on the given call stack. And it indeed fails in ORCA, even though it is not the root cause of the problem.

bandetto:
> just to be completely sure

LGTM. But I'll wait for an definitive answer on how would we fix this for existing tables.

GirgerS:
> > just to be completely sure
> 
> LGTM. But I'll wait for an definitive answer on how would we fix this for existing tables.

I have outlined the steps necessary to fix the issue in the parent ticket.

bandetto (dismissed):

silent-observer (dismissed):

RekGRpth (approved):

silent-observer (approved):

PR #319, opened 2026-03-13
GG-285: Fix SIGSEGV due to subqueries flow temporary nullification

ShapatinAA:
Fix SIGSEGV due to subqueries flow temporary nullification

When subquery gets omitted in `set_subqueryscan_references()`, via
`trivial_subqueryscan()` there's a chance that previously we inserted `Result`
node via `cdb_insert_result_node()`, which may lead to nullification of flow in
firstly `SubqueryScan` node and secondly in node, which gets pulled up in place
of `SubqueryScan`. This happens because of next code part:
```c
/* Honor the flow of the SubqueryScan, by copying it to the subplan. */
		result->flow = plan->scan.plan.flow;
```
Here we assign previously nullified flow to pulled up node. 

So to fix this issue we should assign proper flow to the pulled up node in case
of `SubqueryScan` omition.

Co-authored-by: Georgy Shelkovy <g.shelkovy@arenadata.io>

### DIFF
--- a/src/backend/optimizer/plan/setrefs.c
+++ b/src/backend/optimizer/plan/setrefs.c
@@ -2953,5 +2953,7 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
 		plan->flow = flow;
 	}
 
+	AssertImply(flow && resultplan->lefttree, resultplan->lefttree->flow);
+
     return resultplan;
 }                               /* cdb_insert_result_node */
--- a/src/backend/optimizer/plan/setrefs.c
+++ b/src/backend/optimizer/plan/setrefs.c
@@ -2938,9 +2938,20 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
     /* Fix up the Result node and the Plan tree below it. */
     resultplan = set_plan_refs(root, resultplan, rtoffset);
 
-    /* Reattach the Flow node. */
     resultplan->flow = flow;
-	plan->flow = flow;
+	/* In case we ommited SubqueryScan node in set_subqueryscan_references()
+	 * we need to reatach flow to new outerplan node, because previously we
+	 * copied null into it. Otherwise reatach flow to previous node.
+	 */
+	if (resultplan->lefttree && IsA(plan, SubqueryScan) &&
+		nodeTag(resultplan->lefttree) != nodeTag(plan))
+	{
+		resultplan->lefttree->flow = flow;
+	}
+	else
+	{
+		plan->flow = flow;
+	}
 
     return resultplan;
 }                               /* cdb_insert_result_node */
--- a/src/backend/optimizer/plan/setrefs.c
+++ b/src/backend/optimizer/plan/setrefs.c
@@ -2944,7 +2944,7 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
 	 * copied null into it. Otherwise reatach flow to previous node.
 	 */
 	if (resultplan->lefttree && IsA(plan, SubqueryScan) &&
-		nodeTag(resultplan->lefttree) != nodeTag(plan))
+		resultplan->lefttree != plan)
 	{
 		resultplan->lefttree->flow = flow;
 	}
--- a/src/backend/optimizer/plan/setrefs.c
+++ b/src/backend/optimizer/plan/setrefs.c
@@ -2939,7 +2939,8 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
     resultplan = set_plan_refs(root, resultplan, rtoffset);
 
     resultplan->flow = flow;
-	/* In case we ommited SubqueryScan node in set_subqueryscan_references()
+	/*
+	 * In case we ommited SubqueryScan node in set_subqueryscan_references()
 	 * we need to reatach flow to new outerplan node, because previously we
 	 * copied null into it. Otherwise reatach flow to previous node.
 	 */
--- a/src/test/regress/sql/bfv_subquery.sql
+++ b/src/test/regress/sql/bfv_subquery.sql
@@ -295,20 +295,13 @@ DROP TABLE A;
 --
 -- Test if we reattach flow of omitted SubqueryScan to outerplan.
 --
--- start_ignore
-DROP TABLE IF EXISTS T;
-CREATE TABLE T (I INT) DISTRIBUTED BY (I);
--- end_ignore
 
+EXPLAIN (VERBOSE, COSTS OFF)
 WITH A AS (
-    SELECT I FROM T
+    SELECT * FROM UNNEST(ARRAY[0]) I
 ), B AS (
-    SELECT I, UNNEST(ARRAY[0]) FROM A
-), C AS (
-    SELECT I FROM B
-) SELECT I FROM C JOIN T USING (I) WHERE T.I IN (SELECT T.I FROM C);
-
-DROP TABLE T;
+    SELECT *, UNNEST(ARRAY[1]) FROM A
+) SELECT * FROM A WHERE I IN (SELECT A.I FROM B);
 
 --
 -- Test the ctid in Function and Values Scans
--- a/src/test/regress/sql/bfv_subquery.sql
+++ b/src/test/regress/sql/bfv_subquery.sql
@@ -293,6 +293,24 @@ EXPLAIN SELECT (EXISTS (SELECT UNNEST(X))) AS B FROM A;
 DROP TABLE A;
 
 --
+-- Test if we reattach flow of omitted SubqueryScan to outerplan.
+--
+-- start_ignore
+DROP TABLE IF EXISTS T;
+CREATE TABLE T (I INT) DISTRIBUTED BY (I);
+-- end_ignore
+
+WITH A AS (
+    SELECT I FROM T
+), B AS (
+    SELECT I, UNNEST(ARRAY[0]) FROM A
+), C AS (
+    SELECT I FROM B
+) SELECT I FROM C JOIN T USING (I) WHERE T.I IN (SELECT T.I FROM C);
+
+DROP TABLE T;
+
+--
 -- Test the ctid in Function and Values Scans
 --
 
--- a/src/test/regress/expected/bfv_subquery.out
+++ b/src/test/regress/expected/bfv_subquery.out
@@ -500,6 +500,21 @@ EXPLAIN SELECT (EXISTS (SELECT UNNEST(X))) AS B FROM A;
 
 DROP TABLE A;
 --
+-- Test if we reattach flow of omitted SubqueryScan to outerplan.
+--
+WITH A AS (
+    SELECT I FROM T
+), B AS (
+    SELECT I, UNNEST(ARRAY[0]) FROM A
+), C AS (
+    SELECT I FROM B
+) SELECT I FROM C JOIN T USING (I) WHERE T.I IN (SELECT T.I FROM C);
+ i 
+---
+(0 rows)
+
+DROP TABLE
+--
 -- Test the ctid in Function and Values Scans
 --
 create table t1(a int) ;
--- a/src/test/regress/expected/bfv_subquery.out
+++ b/src/test/regress/expected/bfv_subquery.out
@@ -504,50 +504,30 @@ DROP TABLE A;
 --
 EXPLAIN (VERBOSE, COSTS OFF)
 WITH A AS (
-    SELECT I FROM T
+    SELECT * FROM UNNEST(ARRAY[0]) I
 ), B AS (
-    SELECT I, UNNEST(ARRAY[0]) FROM A
-), C AS (
-    SELECT I FROM B
-) SELECT I FROM C JOIN T USING (I) WHERE T.I IN (SELECT T.I FROM C);
-                                          QUERY PLAN                                           
------------------------------------------------------------------------------------------------
- Gather Motion 3:1  (slice2; segments: 3)
-   Output: b.i
-   ->  Hash Join
-         Output: b.i
-         Hash Cond: (t.i = b.i)
-         ->  Seq Scan on public.t
-               Output: t.i
-               Filter: (SubPlan 1)
-               SubPlan 1  (slice2; segments: 3)
-                 ->  Subquery Scan on c
-                       Output: t.i
-                       ->  Subquery Scan on b_1
-                             Output: b_1.i
-                             ->  Result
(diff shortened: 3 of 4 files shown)


src/test/regress/sql/bfv_subquery.sql
```diff
@@ -292,6 +292,24 @@ EXPLAIN SELECT (EXISTS (SELECT UNNEST(X))) AS B FROM A;
...
 
+--
+-- Test if we reattach flow of omitted SubqueryScan to outerplan.
+--
+-- start_ignore
+DROP TABLE IF EXISTS T;
+CREATE TABLE T (I INT) DISTRIBUTED BY (I);
+-- end_ignore
+
+WITH A AS (
```
RekGRpth, line 300: Since the patch fixes the planner, it makes more sense to test the plan rather than the query execution:
```suggestion
explain (verbose, costs off)
WITH A AS (
```

ShapatinAA: I thought that in this case the fact that the query executes - matters most, and the way how it does it - not. Moreover if we will fix plan right now, in future it might be changed due to some optimization patches and we will add work to future committers to resolve if plan became wrong or not (when actually we don't fix it's behavior).

RekGRpth: > in future it might be changed due to some optimization patches

This is exactly what is required from fixing the plan in the patch, look at other plans.

ShapatinAA: Okay I see now. Test updated.

src/test/regress/sql/bfv_subquery.sql
```diff
@@ -292,6 +292,24 @@ EXPLAIN SELECT (EXISTS (SELECT UNNEST(X))) AS B FROM A;
...
+    SELECT I FROM T
```
RekGRpth: Replace all `SELECT I` with `SELECT *`, because this is another bug where the planner doesn't omit things that aren't used.

src/test/regress/sql/bfv_subquery.sql
```diff
@@ -292,6 +292,24 @@ EXPLAIN SELECT (EXISTS (SELECT UNNEST(X))) AS B FROM A;
...
+), B AS (
+    SELECT I, UNNEST(ARRAY[0]) FROM A
+), C AS (
+    SELECT I FROM B
+) SELECT I FROM C JOIN T USING (I) WHERE T.I IN (SELECT T.I FROM C);
```
RekGRpth: simplify repro
```suggestion
explain (verbose, costs off)
with a as (
    select * from t
), b as (
    select *, unnest(array[0]) from a
) select * from t where t.i in (select t.i from b)
```

RekGRpth: or even more
```suggestion
explain (verbose, costs off)
with a as (
    select * from unnest(array[0]) i
), b as (
    select *, unnest(array[1]) from a
) select * from a where i in (select a.i from b)
```

src/test/regress/sql/bfv_subquery.sql
ShapatinAA: As I understand this is not a bug, but an expected behavior as we have SRF in select section. So if we omit this function we might get lesser number of returned lines.

src/test/regress/sql/bfv_subquery.sql
ShapatinAA: Thank you for finding this out! Updated.

src/backend/optimizer/plan/setrefs.c
```diff
@@ -2938,9 +2938,20 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
...
 
-    /* Reattach the Flow node. */
     resultplan->flow = flow;
-	plan->flow = flow;
+	/* In case we ommited SubqueryScan node in set_subqueryscan_references()
+	 * we need to reatach flow to new outerplan node, because previously we
+	 * copied null into it. Otherwise reatach flow to previous node.
+	 */
+	if (resultplan->lefttree && IsA(plan, SubqueryScan) &&
+		nodeTag(resultplan->lefttree) != nodeTag(plan))
```
red1452: Why do you not compare simple pointers? `resultplan->lefttree != plan`? Is it correct, if nodeTag are same, but it is different nodes?

red1452 (commented):
At description:
> This does happen because

It looks incorrect. There are two variants:
> This happens because
>This does not happen because

src/backend/optimizer/plan/setrefs.c
RekGRpth: Actually, initially I proposed a simpler option.
```diff
diff --git a/src/backend/optimizer/plan/setrefs.c b/src/backend/optimizer/plan/setrefs.c
index 40a086257cf..87bbcf1483b 100644
--- a/src/backend/optimizer/plan/setrefs.c
+++ b/src/backend/optimizer/plan/setrefs.c
@@ -2941,6 +2941,8 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
     /* Reattach the Flow node. */
     resultplan->flow = flow;
 	plan->flow = flow;
+	if (resultplan->lefttree && !resultplan->lefttree->flow)
+		resultplan->lefttree->flow = flow;
 
     return resultplan;
 }                               /* cdb_insert_result_node */                       /* cdb_insert_result_node */
```

ShapatinAA: >Actually, initially I proposed a simpler option.

I made condition more complex to cut out other possibilities as we can apply this code for variety of nodes (we can have `T_IndexScan`, `T_BitmapHeapScan` and so on), but in this patch we're interested only in one case: result+subqueryscan and the case when we're omitting subqueryscan.

ShapatinAA: >Why do you not compare simple pointers? resultplan->lefttree != plan? Is it correct, if nodeTag are same, but it is different nodes?

Yep, you right. It will be more correct way as we're interested in comparison of identities.

RekGRpth: I don't think it will be good if we nullify the flow for other types of nodes.

src/backend/optimizer/plan/setrefs.c
```diff
@@ -2944,7 +2944,7 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
 	 * copied null into it. Otherwise reatach flow to previous node.
 	 */
 	if (resultplan->lefttree && IsA(plan, SubqueryScan) &&
-		nodeTag(resultplan->lefttree) != nodeTag(plan))
+		resultplan->lefttree != plan)
```
RekGRpth: But isn't this condition now unnecessary?
`resultplan->lefttree`

src/backend/optimizer/plan/setrefs.c
ShapatinAA: This is why I kept `IsA(plan, SubqueryScan)` in this condition. As only stepping here from this part is matters
https://github.com/GreengageDB/greengage/blob/a02fcaf6603ff68a68c1239f23b2c0f28e57b0fc/src/backend/optimizer/plan/setrefs.c#L739-L740

ShapatinAA: Ah I got what you meant. Current logic doesn't nullifies flows for other types of plans, as only there we do this
https://github.com/GreengageDB/greengage/blob/a02fcaf6603ff68a68c1239f23b2c0f28e57b0fc/src/backend/optimizer/plan/setrefs.c#L1351-L1352
This happens only when we enter this `set_plan_refs` with SubqueryScanNode.

ShapatinAA: So other plans doesn't affected by this nullification in current logic.

RekGRpth: Then add an assert to check this.

ShapatinAA:
> At description:
> 
> > This does happen because
> 
> It looks incorrect. There are two variants:
> 
> > This happens because
> > This does not happen because

Updated the description (if I got you correctly). Used `This happens because`

src/backend/optimizer/plan/setrefs.c
ShapatinAA: What exactly? `IsA(plan, SubqueryScan)`?

RekGRpth:
```c
AssertImply(resultplan->lefttree, resultplan->lefttree->flow);
```

ShapatinAA: Are we should have `resultplan->lefttree->flow` as non null in any possible case?

RekGRpth: Well, check it out.

ShapatinAA: Maybe then we shouldn't put assert like that one? If we're not sure if flow shouldn't be null at the end of this function.
For example I found this place, which as I understand may be reached after creation of `top_plan`
https://github.com/GreengageDB/greengage/blob/a02fcaf6603ff68a68c1239f23b2c0f28e57b0fc/src/backend/cdb/cdbllize.c#L800

RekGRpth: I don't quite understand: what does zero flow have to do with it?

ShapatinAA: We're checking that if `resltplan->lefttree` exists then flow of this `lefttree` also exists with next assertion, right? `AssertImply(resultplan->lefttree, resultplan->lefttree->flow);`
And the only place where as this assertion can be placed - is the end.

RekGRpth: yes

ShapatinAA: But if initially this is null
https://github.com/GreengageDB/greengage/blob/a02fcaf6603ff68a68c1239f23b2c0f28e57b0fc/src/backend/optimizer/plan/setrefs.c#L2920

Then we'll be having null also here and assert will fail
https://github.com/GreengageDB/greengage/blob/a02fcaf6603ff68a68c1239f23b2c0f28e57b0fc/src/backend/optimizer/plan/setrefs.c#L2946-L2955

RekGRpth:
```c
AssertImply(flow && resultplan->lefttree, resultplan->lefttree->flow);
```

ShapatinAA: IMHO this assert is kinda obvious maybe even verifiable, but added it anyway!

red1452 (dismissed):

RekGRpth:
> subquerys

sub-queries?

src/backend/optimizer/plan/setrefs.c
```diff
@@ -2938,9 +2938,22 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
     /* Fix up the Result node and the Plan tree below it. */
     resultplan = set_plan_refs(root, resultplan, rtoffset);
 
-    /* Reattach the Flow node. */
     resultplan->flow = flow;
-	plan->flow = flow;
+	/* In case we ommited SubqueryScan node in set_subqueryscan_references()
```
RekGRpth:
```suggestion
	/*
	 * In case we ommited SubqueryScan node in set_subqueryscan_references()
```

src/backend/optimizer/plan/setrefs.c
```diff
@@ -2938,9 +2938,22 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
     /* Fix up the Result node and the Plan tree below it. */
     resultplan = set_plan_refs(root, resultplan, rtoffset);
 
-    /* Reattach the Flow node. */
```
RekGRpth, line 2941: Why was this line removed?

src/backend/optimizer/plan/setrefs.c
ShapatinAA: Thank you for noticing! Missed that...

src/backend/optimizer/plan/setrefs.c
ShapatinAA: As I understand this comment was aimed for second line and I decided to move it to new comment
https://github.com/GreengageDB/greengage/blob/ed11dae619bc4482eb3a16d97b9f1a097d53156d/src/backend/optimizer/plan/setrefs.c#L2943-L2945

src/backend/optimizer/plan/setrefs.c
```diff
@@ -2938,9 +2938,23 @@ cdb_insert_result_node(PlannerInfo *root, Plan *plan, int rtoffset)
...
+	if (resultplan->lefttree && IsA(plan, SubqueryScan) &&
+		resultplan->lefttree != plan)
+	{
+		resultplan->lefttree->flow = flow;
+	}
+	else
+	{
+		plan->flow = flow;
+	}
+
```
RekGRpth, line 2956: how about such solution?
```diff
diff --git a/src/backend/optimizer/plan/setrefs.c b/src/backend/optimizer/plan/setrefs.c
index 40a086257cf..74018c16ae7 100644
--- a/src/backend/optimizer/plan/setrefs.c
+++ b/src/backend/optimizer/plan/setrefs.c
@@ -1348,6 +1348,7 @@ set_subqueryscan_references(PlannerInfo *root,
 			ctle->resorigcol = ptle->resorigcol;
 		}
 
+		if (plan->scan.plan.flow)
 		/* Honor the flow of the SubqueryScan, by copying it to the subplan. */
 		result->flow = plan->scan.plan.flow;
 	}
```

ShapatinAA: Then we would not "honor" the flow of `SubqueryScan` in case of it's omition after insertion of `ResultNode` (that is our case). As it will be left with it's own flow.

RekGRpth: Isn't it's own flow the same one that is assigned to the node at the end of the `cdb_insert_result_node` function in your patch?

ShapatinAA: Nope.
In current version we assign what `SubqueryScan` has to `Result` and to pulled up nodes. As we saved it here https://github.com/GreengageDB/greengage/blob/a02fcaf6603ff68a68c1239f23b2c0f28e57b0fc/src/backend/optimizer/plan/setrefs.c#L2920
In this suggestion (`if (plan->scan.plan.flow)`) in case of omitting `SubqueryScan` after insertion of `Result` node we'll be having null in `plan->scan.plan.flow` and skipping this assignation, so letting pulled up node have it's own flow.

RekGRpth (approved):

red1452 (approved):

PR #320, opened 2026-03-13
CI-5245: Migrate default branch docs before switching

Stolb27:
We are going to update default branch to 7.x. So we should actualize documents that user see when open repo page. The given docs extracted from 6.x as is.

Patch should be merged by rebase.


magf:
Also, may be copy [greengage-sql-dump.yml](https://github.com/GreengageDB/greengage/blob/6.x/.github/workflows/greengage-sql-dump.yml) to 7.x in this patch?

magf:
> Also, may be copy [greengage-sql-dump.yml](https://github.com/GreengageDB/greengage/blob/6.x/.github/workflows/greengage-sql-dump.yml) to 7.x in this patch?

or I'll do it as part of another task where I'm currently working on changing the dump algorithm to include the OS version

Stolb27:
> Also, may be copy [greengage-sql-dump.yml](https://github.com/GreengageDB/greengage/blob/6.x/.github/workflows/greengage-sql-dump.yml) to 7.x in this patch?

@magf, you may push the needed commit to this branch. As far I remember, the branch switching has priority

GirgerS:
`CODE_OF_CONDUCT` and `pull_request_template.md` have different directories, `NOTICE` is quite different but it all looks fine. 

There is also an additional `.mailmap` file, and it seems dated. Maybe we want to remove/modify it?

Stolb27:
> CODE_OF_CONDUCT and pull_request_template.md have different directories, 

Yes. The upstream moved this files years ago. But both locations should be handled by GitHub correctly.

> NOTICE is quite different but it all looks fine.

In my opinion, it's expected because of differences in dependencies (as well in licences of them).

> There is also an additional .mailmap file, and it seems dated. Maybe we want to remove/modify it?

Yeah. We can append our existing and previous teammates here. But in my opinion, it's a separate task. I'll add it to to minor fix list.

GirgerS (approved):

PR #321, opened 2026-03-16
GG-202: Protect replicated tables with volatile functions from rewrite (7.x)

VoidZeroNull0:
What happens?
Distribution policy of replicated tables can be violated via usage of ALTER
TABLE (or CREATE TABLE) and a volatile expression as a default
value.

Why it happens?
No check for this case is made anywhere in the code, thus different data may
appear on segment's table.

How do we fix this?
Main check for expression volatility and table's distribution policy
inside AddRelationNewConstraintsn(), additional ones in
ATExecCookedColumnDefault()  and ATExecSetDistributedBy() are made, to output
an error before table is altered or created.

Tests?
Two tests were developed. One for volatile case, which is prohibited. Another
one for immutable - it works. These cases applied to both ALTER TABLE (ADD
COLUMN, ALTER COLUMN, SET DISTRIBUTED) and CREATE TABLE.


src/backend/commands/tablecmds.c
```diff
@@ -7626,6 +7626,28 @@ ATExecAddColumn(List **wqueue, AlteredTableInfo *tab, Relation rel,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("cannot add column to a partition")));
 
+	/* Protect replicated tables from volatile functions as default value */
+	if (colDef->raw_default != NULL && colDef->raw_default->type == T_FuncCall)
```
RekGRpth:
```suggestion
	if (colDef->raw_default != NULL && colDef->raw_default->type == T_FuncCall && GpPolicyIsReplicated(rel->rd_cdbpolicy))
```

src/backend/commands/tablecmds.c
```diff
@@ -7626,6 +7626,28 @@ ATExecAddColumn(List **wqueue, AlteredTableInfo *tab, Relation rel,
...
+			funcOid = InvalidOid;
+		free_parsestate(pstate);
+
+		if (funcOid != InvalidOid &&
+			func_volatile(funcOid) == PROVOLATILE_VOLATILE && 
+			GpPolicyIsReplicated(rel->rd_cdbpolicy))
+			ereport(ERROR,
+					(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+					 errmsg("tried to add volatile function as default")));		
+	}
```
RekGRpth: why not something like
```c
CatCList   *catlist = SearchSysCacheList1(PROCNAMEARGSNSP, CStringGetDatum(((FuncCall *)colDef->raw_default))->funcname);
for (int i = 0; i < catlist->n_members; i++)
{
	HeapTuple	proctup = &catlist->members[i]->tuple;
	Form_pg_proc procform = (Form_pg_proc) GETSTRUCT(proctup);
	if (procform->provolatile == PROVOLATILE_VOLATILE)
	...
```

VoidZeroNull0: > why not something like

Don't know. Is it faster, because of cache usage? Won't it also return multiple functions with the same name (if they are present)? Gonna test.

RekGRpth: > Is it faster, because of cache usage?

The `func_volatile` function also takes data from the cache, but calling it requires a lot of extra work.

RekGRpth: > Won't it also return multiple functions with the same name (if they are present)?

The `FuncCall` structure contains a list of arguments, and the `PROCNAMEARGSNSP` cache can be searched based on the function's arguments, and a function is uniquely identified by its (full, including schema) name and argument types.

RekGRpth:
With the patch, it is still possible to CREATE a replicated table with a volatile function as the default column value.
```sql
create table t (i int, t text default random()) distributed replicated;
CREATE TABLE
alter table t add column d text default random();
ERROR:  tried to add volatile function as default
insert into t (i) select i from generate_series(1, 10) i;
INSERT 0 10
select * from t;
 i  |          t           
----+----------------------
  1 | 0.690699301056334
  2 | 0.674519328932849
  3 | 0.20110692446836254
  4 | 0.5131376430623185
  5 | 0.012201621028303578
  6 | 0.8603214449746339
  7 | 0.7699967914049921
  8 | 0.6751674667284924
  9 | 0.4432294718887775
 10 | 0.8581767201296664
(10 rows)
```

src/backend/commands/tablecmds.c
```diff
@@ -7626,6 +7626,28 @@ ATExecAddColumn(List **wqueue, AlteredTableInfo *tab, Relation rel,
...
+		else
+			funcOid = InvalidOid;
+		free_parsestate(pstate);
+
+		if (funcOid != InvalidOid &&
+			func_volatile(funcOid) == PROVOLATILE_VOLATILE && 
+			GpPolicyIsReplicated(rel->rd_cdbpolicy))
+			ereport(ERROR,
+					(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+					 errmsg("tried to add volatile function as default")));
```
RekGRpth: The error message isn't very informative. Maybe something like "volatile functions are not supported as default values ​​for columns in replicated tables."

src/backend/commands/tablecmds.c
VoidZeroNull0: Done.

src/backend/commands/tablecmds.c
VoidZeroNull0: Rewritten to use `SearchSysCacheList1()`.

src/backend/commands/tablecmds.c
VoidZeroNull0: Made it:
`volatile expressions are not supported as default values ​for columns in replicated tables`

VoidZeroNull0:
> With the patch, it is still possible to CREATE a replicated table with a volatile function as the default column value.

Fixed by adding check inside `DefineRelation()`. Relevant tests are also added.

RekGRpth:
> Distribution policy of replicated tables can be violated via usage of ALTER
> TABLE ADD COLUMN (or CREATE TABLE) and a volatile expression as a default
> value.

Is it possible to add functionality to `gpcheckcat` to correct such bad tables that already exist for users?

src/backend/commands/tablecmds.c
```diff
@@ -978,6 +980,13 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
 
 			Assert(colDef->cooked_default == NULL);
 
+			/* Protect replicated tables from volatile expressions as default value */
+			if (GpPolicyIsReplicated(policy) && volatility_walker(colDef->raw_default))
```
RekGRpth: Won't the function `contain_volatile_functions` or `contain_volatile_functions_not_nextval` help?

bimboterminator1: Yep, Is there a way to avoid writing a separate walker?

VoidZeroNull0: Rewritten to use `contain_volatile_functions()` in cd849d2. Although, some work has to be done - it cannot process raw expressions (e.g. `T_A_Expr` and `T_FuncCall`), so `TransformExpr()` is needed.

VoidZeroNull0:
> Is it possible to add functionality to `gpcheckcat` to correct such bad tables that already exist for users?

I think it is possible, but has its own problems. I was able to search for replicated tables with volatile expressions using sql, but how do we synchronize data among segments? Only thing that comes to my mind is `COPY`ing data from _reference_ segment to the file, truncating whole table and then again `COPY`ing, but now from the file to the table.  Anyway, I think this addition worth its own small ticket.

bimboterminator1 (dismissed):

src/backend/commands/tablecmds.c
```diff
@@ -85,6 +85,7 @@
 #include "nodes/nodeFuncs.h"
 #include "nodes/parsenodes.h"
 #include "optimizer/optimizer.h"
+#include "optimizer/clauses.h"
```
RekGRpth:
```suggestion
#include "optimizer/clauses.h"
#include "optimizer/optimizer.h"
```

src/backend/optimizer/util/clauses.c
```diff
@@ -963,6 +963,27 @@ contain_volatile_functions_not_nextval_walker(Node *node, void *context)
...
+	ParseState *pstate;
+    Node       *analyzed_expr;
+    bool        result;
+    
+    pstate = make_parsestate(NULL);
+    analyzed_expr = transformExpr(pstate, raw_expr, expr_kind);
+    result = contain_volatile_functions(analyzed_expr);
+    free_parsestate(pstate);
+    
+    return result;
```
RekGRpth:
```suggestion
contain_volatile_functions_raw(Node* raw_expr)
{
	ParseState *pstate = make_parsestate(NULL);
    Node       *analyzed_expr = transformExpr(pstate, raw_expr, EXPR_KIND_OTHER);
    bool        result = contain_volatile_functions(analyzed_expr);
    
    free_parsestate(pstate);
    
    return result;
```

src/backend/commands/tablecmds.c
```diff
@@ -978,6 +979,15 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
...
 			Assert(colDef->cooked_default == NULL);
 
+			/* Protect replicated tables from volatile expressions as default value */
+			if (!colDef->generated &&
+				GpPolicyIsReplicated(policy) &&
+				contain_volatile_functions_raw(colDef->raw_default, EXPR_KIND_OTHER))
+				ereport(ERROR,
+					(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+					 errmsg("volatile expressions are not supported as "
+							"default values ​for columns in replicated tables")));
```
RekGRpth:
```suggestion
				ereport(ERROR,
						(errcode(ERRCODE_WRONG_OBJECT_TYPE),
					 	 errmsg("volatile expressions are not supported as "
								"default values ​for columns in replicated tables")));
```

src/backend/commands/tablecmds.c
```diff
@@ -7626,6 +7636,14 @@ ATExecAddColumn(List **wqueue, AlteredTableInfo *tab, Relation rel,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("cannot add column to a partition")));
 
+	/* Protect replicated tables from volatile expressions as default value */
+    if (GpPolicyIsReplicated(rel->rd_cdbpolicy) &&
+		contain_volatile_functions_raw(colDef->raw_default, EXPR_KIND_OTHER))
+		ereport(ERROR,
+			   (errcode(ERRCODE_WRONG_OBJECT_TYPE),
+				errmsg("volatile expressions are not supported as "
+					   "default values ​for columns in replicated tables")));
```
RekGRpth:
```suggestion
		ereport(ERROR,
				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
				 errmsg("volatile expressions are not supported as "
						"default values ​for columns in replicated tables")));
```

src/backend/commands/tablecmds.c
```diff
@@ -8745,6 +8763,14 @@ ATExecColumnDefault(Relation rel, const char *colName,
 	AttrNumber	attnum;
 	ObjectAddress address;
 
+	/* Protect replicated tables from volatile expressions as default value */
+	if (GpPolicyIsReplicated(rel->rd_cdbpolicy) &&
+		contain_volatile_functions_raw(newDefault, EXPR_KIND_OTHER))
+		ereport(ERROR,
+			(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+				errmsg("volatile expressions are not supported as "
+					"default values ​for columns in replicated tables")));
```
RekGRpth:
```suggestion
		ereport(ERROR,
				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
				 errmsg("volatile expressions are not supported as "
						"default values ​for columns in replicated tables")));
```

RekGRpth:
tests failed with
```diff
create table t_replicate_dst(id serial, i integer) distributed replicated;
+ERROR:  volatile expressions are not supported as default values ​for columns in replicated tables
```
Maybe we should use the `contain_volatile_functions_not_nextval` function?

VoidZeroNull0:
> Maybe we should use the `contain_volatile_functions_not_nextval` function?

Yup. Changed to use `contain_volatile_functions_not_nextval()`.

GirgerS (commented):
We have another edge-case, the latter `create table` shouldn't succeed:
```
create table parent (a float default random()) distributed by (a);
create table child (like parent including defaults) distributed replicated;
```

Something like this fixes the issue:

```diff
diff --git a/src/backend/commands/tablecmds.c b/src/backend/commands/tablecmds.c
index 9db943259d..959274c22b 100644
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -8859,6 +8859,24 @@ ATExecCookedColumnDefault(Relation rel, AttrNumber attnum,

        /* We assume no checking is required */

+       /*
+        *  ... except than in ggdb we need to perform an additional check.
+        *
+        *  The parent table might have a volatile function as the default value
+        *  for a column, for example if it is distributed by a key,
+        *  and the target table might be replicated. Such columns are prohibited
+        *  for replicated tables, so enforce this rule here as well.
+        */
+       if (GpPolicyIsReplicated(rel->rd_cdbpolicy) &&
+               contain_volatile_functions_not_nextval(newDefault))
+       {
+               ereport(ERROR,
+                       (errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
+                        errmsg("volatile expressions are not supported as "
+                                       "default values for columns in replicated tables"),
+                        errhint("Volatile expression(s) originate from the table(s) specified in the LIKE clause(s)")));
+       }
+

```

I also thought to check inherited tables, but it seems like we can't have replicated policy and inheritance at the same time (see commit 7efe320489d4f916d547e44b46d0c760db474c36).

src/backend/commands/tablecmds.c
```diff
@@ -978,6 +979,15 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
 
 			Assert(colDef->cooked_default == NULL);
 
+			/* Protect replicated tables from volatile expressions as default value */
+			if (!colDef->generated &&
+				GpPolicyIsReplicated(policy) &&
+				contain_volatile_functions_raw(colDef->raw_default))
+				ereport(ERROR,
+						(errcode(ERRCODE_WRONG_OBJECT_TYPE),
```
GirgerS: Do we use the correct error code? 
https://github.com/GreengageDB/greengage/blob/09906641542d68f290262252c11ef65d5373bb84/src/backend/catalog/heap.c#L3710-L3715

src/backend/commands/tablecmds.c
```diff
@@ -7626,6 +7636,14 @@ ATExecAddColumn(List **wqueue, AlteredTableInfo *tab, Relation rel,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("cannot add column to a partition")));
 
+	/* Protect replicated tables from volatile expressions as default value */
```
GirgerS: It seems like we interfere with the logic handling generated columns

Before the patch:
``` 
postgres=# alter table test add column c float generated always as (random()) stored;
ERROR:  generation expression is not immutable
```

After the patch:

```
postgres=# alter table test add column c float generated always as (random()) stored;
ERROR:  volatile expressions are not supported as default values for columns in replicated tables
```

And we explicitly exclude generated columns when [creating](https://github.com/GreengageDB/greengage/pull/321/changes#diff-d373fb790605aded212c2e3a0fbe8c015d110077ff46c07d0a6d6647a2d7e928R983-R985) a table.

src/backend/commands/tablecmds.c
```diff
@@ -978,6 +979,15 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
 
 			Assert(colDef->cooked_default == NULL);
 
+			/* Protect replicated tables from volatile expressions as default value */
+			if (!colDef->generated &&
```
GirgerS: Do generated columns cause the same error? If so, maybe we should mention that they are [handled](https://github.com/GreengageDB/greengage/blob/09906641542d68f290262252c11ef65d5373bb84/src/backend/catalog/heap.c#L3710-L3715) separately.

src/backend/optimizer/util/clauses.c
```diff
@@ -963,6 +963,25 @@ contain_volatile_functions_not_nextval_walker(Node *node, void *context)
...
+ *
+ * This is a wrapper for contain_volatile_functions_not_nextval() that 
+ * is for expressions that are raw and yet have been analyzed 
+ * (e.g. ones for DEFAULT COLUMN).
+ */
+bool
+contain_volatile_functions_raw(Node* raw_expr)
+{
+	ParseState *pstate = make_parsestate(NULL);
+    Node       *analyzed_expr = transformExpr(pstate, raw_expr, EXPR_KIND_OTHER);
```
GirgerS: Looks like we do cook the raw expression tree by calling [AddRelationNewConstraints](https://github.com/GreengageDB/greengage/blob/1d1dd7d55e9c15795adadbbe6bf62dc296eabc38/src/backend/catalog/heap.c#L3234-L3286) already:
	[DefineRelation](https://github.com/GreengageDB/greengage/blob/1d1dd7d55e9c15795adadbbe6bf62dc296eabc38/src/backend/commands/tablecmds.c#L1192-L1195), [ATExecAddColumn](https://github.com/GreengageDB/greengage/blob/1d1dd7d55e9c15795adadbbe6bf62dc296eabc38/src/backend/commands/tablecmds.c#L7832-L7834), [ATExecColumnDefault](https://github.com/GreengageDB/greengage/blob/1d1dd7d55e9c15795adadbbe6bf62dc296eabc38/src/backend/commands/tablecmds.c#L8831-L8833).
	
Maybe we can leverage this fact to skip additional parsing? It seems that with the current approach we do it two times.

src/backend/commands/tablecmds.c
VoidZeroNull0: I agree. Thanks for noticing!

src/backend/optimizer/util/clauses.c
VoidZeroNull0: Yup, thanks much! Seems like a very good idea. Moved check there.

src/backend/commands/tablecmds.c
VoidZeroNull0: They used to, because parser would output an error stating that columns do not exists (as we have not made a table yet). This is no longer the problem after moving `AddRelationNewConstraints()`.

VoidZeroNull0:
> We have another edge-case, the latter create table shouldn't succeed:

Added your solution and a test to `create_table.sql`.

src/backend/commands/tablecmds.c
```diff
@@ -977,7 +978,7 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
 			RawColumnDefault *rawEnt;
 
 			Assert(colDef->cooked_default == NULL);
-
+
```
RekGRpth: remove this change

src/backend/commands/tablecmds.c
VoidZeroNull0: There was such thing. Now returned back to `ERROR:  generation expression is not immutable`.

src/backend/optimizer/util/clauses.c
GirgerS: Let's wait for the test results to make sure that there is no subtle behavior. If there isn't any, feel free to close this thread on my behalf.

GirgerS:
Looks like we also didn't consider the workaround from the parent ticket:

```
create table alter_policy (a float default random()) distributed by (a);
alter table alter_policy set distributed replicated;
CREATE TABLE 
ALTER TABLE
```

src/backend/catalog/heap.c
```diff
@@ -3242,6 +3242,14 @@ AddRelationNewConstraints(Relation rel,
 						   NameStr(atp->attname),
 						   atp->attgenerated);
 
+		/* Protect replicated tables from volatile expressions as DEFAULT value */
+		if (GpPolicyIsReplicated(rel->rd_cdbpolicy) &&
+			contain_volatile_functions_not_nextval(expr))
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
+					errmsg("volatile expressions are not supported as "
```
RekGRpth:
```suggestion
					 errmsg("volatile expressions are not supported as "
```

src/backend/catalog/heap.c
```diff
@@ -3242,6 +3242,14 @@ AddRelationNewConstraints(Relation rel,
 						   NameStr(atp->attname),
 						   atp->attgenerated);
 
+		/* Protect replicated tables from volatile expressions as DEFAULT value */
```
RekGRpth: Can you place your new condition below after `continue`?

src/backend/commands/tablecmds.c
```diff
@@ -8833,6 +8834,25 @@ ATExecCookedColumnDefault(Relation rel, AttrNumber attnum,
...
+	*  for replicated tables, so enforce this rule here as well.
+	*/
+	if (GpPolicyIsReplicated(rel->rd_cdbpolicy) &&
+		contain_volatile_functions_not_nextval(newDefault))
+	{
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
+					 errmsg("volatile expressions are not supported as "
+							"default values for columns in replicated tables"),
+					 errhint("Volatile expression(s) originate from the table(s) "
```
RekGRpth: maybe `errdetail`?

src/backend/commands/tablecmds.c
```diff
@@ -8833,6 +8834,25 @@ ATExecCookedColumnDefault(Relation rel, AttrNumber attnum,
...
 	/* We assume no checking is required */
 
+	/*
+	*  ... except than in ggdb we need to perform an additional check.
+	*
+	*  The parent table might have a volatile function as the default value
+	*  for a column, for example if it is distributed by a key,
+	*  and the target table might be replicated. Such columns are prohibited
+	*  for replicated tables, so enforce this rule here as well.
+	*/
```
RekGRpth:
```suggestion
	 *  ... except than in ggdb we need to perform an additional check.
	 *
	 *  The parent table might have a volatile function as the default value
	 *  for a column, for example if it is distributed by a key,
	 *  and the target table might be replicated. Such columns are prohibited
	 *  for replicated tables, so enforce this rule here as well.
	 */
```

src/backend/commands/tablecmds.c
```diff
@@ -18299,6 +18319,27 @@ ATExecSetDistributedBy(Relation rel, Node *node, AlterTableCmd *cmd)
...
+				if (contain_volatile_functions_not_nextval(expr))
+				{
+					ereport(ERROR,
+							(errcode(ERRCODE_INVALID_TABLE_DEFINITION),
+							 errmsg("volatile expressions are not supported as "
+									"default values for columns in replicated tables"),
+							 errhint("Cannot change policy as some of the columns have volatile "
+									 "expressions as defaults.")));
+				}
+			}
```
RekGRpth, line 18338:
```suggestion
			TupleConstr *constr = rel->rd_att->constr;
			if (constr)
			{
				for (uint16 i = constr->num_defval - 1; i >= 0; i--)
				{
					char *adbin = constr->defval[i].adbin;
					if (adbin &&
						contain_volatile_functions_not_nextval(stringToNode(adbin)))
						ereport(ERROR, ...
				}
			}
```

src/backend/commands/tablecmds.c
```diff
@@ -8842,15 +8842,22 @@ ATExecCookedColumnDefault(Relation rel, AttrNumber attnum,
...
+		{
+			char *adbin = constr->defval[i].adbin;
+			if (adbin && contain_volatile_functions_not_nextval(stringToNode(adbin)))
+			{
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_TABLE_DEFINITION),
+						 errmsg("volatile expressions are not supported as "
+							   "default values for columns in replicated tables"),
+						 errdetail("Cannot change policy as some of the columns have volatile "
+								  "expressions as defaults.")));
```
RekGRpth:
```suggestion

						 errmsg("volatile expressions are not supported as "
							    "default values for columns in replicated tables"),
						 errdetail("Cannot change policy as some of the columns have volatile "
								   "expressions as defaults.")));
```

src/backend/commands/tablecmds.c
RekGRpth: I suggest it only here

src/backend/commands/tablecmds.c
```diff
@@ -8842,15 +8842,22 @@ ATExecCookedColumnDefault(Relation rel, AttrNumber attnum,
...
+			}
+		}
```
RekGRpth: why removed?
```suggestion
	if (GpPolicyIsReplicated(rel->rd_cdbpolicy) &&
		contain_volatile_functions_not_nextval(newDefault))
	{
			ereport(ERROR,
					(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
					 errmsg("volatile expressions are not supported as "
							"default values for columns in replicated tables"),
					 errhint("Volatile expression(s) originate from the table(s) "
							 "specified in the LIKE clause(s)")));
	}
```

VoidZeroNull0: Accidentally refactored wrong place.

src/backend/commands/tablecmds.c
VoidZeroNull0: Done in 6b74d2e.

RekGRpth:
actuallize pr descr

VoidZeroNull0:
I have found another loophole:
```SQL
CREATE FUNCTION fn_val()
        RETURNS int
        LANGUAGE plpgsql
        IMMUTABLE
AS $$
BEGIN
	RETURN 42;
END;
$$
EXECUTE ON ANY;

CREATE TABLE dist_replicated(x int DEFAULT fn_val()) DISTRIBUTED REPLICATED;

INSERT INTO dist_replicated DEFAULT VALUES;
INSERT INTO dist_replicated DEFAULT VALUES;
INSERT INTO dist_replicated DEFAULT VALUES;

CREATE OR REPLACE FUNCTION  fn_val()
        RETURNS int
        LANGUAGE plpgsql
        VOLATILE
AS $$
BEGIN
	RETURN floor(random() * 100 + 1);
END;
$$
EXECUTE ON ANY;

INSERT INTO dist_replicated DEFAULT VALUES;
INSERT INTO dist_replicated DEFAULT VALUES;
INSERT INTO dist_replicated DEFAULT VALUES;

SELECT DISTINCT x FROM dist_replicated;
 x  
----
  2
  6
 42
 69
(4 rows)
```

src/test/regress/expected/alter_table_ao.out
```diff
@@ -1137,7 +1139,12 @@ end;
 execute checkattributeencoding('t_addcol');
  attnum | filenum | lastrownums | attoptions 
 --------+---------+-------------+------------
-(0 rows)
```
GirgerS, line 1140: One last thing, what caused these changes?

VoidZeroNull0:
> I have found another loophole:

Generated columns also have this issue, so I think it is redundant to track volatility for replicated tables.

```SQL
CREATE FUNCTION get_num() RETURNS INT
LANGUAGE sql IMMUTABLE AS $$ SELECT 1; $$;

CREATE TABLE test (a int, num int GENERATED ALWAYS AS (get_num()) STORED) DISTRIBUTED BY (a);

INSERT INTO test DEFAULT VALUES;
INSERT INTO test DEFAULT VALUES;
INSERT INTO test DEFAULT VALUES;

SELECT * FROM test;
 a | num 
---+-----
   |   1
   |   1
   |   1

CREATE OR REPLACE FUNCTION get_num() RETURNS INT
LANGUAGE sql VOLATILE AS $$ SELECT floor(random()*1000)::int; $$;

INSERT INTO test DEFAULT VALUES;
INSERT INTO test DEFAULT VALUES;
INSERT INTO test DEFAULT VALUES;

SELECT * FROM test;
 a | num 
---+-----
   |   1
   |   1
   |   1
   | 512
   | 137
   | 243
(6 rows)

ALTER TABLE test 
ADD COLUMN nu1m INT GENERATED ALWAYS AS (get_num()) STORED;
ERROR:  generation expression is not immutable
```

src/test/regress/expected/alter_table_ao.out
VoidZeroNull0: When we added a column with volatile default before, we initiated a [rewrite](https://github.com/GreengageDB/greengage/blob/55ac8efbfbc50ba4a85ac1f9b4e4e1772ed6fa0f/src/test/regress/expected/alter_table_ao.out#L1092-L1096). Because of this, AO table was fully rewritten and info about columns `pg_attribute_encoding` was [deleted](https://github.com/GreengageDB/greengage/blob/55ac8efbfbc50ba4a85ac1f9b4e4e1772ed6fa0f/src/backend/commands/cluster.c#L1270-L1272). I suppose it is not kept as it is obsolete. So, as we no longer add column with volatile default, we keep `pg_attribute_encoding` .

GirgerS: Thanks, I understand it more or less now.

GirgerS (approved):

bimboterminator1 (approved):

PR #322, opened 2026-03-16
Add new GUC gp_autostats_on_change_ratio_threshold

red1452:
Add new GUC gp_autostats_on_change_ratio_threshold

The gp_autostats_on_change_ratio_threshold is GUC for improving auto-analyze
tables. It allows for less frequent analysis of large tables when adding a large
number of rows, which is more than gp_autostats_on_change_threshold, but too
small for the table.
By default, gp_autostats_on_change_ratio_threshold is "0". To enable GUC, it
should be greater than "0". Value is a fraction of tuples from statistics.
Auto-analyze will be executed, if number of added rows will be enough for both
of thresholds.

Also, new value for gp_autostats_mode was added: "ON_CHANGE_AND_NO_STATS". It
enables auto-analyzing for both of cases: "no stats" and "on change".


whitehawk:
Please check ABI and behave failures.

src/backend/postmaster/autostats.c
```diff
@@ -104,6 +108,34 @@ autostats_on_change_check(AutoStatsCmdType cmdType, uint64 ntuples)
 			break;
 	}
 
+	if (result && gp_autostats_on_change_ratio_threshold > 0)
+	{
+		HeapTuple	tuple;
+		Form_pg_class classForm;
+		float4 reltuples;
```
whitehawk:
```
autostats.c: In function ‘autostats_on_change_check’:
autostats.c:115:24: warning: unused variable ‘reltuples’ [-Wunused-variable]
  115 |                 float4 reltuples;
      |                        ^~~~~~~~~

```

src/backend/postmaster/autostats.c
```diff
@@ -104,6 +108,34 @@ autostats_on_change_check(AutoStatsCmdType cmdType, uint64 ntuples)
...
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_TABLE),
+					 errmsg("relation with OID %u does not exist",
+							relationOid)));
+			return false;
+		}
+		classForm = (Form_pg_class) GETSTRUCT(tuple);
+
+		if (ntuples / classForm->reltuples < gp_autostats_on_change_ratio_threshold)
```
whitehawk: In case the table is already empty, `classForm->reltuples` will be 0. That should be handled to avoid division by 0.
For ex, with added some debug logs:
```
postgres=# create table test(a int);
CREATE TABLE
postgres=# insert into test select generate_series(1, 10);
INSERT 0 10
postgres=# delete from test;
WARNING:  [RELOG] autostats_on_change_check, cmdType 3, ntuples 10
WARNING:  [RELOG] classForm->reltuples 10.000000
WARNING:  [RELOG] ntuples / classForm->reltuples = 1.000000
DELETE 10
postgres=# insert into test select generate_series(1, 10);
WARNING:  [RELOG] autostats_on_change_check, cmdType 2, ntuples 10
WARNING:  [RELOG] classForm->reltuples 0.000000
WARNING:  [RELOG] ntuples / classForm->reltuples = inf
INSERT 0 10
postgres=# show gp_autostats_mode;
   gp_autostats_mode
------------------------
 on_change_and_no_stats
(1 row)

postgres=# show gp_autostats_on_change_ratio_threshold;
 gp_autostats_on_change_ratio_threshold
----------------------------------------
 0.1
(1 row)

postgres=# show gp_autostats_on_change_threshold;
 gp_autostats_on_change_threshold
----------------------------------
 0
(1 row)

```

whitehawk: It looks that `ntuples / classForm->reltuples` will always be in range of [0; 1] when we execute `UPDATE <table> ...` operation, while  gp_autostats_on_change_ratio_threshold can be in [0; 100].

On the other hand, if we consider an insert operation. For ex., we have a table that have already 10 rows, and insert another 10 rows.  In this case `classForm->reltuples = 10.000000` and `ntuples = 10`. So, `ntuples / classForm->reltuples = 1.000000`, which I interpet as 100% rows were affected. While in reality only 50% of rows are new.

src/test/regress/sql/autostats.sql
```diff
@@ -149,3 +149,43 @@ reset gp_autostats_mode_in_functions;
 reset gp_autostats_on_change_threshold;
 reset log_autostats;
 reset gp_autostats_allow_nonowner;
+
+/*
+ * Test for gp_autostats_mode=on_change_and_no_stats and gp_autostats_on_change_ratio_threshold > 0
+ */
+set gp_autostats_mode=on_change_and_no_stats;
```
whitehawk:
```suggestion
set gp_autostats_mode = on_change_and_no_stats;
```

whitehawk:
In the description:
>  It makes not often analyzing for huge tables, when adding number of
rows, which is more than gp_autostats_on_change_threshold, but is too little for
table.

sounds a bit awkwardly. I suggest to re-phrase, for ex, smth like:

> It allows for less frequent analysis of large tables when adding a large number of rows, which is more than gp_autostats_on_change_threshold, but too small for the table.

src/backend/postmaster/autostats.c
red1452: Values for `gp_autostats_on_change_ratio_threshold` was taken from [doc](https://gitlab.adsw.io/arenadata/development/adb-requirements/-/blob/master/greengage/statistics/gp_autostats_mode.md)

whitehawk: > Values for gp_autostats_on_change_ratio_threshold was taken from [doc](https://gitlab.adsw.io/arenadata/development/adb-requirements/-/blob/master/greengage/statistics/gp_autostats_mode.md)

The design document can also have issues. Please check it with the architect.

From the user point of view, let's imagine following situation. I have a table with 10 rows, and I want auto stats to be triggered when more than half of the table's content is changed. So I set `gp_autostats_on_change_ratio_threshold` to 51 (meaning 51%).
Then 
1. Perform 'UPDATE <my table> SET <some_value> WHERE TRUE;' - it will alter the full content of the table. But the formula will give `ntuples / classForm->reltuples = 1`, which is less than 51. And auto stats will not be triggered.
or 
2. Insert new, let's say, 20 new rows. The delta of table's content will be 20/(10+20) = 0.67 (meaning 67%). But we will calculate it as 20/10 = 2 (meaning 200%, which is still actually wrong from my point of view), and compare 2 to 51.

So, I think there are 2 issues here:
1. We need to normalize the comparison of ratio with percents;
2. Need to adjust the calculation for the insert opeations.

src/backend/utils/misc/guc_gp.c
```diff
@@ -5001,6 +5003,16 @@ struct config_real ConfigureNamesReal_gp[] =
 		NULL, NULL, NULL
 	},
 
+	{
+		{"gp_autostats_on_change_ratio_threshold", PGC_USERSET, DEVELOPER_OPTIONS,
+			gettext_noop("Threshold for number of tuples added to table by CTAS or Insert-to to trigger autostats in on_change mode. See gp_autostats_mode."),
```
whitehawk: > added to table by CTAS or Insert-to

But the UPDATE/DELETE/COPY operations are also affected, no?

src/backend/postmaster/autostats.c
```diff
@@ -259,6 +293,8 @@ autostats_get_cmdtype(QueryDesc *queryDesc, AutoStatsCmdType * pcmdType, Oid *pr
  * on_change	:	if the number of modified tuples > gp_onchange_threshold, then an automatic analyze is issued.
  * on_no_stats	:	if the operation is a ctas/insert-select and there are no stats on the modified table,
  *					an automatic analyze is issued.
+ * on_change_and_no_stats: if one of on_change or on_no_stats is returns true,
```
GirgerS: This line sounds a bit off to me

```suggestion
 * on_change_and_no_stats: if either on_change or on_no_stats returns true,
```

src/backend/postmaster/autostats.c
GirgerS: > So, I think there are 2 issues here:
    We need to normalize the comparison of ratio with percents;
    Need to adjust the calculation for the insert opeations.

I also agree with the mentioned issues. 

There is one more thing: if we have a large number of tuples, the ratio might not be within the bounds of the guc:

For `INSERT`, if `num_tuples = 101` and `table_size = 1` then the ratio is `101 / 1 = 101`, it is outside of the current guc range, and we would not be able to prohibit statistics update in this case.

Of course this is theoretical, but still.

red1452: `gp_autostats_on_change_ratio_threshold` is not a percent, it is fraction. value 51 is the same as 5100%. You may set 0.51 to `gp_autostats_on_change_ratio_threshold`.

whitehawk: In this case why do we limit it by `100.0`? And please explicitly specify it in the help of the GUC.

And in this case there is still descripancy between UPDATE/DELETE and INSERT cases.
Another use case - let's say we, as the users, want to the statistics to be updated when 90% of rows are contents is altered. We set `gp_autostats_on_change_ratio_threshold` to 0.9.
Given a table with 10 rows, we:
1. Insert 9 new rows. Following current calculations, 9/10=0.9 - we will see the statistics update;
2. Update or delete the same 9 rows (that we've previously inserted). Now, according to current calculations, 9/19=0.47, and we will not have statistics update. Which for me looks inconsistent, as we've changed the same amount or rows.

GirgerS: > gp_autostats_on_change_ratio_threshold is not a percent, it is fraction. value 51 is the same as 5100%. You may set 0.51 to gp_autostats_on_change_ratio_threshold.

You are right, considering the way how it is specified in the doc. But I'd suggest to ask the architect if it is really expected, at least because

> In this case why do we limit it by 100.0? And please explicitly specify it in the help of the GUC.`

The logic doesn't seem to be clear, and we might unintentionally confuse our users making them thing that the value they specified is a percentage, not a fraction.

whitehawk (approved):

GirgerS (approved):
Turn out it works similarly to `autovacuum_analyze_scale_factor`.

PR #323, opened 2026-03-17
Adjust  GP_HLL_ERROR_MARGIN

dkovalev1:
In the ANALYZE command, GP_HLL_ERROR_MARGIN is used as a threshold to determine 
if all values are distinct. It was set too low, leading to incorrect statistics 
calculation and the possible generation of a suboptimal plan.

The problem has a probabilistic nature and does not appear on every run. 
Therefore, a regression test was not provided, as it would be a flaky test. 
However, even with the fix, due to the probabilistic nature, we cannot guarantee
100% accuracy.
Checks were performed by using the introduced unit test and running ANALYZE 
several times to inspect n_distinct.

GG-298


ShapatinAA:
Is there a reason on why we can't declare that root has -1 as ndistinct when all data in partitions are calculated to be distinct?

dkovalev1:
> Is there a reason on why we can't declare that root has -1 as ndistinct when all data in partitions are calculated to be distinct?

Not quite understood the idea. The "calculated" - what do you mean? 

In current architecture executor does not deal with statistics, and analyze deals with data but not how it was created.

ShapatinAA:
> > Is there a reason on why we can't declare that root has -1 as ndistinct when all data in partitions are calculated to be distinct?
> 
> Not quite understood the idea. The "calculated" - what do you mean?
> 
> In current architecture executor does not deal with statistics, and analyze deals with data but not how it was created.

As I understand: firstly we get to count statistics for all partitions (leafs). And then, when we count statistic for root, we can use already calculated statistic for leafs. If this is right: what is stopping us from addressing all `n_distinct` values from all leafs and declaring that root also has -1 if all leafs are unique on chosen attribute?

src/backend/commands/analyze.c
whitehawk: The comment now requires adjustment as well.

whitehawk:
In the description:

1. First sentense is pretty long and, without commas, a bit hard to understand. I suggest to re-phrase it a little:
> In the ANALYZE define GP_HLL_ERROR_MARGIN used as a threshold to determine if
all values are distinct is too low leading to incorrect statistic calculation
and possible generation of not optimal plan.

-> 
> In the ANALYZE define GP_HLL_ERROR_MARGIN is used as a threshold to determine if
all values are distinct. It was too low leading to incorrect statistic calculation
and possible generation of not optimal plan.

2. "test was not provided is even without" -> "test was not provided as even without"?
3. Please wrap the text into 80-char limit.

whitehawk:
>  test was not provided

But there is a test in the patch... The description doesn't match well with the contents of the patch.

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,252 @@
+#include <stdarg.h>
```
whitehawk, line 15: Please add a proper copyright header to the new file

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,252 @@
...
+     * Expected HLL error
+     */
+    double expected_error = 1.04 / sqrt(POW2(hll->b));
+
+    /*
+     * Allow some safety margin (2x theoretical error)
+     */
+    assert_true(rel_error < expected_error * 2);
+
+    run_count++;
```
whitehawk: What is the purpose of `run_count`? It seems we don't use it in other places....

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,252 @@
...
+
+    printf("actual=%d estimate=%f rel_error=%f\n",
+           nitems,
+           estimate,
+           rel_error);
+
+	/*
+     * Expected HLL error
+     */
+    double expected_error = 1.04 / sqrt(POW2(hll->b));
```
whitehawk: `1.04` - what is this magic number?

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
whitehawk, line 1: The indentation is not consistent in the file. Motly 4 spaces are used, but in some places tabs are used. Please make it consistent.

whitehawk:
>  even without a fix it will be flaky test which works in ~70% of cases.

As I understand the nature of the problem:
 - without the fix in 30% of cases ANALYZE will give `n_distinct != -1` result.
 - with the fix ANALYZE will always give `n_distinct == -1` result.

So, why don't we conduct a test that will runs ANALYZE 10 times in a loop (maybe in a SQL function), and checks the result is always -1? In this case it should always pass with the fix, and almost always not pass without the fix (which I think is acceptable). What do you think?

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
ShapatinAA: This is variable βₘ described in section 3.1 of original paper. Later they "calculate" it to be 1,03896.

dkovalev1:
> In the description:
> 
>     1. First sentense is pretty long and, without commas, a bit hard to understand. I suggest to re-phrase it a little:
>     2. "test was not provided is even without" -> "test was not provided as even without"?
>     3. Please wrap the text into 80-char limit.

Done

dkovalev1:
> > test was not provided
> 
> But there is a test in the patch... The description doesn't match well with the contents of the patch.

description updated

src/backend/commands/analyze.c
dkovalev1: adjusted

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: Interesting that only 10 test files out of 1292 had copyrights, which makes think that this is not a rule for a tests. However added 11-th.

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: removed

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: added comment with a link to a header where the paper mentioned

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: reindented

dkovalev1:
> > even without a fix it will be flaky test which works in ~70% of cases.
> 
> As I understand the nature of the problem:
> 
>     * without the fix in 30% of cases ANALYZE will give `n_distinct != -1` result.
> 
>     * with the fix ANALYZE will always give `n_distinct == -1` result.
> 
> 
> So, why don't we conduct a test that will runs ANALYZE 10 times in a loop (maybe in a SQL function), and checks the result is always -1? In this case it should always pass with the fix, and almost always not pass without the fix (which I think is acceptable). What do you think?

This is a good idea, but it actually shows stochastic nature of the algorithm. I tried this test and it shown that even with the patch it's possible to have ~1% of false detection, compared to ~34% before the patch. In fact with the test data we still will have flaky test (or useless if we lower dataset), but its rather lengthy. Please see the issue for more details.
I vote for not including this test into the regression.

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,268 @@
+/*-------------------------------------------------------------------------
+ *
+ * gp_hyperloglog_test.c
+ *	  HyperLogLog algorithm test program
+ *
+ * Portions Copyright (c) 2012, Greenplum Inc.
+ * Portions Copyright (c) 2012-Present Pivotal Software, Inc.
```
RekGRpth: Greenplum and Pivotal?

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,268 @@
...
+
+static void
+hll_statistical_accuracy(size_t cardinality, size_t step)
+{
+	GpHLLCounter hll;
+
+	double		estimate;
+	double		rel_error;
+
+	hll = gp_hyperloglog_init_def();
```
RekGRpth:
```suggestion
	GpHLLCounter hll = gp_hyperloglog_init_def();
```

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,268 @@
...
+		hll = gp_hyperloglog_add_item(
+									  hll,
+									  d,
+									  8,		/* typlen */
+									  true,		/* typbyval */
+									  'd');		/* typalign */
+		nitems++;
+	}
+
+	estimate = gp_hyperloglog_estimate(hll);
```
RekGRpth:
```suggestion
	double		estimate = gp_hyperloglog_estimate(hll);
```

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,268 @@
...
+
+
+	rel_error = fabs(estimate - nitems) / nitems;
```
RekGRpth:
```suggestion
	double		rel_error = fabs(estimate - nitems) / nitems;
```

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: I just took the comment from adjacent file. It is definitely uses code under the test, so Greenplum and Pivotal are relevant here. And it's remarkable that there is no Arenadata in copyrights, did not want to be the first.

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: I followed ANSI C style, which appeared in the tested code to be consistent with it. According to the our patch requirements.

RekGRpth: > I followed ANSI C style

https://github.com/GreengageDB/greengage/pull/323/changes#diff-333b1f8067cebb6a06c5d2f79646ff2c4eb44c52a9bff7f2a5dbae1b34c3ef67R85

RekGRpth: > I followed ANSI C style

Why? It doesn't make any sense.

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: I followed ANSI C style, which appeared in the tested code to be consistent with it. According to the our patch requirements.

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: I followed ANSI C style, which appeared in the tested code to be consistent with it. According to the our patch requirements.

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: okay, moved

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
whitehawk: Why is the copyright of Pivotal? It should be Greengage

dkovalev1: test code for sure uses code under test, so we can not remove Pivotal. I've added Greengage.

whitehawk:
```
 * Portions Copyright (c) 2012, Greenplum Inc.
 * Portions Copyright (c) 2012-Present Pivotal Software, Inc.
```
These lines should be removed. This source file didn't exist in 2012. Its history has started in 2026, and it is authored by Greengage. Please refer to [example](https://github.com/GreengageDB/greengage/blob/c29d59c8bd8ff4cade46dd08ea416a3b74fc0589/src/backend/catalog/storage_pending_deletes_redo.c#L6)

dkovalev1: okay

whitehawk:
> > So, why don't we conduct a test that will runs ANALYZE 10 times in a loop (maybe in a SQL function), and checks the result is always -1? In this case it should always pass with the fix, and almost always not pass without the fix (which I think is acceptable). What do you think?
> 
> This is a good idea, but it actually shows stochastic nature of the algorithm. I tried this test and it shown that even with the patch it's possible to have ~1% of false detection, compared to ~34% before the patch. In fact with the test data we still will have flaky test (or useless if we lower dataset), but its rather lengthy. Please see the issue for more details. I vote for not including this test into the regression.

Just to clarify - does "even with the patch it's possible to have ~1% of false detection" mean that with the fix in 1% of cases ANALYZE will give n_distinct != -1 result?

dkovalev1:
> Just to clarify - does "even with the patch it's possible to have ~1% of false detection" mean that with the fix in 1% of cases ANALYZE will give n_distinct != -1 result?

yes

whitehawk (dismissed):

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,261 @@
...
+#include <stdio.h>
+
+#include "../gp_hyperloglog.c"
+
+/*
+	Uncomment the following to check a lot of variants. It will take a long time.
+*/
+/*	#define TEST_ACCURACY_LONG 1 */
+
+float		max_error = 0;
```
ShapatinAA, line 45: for what do we need max_error?

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
```diff
@@ -0,0 +1,261 @@
...
+	for (uint64 i = 0; i < cardinality; i += step)
+	{
+		Datum		d = Int64GetDatum(i);
+
+		hll = gp_hyperloglog_add_item(
+									  hll,
+									  d,
+									  8,		/* typlen */
+									  true,		/* typbyval */
+									  'd');		/* typalign */
```
ShapatinAA:
```suggestion
									  sizeof(uint64),	/* typlen */
									  true,				/* typbyval */
									  'd');				/* typalign */
```

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: This is a remains from the test code which calculated maximum of appeared error. I've restored this code.

src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
dkovalev1: done

ShapatinAA (approved):

whitehawk (approved):

ShapatinAA:
Btw do we have a reason on why unit tests are turned off while building?

PR #327, opened 2026-03-17
ADBDEV-7532: Track PQresult allocations in server code [7X]

bandetto:
Track PQresult allocations in server code

8a2a471 was adopted for 7X.

Fixed palloc() for events. Some palloc() wrappers were missing--this is a bug in the original commit, which will be addressed separately.

There is no 'ERROR' part of the test, and no parallel retrieve cursor tests, since 7X is not affected by the original problem and the it's impossible to set shared quota for a resource group.

Please note that this PR needs to be rebased.

### DIFF
--- a/src/backend/cdb/dispatcher/cdbdisp_async.c
+++ b/src/backend/cdb/dispatcher/cdbdisp_async.c
@@ -687,6 +687,8 @@ checkDispatchResult(CdbDispatcherState *ds, int timeout_sec)
 
 	pfree(revents);
 	pfree(added);
+
+	SIMPLE_FAULT_INJECTOR("check_dispatch_result_end");
 }
 
 /*
--- a/src/interfaces/libpq/fe-connect.c
+++ b/src/interfaces/libpq/fe-connect.c
@@ -102,13 +102,31 @@ static int	ldapServiceLookup(const char *purl, PQconninfoOption *options,
  * than looking into errcodes.h since it reflects historical behavior
  * rather than that of the current code.
  */
+#ifdef ERRCODE_APPNAME_UNKNOWN
+#undef ERRCODE_APPNAME_UNKNOWN
+#endif
+
 #define ERRCODE_APPNAME_UNKNOWN "42704"
 
 /* This is part of the protocol so just define it */
+#ifdef ERRCODE_INVALID_PASSWORD
+#undef ERRCODE_INVALID_PASSWORD
+#endif
+
 #define ERRCODE_INVALID_PASSWORD "28P01"
+
 /* This too */
+#ifdef ERRCODE_CANNOT_CONNECT_NOW
+#undef ERRCODE_CANNOT_CONNECT_NOW
+#endif
+
 #define ERRCODE_CANNOT_CONNECT_NOW "57P03"
+
 /* And this GPDB-specific one, too */
+#ifdef ERRCODE_MIRROR_READY
+#undef ERRCODE_MIRROR_READY
+#endif
+
 #define ERRCODE_MIRROR_READY "57M02"
 
 /*
--- a/src/interfaces/libpq/fe-exec.c
+++ b/src/interfaces/libpq/fe-exec.c
@@ -145,13 +145,18 @@ static int	check_field_number(const PGresult *res, int field_num);
  *	 returns a newly allocated, initialized PGresult with given status.
  *	 If conn is not NULL and status indicates an error, the conn's
  *	 errorMessage is copied.  Also, any PGEvents are copied from the conn.
+ *
+ * GPDB: most of the libpq's memory allocations happen here--PQresults store
+ *       all retrieved rows. Thus we use a palloc() wrapper instead of bare
+ *       malloc() for vmtracker to see the allocations, in case we're the
+ *       backend.
  */
 PGresult *
 PQmakeEmptyPGresult(PGconn *conn, ExecStatusType status)
 {
 	PGresult   *result;
 
-	result = (PGresult *) malloc(sizeof(PGresult));
+	result = (PGresult *) pqPalloc(sizeof(PGresult));
 	if (!result)
 		return NULL;
 
@@ -402,6 +407,9 @@ PQcopyResult(const PGresult *src, int flags)
  * Does not duplicate the event instance data, sets this to NULL.
  * Also, the resultInitialized flags are all cleared.
  * The total space allocated is added to *memSize.
+ *
+ * GPDB: backends use palloc() for allocations here. Make sure this function
+ *       is only used for PGresult->events!
  */
 static PGEvent *
 dupEvents(PGEvent *events, int count, size_t *memSize)
@@ -414,7 +422,7 @@ dupEvents(PGEvent *events, int count, size_t *memSize)
 		return NULL;
 
 	msize = count * sizeof(PGEvent);
-	newEvents = (PGEvent *) malloc(msize);
+	newEvents = (PGEvent *) pqPalloc(msize);
 	if (!newEvents)
 		return NULL;
 
@@ -424,12 +432,12 @@ dupEvents(PGEvent *events, int count, size_t *memSize)
 		newEvents[i].passThrough = events[i].passThrough;
 		newEvents[i].data = NULL;
 		newEvents[i].resultInitialized = false;
-		newEvents[i].name = strdup(events[i].name);
+		newEvents[i].name = pqPstrdup(events[i].name);
 		if (!newEvents[i].name)
 		{
 			while (--i >= 0)
-				free(newEvents[i].name);
-			free(newEvents);
+				pqPfree(newEvents[i].name);
+			pqPfree(newEvents);
 			return NULL;
 		}
 		msize += strlen(events[i].name) + 1;
@@ -596,7 +604,7 @@ pqResultAlloc(PGresult *res, size_t nBytes, bool isBinary)
 	{
 		size_t		alloc_size = nBytes + PGRESULT_BLOCK_OVERHEAD;
 
-		block = (PGresult_data *) malloc(alloc_size);
+		block = (PGresult_data *) pqPalloc(alloc_size);
 		if (!block)
 			return NULL;
 		res->memorySize += alloc_size;
@@ -621,7 +629,7 @@ pqResultAlloc(PGresult *res, size_t nBytes, bool isBinary)
 	}
 
 	/* Otherwise, start a new block. */
-	block = (PGresult_data *) malloc(PGRESULT_DATA_BLOCKSIZE);
+	block = (PGresult_data *) pqPalloc(PGRESULT_DATA_BLOCKSIZE);
 	if (!block)
 		return NULL;
 	res->memorySize += PGRESULT_DATA_BLOCKSIZE;
@@ -730,22 +738,22 @@ PQclear(PGresult *res)
 			(void) res->events[i].proc(PGEVT_RESULTDESTROY, &evt,
 									   res->events[i].passThrough);
 		}
-		free(res->events[i].name);
+		pqPfree(res->events[i].name);
 	}
 
 	if (res->events)
-		free(res->events);
+		pqPfree(res->events);
 
 	/* Free all the subsidiary blocks */
 	while ((block = res->curBlock) != NULL)
 	{
 		res->curBlock = block->next;
-		free(block);
+		pqPfree(block);
 	}
 
 	/* Free the top-level tuple pointer array */
 	if (res->tuples)
-		free(res->tuples);
+		pqPfree(res->tuples);
 
 	/* zero out the pointer fields to catch programming errors */
 	res->attDescs = NULL;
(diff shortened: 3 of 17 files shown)


RekGRpth:
> Prior to 3ce2e6a

may be 513b8d1?

RekGRpth:
> Ticket: ADBDEV-7145

I think this can be deleted, but at the very end you can add the current ticket.

src/interfaces/libpq/fe-exec.c
```diff
@@ -730,22 +730,22 @@ PQclear(PGresult *res)
 			(void) res->events[i].proc(PGEVT_RESULTDESTROY, &evt,
 									   res->events[i].passThrough);
 		}
-		free(res->events[i].name);
+		pqPfree(res->events[i].name);
```
RekGRpth, line 741: Why is this change not in the original patch and not in 6X?

src/interfaces/libpq/libpq-int.h
```diff
@@ -81,6 +81,34 @@ typedef struct
...
+/*
+ * Definitions meant to lessen the malloc() usage for CDB routines in
+ * server-sided code for struct PGresult allocations.
+ */
+#ifndef FRONTEND
+#include "postgres.h"
+#include "utils/memutils.h"
+
+/*
+ * CurTransactionContext's lifetime lasts until the end of the current query,
```
RekGRpth:
```suggestion
 * TopTransactionContext's lifetime lasts until the end of the current query,
```

src/test/isolation2/expected/resgroup/resgroup_oom.out
```diff
@@ -0,0 +1,66 @@
...
+-- s/ERROR:  Out of memory.*/ERROR:  Out of memory/
+-- m/ERROR:  Canceling query because of high VMEM usage.*/
+-- s/ERROR:  Canceling query because of high VMEM usage.*/ERROR:  Out of memory/
+-- end_matchsubs
+
+CREATE RESOURCE GROUP rg_oom_test WITH (cpu_max_percent=20, memory_limit=20);
+CREATE RESOURCE GROUP
+CREATE ROLE role_oom_test RESOURCE GROUP rg_oom_test;
+CREATE ROLE
+CREATE OR REPLACE FUNCTION gp_mock_cdbdispatchcommand(amount int) RETURNS SETOF text AS '/home/gpadmin/gpdb_src/src/test/isolation2/../regress/regress.so', 'gp_mock_cdbdispatchcommand' LANGUAGE C;
```
RekGRpth: Why isn't this parameterized like in the original patch?

src/test/isolation2/sql/resgroup/resgroup_parallel_retrieve.sql
```diff
@@ -0,0 +1,192 @@
+-- start_ignore
```
RekGRpth:
```suggestion
-- start_ignore
CREATE EXTENSION IF NOT EXISTS gp_parallel_retrieve_cursor;

```
as in original patch

src/test/isolation2/sql/resgroup/resgroup_oom.sql
```diff
@@ -0,0 +1,55 @@
+-- start_matchsubs
```
RekGRpth: Why do we need this file if we already have `resgroup_oom.source`?

src/test/isolation2/input/resgroup/resgroup_oom.source
```diff
@@ -0,0 +1,42 @@
...
+  FROM gp_toolkit.gp_resgroup_status_per_host WHERE groupname = 'rg_oom_test'
+)
+SELECT r.mb < 150 AS "under 150 MB", r.mb > 90 AS "above 90 MB"
+FROM r;
+
+SELECT gp_inject_fault('check_dispatch_result_end', 'reset', dbid) FROM
+gp_segment_configuration WHERE role = 'p' AND content = -1;
+
+1<:
+
```
RekGRpth, line 39: where is test
```sql

-- And finally, make sure we don't enter error recursion on fail.
ALTER RESOURCE GROUP rg_oom_test SET memory_shared_quota 0;
1: SELECT count(*) FROM gp_mock_cdbdispatchcommand(10000000);
```
?

src/test/isolation2/sql/resgroup/resgroup_parallel_retrieve.sql
RekGRpth: Why do we need this file if we already have `resgroup_parallel_retrieve.source`?

src/interfaces/libpq/fe-connect.c
```diff
@@ -102,13 +102,31 @@ static int	ldapServiceLookup(const char *purl, PQconninfoOption *options,
  * than looking into errcodes.h since it reflects historical behavior
  * rather than that of the current code.
  */
+#ifdef ERRCODE_APPNAME_UNKNOWN
+#undef ERRCODE_APPNAME_UNKNOWN
+#endif
+
```
VoidZeroNull0, line 108: Why do we do some `#undef`s here? Is it because those are defined in some other place and then included here? Does it apply to 6.x?

RekGRpth:
> This is a cherry pick from commit [52dbaa7](https://github.com/GreengageDB/greengage/commit/52dbaa705a0dd4ed5c7b813aeeea807ebc833c1f).

why not 8a2a471 ?

src/interfaces/libpq/fe-connect.c
RekGRpth: > Does it apply to 6.x?

yes, see 8a2a471

bandetto:
> why not https://github.com/GreengageDB/greengage/commit/8a2a4716b8eb2d40a14ea0c727835f38b97cbadf ?

Why not not?

@RekGRpth regarding most of review: sorry, I forgot to push commited changes. Everything should be fine now.

src/interfaces/libpq/fe-exec.c
bandetto: Removed `palloc()` for events. They are a public API, expecting `malloc()`. Good for us, they aren't used anywhere in the greengage codebase.

I'll create a ticket to remove part regarding the `events` for 6X.

src/test/isolation2/input/resgroup/resgroup_oom.source
bandetto: Added a reason in the PR description.

VoidZeroNull0 (dismissed):
LGTM.

bandetto:
Last force-push: actualized the test again: increased the range of memory allowed for the test query and removed the 'ERROR' part that got from the old version again.

After the tests pass, I'll remove the temporary commit and ask for approvals again.

RekGRpth:
> Everything should be fine now.

Nothing has changed in the commit description:

1) Commit 3ce2e6a in 6X is still listed instead of commit 513b8d1 in 7X.

2) Commit 52dbaa7 is still listed instead of the correct 8a2a471.

3) Ticket 7145 is still listed instead of 7532.

src/interfaces/libpq/fe-exec.c
RekGRpth: I don't quite understand why there are no changes for `newEvents` and `newEvents[i].name`, like in the original patch?

src/interfaces/libpq/fe-exec.c
RekGRpth: still actual

RekGRpth: So, in 6X for `events[i].name` we use `pallock` and `free`?

bandetto: Yes, and for dupEvents() too. which mixes palloc() and malloc() if used with public-facing functions that modify events field.

bandetto:
Updated related commit hashes and the commit description.

RekGRpth:
The commit description doesn't indicate why the `parallel_retrieve_cursor/fault_inject` tests from the original patch weren't taken?

bandetto:
@RekGRpth updated the description:

> There is no 'ERROR' part of the test, and no parallel retrieve cursor tests, since 7X is not affected by the original problem and the it's impossible to set shared quota for a resource group.

RekGRpth:
> @RekGRpth updated the description:
> 
> > There is no 'ERROR' part of the test, and no parallel retrieve cursor tests, since 7X is not affected by the original problem and the it's impossible to set shared quota for a resource group.

I mean in commit descr

RekGRpth:
remove last commit

src/interfaces/libpq/fe-exec.c
VoidZeroNull0: Me too.

bandetto: > Removed palloc() for events. They are a public API, expecting malloc(). This is a bug in the original commit, which will be addressed separately.

Also, see #377

VoidZeroNull0 (dismissed):

bandetto:
> I mean in commit descr

Was fixed.

RekGRpth:
> > I mean in commit descr
> 
> Was fixed.

screen

src/interfaces/libpq/fe-exec.c
RekGRpth: still actual

bandetto: #377

RekGRpth: Check my question more carefully: in the current patch, you allocate memory for events using the regular allocator and free it using the Postgres allocator.

RekGRpth: https://github.com/GreengageDB/greengage/pull/327/changes#diff-d9aebfcab0a1e118492867722a6b344e9cec311fae381e049d862196d715657aR417

RekGRpth: Image

src/interfaces/libpq/fe-exec.c
```diff
@@ -402,6 +407,8 @@ PQcopyResult(const PGresult *src, int flags)
  * Does not duplicate the event instance data, sets this to NULL.
  * Also, the resultInitialized flags are all cleared.
  * The total space allocated is added to *memSize.
+ *
+ * GPDB: avoid using palloc() for events, since they may be accessed publicly.
```
RekGRpth: I'm not quite sure: here you're saying not to use `palloc`, but below you're using it.

src/interfaces/libpq/fe-exec.c
bandetto: Tried the API and looked at allocations more closely. Actually, PGresults can be allocated via palloc(), but it required moving an additional object file into backend. I wrapped the remaining PGresult fields into palloc(), so now all it's fields should be allocated via vmtracker.

See updates to #377 as well

src/interfaces/libpq/fe-exec.c
bandetto: Forgot to remove. Thanks.

RekGRpth:
Why did the tests fail?

src/interfaces/libpq/libpq-events.c
```diff
@@ -59,9 +59,9 @@ PQregisterEventProc(PGconn *conn, PGEventProc proc,
 
 		newSize = conn->eventArraySize ? conn->eventArraySize * 2 : 8;
 		if (conn->events)
-			e = (PGEvent *) realloc(conn->events, newSize * sizeof(PGEvent));
+			e = (PGEvent *) pqRepalloc(conn->events, newSize * sizeof(PGEvent));
```
RekGRpth: I think that in `conn` the `events` there is no need to allocate in a new way, but only when copying to the `result` through `dupEvents`.

src/interfaces/libpq/libpq-events.c
```diff
@@ -12,7 +12,7 @@
  *
  *-------------------------------------------------------------------------
  */
-#include "postgres_fe.h"
+#include "c.h"
```
RekGRpth, line 15: ???

bandetto: Since this file is now linked to the backend, to be consistent with:
https://github.com/GreengageDB/greengage/blob/f65088c1968d335617bde400855b55427c354f34/src/interfaces/libpq/fe-connect.c#L21-L22

src/backend/cdb/dispatcher/cdbdisp_async.c
```diff
@@ -687,6 +687,8 @@ checkDispatchResult(CdbDispatcherState *ds, int timeout_sec)
 
 	pfree(revents);
 	pfree(added);
+
+	SIMPLE_FAULT_INJECTOR("check_dispatch_result_end");
```
RekGRpth, line 691: Why was this line added before the `pfree` in the original patch, and here after?

bandetto: Moved it to the end of the function due to stylistic preference. There are no functional differences.

RekGRpth (approved):

VoidZeroNull0 (approved):

PR #328, opened 2026-03-17
GG-284: Fix distribution by resjunk column (7.x)

GirgerS:
In the PostgreSQL planner, some queries require auxiliary resjunk columns in
the targetlist.

These columns are necessary for query execution but do not appear in the
resulting relation.

Our logic for CTAS expressions, however, did not account this case.
If a query didn't contain an explicit DISTRIBUTED BY clause,
and other columns in the targetlist were not hashable,
the planner could use resjunk column in the distribution policy,
resulting in a segmentation fault.

To fix this issue, it is necessary to filter out resjunk columns.

This presents some complications, because it was decided that 
the distribution key for tables created with CTAS should be chosen 
at the Path tree. 

And the Paths tree don't contain information whether a pathtarget
node is a resjunk or not. 

To overcome this problem, use root->proccessed_tlist instead of
path->pathtarget to select the distribution key.


RekGRpth:
Why isn't this a cherry-pick of the patch https://github.com/GreengageDB/greengage/pull/318 with the changes? The tests are completely identical.

GirgerS:
> Why isn't this a cherry-pick of the patch #318 with the changes? The tests are completely identical.

Because the changes to `cdbllize.c` are trickier to verify compared to those in 6.x. So it not a plain port. And they are the main part of the patch.

src/backend/cdb/cdbllize.c
```diff
@@ -439,12 +439,16 @@ cdbllize_adjust_top_path(PlannerInfo *root, Path *best_path,
 				int			i;
 				List	   *policykeys = NIL;
 				List	   *policyopclasses = NIL;
-				ListCell   *lc;
 
-				i = 0;
-				foreach(lc, best_path->pathtarget->exprs)
+				for (i = 0; i < list_length(root->processed_tlist); i++)
```
RekGRpth: why not something like https://github.com/GreengageDB/greengage/blob/8e42739e4d06ce4c721973477470f0037368a1d8/src/backend/cdb/cdbllize.c#L208-L217

GirgerS: It originates from [6.x](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/cdb/cdbmutate.c#L489-L492). I believe that the intent behind this code is described [here](https://github.com/GreengageDB/greengage/blob/6.x/src/include/nodes/primnodes.h#L1300-L1312). That is, this way of iterating over the targetlist is preferred. 

I also believe that in our case it is excessive, based on the same commit and my research. But I left it that way anyway to keep 6.x and 7.x version similar.

RekGRpth: The `get_tle_by_resno` function does another loop over the list, so you end up with TWO loops over the same list!

GirgerS: > The get_tle_by_resno function does another loop over the list, so you end up with TWO loops over the same list!

Yes, so we get `O(N^2)` complexity for no reason. I doubt that it is that slow, as the `N` itself shouldn't be big enough to cause any problems, as it is basically `number of attributes in the table + resjunk`. 

Anyway, I don't like it either. But my reasoning stays the same, `6.x` (and maybe even earlier versions) has been doing the same thing for years, if I'm not missing anything. And if we are going to change this, then it should be done for both versions.

red1452:
```c
int attno; 
foreach_with_count(tle_cell, root->processed_tlist, attno)
```

src/backend/cdb/cdbllize.c
```diff
@@ -466,7 +470,6 @@ cdbllize_adjust_top_path(PlannerInfo *root, Path *best_path,
 						policyopclasses = lappend_oid(policyopclasses, opclass);
 						break;
 					}
-					i++;
 				}
 				targetPolicy = createHashPartitionedPolicy(policykeys,
 														   policyopclasses,
```
red1452, line 474: May we get a case, when `policykeys` and `policyopclasses` are NULL? May it lead to segfault?

src/backend/cdb/cdbllize.c
GirgerS: > int attno; 
foreach_with_count(tle_cell, root->processed_tlist, attno)

We fear that the list entry at the position `N` would have a different position in the final table, lets say `M`.  So a call to the `get_tle_by_resno` is performed to make sure that is is not the case. It is partially described [here](https://github.com/GreengageDB/greengage/blob/6.x/src/include/nodes/primnodes.h#L1300-L1312).

For my opinion on this, please refer to other replies.

src/backend/cdb/cdbllize.c
GirgerS: We actually want this to happen in the added test, as there is no column to distribute table by.

See https://github.com/GreengageDB/greengage/blob/7.x/src/include/nodes/pg_list.h#L88-L92

And https://github.com/GreengageDB/greengage/blob/8e42739e4d06ce4c721973477470f0037368a1d8/src/backend/cdb/cdbcat.c#L135-L140

GirgerS: So we do the same thing as if  `create table ... distributed randomly;` was specified.

src/backend/cdb/cdbllize.c
RekGRpth: > number of attributes in the table

may be up to `MaxHeapAttributeNumber` == `1600`

RekGRpth: > It originates from [6.x](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/cdb/cdbmutate.c#L489-L492). I believe that the intent behind this code is described [here](https://github.com/GreengageDB/greengage/blob/6.x/src/include/nodes/primnodes.h#L1300-L1312). That is, this way of iterating over the targetlist is preferred.
> 
> I also believe that in our case it is excessive, based on the same commit and my research. But I left it that way anyway to keep 6.x and 7.x version similar.

And what's the problem with optimizing 6X too, especially since a similar patch https://github.com/GreengageDB/greengage/pull/318 hasn't been merged yet?

GirgerS: > And what's the problem with optimizing 6X too, especially since a similar patch https://github.com/GreengageDB/greengage/pull/318 hasn't been merged yet?

It's not, lets do it! I just wanted to clarify that it is not entirely my idea to make O(N^2) algorithms out of blue.

GirgerS: And what do we do here?
 
https://github.com/GreengageDB/greengage/blob/8e42739e4d06ce4c721973477470f0037368a1d8/src/backend/cdb/cdbllize.c#L194-L214

RekGRpth: > And what do we do here?

What's there and how does it relate to the current patch?

GirgerS: > What's there

It also does O(N*M) lookup, though in this case is doesn't seem to be avoidable that easily. 

> and how does it relate to the current patch?

We also consider it when choosing a distribution key:
https://github.com/GreengageDB/greengage/blob/8e42739e4d06ce4c721973477470f0037368a1d8/src/backend/cdb/cdbllize.c#L428-L474

RekGRpth: I'm just suggesting making the NEW code MORE OPTIMAL.

RekGRpth (approved):

red1452 (approved):

PR #332, opened 2026-03-18
GG-306: Fix cost estimation of bitmap scans in ORCA

silent-observer:
Fix cost estimation of bitmap scans in ORCA

Previously ORCA had issues estimating cost of bitmap scans on indexes. ORCA
has a cost component proportional to the number of distinct values passing the
filter condition. This makes sense for bitmap and hash indexes, but not for
Btree, BRIN, GIST or GIN indexes. This component was already disabled for
Btree, and here we disable it for other index types as well.

The focus of this patch is to fix BRIN indexes specifically, so the test is
added only for them. GIST and GIN index cost estimation is improved in some
cases, but ORCA still often chooses Seq Scan instead of them because it
struggles to estimate the selectivity of conditions with custom operators.
Fixing them completely is a matter of a separate patch.


src/backend/gporca/libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -2153,7 +2153,8 @@ CCostModelGPDB::CostBitmapTableScan(CMemoryPool *mp, CExpressionHandle &exprhdl,
 			CDouble bitmap_union_cost_per_distinct_value(0.000027);
 			CDouble init_cost_advantage_for_bitmap_scan(0.9);
 
-			if (IMDIndex::EmdindBtree == indexType)
+			if (IMDIndex::EmdindBitmap != indexType &&
+				IMDIndex::EmdindHash != indexType)
 			{
 				// btree indexes are not sensitive to the NDV, since they don't have any bitmaps
```
whitehawk: Please update the comment here as well

whitehawk:
In the description:
>  but ORCA still often chooses Seq Scan instead of them because it
struggles to estimate the selectivity of conditions with custom operators.

Please add a note that it is a matter of a separate patch

src/test/regress/output/uao_dml/ao_covering_index_optimizer.source
```diff
@@ -463,13 +463,16 @@ VACUUM ANALYZE test_index_types_@amname@;
...
 EXPLAIN (ANALYZE, COSTS OFF, TIMING OFF, SUMMARY OFF)
 SELECT a, b FROM test_index_types_@amname@ WHERE a<@ box '(0,0,3,3)';
-                                QUERY PLAN                                
---------------------------------------------------------------------------
+                                       QUERY PLAN                                       
+----------------------------------------------------------------------------------------
  Gather Motion 3:1  (slice1; segments: 3) (actual rows=40000 loops=1)
-   ->  Seq Scan on test_index_types_@amname@ (actual rows=40000 loops=1)
-         Filter: (a <@ '(3,3),(0,0)'::box)
+   ->  Bitmap Heap Scan on test_index_types_@amname@ (actual rows=40000 loops=1)
```
whitehawk, line 469: Here the old plan with sequential scan was a bit better for both row and column tables - plan with sequential scan in my experiments executed for ~34-37ms, while the new plan executed ~44-47ms.

src/backend/gporca/libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -2153,7 +2153,8 @@ CCostModelGPDB::CostBitmapTableScan(CMemoryPool *mp, CExpressionHandle &exprhdl,
 			CDouble bitmap_union_cost_per_distinct_value(0.000027);
 			CDouble init_cost_advantage_for_bitmap_scan(0.9);
 
-			if (IMDIndex::EmdindBtree == indexType)
+			if (IMDIndex::EmdindBitmap != indexType &&
+				IMDIndex::EmdindHash != indexType)
```
GirgerS, line 2157: Just to be completely sure, did we check that all other indexes are not slowed down by the number of distinct values? BRIN seems fine from my research, what about GIN, Gist?

src/backend/gporca/libgpdbcost/src/CCostModelGPDB.cpp
GirgerS, line 2142: The name of this value is misleading. I don't know what was the original intention behind it, but now it acts as a `PenaltyForAmountOfDistinctValues` for bitmap index.

Do we want to change it now, or it is better done during dedicated bitmap index adjustments?   

It also means that we don't cost the bitmap required by the scan itself, but it is fine as we didn't find any noticeable performance issues related to it.

GirgerS: We also adjust the cost of the bitmap index here:

 https://github.com/GreengageDB/greengage/pull/332/changes#diff-0848114b5540cb3c740a60679523e54a2a139e871ea3d9479c8c175732e89cfdR2182-R2188

And I don't know what to make of it.

silent-observer: Could that be that quadratic component you found? It has dNDV * rows in it, so it's probably going to get large compared to other parts of the cost as NDV and rows are increased

src/test/regress/output/uao_dml/ao_covering_index_optimizer.source
silent-observer: I think this is just the case of Seq Scans being really cheap on small tables, but optimizers have minimum cost for it (~431 for ORCA). This also matches behavior on Postgres (Seq Scan has cost 187 and Bitmap Heap Scan has cost 23.94), so it's probably fine

src/backend/gporca/libgpdbcost/src/CCostModelGPDB.cpp
silent-observer: I did check it, and also it seems reasonable from their structure that they shouldn't have this dependence (both are based on Btrees)

src/backend/gporca/libgpdbcost/src/CCostModelGPDB.cpp
silent-observer: As for `PenaltyForAmountOfDistinctValues`, that depends on whether we will even *need* dedicated bitmap adjustments. If we conclude they are working fine already, then we can change it here.

src/backend/gporca/libgpdbcost/src/CCostModelGPDB.cpp
silent-observer: Fixed

src/backend/gporca/libgpdbcost/src/CCostModelGPDB.cpp
GirgerS: > Could that be that quadratic component you found? It has dNDV * rows in it, so it's probably going to get large compared to other parts of the cost as NDV and rows are increased

It is. But what is the purpose of the `c3_dBitmapPageCost` then? Represent the cost of the bitmap itself?

And because I can't think of a reasonable explanation right now, let's postpone this question for the further patch. Of course if you don't have any objections or ideas.

src/test/regress/output/uao_dml/ao_covering_index_optimizer.source
GirgerS: Also, cost of the bitmap scan is explicitly made cheaper: 

https://github.com/GreengageDB/greengage/blob/GG-306/src/backend/gporca/libgpdbcost/src/CCostModelGPDB.cpp#L2163-L2166

So this piece of code is finally working, unlike before.

whitehawk (approved):

GirgerS (approved):

PR #337, opened 2026-03-19
CI-5400 Collect logs for tests even if they hang (6.x)

uaqq:
### Collect logs for tests even if they hang (#337)

- update behave and regression tests workflow;
- add CI condition for behave tests (do not stop containers in CI).
- remove legacy log collection from `ci/scripts/run_behave_tests.bash`;
- remove `behave_collect_logs.bash` script;
- fix sql dump artifact name.

Introduce a composite action to collect logs from containers,
even for failed or canceled test runs.

### Implementation details

- iterate over Docker containers on the host;
- execute log collection inside each container.

### Tests

Canceled:
- behave: https://github.com/arenadata/greengage/actions/runs/23237137733/job/67545465246?pr=4
- regression: https://github.com/arenadata/greengage/actions/runs/23237137733/job/67545393363?pr=4

Task: CI-5283


silent-observer:
This isn't an issue of this specific patch, but not all logs are currently collected. In particular
1. Extensions placed in contrib or gpcontrib can also have `results` and `regression.diffs`. I think it's better to look for them in `gpdb_src`, not `gpdb_src/src/test`.
2. Behave tests running on concourse cluster place their logs (`log`, `pg_log`) in `/data` instead of `gpdb_src/gpAux/gpdemo/datadirs`, these should be collected as well.
Should this be resolved as a separate task?

magf:
> This isn't an issue of this specific patch, but not all logs are currently collected. In particular
>  ...
>        Should this be resolved as a separate task?

A more comprehensive approach is needed.
A YAML configuration file describing all the tests, for example.
I'm ready to develop this.

magf (approved):

silent-observer (approved):

PR #342, opened 2026-03-20
GG-318: Use gpsubprocess instead of subprocess

KnightMurloc:
Decoding the results of the subcommands has negative consequences in the
Python 2 environment. In the Python 2, decode returns a unicode string whereas
previously it returned a regular string. This leads to unexpected errors when
trying to convert a unicode string to a byte string if it contains non-ascii
characters. This patch changes the approach. Instead of decoding the results of
the subcommands, the Python 3 environment now uses text mode to work with
subprocesses and files. In most uses of subprocess.Popen has been replaced with
the already existing wrapper gpsubprocess.Popen. A number of improvements have
been made to this end.
It now use a text mode in the Python 3 environment. There is a new wrapper for
check_output that also uses text mode for Python 3. Remove the futurize changes
from the pexpect module since this module already supports Python 3. In the
gpssh module, decode the result only for Python 3 to avoid the error of
converting unicode to ascii when printing.

The gpoperation mechanism has been reworked. Instead of sending raw pickled
serialized data, this data is now sent in base64 format. This was done so that
the data could be sent in text mode.
In gpssh, the decoding of the result is only for Python 3. In Python 2,
outputting unicode strings via print can lead to an error.


src/test/isolation2/sql_isolation_testcase.py
ShapatinAA: Are we still getting here bytes? I removed decode and tried to print it but got bytes - is this right behavior?

gpcontrib/gp_replica_check/gp_replica_check.py
```diff
@@ -74,7 +75,7 @@ def run(self):
             print("Primary segment for content %d is down" % self.content)
         else:
             try:
-                res = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True).decode('utf-8')
+                res = gpsubprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
```
ShapatinAA, line 78: should we use gpsubprocess for stredd/stdin also?

KnightMurloc: You are about `subprocess.STDOUT`? If yes, then no. These constants are not defined in gpsubprocess.

src/test/isolation2/sql_isolation_testcase.py
KnightMurloc: `os.read` returns bytes regardless of the text mode.

gpMgmt/bin/gppylib/operations/utils.py
```diff
@@ -46,10 +46,9 @@ def __init__(self, operation, host, msg_ctx=""):
         
     def execute(self):
         execname = os.path.split(sys.argv[0])[-1]
-        pickled_execname = pickle.dumps(execname) 
-        pickled_operation = pickle.dumps(self.operation)
+        pickled_data = base64.urlsafe_b64encode(pickle.dumps((execname, self.operation))).decode('ascii')
```
ShapatinAA, line 49: `dumps()` method says, that it uses default protocol if not specified and depending on python version protocols may be not compatible with each other. Should we consider that?

KnightMurloc: Data is serialized and deserialized in the single Python version. Running different utilities in different versions of Python is not the norm and we does not support it.

ShapatinAA:
Check gpMgmts unit tests on python utilities on ubuntu24. In docker they failed with 
```python
ERROR: Test Suite Name|commands.test.unit.test_unit_base.WorkerPoolTestCase|Test Case Name|test_RemoteExecutionContext_uses_default_gphome|Test Details|
----------------------------------------------------------------------
  File "/home/gpadmin/gpdb_src/gpMgmt/bin/gppylib/commands/test/unit/test_unit_base.py", line 18, in test_RemoteExecutionContext_uses_default_gphome
    self.subject.execute(cmd)
  File "/home/gpadmin/gpdb_src/gpMgmt/bin/gppylib/commands/base.py", line 512, in execute
    LocalExecutionContext.execute(self, cmd)
  File "/home/gpadmin/gpdb_src/gpMgmt/bin/gppylib/commands/base.py", line 462, in execute
    (rc, stdout_value, stderr_value) = self.proc.communicate2(input=self.stdin)
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gpadmin/gpdb_src/gpMgmt/bin/gppylib/gpsubprocess.py", line 64, in communicate2
    self.stdin.write(input)
TypeError: write() argument must be str, not bytes
```

gpMgmt/bin/gpload.py
ShapatinAA, line 1345: While we at it, can we swap this conditions?

KnightMurloc:
> Check gpMgmts unit tests on python utilities on ubuntu24. In docker they failed with

Fixed

gpMgmt/bin/gpload.py
KnightMurloc: what do you mean?

ShapatinAA: Ain't this will result an error on python3? `NameError: name 'unicode' is not defined`

ShapatinAA, line 2901: Python 2 - is it correct?

KnightMurloc: We define unicode for Python 3 at the beginning of the file:
https://github.com/GreengageDB/greengage/blob/04e4184f9431e8b522f3f14a3ee8ee8501e0072a/gpMgmt/bin/gpload.py#L32-L36

gpMgmt/bin/gpmemreport
```diff
@@ -428,7 +427,7 @@ if __name__ == '__main__':
     (options, args) = parseCmdLine()
     start = datetime.strptime(options.start_time, datetime_format) if options.start_time else None
     end = datetime.strptime(options.end_time, datetime_format) if options.end_time else None
-    f = gzip.open(args[0], 'r')
+    f = gzip.open(args[0], 'rt')
```
ShapatinAA, line 430: Why don't we need to open it in text anymore?

KnightMurloc: What do you mean? I added opening in the text mode.

gpMgmt/sbin/packcore
ShapatinAA: We can also remove decode from here?

gpMgmt/bin/gpmemreport
ShapatinAA: Ah, sorry, read diff wrongfully.

src/backend/gporca/scripts/get_debug_event_counters.py
ShapatinAA:
```suggestion
		utf8_line = line
```

ShapatinAA: If change interpreter and remove version=1.0, then it will error out on this line

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -2120,6 +2122,8 @@ def impl(context, filename, contain, output):
         cmd.run(validateAfter=True)
 
         actual = cmd.get_stdout()
+        if sys.version_info[0] == 2:
+            actual = actual.decode('utf-8')
```
ShapatinAA, line 2126: why we're decoding it here, but a little above - we're not doing it?

ShapatinAA:
https://github.com/GreengageDB/greengage/blob/372243324133b66f77d852f9977ea0d6d983aef5/gpMgmt/test/behave/mgmt_utils/steps/mirrors_mgmt_utils.py#L95

Remove here too?

ShapatinAA:
I think we need to change this kind of lines also.
https://github.com/GreengageDB/greengage/blob/372243324133b66f77d852f9977ea0d6d983aef5/gpAux/gpdemo/gpsegwalrep.py#L53

ShapatinAA:
Can we also update title of PR or change scope of GG-318 task?

gpMgmt/bin/gpload.py
KnightMurloc: Yes. Decode was here even before the addition of Python 3 support. We are doing decode only for Python 2, since in Python 3 it is already unicode.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
KnightMurloc: Decode was here even before Python 3 support was added. We only do decode for Python 2 since it's already unicode in Python 3.

KnightMurloc:
> https://github.com/GreengageDB/greengage/blob/372243324133b66f77d852f9977ea0d6d983aef5/gpMgmt/test/behave/mgmt_utils/steps/mirrors_mgmt_utils.py#L95
> 
> Remove here too?

This code was like this even before Python 3 was supported. I'm not sure if it's worth changing.

KnightMurloc:
> I think we need to change this kind of lines also.
> 
> https://github.com/GreengageDB/greengage/blob/372243324133b66f77d852f9977ea0d6d983aef5/gpAux/gpdemo/gpsegwalrep.py#L53

It is a 5X utility and does not work in 6X.

ShapatinAA:
What about here? Seems like this commit suits to be changed. https://github.com/GreengageDB/greengage/commit/3a7bd87028048b76139857b6e9211c391270aaa3
https://github.com/GreengageDB/greengage/blob/04e4184f9431e8b522f3f14a3ee8ee8501e0072a/concourse/pipelines/gen_pipeline.py#L62

KnightMurloc:
> What about here? Seems like this commit suits to be changed. [3a7bd87](https://github.com/GreengageDB/greengage/commit/3a7bd87028048b76139857b6e9211c391270aaa3)
> 
> https://github.com/GreengageDB/greengage/blob/04e4184f9431e8b522f3f14a3ee8ee8501e0072a/concourse/pipelines/gen_pipeline.py#L62

We don't use concourse pipelines.

gpMgmt/sbin/packcore
KnightMurloc: fixed

src/backend/gporca/scripts/get_debug_event_counters.py
KnightMurloc: fixed

src/test/isolation2/sql_isolation_testcase.py
ShapatinAA, line 162: do we need to add decode here [then](https://github.com/GreengageDB/greengage/pull/342#discussion_r3020908363)?

KnightMurloc: fixed

gpMgmt/bin/gppylib/operations/utils.py
dkovalev1, line 49: Could you explain in plain text what happens with `execname` here and why is it best transformation:
dumps->b64encode->decode ?

gpMgmt/bin/gppylib/test/unit/test_unit_gpinitsystem.py
```diff
@@ -4,9 +4,10 @@
 from .gp_unittest import *
 from io import StringIO
 try:
-    from subprocess32 import Popen, PIPE
+    from subprocess32 import PIPE
 except:
-    from subprocess import Popen, PIPE
+    from subprocess import PIPE
+from gppylib.gpsubprocess import Popen
```
dkovalev1: Rule 2 from "The Zen of Python":
`Explicit is better than implicit.`

IMHO if remove this line and use `gpsubprocess.Popen()` over the code, it is more clearly states the intent of using custom `Popen()`

gpMgmt/bin/gppylib/util/ssh_utils.py
```diff
@@ -302,7 +302,10 @@ def executeCommand(self, command):
         for s in self.pxssh_list:
             # Split the output into an array of lines so that we can add text to the beginning of
             #    each line
-            output = s.before.decode('utf-8').split('\n')
+            output_raw = s.before
```
dkovalev1, line 305: `s.before` comes from `gppxssh_wrapper.PxsshWrapper` which is inherited from `pxssh.pxssh` which in its turn has a parameter `encoding` to deal with different text encodings, and it has py3 support somehow.

Have you considered tweaking `encoding` parameter of `pxssh.pxssh` to have strings in both pythons?

src/backend/gporca/scripts/get_debug_event_counters.py
```diff
@@ -126,7 +127,9 @@ def processLogFile(logFileLines, allruns):
 	current_output = []
 
 	for line in logFileLines:
-		utf8_line = line.decode('utf-8')
+		utf8_line = line
+		if sys.version_info[0] == 2:
+			utf8_line = utf8_line.decode('utf-8')
```
dkovalev1, line 132: `gpsubprocess.Popen` now works in the text mode, do we still need to decode it?

gpMgmt/bin/gppylib/operations/utils.py
KnightMurloc: To perform an operation on a remote host, we run on it `gpoperation.py ` and we pass to its stdin the serialized `Operation` object that needs to be performed. Since we enabled a text mode in subprocess, stdin has also become text-based and it should be possible to encode it in utf-8. `pickle.dumps` returns bytes that cannot be represented as utf-8. By encoding it in base64, we make sure that the data can be represented as utf-8.

src/backend/gporca/scripts/get_debug_event_counters.py
KnightMurloc: It work in text mode only for Python 3.

gpMgmt/bin/gppylib/util/ssh_utils.py
KnightMurloc: It seems that `pxssh` uses spawn as the parent class, whereas encoding support is only available in `spawnu`. It seems to me that we don't have a simple way to enable a text mode in ssh.

ShapatinAA (approved):

RekGRpth (approved):

ShapatinAA:
BTW fix PR title.

KnightMurloc:
> BTW fix PR title.

Done

dkovalev1 (approved):

PR #343, opened 2026-03-23
GG-323: Fix potential catalog corruption with temporary identity columns

RekGRpth:
Fix potential catalog corruption with temporary identity columns

If a temporary table with an identity column and ON COMMIT DROP is
created in a single-statement transaction (not useful, but allowed),
it would leave the catalog corrupted.  We need to add a
CommandCounterIncrement() so that PreCommit_on_commit_actions() sees
the created dependency between table and sequence and can clean it
up.

The analogous and more useful case of doing this in a transaction
block already runs some CommandCounterIncrement() before it gets to
the on-commit cleanup, so it wasn't a problem in practical use.

Several locations for placing the new CommandCounterIncrement() call
were discussed.  This patch places it at the end of
standard_ProcessUtility().  That would also help if other commands
were to create catalog entries that some on-commit action would like
to see.

Bug: #15631
Reported-by: Serge Latyntsev <dnsl48@gmail.com>
Author: Peter Eisentraut <peter.eisentraut@2ndquadrant.com>
Reviewed-by: Michael Paquier <michael@paquier.xyz>
(cherry picked from commit cd3e27464cca40664c54fb8cd10454f979c1db4e)

changes from original commit
- Add a test to check that the sequence does not exist after the table
  auto-drop.

Ticket: GG-323

### DIFF
--- a/src/backend/tcop/utility.c
+++ b/src/backend/tcop/utility.c
@@ -1096,6 +1096,13 @@ standard_ProcessUtility(Node *parsetree,
 							   dest, completionTag);
 			break;
 	}
+
+	/*
+	 * Make effects of commands visible, for instance so that
+	 * PreCommit_on_commit_actions() can see them (see for example bug
+	 * #15631).
+	 */
+	CommandCounterIncrement();
 }
 
 /*
--- a/src/test/regress/expected/temp.out
+++ b/src/test/regress/expected/temp.out
@@ -113,6 +113,13 @@ SELECT * FROM temptest;
 ERROR:  relation "temptest" does not exist
 LINE 1: SELECT * FROM temptest;
                       ^
+-- Check that the sequence does not exist after auto-dropping the table.
+CREATE TEMP TABLE temptest(col serial) ON COMMIT DROP;
+SELECT relname, relkind, relpersistence FROM pg_class WHERE relname = 'temptest_col_seq';
+ relname | relkind | relpersistence 
+---------+---------+----------------
+(0 rows)
+
 -- ON COMMIT is only allowed for TEMP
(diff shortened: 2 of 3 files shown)


VoidZeroNull0:
Maybe, add a little test to be sure we won't have dangling sequence in future (even though it is extremely rare and artificial case)? Or is it an overkill?

RekGRpth:
> Maybe, add a little test to be sure we won't have dangling sequence in future (even though it is extremely rare and artificial case)? Or is it an overkill?

Can you provide an example of a test? It's not so easy to reproduce the original problem on the 6X because it doesn't support identity column.

VoidZeroNull0:
> Can you provide an example of a test? It's not so easy to reproduce the original problem on the 6X because it doesn't support identity column.

Is not `SERIAL` the analogue of `AS IDENTITY` in Postgres 9?

```PostgresSQL
postgres=# CREATE TEMP TABLE test (id SERIAL) ON COMMIT DROP DISTRIBUTED BY (id);
SELECT relname, relkind, relpersistence 
FROM pg_class 
WHERE relname LIKE 'test%';
CREATE TABLE
   relname    | relkind | relpersistence 
--------------+---------+----------------
 test_id_seq  | S       | t
 test_id_seq1 | S       | t
 test_id_seq2 | S       | t
(3 rows)

```

RekGRpth:
> SELECT relname, relkind, relpersistence 
> FROM pg_class 
> WHERE relname LIKE 'test%';

added

VoidZeroNull0 (approved):

red1452 (approved):

PR #352, opened 2026-03-27
GG-326: Add separate queue support for the communication channel

silent-observer:
Add separate queue support for the communication channel

Multiple extensions might use the metadata communication channel, possibly in
the same session and at the same time. To avoid metadata messages mixing
between extensions, we should add an API to create/delete separate message
queues for extensions. Existing APIs should also be adjusted to work with
separate queues instead of a global one.

Each queue is identified with its own queue_id. Since extensions might want to
preallocate this id in advance (for example, to pass it from master to
segments during planning stage), allocating a queue_id is separate from
actually creating a queue. All actions with metadata queue now require a
queue_id.

### DIFF
--- a/src/backend/cdb/dispatcher/cdbconn.c
+++ b/src/backend/cdb/dispatcher/cdbconn.c
@@ -36,7 +36,7 @@ static uint32 cdbconn_get_motion_listener_port(PGconn *conn);
 static void cdbconn_disconnect(SegmentDatabaseDescriptor *segdbDesc);
 
 static void MPPnoticeReceiver(void *arg, const PGresult *res);
-static void MPPmetadataReceiver(void *arg, ggMetadataChunk *, ggMetadataQueueId);
+static void MPPmetadataReceiver(void *arg, ggMetadataChunk *metadata_chunk, ggMetadataQueueId queue_id);
 
 static const char *
 transStatusToString(PGTransactionStatusType status)
@@ -873,7 +873,8 @@ forwardQENotices(void)
 		pq_flush();
 }
 
-typedef struct ggMetadataQueue {
+typedef struct ggMetadataQueue
+{
 	ggMetadataQueueId id;
 	int count;
 	ggMetadataChunk *chunks;
--- a/src/backend/cdb/dispatcher/cdbconn.c
+++ b/src/backend/cdb/dispatcher/cdbconn.c
@@ -915,7 +915,7 @@ PQMetadataWalk(ggMetadataQueueId queue_id)
 {
 	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
 	if (!queue)
-		elog(ERROR, "No metadata queue with id %d", queue_id);
+		elog(ERROR, "No metadata queue with id %u", queue_id);
 	return queue->chunks;
 }
 
@@ -947,7 +947,7 @@ PQgetMetadataCount(ggMetadataQueueId queue_id)
 {
 	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
 	if (!queue)
-		elog(ERROR, "No metadata queue with id %d", queue_id);
+		elog(ERROR, "No metadata queue with id %u", queue_id);
 	return queue->count;
 }
 
@@ -1003,7 +1003,7 @@ PQDeleteMetadataQueue(ggMetadataQueueId queue_id)
 {
 	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
 	if (!queue)
-		elog(ERROR, "No metadata queue with id %d", queue_id);
+		elog(ERROR, "No metadata queue with id %u", queue_id);
 
 	PQCleanMetadataInternal(queue);
 
--- a/src/backend/cdb/dispatcher/cdbconn.c
+++ b/src/backend/cdb/dispatcher/cdbconn.c
@@ -36,7 +36,7 @@ static uint32 cdbconn_get_motion_listener_port(PGconn *conn);
 static void cdbconn_disconnect(SegmentDatabaseDescriptor *segdbDesc);
 
 static void MPPnoticeReceiver(void *arg, const PGresult *res);
-static void MPPmetadataReceiver(void *arg, ggMetadataChunk *);
+static void MPPmetadataReceiver(void *arg, ggMetadataChunk *, ggMetadataQueueId);
 
 static const char *
 transStatusToString(PGTransactionStatusType status)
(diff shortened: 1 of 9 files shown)


src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2381,6 +2381,8 @@ pgGetMetadataMessage(PGconn *conn, int length)
 		return pqSkipnchar(length, conn);
 	}
 
+	int payload_len = length - 4;
```
whitehawk:
```suggestion
	int payload_len = length - sizeof(ggMetadataQueueId);
```
?

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2389,21 +2391,29 @@ pgGetMetadataMessage(PGconn *conn, int length)
...
 	if (chunk == NULL)
 		return 1;
 
 	chunk->next = NULL;
-	chunk->metadataLen = length;
+	chunk->metadataLen = payload_len;
+
+	int32 queue_id;
+
+	if (pqGetInt(&queue_id, 4, conn))
```
whitehawk:
```suggestion
	if (pqGetInt(&queue_id, sizeof(ggMetadataQueueId), conn))
```
?

whitehawk:
> Since extensions might want to preallocate this id in advance (for example, to pass it from master to segments during planning stage), allocating a queue_id is separate from actually creating a queue.

Is it really needed? Why can't the pre-allocation be done together with the queue creation? What is the benefit in postponing the queue creatio in this case?

Also, what please specify what will be the proper way to pass the queue id from coordinator to the segments.

silent-observer:
> Is it really needed? Why can't the pre-allocation be done together with the queue creation? What is the benefit in postponing the queue creatio in this case?

Yes. For proper communication, both master and segments should know queue_id. Currently used method of communication for FDWs is passing data in fdw_private during the planning stage: master fills in fdw_private and it is passed to segments as part of the plan. This mechanism is already used, for example, in adb_fdw to pass the name of the parallel cursor. There aren't really any mechanisms to achieve that communication between master and segments during query execution. That's why queue allocation must be done in the planning stage.

However, queue creation can't be done in the planning stage, since sometimes the plans are cached and so the planning routines of the FDWs are not called again. In the current implementation it simply reuses the same queue_id as before (which is acceptable since it won't overlap with other extensions anyway), but if queue was created in the planning stage, it wouldn't be created at all, leading to errors. In short, we should know queue_id during planning to pass it in fdw_private, but we should create the queue during execution.

> Also, what please specify what will be the proper way to pass the queue id from coordinator to the segments.

That will depend on the exact usecase, but since we intend to use this for FDWs, it will be passed in fdw_private. If the usecase allows to pre-allocate the queue at the same time as create it, no passing is needed.

src/interfaces/libpq/fe-protocol3.c
silent-observer: Fixed

src/interfaces/libpq/fe-protocol3.c
silent-observer: Fixed

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -917,21 +942,70 @@ PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor *out_desc)
...
+
+	PQCleanMetadataInternal(queue);
+}
+
+
+ggMetadataQueueId
+PQMetadataNextQueueId()
+{
+	static ggMetadataQueueId queueId = 0;
+	return queueId++;
```
whitehawk, line 983: what about risk of wraparound? In this case we might face collisions.

silent-observer: This queueId is local for each master connection. The only way to achieve collision here is to open a cursor with FDW, then do 4 billion commands using FDW in the same session *without closing the cursor*, and only then we encounter a collision. Reusing the same queueId after the query is complete (and queue is deleted) is safe.

P.S. Similar logic was already used in adb_fdw:
```c
 * Note that even if wraparound happens in a very long transaction, actual
 * collisions are highly improbable; just be sure to use %u not %d to print.
 */
unsigned int
GetCursorNumber(PGconn *conn)
{
	return ++cursor_number;
}
```

whitehawk (approved):

whitehawk:
LGTM for the MVP level.

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -36,7 +36,7 @@ static uint32 cdbconn_get_motion_listener_port(PGconn *conn);
 static void cdbconn_disconnect(SegmentDatabaseDescriptor *segdbDesc);
 
 static void MPPnoticeReceiver(void *arg, const PGresult *res);
-static void MPPmetadataReceiver(void *arg, ggMetadataChunk *);
+static void MPPmetadataReceiver(void *arg, ggMetadataChunk *, ggMetadataQueueId);
```
RekGRpth: Why not add argument names?
```suggestion
static void MPPmetadataReceiver(void *arg, ggMetadataChunk *metadata_chunk, ggMetadataQueueId queue_id);
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -873,24 +873,49 @@ forwardQENotices(void)
 		pq_flush();
 }
 
-static ggMetadataChunk *ggMetadataList = NULL;
-static int ggMetadataCount = 0;
+typedef struct ggMetadataQueue {
```
RekGRpth:
```suggestion
typedef struct ggMetadataQueue
{
```

src/backend/cdb/dispatcher/cdbconn.c
silent-observer: Fixed

src/backend/cdb/dispatcher/cdbconn.c
silent-observer: Fixed

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -917,21 +943,70 @@ PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor *out_desc)
...
+	queue->chunks = NULL;
+	queue->count = 0;
+}
+
+void
+PQCleanMetadata(ggMetadataQueueId queue_id)
+{
+	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
+	if (!queue)
+		elog(ERROR, "No metadata queue with id %d", queue_id);
```
dkovalev1: queue_id is unsigned, %u format is better

silent-observer: Fixed

dkovalev1 (approved):

PR #353, opened 2026-03-27
GG-259: Orca always scans replicated tables on the 0 segment for singleton slices

GirgerS:
Previously, inputs for duplicate motion hazard, tainted replicated,  
and non-coordinator singleton distributions were always taken 
from the first segment (with id 0). 

At the same time, in the PostgreSQL-based planner this value 
is computed using current session id. 

This patch ports aforementioned PostgreSQL planner logic to ORCA 
by making CTranslatorDXLToExpr class store session id. 
However, the instance of this class, created inside CConstExprEvaluatorDXL 
is left with a placeholder session id equal to zero, 
since it is currently used only for scalar operator evaluation. 
We can always change it when we need to. 

Other notable change is updating ORCA unit tests 
(as they use a hardcoded session id that is equal to '1')

Ticket: GG-259

### DIFF
--- a/src/backend/gpopt/utils/COptTasks.cpp
+++ b/src/backend/gpopt/utils/COptTasks.cpp
@@ -798,12 +798,16 @@ COptTasks::OptimizeTask(void *ptr)
 			CMDAccessor mda(mp, CMDCache::Pcache(), default_sysid,
 							relcache_provider);
 
+			// For some unknown reason the code below assumes that the following values are
+			// unsigned ints, when in reality they are signed.
+			// It easier to do calculation this way, though, so let it be.
 			ULONG num_segments = gpdb::GetGPSegmentCount();
 			ULONG num_segments_for_costing = optimizer_segments;
 			if (0 == num_segments_for_costing)
 			{
 				num_segments_for_costing = num_segments;
 			}
+			ULONG session_id = gpdb::GetGPSessionId();
 
 			CAutoP<CTranslatorQueryToDXL> query_to_dxl_translator;
 			query_to_dxl_translator = CTranslatorQueryToDXL::QueryToDXLInstance(
@@ -839,7 +843,7 @@ COptTasks::OptimizeTask(void *ptr)
 
 			plan_dxl = COptimizer::PdxlnOptimize(
 				mp, &mda, query_dxl, query_output_dxlnode_array,
-				cte_dxlnode_array, expr_evaluator, num_segments, gp_session_id,
+				cte_dxlnode_array, expr_evaluator, num_segments, session_id,
 				MyProc->queryCommandId, search_strategy_arr, optimizer_config);
 
 			if (opt_ctxt->m_should_serialize_plan_dxl)
--- a/src/backend/gpopt/utils/COptTasks.cpp
+++ b/src/backend/gpopt/utils/COptTasks.cpp
@@ -807,7 +807,6 @@ COptTasks::OptimizeTask(void *ptr)
 			{
 				num_segments_for_costing = num_segments;
 			}
-			ULONG session_id = (ULONG) gpdb::GetGPSessionId();
 
 			CAutoP<CTranslatorQueryToDXL> query_to_dxl_translator;
 			query_to_dxl_translator = CTranslatorQueryToDXL::QueryToDXLInstance(
@@ -843,7 +842,7 @@ COptTasks::OptimizeTask(void *ptr)
 
 			plan_dxl = COptimizer::PdxlnOptimize(
 				mp, &mda, query_dxl, query_output_dxlnode_array,
-				cte_dxlnode_array, expr_evaluator, num_segments, session_id,
+				cte_dxlnode_array, expr_evaluator, num_segments, (ULONG) gp_session_id,
 				MyProc->queryCommandId, search_strategy_arr, optimizer_config);
 
 			if (opt_ctxt->m_should_serialize_plan_dxl)
--- a/src/backend/gpopt/utils/COptTasks.cpp
+++ b/src/backend/gpopt/utils/COptTasks.cpp
@@ -807,7 +807,7 @@ COptTasks::OptimizeTask(void *ptr)
 			{
 				num_segments_for_costing = num_segments;
 			}
-			ULONG session_id = gpdb::GetGPSessionId();
+			ULONG session_id = (ULONG) gpdb::GetGPSessionId();
 
 			CAutoP<CTranslatorQueryToDXL> query_to_dxl_translator;
 			query_to_dxl_translator = CTranslatorQueryToDXL::QueryToDXLInstance(
--- a/src/backend/gpopt/utils/COptTasks.cpp
+++ b/src/backend/gpopt/utils/COptTasks.cpp
@@ -842,8 +842,9 @@ COptTasks::OptimizeTask(void *ptr)
 
 			plan_dxl = COptimizer::PdxlnOptimize(
 				mp, &mda, query_dxl, query_output_dxlnode_array,
-				cte_dxlnode_array, expr_evaluator, num_segments, (ULONG) gp_session_id,
-				MyProc->queryCommandId, search_strategy_arr, optimizer_config);
+				cte_dxlnode_array, expr_evaluator, num_segments,
+				(ULONG) gp_session_id, MyProc->queryCommandId,
+				search_strategy_arr, optimizer_config);
 
 			if (opt_ctxt->m_should_serialize_plan_dxl)
 			{
--- a/src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp
+++ b/src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp
@@ -43,6 +43,7 @@ CConstExprEvaluatorDXL::CConstExprEvaluatorDXL(
 	IConstDXLNodeEvaluator *pconstdxleval)
 	: m_pconstdxleval(pconstdxleval),
 	  m_trexpr2dxl(mp, md_accessor, nullptr /*pdrgpiSegments*/,
+				   0, /* ulSegmentId*/
 				   false /*fInitColumnFactory*/),
 	  m_trdxl2expr(mp, md_accessor, false /*fInitColumnFactory*/)
 {
--- a/src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp
+++ b/src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp
@@ -43,7 +43,7 @@ CConstExprEvaluatorDXL::CConstExprEvaluatorDXL(
 	IConstDXLNodeEvaluator *pconstdxleval)
 	: m_pconstdxleval(pconstdxleval),
 	  m_trexpr2dxl(mp, md_accessor, nullptr /*pdrgpiSegments*/,
-				   0, /* ulSegmentId*/
+				   0, /* ulSessionId*/
 				   false /*fInitColumnFactory*/),
 	  m_trdxl2expr(mp, md_accessor, false /*fInitColumnFactory*/)
 {
--- a/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
+++ b/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
@@ -169,7 +169,8 @@ CMiniDumperDXLTest::EresUnittest_Basic()
 			pdrgpiSegments->Append(GPOS_NEW(mp) INT(ul));
 		}
 
-		CTranslatorExprToDXL ptrexprtodxl(mp, &mda, pdrgpiSegments);
+		CTranslatorExprToDXL ptrexprtodxl(mp, &mda, pdrgpiSegments,
+										  1 /* ulSessionId */);
 		CDXLNode *pdxlnPlan = ptrexprtodxl.PdxlnTranslate(
 			pexprPlan, pqc->PdrgPcr(), pqc->Pdrgpmdname());
 		GPOS_UNITTEST_ASSERT(nullptr != pdxlnPlan);
--- a/src/backend/gpopt/gpdbwrappers.cpp
+++ b/src/backend/gpopt/gpdbwrappers.cpp
@@ -1100,17 +1100,6 @@ gpdb::GetGPSegmentCount(void)
 	return 0;
 }
 
-int
-gpdb::GetGPSessionId(void)
-{
-	GP_WRAP_START;
-	{
-		return gp_session_id;
-	}
-	GP_WRAP_END;
-	return 0;
-}
-
 bool
 gpdb::HeapAttIsNull(HeapTuple tup, int attno)
 {
--- a/src/backend/gpopt/gpdbwrappers.cpp
+++ b/src/backend/gpopt/gpdbwrappers.cpp
@@ -1100,6 +1100,17 @@ gpdb::GetGPSegmentCount(void)
 	return 0;
 }
 
+int
+gpdb::GetGPSessionId(void)
+{
+	GP_WRAP_START;
+	{
+		return gp_session_id;
+	}
+	GP_WRAP_END;
+	return 0;
+}
+
 bool
 gpdb::HeapAttIsNull(HeapTuple tup, int attno)
 {
--- a/src/backend/gporca/data/dxl/minidump/ArrayCoerceCast.mdp
+++ b/src/backend/gporca/data/dxl/minidump/ArrayCoerceCast.mdp
@@ -311,7 +311,7 @@
               </dxl:ProjList>
               <dxl:Filter/>
               <dxl:OneTimeFilter/>
-              <dxl:RandomMotion InputSegments="0" OutputSegments="0,1,2" DuplicateSensitive="true">
+              <dxl:RandomMotion InputSegments="1" OutputSegments="0,1,2" DuplicateSensitive="true">
                 <dxl:Properties>
                   <dxl:Cost StartupCost="0" TotalCost="0.000004" Rows="1.000000" Width="1"/>
                 </dxl:Properties>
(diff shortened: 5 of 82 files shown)


src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp
```diff
@@ -43,6 +43,7 @@ CConstExprEvaluatorDXL::CConstExprEvaluatorDXL(
 	IConstDXLNodeEvaluator *pconstdxleval)
 	: m_pconstdxleval(pconstdxleval),
 	  m_trexpr2dxl(mp, md_accessor, nullptr /*pdrgpiSegments*/,
+				   0, /* ulSegmentId*/
```
RekGRpth: maybe
```suggestion
				   0, /* ulSessionId*/
```
?

GirgerS: Thanks, I guess changing comments at 4 am is not the greatest idea.

GirgerS:
Somewhat noteworthy, outputs of singleton slices are still [hardcoded](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp#L7616-L7622) to the first segment. I decided to leave it as it is, as the ticket is mainly concerned with inputs. But maybe we want to do something with it too.

src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp
RekGRpth: Could you explain why there is still hardcode here?

RekGRpth:
> Somewhat noteworthy, outputs of singleton slices are still [hardcoded](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp#L7616-L7622) to the first segment. I decided to leave it as it is, as the ticket is mainly concerned with inputs. But maybe we want to do something with it too.

And how does the Postgres planner handle them?

src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp
GirgerS: For the same reason why `pdrgpiSegments` is set to null. The only meaningful method of this class is [CConstExprEvaluatorDXL::PexprEval](https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp#L71-L93), and it goes through `CTranslatorDXLToExpr::PexprTranslateScalar`
to [CTranslatorDXLToExpr::PexprScalar](https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/gporca/libgpopt/src/translate/CTranslatorDXLToExpr.cpp#L2591-L2680). And from my brief overview, this method doesn't care about segments (and probably never should, as motions are represented as physical operators, not scalar). But I didn't go as far as verifying each individual switch case.

RekGRpth: I think it's worth somehow reflecting this in the PR description.

RekGRpth:
The tests did not fail with the diff
```diff
diff --git a/src/backend/gporca/server/src/unittest/CTestUtils.cpp b/src/backend/gporca/server/src/unittest/CTestUtils.cpp
index b76cd611282..50ca1c6e735 100644
--- a/src/backend/gporca/server/src/unittest/CTestUtils.cpp
+++ b/src/backend/gporca/server/src/unittest/CTestUtils.cpp
@@ -2890,7 +2890,7 @@ CTestUtils::EresTranslate(CMemoryPool *mp, const CHAR *szQueryFileName,
 
        // translate plan back to DXL
        CTranslatorExprToDXL ptrexpr2dxl(mp, md_accessor, PdrgpiSegments(mp),
-                                                                        1 /* ulSessionId */);
+                                                                        0 /* ulSessionId */);
        CDXLNode *pdxlnPlan = ptrexpr2dxl.PdxlnTranslate(pexprPlan, pqc->PdrgPcr(),
                                                                                                         pqc->Pdrgpmdname());
        GPOS_UNITTEST_ASSERT(nullptr != pdxlnPlan);
diff --git a/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp b/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
index 68e76653c8b..e3788be6736 100644
--- a/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
+++ b/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
@@ -170,7 +170,7 @@ CMiniDumperDXLTest::EresUnittest_Basic()
                }
 
                CTranslatorExprToDXL ptrexprtodxl(mp, &mda, pdrgpiSegments,
-                                                                                 1 /* ulSessionId */);
+                                                                                 0 /* ulSessionId */);
                CDXLNode *pdxlnPlan = ptrexprtodxl.PdxlnTranslate(
                        pexprPlan, pqc->PdrgPcr(), pqc->Pdrgpmdname());
                GPOS_UNITTEST_ASSERT(nullptr != pdxlnPlan);
```

src/backend/gporca/libgpopt/src/eval/CConstExprEvaluatorDXL.cpp
GirgerS: Agreed, I updated the description.

src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
```diff
@@ -169,7 +169,8 @@ CMiniDumperDXLTest::EresUnittest_Basic()
 			pdrgpiSegments->Append(GPOS_NEW(mp) INT(ul));
 		}
 
-		CTranslatorExprToDXL ptrexprtodxl(mp, &mda, pdrgpiSegments);
+		CTranslatorExprToDXL ptrexprtodxl(mp, &mda, pdrgpiSegments,
+										  1 /* ulSessionId */);
```
RekGRpth, line 173: Isn't `ulSessionId` being hardcoded too many times in tests? Wouldn't it be easier to set it in just one place?

GirgerS:
> The tests did not fail with the diff
> 
> ```diff
> diff --git a/src/backend/gporca/server/src/unittest/CTestUtils.cpp b/src/backend/gporca/server/src/unittest/CTestUtils.cpp
> index b76cd611282..50ca1c6e735 100644
> --- a/src/backend/gporca/server/src/unittest/CTestUtils.cpp
> +++ b/src/backend/gporca/server/src/unittest/CTestUtils.cpp
> @@ -2890,7 +2890,7 @@ CTestUtils::EresTranslate(CMemoryPool *mp, const CHAR *szQueryFileName,
>  
>         // translate plan back to DXL
>         CTranslatorExprToDXL ptrexpr2dxl(mp, md_accessor, PdrgpiSegments(mp),
> -                                                                        1 /* ulSessionId */);
> +                                                                        0 /* ulSessionId */);
>         CDXLNode *pdxlnPlan = ptrexpr2dxl.PdxlnTranslate(pexprPlan, pqc->PdrgPcr(),
>                                                                                                          pqc->Pdrgpmdname());
>         GPOS_UNITTEST_ASSERT(nullptr != pdxlnPlan);
> diff --git a/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp b/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
> index 68e76653c8b..e3788be6736 100644
> --- a/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
> +++ b/src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
> @@ -170,7 +170,7 @@ CMiniDumperDXLTest::EresUnittest_Basic()
>                 }
>  
>                 CTranslatorExprToDXL ptrexprtodxl(mp, &mda, pdrgpiSegments,
> -                                                                                 1 /* ulSessionId */);
> +                                                                                 0 /* ulSessionId */);
>                 CDXLNode *pdxlnPlan = ptrexprtodxl.PdxlnTranslate(
>                         pexprPlan, pqc->PdrgPcr(), pqc->Pdrgpmdname());
>                 GPOS_UNITTEST_ASSERT(nullptr != pdxlnPlan);
> ```

These are not the only places where the session id is hardcoded (which I don't like, an maybe it is a good idea to at least factor out this value to a constant). Try changing the following code, as it runs most of the tests: https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/gporca/server/src/unittest/CTestUtils.cpp#L3303-L3306

src/backend/gporca/server/src/unittest/gpopt/minidump/CMiniDumperDXLTest.cpp
GirgerS: Totally agree, there are a lot of functions to start a unit test (for some reason) and each of them does it separately.

Let's make a cleanup as a part of this patch, then?

RekGRpth: Looks like a separate task with low priority.

src/backend/gpopt/utils/COptTasks.cpp
```diff
@@ -839,7 +843,7 @@ COptTasks::OptimizeTask(void *ptr)
 
 			plan_dxl = COptimizer::PdxlnOptimize(
 				mp, &mda, query_dxl, query_output_dxlnode_array,
-				cte_dxlnode_array, expr_evaluator, num_segments, gp_session_id,
```
red1452, line 842: Why not `(ULONG) gp_session_id`?

GirgerS:
> And how does the Postgres planner handle them?

Great question, turn out that it was a blind spot in my understanding. It is done in the same [place](https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/optimizer/plan/createplan.c#L3137-L3197) where outputs are selected, because the algorithm is quite different.

It only computes outputs, because inputs of a motions are equal to the outputs of the previous one. These previous(parent) slices are set a bit [later](https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/executor/execUtils.c#L1161-L1274).

What is scary, is that having mismatch between inputs and outputs of different segments would likely result in an invalid plan. 

And as I haven't changed the outputs [here](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp), it is possible that I have introduced a silent bug. I'll sleep on this until tomorrow.

src/backend/gpopt/utils/COptTasks.cpp
GirgerS: To clarify: `(ULONG) gp_session_id` or `(ULONG) session_id`?

I though that factoring `gp_session_id` to a function is the right way to go, as it comes from outside of ORCA (even though ORCA is far from being 'modular', so there is no benefit from doing so. I just like it more).

Regarding an explicit cast, I'll add it.

RekGRpth:
> > And how does the Postgres planner handle them?
> 
> Great question, turn out that it was a blind spot in my understanding. It is done in the same [place](https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/optimizer/plan/createplan.c#L3137-L3197) where outputs are selected, because the algorithm is quite different.
> 
> It only computes outputs, because inputs of a motions are equal to the outputs of the previous one. These previous(parent) slices are set a bit [later](https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/executor/execUtils.c#L1161-L1274).
> 
> What is scary, is that having mismatch between inputs and outputs of different segments would likely result in an invalid plan.
> 
> And as I haven't changed the outputs [here](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp), it is possible that I have introduced a silent bug. I'll sleep on this until tomorrow.

That's why I suggest fixing it for the output as well.

src/backend/gpopt/utils/COptTasks.cpp
red1452: I said correctly. Could you such to remove extra variable (`session_id`)? Your function returns signed int, which is the same as `gp_session_id`. I can not understand, why does your function do? What will be changed if you remove function?

GirgerS:
> > > And how does the Postgres planner handle them?
> > 
> > 
> > Great question, turn out that it was a blind spot in my understanding. It is done in the same [place](https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/optimizer/plan/createplan.c#L3137-L3197) where outputs are selected, because the algorithm is quite different.
> > It only computes outputs, because inputs of a motions are equal to the outputs of the previous one. These previous(parent) slices are set a bit [later](https://github.com/GreengageDB/greengage/blob/b10c04607b99ae323d43489d26d33df72dd9cd36/src/backend/executor/execUtils.c#L1161-L1274).
> > What is scary, is that having mismatch between inputs and outputs of different segments would likely result in an invalid plan.
> > And as I haven't changed the outputs [here](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp), it is possible that I have introduced a silent bug. I'll sleep on this until tomorrow.
> 
> That's why I suggest fixing it for the output as well.

Ok, it wansn't hard to find a query that triggers both [GetOutputSegIdsArray](https://github.com/GreengageDB/greengage/blob/eefafdf22736b8fbd5697b45d27f10d8f23d1505/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp#L7618-L7627) and [GetInputSegIdsArray](https://github.com/GreengageDB/greengage/blob/eefafdf22736b8fbd5697b45d27f10d8f23d1505/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp#L7674-L7679) conditions after each other:

```
create table t1 as (select * from generate_series(1, 10) a) distributed by (a);
explain create table t2 as (select a, sum(a) over (order by a) from t1) distributed by (a);
                                           QUERY PLAN
------------------------------------------------------------------------------------------------
 Result  (cost=0.00..0.00 rows=0 width=0)
   ->  Redistribute Motion 1:3  (slice1; segments: 1)  (cost=0.00..431.02 rows=1 width=12)
         Hash Key: a
         ->  WindowAgg  (cost=0.00..431.00 rows=1 width=12)
               Order By: a
               ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
                     Merge Key: a
                     ->  Sort  (cost=0.00..431.00 rows=1 width=4)
                           Sort Key: a
                           ->  Seq Scan on t1  (cost=0.00..431.00 rows=1 width=4)
 Optimizer: GPORCA
(11 rows)
```

And the worst thing is that is works! 

```
create table t2 as (select a, sum(a) over (order by a) from t1) distributed by (a);
select * from t2;
 a  | sum
----+-----
  5 |  15
  6 |  21
  9 |  45
 10 |  55
  2 |   3
  3 |   6
  4 |  10
  7 |  28
  8 |  36
  1 |   1
(10 rows)
```

Turns out that we don't care about motion outputs in when physical operators are [translated](https://github.com/GreengageDB/greengage/blob/55ac8efbfbc50ba4a85ac1f9b4e4e1772ed6fa0f/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp#L2399-L2436) into PostgreSQL plans. So at the end this logic is very similar to the one in the PostgreSQL planner.

Also, motion outputs are used only in [here](https://github.com/GreengageDB/greengage/blob/55ac8efbfbc50ba4a85ac1f9b4e4e1772ed6fa0f/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp#L2570-L2574), if clangd isn't lying to me.

Regarding the patch, I've changed it anyway, as leaving outputs out of sync with inputs doesn't look right regardless of the implementation.

RekGRpth:
> motion outputs are used only in [here](https://github.com/GreengageDB/greengage/blob/55ac8efbfbc50ba4a85ac1f9b4e4e1772ed6fa0f/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp#L2570-L2574)

https://github.com/GreengageDB/greengage/blob/55ac8efbfbc50ba4a85ac1f9b4e4e1772ed6fa0f/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp#L4896

GirgerS:
> > motion outputs are used only in [here](https://github.com/GreengageDB/greengage/blob/55ac8efbfbc50ba4a85ac1f9b4e4e1772ed6fa0f/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp#L2570-L2574)
> 
> https://github.com/GreengageDB/greengage/blob/55ac8efbfbc50ba4a85ac1f9b4e4e1772ed6fa0f/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp#L4896

True, that's where we set them and what we are changing in this patch. What I meant is something like `reading/using their value`.

P.S: it is also a different method, the one you mentioned belong to `CTranslatorExprToDXL`, when mine is a part of `CDXLPhysicalMotion`

src/backend/gpopt/utils/COptTasks.cpp
GirgerS: Well, then the idea remains the same. But after taking a closer look at `gpdbwrappes.cpp`, I might say that it is indeed not a common practice to wrap a guc, so I reverted the changes.

red1452 (approved):

RekGRpth (approved):

PR #356, opened 2026-03-31
Implement duration control for ggrebalance

whitehawk:
Implement duration control for ggrebalance

This patch adds support for '--duration' and '--end' options. The underlying
implementation for both options is the same - a timer is started, and on its
expiration, it sets the shutdown type to 'soft' (also added by this patch) and
sends SIGINT to the tool. In the signal handler we shut down the rebalance
state machine with soft shutdown. It means that the shutdown will wait for the
critical operations (like role switchovers and mirror movements), that are
currently being executed, to finish, and then will halt the execution.
Hard shutdown (without waiting for role switchovers and mirror movements) is
still used for the abort operation performed by the user.
Also, the new options were removed from the list of incompatible options for
the 'rollback' operation, as they are compatible according to the requirements.

Plus, some redundant code was cleaned up:
 - removed cmd_recoverseg as not used;
 - removed configuration provider code as not used;
 - removed unused imports and reorganized existing ones.

### DIFF
--- a/gpMgmt/bin/ggrebalance
+++ b/gpMgmt/bin/ggrebalance
@@ -1,8 +1,5 @@
 #!/usr/bin/env python3
 
-from gppylib.mainUtils import getProgramName
-
-from psycopg2 import DatabaseError
 try:
     from gppylib.commands.unix import *
     from gppylib.commands.gp import *
@@ -19,6 +16,9 @@ try:
     from gppylib.programs.clsRecoverSegment_triples import get_segments_with_running_basebackup
 except ImportError as e:
     sys.exit('ERROR: Cannot import modules.  Please check that you have sourced greenplum_path.sh.  Detail: ' + str(e))
+from datetime import datetime, time
+from threading import Timer
+from psycopg2 import DatabaseError
 
 _usage = """
     Use 'ggrebalance --help' to get list of available options.
@@ -82,6 +82,10 @@ def parseargs():
                       help='Use host names when updating the pg_hba.conf file.')
     parser.add_option('--replay-lag', dest="replay_lag", type="float",
                       help='Replay lag(in GBs) allowed on mirror when rebalancing the segments.')
+    parser.add_option('-T', '--duration', type='duration', metavar='[h][:m[:s]]',
+                      help='duration from beginning to end.')
+    parser.add_option('-E', '--end', type='datetime', metavar='datetime',
+                      help="ending date and time in the format 'YYYY-MM-DD hh:mm:ss'.")
     parser.add_option('-v', '--verbose', action='store_true', default=False,
                       help='debug output.')
     parser.add_option('--inplace-swap-roles', dest="inplace_swap_roles", action='store_true', default=False,
@@ -135,6 +139,7 @@ def validate_options(options, args, parser):
         ('add_hosts', ['add_hosts_file', 'remove_hosts', 'remove_hosts_file']),
         ('add_hosts_file', ['remove_hosts', 'remove_hosts_file']),
         ('remove_hosts', ['remove_hosts_file']),
+        ('duration', ['end']),
 
         ('target_datadirs', ['target_datadirs_file']),
 
@@ -144,14 +149,15 @@ def validate_options(options, args, parser):
             'target_hosts', 'target_hosts_file', 'add_hosts', 'add_hosts_file',
             'remove_hosts', 'remove_hosts_file', 'target_datadirs', 'target_datadirs_file',
             'target_segment_count', 'mirror_mode', 'skip_rebalance', 'show_plan',
-            'skip_resource_estimation', 'analyze', 'replay_lag', 'hba_hostnames', 'inplace_swap_roles'
+            'skip_resource_estimation', 'analyze', 'replay_lag', 'hba_hostnames', 'inplace_swap_roles',
+            'duration', 'end'
         ]),
 
         ('rollback_required', [
             'target_hosts', 'target_hosts_file', 'add_hosts', 'add_hosts_file',
             'remove_hosts', 'remove_hosts_file', 'target_datadirs', 'target_datadirs_file',
             'target_segment_count', 'mirror_mode', 'skip_rebalance', 'show_plan',
-            'skip_resource_estimation', 'duration', 'end', 'inplace_swap_roles'
+            'skip_resource_estimation', 'inplace_swap_roles'
         ]),
     ]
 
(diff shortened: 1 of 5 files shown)


gpMgmt/bin/ggrebalance
```diff
@@ -250,14 +267,13 @@ def remove_pid_file(coordinator_data_directory: str):
         pass
 
 gg_main_rebalance_SM = None
-cmd_recoverseg = None
 
 def sig_handler(sig, arg):
     if gg_main_rebalance_SM is not None:
         gg_main_rebalance_SM.shutdown()
+        # Set shutdown type back to 'hard' for the case we are here due to timeout
```
dkovalev1, line 274: How does it know that a timeout happened here and not some external signal?

gpMgmt/bin/ggrebalance
```diff
@@ -250,14 +267,13 @@ def remove_pid_file(coordinator_data_directory: str):
...
+        gg_main_rebalance_SM.set_hard_shutdown(True)
```
dkovalev1, line 275: Am I understand correctly that call to `shutdown()` is not blocking and we set hasd_shutdown here so next signal will abort process?

Am I understand correctly that if the user sends SIGTERM (no timeout yet), we start soft shutdown and if then timeout happens then hard shutdown starts violating REQ.3.1.23?

gpMgmt/test/behave/mgmt_utils/ggrebalance_misc_options.feature
dkovalev1, line 1: Can we not skip this feature?

whitehawk: Currently all `ggrebalance` tests are skipped by default, because they take a pretty long time on CI. For manual testing the skip flag should be removed. We'll need to define a subset of tests to run at CI when we finally integrate the feature into the main branch.

gpMgmt/bin/ggrebalance
whitehawk: The flow is following:
1) The default shutdown type is hard shutdown.
2) If user sends SIGTERM (no timeout yet), hard shutdown will be triggered.
3) In case timeout happens, `timeout_handler()` is called. It sets `hard_shutdown` to `False` and sends SIGINT to self. So, sig_handler is invoked, and soft shutdown is triggered. And we set  `hard_shutdown` back to `True` right after it. Therefore, if user doesn't want to wait for soft shutdown to finish properly, he can send signal by himself and it will trigger hard shutdown.

gpMgmt/bin/ggrebalance
whitehawk: Refer to [my comment](https://github.com/GreengageDB/greengage/pull/356/changes#r3018877062)

gpMgmt/test/behave/mgmt_utils/ggrebalance_misc_options.feature
dkovalev1: I thought people use CI to not run tests manually, not vise versa...

dkovalev1 (approved):

gpMgmt/bin/ggrebalance
```diff
@@ -82,6 +82,10 @@ def parseargs():
                       help='Use host names when updating the pg_hba.conf file.')
     parser.add_option('--replay-lag', dest="replay_lag", type="float",
                       help='Replay lag(in GBs) allowed on mirror when rebalancing the segments.')
+    parser.add_option('-T', '--duration', type='duration', metavar='[h][:m[:s]]',
+                      help='duration from beginning to end.')
```
bandetto, line 86:
```suggestion
                      help='limit execution time.')
```

gpMgmt/bin/ggrebalance
```diff
@@ -82,6 +82,10 @@ def parseargs():
...
+    parser.add_option('-E', '--end', type='datetime', metavar='datetime',
```
bandetto, line 87: How does 'datetime' metavar work?

Should we write something like:
```suggestion
    parser.add_option('-E', '--end', type='YYYY-MM-DD hh:mm:ss', metavar='datetime',
```

gpMgmt/bin/ggrebalance
```diff
@@ -82,6 +82,10 @@ def parseargs():
...
+                      help="ending date and time in the format 'YYYY-MM-DD hh:mm:ss'.")
```
bandetto, line 88:
```suggestion
                      help="future date and time to prematurely stop execution if exceeded.")
```

gpMgmt/bin/ggrebalance
```diff
@@ -172,7 +178,18 @@ def validate_options(options, args, parser):
...
+
+    # OptParse can return date instead of datetime so we might need to convert
+    if options.end and not isinstance(options.end, datetime):
+        options.end = datetime.combine(options.end, time(0))
+
+    if options.end and options.end < datetime.now():
+        logger.error('End time occurs in the past')
+        parser.exit(1)
+
+    if options.duration:
```
bandetto, line 190: What if both '--duration' and '-end' are set?

whitehawk: It is prohibited and tested in the new test case added to the `test 1. validate incompatible option combinations`.

gpMgmt/bin/ggrebalance
whitehawk: Current help value is consistent to the [same option in gpexpand](https://github.com/GreengageDB/greengage/blob/7e483fe182428fc3415f54a28b889881127ca3ad/gpMgmt/bin/gpexpand#L132)

gpMgmt/bin/ggrebalance
whitehawk: Current help value is consistent to the [same option in gpexpand](https://github.com/GreengageDB/greengage/blob/7e483fe182428fc3415f54a28b889881127ca3ad/gpMgmt/bin/gpexpand#L134)

gpMgmt/bin/ggrebalance
whitehawk: From my current understanding, there is no such need. We have our datetime addition to the optparse, that follows [this rule](https://github.com/GreengageDB/greengage/blob/7e483fe182428fc3415f54a28b889881127ca3ad/gpMgmt/bin/gppylib/datetimeutils.py#L90). 
And, gpexpand also follow the same definition for [the similar option](https://github.com/GreengageDB/greengage/blob/7e483fe182428fc3415f54a28b889881127ca3ad/gpMgmt/bin/gpexpand#L133)

bandetto (approved):
LGTM, but I don't really like the confusing help messages, even if they are consistent.

PR #363, opened 2026-04-02
Fix skip of table rebalance on shrink re-enter if ALTER TABLE failed on COMMIT

whitehawk:
Fix skip of table rebalance on shrink re-enter if ALTER TABLE failed on COMMIT

Problem description:
During table redistribution at shrink, if the COMMIT operation for the
transaction with ALTER TABLE ... REBALANCE failed, the table was not
redistributed at shrink operation re-enter, and was be not accessible after
the shrink, though ggrebalance reported that shrink process had finished Ok.

Root cause:
The table status was already set to Done in the ggrebalance schema, as the
status setting is done in a separate transaction (because the ggrebalance
schema resides in the different database for the most cases).

Fix:
We do not have transactions that can handle 2 different databases, so we need
to handle the possible misalignment of table status in the ggrebalance schema
and real status of the table by the means of ggrebalance application.
This patch moves the status setting after the COMMIT operation, therefore, if
COMMIT fails, the status will not be set to Done. But it opens another issue -
possible failure when setting the status, while the table is actually already
rebalanced. It led to an attempt to rebalance the already rebalanced table on
re-enter. To fix it, we now check if the table is rebalanced (according to
the catalog) before rebalancing it. A new function 'table_is_rebalanced()' is
introduced for that purpose.
Also, this patch simplifies the existing similar functions 'table_exists()' and
'db_exists()' (by removing the function parameters, that are replaced with the
object attributes, and by removing the redundant filter condition in the query
which duplicates the join condition), and adds 'rebalance_table()' function
to improve readability of the modified piece of code.

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/shrink.py
+++ b/gpMgmt/bin/ggrebalance_modules/shrink.py
@@ -262,12 +262,6 @@ class GGShrink:
                     self.trigger('move_to_STATE_END_FROM_ROLLBACK')
                     return
 
-        # In some rare cases there can be tables left, that
-        # have been already redistributed, but the status is still 'none'.
-        # To handle it, we cleanup all tables with status 'none' at the first
-        # enter of rollback operation.
-        self.rebalance_schema.clearTablesToRebalanceWithStatus('none')
-
         self.trigger('start')
 
     def get_state_after_interrupt(self, prev_state) -> str:
@@ -527,6 +521,12 @@ class GGShrink:
 
     @wrap_func_with_faults
     def on_enter_STATE_SHRINK_ROLLBACK_RESTORE_TARGET_SEGMENT_COUNT_DONE(self) -> None:
+        # In some rare cases there can be tables left, that
+        # have been already redistributed, but the status is still 'none'.
+        # To handle it, we cleanup all tables with status 'none' at the first
+        # enter of rollback operation.
+        self.rebalance_schema.clearTablesToRebalanceWithStatus('none')
+
         self.trigger('move_to_STATE_SHRINK_ROLLBACK_PREPARE_SCHEMA_START')
 
     @wrap_func_with_faults
--- a/gpMgmt/bin/ggrebalance_modules/shrink.py
+++ b/gpMgmt/bin/ggrebalance_modules/shrink.py
@@ -18,7 +18,7 @@ try:
     from gppylib.system.environment import *
     from gppylib.utils import escape_string, escapeDoubleQuoteInSQLString
     from ggrebalance_modules.planner import *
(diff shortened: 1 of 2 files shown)


red1452 (approved):

gpMgmt/bin/ggrebalance_modules/shrink.py
```diff
@@ -636,41 +636,53 @@ def func_with_faults(self, attempt: int):
...
                         if self.shrink.options.analyze:
                             dbconn.execSQL(conn,
                                            f'''ANALYZE {escapeDoubleQuoteInSQLString(self.schema_name, True)}.{escapeDoubleQuoteInSQLString(self.rel_name, True)}''')
                     else:
                         self.shrink.logger.info(f'''Table "{self.db_name}"."{self.schema_name}"."{self.rel_name}" doesn't exist, skipping actual rebalance''')
 
-                    self.shrink.rebalance_schema.setStatusForTableToRebalance(self.db_name, self.schema_name, self.rel_name, self.table_status_after_rebalance)
                     dbconn.execSQL(conn, 'COMMIT')
+
+                    inject_fault(f'before_set_status_{self.db_name}.{self.schema_name}.{self.rel_name}')
```
hilltracer, line 690: Did you test a subsequent ggrebalance -r after failure here?

Looks like after patch, if COMMIT already succeeded but setStatusForTableToRebalance() fails, the table may already be rebalanced while its status row is still none. Then rollback may rebuild the same table as [done](https://github.com/GreengageDB/greengage/blob/d40d2c06ad4ec54e1e006253004fbf4f40a5eafb/gpMgmt/bin/ggrebalance_modules/shrink.py#L711) again in prepare_shrink_schema(True).

whitehawk: Good point! Fixed it and added test cases.

hilltracer: Is it ok to clear all none rows in rollback()? May we drop some legitimate entries? Would it be safer to reconcile none -> done in prepare_shrink_schema(True) only for tables that are already shrunk in catalog?

whitehawk: > Is it ok to clear all none rows in rollback()? May we drop some legitimate entries?

I think it should be Ok, for now in the rollback flow we do not need these entries with `none` status, and we can remove them all, incuding  legitimate entries. 

> Would it be safer to reconcile none -> done in prepare_shrink_schema(True) only for tables that are already shrunk in catalog?

At this moment it looks to me that this approach will require adding more complicated logic than the approach I've currenlty used, and in this way it doesn't seem safer to me.

hilltracer (approved):

red1452 (approved):

PR #365, opened 2026-04-02
GG-229: Switch regression testing to ubuntu24

ShapatinAA:
Switch regression testing to ubuntu24

This patch changes `plpythonu` to `plpython3u` almost everywhere, except for
places where import of greengage internal libraries happens. There we need to
address libraries with version of plpython this libraries were build to. Also
some tests were update were taken as subprocesses module acts different in
python 3 due to difference in strings.

With this `sql_isolation_testcase.py` were updated: now, when creating
psql-like output, for length of column we count not a number of characters but
a number of bytes needed for representation of result

Mapred test were redacted also, as newer version of perl on ubuntu24 gives
different output.


src/test/isolation2/sql/setup.sql
```diff
@@ -1,4 +1,8 @@
-CREATE OR REPLACE LANGUAGE plpythonu;
+-- start_ignore
+! gpconfig -c plpython3.python_path -v "'$GPHOME/lib/python'" --skipvalidation;
```
silent-observer, line 2: I added this as an attempt to fix import issues, is this still necessary?

contrib/amcheck/sql/check_btree.sql
```diff
@@ -136,11 +136,11 @@ RETURNS integer as $$
       else:
         f.seek(corruption_offset, 2)
 
-      f.write(write_char)
+      f.write(bytes(write_char.encode('utf-8')))
```
KnightMurloc: why `encode` is not enough?

src/test/isolation2/expected/resgroup/resgroup_query_mem.out
```diff
@@ -13,8 +13,8 @@
 
 create extension if not exists gp_inject_fault;
 CREATE
-create or replace language plpythonu;
-CREATE
+do $$ begin /*in func*/ execute $func$ /*in func*/ drop language if exists plpythonu cascade; /*in func*/ $func$; /*in func*/ begin /*in func*/ execute $func$ /*in func*/ create language plpythonu; /*in func*/ $func$; /*in func*/ exception /*in func*/ when others then /*in func*/ if SQLERRM = 'could not access file "$libdir/plpython2": No such file or directory' then /*in func*/ execute $func$ /*in func*/ drop language if exists plpython3u cascade; /*in func*/ $func$; /*in func*/ begin /*in func*/ execute $func$ /*in func*/ create language plpython3u; /*in func*/ alter language plpython3u rename to plpythonu; /*in func*/ $func$; /*in func*/ exception /*in func*/ when others then /*in func*/ raise exception 'Could not create or rename PL/Python language: %', SQLERRM; /*in func*/ end; /*in func*/ else /*in func*/ raise exception 'Failed to create PL/Pythonu: %', SQLERRM; /*in func*/ end if; /*in func*/ end; /*in func*/ end; /*in func*/ $$;
+DO
```
KnightMurloc: Can you please explain this diff?

src/test/isolation2/input/resgroup/enable_resgroup.source
```diff
@@ -1,5 +1,33 @@
...
+        exception /*in func*/
+          when others then /*in func*/
+            raise exception 'Could not create or rename PL/Python language: %', SQLERRM; /*in func*/
+        end; /*in func*/
+      else /*in func*/
+        raise exception 'Failed to create PL/Pythonu: %', SQLERRM; /*in func*/
+      end if; /*in func*/
+  end; /*in func*/
+end; /*in func*/
+$$;
```
KnightMurloc: Please explain this diff.

src/test/isolation2/input/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -95,7 +125,7 @@ CREATE RESOURCE GROUP rg2_cpu_test WITH (concurrency=5, cpu_rate_limit=20, memor
 --
 -- check gpdb cgroup configuration
 --
-DO LANGUAGE PLPYTHONU $$
+DO LANGUAGE plpythonu $$
```
KnightMurloc: Are we using Python 2 here?

src/test/isolation2/sql/setup.sql
ShapatinAA: Yes, as `plpython3.python_path` is the only way to gain access to python libraries from greengage build.

silent-observer (dismissed):

src/test/isolation2/sql_isolation_testcase.py
```diff
@@ -480,15 +480,24 @@ def printout_result(self, r):
             # Figure out the widths of each column.
             fields = r.listfields()
             for f in fields:
-                widths.append(len(str(f)))
+                if sys.version_info[0] < 3:
```
KnightMurloc: We usually use equal

```suggestion
                if sys.version_info[0] == 2:
```

src/test/isolation2/sql/setup.sql
bimboterminator1: And the PYTHONPATH enviromental variable?

src/test/regress/output/temp_tablespaces.source
```diff
@@ -98,9 +98,9 @@ process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
 rc = process.wait()
 output, error = process.communicate()
 if rc != 0:
-    raise Exception("error=%s" % str(error))
-return int(str(output).strip())
-$$ LANGUAGE plpythonu;
+    raise Exception("error=%s" % str(error.decode('utf-8')))
```
KnightMurloc: Do we still need `str` here?

src/test/isolation2/sql_isolation_testcase.py
```diff
@@ -748,6 +759,7 @@ def process_command(self, command, output_file, global_sh_executor):
                 cmd_output = gpsubprocess.Popen(sql.strip(), stderr=subprocess.STDOUT, stdout=subprocess.PIPE, shell=True)
                 if not bg_mode:
                     stdout, _ = cmd_output.communicate()
+                    stdout = stdout.decode('utf-8')
```
silent-observer: This is not needed anymore

src/test/isolation2/sql/setup.sql
ShapatinAA: When working with python3 it gets unsetted here:
https://github.com/GreengageDB/greengage/blob/cf47aac9d856b9ec23b60a33848ef9ae45dbe26b/src/pl/plpython/plpy_main.c#L195-L209

src/test/isolation2/input/resgroup/enable_resgroup.source
ShapatinAA: This is do-block for deciding which version of plpython should we use. On systems built for python2 `create language plpythonu` will succeed. Otherwise we'll use plpython3u (which we are renaming to `plpythonu`, so we don't need to change functions)

src/test/isolation2/input/resgroup/resgroup_cpu_rate_limit.source
ShapatinAA: We renaming plpython3u to plpythonu previously, so we don't need to change this with same do-block.

src/test/isolation2/sql_isolation_testcase.py
ShapatinAA: Thank you! Fixed.

src/test/regress/output/temp_tablespaces.source
ShapatinAA: Hmm... I'm afraid we need to redo all exceptions like that one, thank you for noticing!
Fixed it.

src/test/isolation2/sql_isolation_testcase.py
ShapatinAA: Fixed

contrib/amcheck/sql/check_btree.sql
ShapatinAA: Thank you! Fixed.

ci/Dockerfile.ubuntu
```diff
@@ -42,7 +42,8 @@ COPY . gpdb_src
 RUN mkdir bin_gpdb
 
 ENV TARGET_OS=ubuntu \
-    OUTPUT_ARTIFACT_DIR=bin_gpdb
+    OUTPUT_ARTIFACT_DIR=bin_gpdb \
+    SKIP_UNITTESTS=1
```
KnightMurloc: Remove `SKIP_UNITTESTS`.

gpcontrib/gpmapreduce/input/mapred.source
```diff
@@ -7,9 +7,17 @@
...
+-- m|STDERR>  }"|
+-- m|STDERR> syntax error at line \d+, near ";|
+-- m|STDERR> Execution of aborted due to compilation errors.|
+-- end_matchignore
+
 -- start_matchsubs
 -- m|mapreduce_\d+_run|
 -- s|mapreduce_\d+_run|mapreduce_PID_run|
+-- m|STDERR> CONTEXT:  compilation of PL/Perl function "mapreduce_\d+_[a-z_]+"|
+-- s|mapreduce_\d+|mapreduce_X|
```
KnightMurloc, line 20: Why do we need this?

src/test/isolation2/input/resgroup/enable_resgroup.source
KnightMurloc: Can we use something like `include: helpers/server_helpers.sql;` to have this thing in only one place?

src/test/isolation2/sql/packcore.sql
```diff
@@ -19,12 +19,18 @@ DO LANGUAGE plpythonu $$
                                stderr=subprocess.PIPE)
         out = ret.communicate()
         if ret.returncode != 0:
-            raise SystemError('''\
+            if sys.version_info[0] == 2:
+                raise SystemError('''\
```
KnightMurloc: This block is running with `plpython3u` in which case we can have Python 2 here?

src/test/isolation2/sql/packcore.sql
```diff
@@ -19,12 +19,18 @@ DO LANGUAGE plpythonu $$
...
 stdout: {stdout}
 stderr: {stderr}
 '''.format(cmds=cmds, retcode=ret.returncode, stdout=out[0], stderr=out[1]))
-
+            else:
+                raise SystemError('''\
+Command {cmds} returned non-zero exit status {retcode}
+stdout: {stdout}
+stderr: {stderr}
+'''.format(cmds=cmds, retcode=ret.returncode, stdout=out[0].decode('utf-8'), stderr=out[1].decode('utf-8')))
```
KnightMurloc, line 26: Can we bring decoding outside of `if` to not duplicate the message string?

src/test/isolation2/sql_isolation_testcase.py
```diff
@@ -485,15 +485,24 @@ def printout_result(self, r):
             # Figure out the widths of each column.
             fields = r.listfields()
             for f in fields:
-                widths.append(len(str(f)))
+                if sys.version_info[0] == 2:
+                    width = len(str(f))
+                else:
+                    width = len(str(f).encode('utf-8'))
+                widths.append(width)
```
KnightMurloc, line 492: Why do we need length in bytes?

src/test/regress/input/temp_tablespaces.source
```diff
@@ -53,8 +53,11 @@ process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
 rc = process.wait()
 if rc != 0:
     output, error = process.communicate()
-    raise Exception("error=%s" % str(error))
-$$ LANGUAGE plpythonu;
+    if sys.version_info[0] == 2:
+        raise Exception("error=%s" % error)
+    else:
+        raise Exception("error=%s" % error.decode('utf-8'))
+$$ LANGUAGE plpython3u;
```
KnightMurloc: > LANGUAGE plpython3u

How it can be Python 2?

gpcontrib/gpmapreduce/input/mapred.source
silent-observer: This is from Perl version change on Ubuntu 24, the compilation errors use different format. In general I think this patch should be named "add Ubuntu 24 support for tests", even though most of changes have to do with Python 3

src/test/isolation2/sql_isolation_testcase.py
ShapatinAA: As @silent-observer found out `len()` returns number of characters and in python2 it is number of bytes and in python3 it is number of chars (len('资源组') gives 3 in python3, but 9 in python2). So to unify output we decided to use number of bytes in both cases.

src/test/isolation2/sql/packcore.sql
ShapatinAA: Thank you! Totally missed that.

src/test/isolation2/sql/packcore.sql
ShapatinAA: Removed unnecessary branching.

src/test/regress/input/temp_tablespaces.source
ShapatinAA: Thank you! Fixed.

src/test/isolation2/input/resgroup/enable_resgroup.source
ShapatinAA: Done

ci/Dockerfile.ubuntu
ShapatinAA: Fixed

src/test/isolation2/input/resgroup/disable_resgroup.source
```diff
@@ -8,6 +8,8 @@
 ! gpconfig -r gp_resource_group_memory_limit;
 ! gpconfig -r gp_resource_group_bypass_catalog_query;
 ! gpstop -rai;
+drop language if exists plpythonu cascade;
+drop language if exists plpython3u cascade;
```
KnightMurloc, line 12: Why do we need to drop plpython now when we didn't need it before?

src/test/isolation2/input/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -6,10 +6,10 @@ DROP RESOURCE GROUP rg1_cpu_test;
 DROP RESOURCE GROUP rg2_cpu_test;
 DROP VIEW IF EXISTS busy;
 DROP TABLE IF EXISTS bigtable;
-
-CREATE LANGUAGE plpythonu;
 -- end_ignore
 
+include: helpers/decide_plpython.sql;
```
KnightMurloc: Why we don't ignore it now?

src/test/isolation2/input/resgroup/resgroup_cpuset.source
```diff
@@ -306,7 +305,10 @@ DROP RESOURCE GROUP rg1_cpuset_test;
 DROP FUNCTION check_rules();
 DROP FUNCTION check_cpuset(TEXT, TEXT);
 DROP FUNCTION create_allcores_group(TEXT);
-DROP LANGUAGE plpythonu CASCADE;
+-- start_ignore
+2:DROP LANGUAGE IF EXISTS plpythonu CASCADE;
+2:DROP LANGUAGE IF EXISTS plpython3u CASCADE;
```
KnightMurloc: Why we do it in separate session?

src/test/isolation2/output/resgroup/resgroup_cpuset.source
```diff
@@ -315,8 +315,6 @@ DROP FUNCTION check_cpuset(TEXT, TEXT);
 DROP
 DROP FUNCTION create_allcores_group(TEXT);
 DROP
-DROP LANGUAGE plpythonu CASCADE;
-DROP
```
KnightMurloc, line 319: Why?

src/test/isolation2/sql/motion_socket.sql
```diff
@@ -77,3 +79,7 @@ $$ LANGUAGE plpythonu EXECUTE ON MASTER;
 
 !\retcode gpconfig -r gp_interconnect_address_type;
 !\retcode gpstop -au;
+-- start_ignore
+drop language if exists plpythonu cascade;
+drop language if exists plpython3u cascade;
```
KnightMurloc: Why do we need to drop plpython now when we didn't need it before?

src/test/isolation2/sql/setup.sql
KnightMurloc: To be honest, I'm a little worried about this condition. Don't we need to invert it if we have Python 3 by default?

src/test/regress/input/alter_db_set_tablespace.source
```diff
@@ -1,6 +1,8 @@
 -- start_matchsubs
 -- m/WARNING:.*Any temporary tables for this session have been dropped because the gang was disconnected/
 -- s/session id \=\s*\d+/session id \= DUMMY/gm
+-- m/dboid dir for database alter_db does not exist on dbid = \d+/
+-- s/dbid = \d+/dbid = N/
```
KnightMurloc, line 5: Where did this error(?) come from?

src/test/regress/input/temp_tablespaces.source
ShapatinAA: sry didn't catch more of them. now did.

src/test/isolation2/sql/setup.sql
ShapatinAA: In gg-7 we don't have this code.
In python3-only system: we talked it out, that it is behavior that client should expect. And it will not be good idea to change it within stable release at least.

src/test/isolation2/input/resgroup/disable_resgroup.source
ShapatinAA: Because it might break pg_upgrade tests (as we're renaming `plpython3u` to `plpythonu`). Maybe it will be enough to drop only `plpythonu`, but to cleanup as much as you can after tests - should be also right.

src/test/isolation2/sql/motion_socket.sql
ShapatinAA: Cleanup after test. Renaming languages causes pg_upgrade test to fail, so after doing renaming and using plpythonu we should cleanup.

src/test/isolation2/input/resgroup/resgroup_cpuset.source
ShapatinAA: Thank you for noticing! Copied wrongfully.

src/test/isolation2/output/resgroup/resgroup_cpuset.source
ShapatinAA: Thank you for noticing - returned old drops in case of 2 pythons usage.

src/test/isolation2/input/resgroup/resgroup_cpu_rate_limit.source
ShapatinAA: fixed

src/test/regress/input/alter_db_set_tablespace.source
ShapatinAA: It started to came unsorted as I recall

gpcontrib/gpmapreduce/input/mapred.source
ShapatinAA: There was this error
https://github.com/GreengageDB/greengage/actions/runs/23849127059/job/69525961774

KnightMurloc: I think such changes should be described in the PR description.

src/test/regress/input/alter_db_set_tablespace.source
ShapatinAA: Couldn't catch same behavior as once got, will update this test and check out what it gives on CI.

gpcontrib/gpmapreduce/input/mapred.source
ShapatinAA: Done.

src/test/regress/input/alter_db_set_tablespace.source
KnightMurloc: You updated only output.

.github/workflows/greengage-ci.yml
```diff
@@ -62,7 +62,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target_os: [ubuntu]
+        include:
+          - target_os: ubuntu
+          - target_os: ubuntu
+            target_os_version: "24.04"
```
KnightMurloc, line 67: Why do we need to specify `target_os` twice?

ShapatinAA: So we're testing 2 versions: 22 and 24. But soon will exclude regression on ubuntu22.

KnightMurloc: > But soon will exclude regression on ubuntu22.

Will it be done in another PR? If so, why?

ShapatinAA: No, I reconsidered my decision and it will be done in this one.

src/test/regress/input/alter_db_set_tablespace.source
ShapatinAA: Yep, as you can see it prints dbids randomly from now.

KnightMurloc: Have you researched why this is happening?

ShapatinAA: Well, as I understand, it is happening because of how we iterate through `db_instances.items()`. As everything before that part seems to be the same, I don't think that it can worry us much. Should I investigate it?

KnightMurloc: It seems to me that it is worth investigating. To make sure that these are not symptoms of some kind of problem.

ShapatinAA: In python3 
>Keys and values are iterated over in insertion order.

Which matches exactly order of insertions in `db_instances` dictionary in both versions.

So this piece of code `for dbid, datadir in db_instances.items():` makes the difference in output.

.github/workflows/greengage-ci.yml
ShapatinAA: done

KnightMurloc (dismissed):

bimboterminator1 (dismissed):

.github/workflows/greengage-ci.yml
magf: > Why do we need to specify `target_os` twice?

This is matrix.
`target_os` without `target_os_version` means `target_os_version: 22.04` and should not be defined as described [Important Notes on target_os_version](https://github.com/GreengageDB/greengage/blob/6.x/.github/workflows/README.md#important-notes-on-target_os_version) in README.

Both are needed, otherwise the tests for `ubuntu:22.04` won't run.

.github/workflows/greengage-ci.yml
```diff
@@ -62,7 +62,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target_os: [ubuntu]
+        include:
+          - target_os: ubuntu
+            target_os_version: "24.04"
```
magf, line 67:
```suggestion
        include:
          - target_os: ubuntu
          - target_os: ubuntu
            target_os_version: "24.04"
```

.github/workflows/greengage-ci.yml
```diff
@@ -98,7 +101,9 @@ jobs:
...
```
magf, line 106:
```suggestion
        include:
          - target_os: ubuntu
          - target_os: ubuntu
            target_os_version: "24.04"
```

magf (dismissed):

.github/workflows/greengage-ci.yml
ShapatinAA: After the discussion, we decided to switch completely to testing on Ubuntu 24, so the other target (with default value translated to Ubuntu 22) isn’t needed.

.github/workflows/greengage-ci.yml
ShapatinAA: After the discussion, we decided to switch completely to testing on Ubuntu 24, so the other target (with default value translated to Ubuntu 22) isn’t needed.

.github/workflows/greengage-ci.yml
magf: OK, NP.
Resolved.

.github/workflows/greengage-ci.yml
magf: OK, NP.
Resolved.

.github/workflows/greengage-ci.yml
magf: After the discussion, we decided to switch completely to testing on Ubuntu 24, so the other target (with default value translated to Ubuntu 22) isn’t needed.

magf (dismissed):

magf (dismissed):
LGTM

bimboterminator1 (dismissed):

src/test/isolation2/input/resgroup/enable_resgroup.source
```diff
@@ -1,4 +1,5 @@
-CREATE LANGUAGE plpythonu;
+DROP LANGUAGE IF EXISTS plpython3u CASCADE;
```
KnightMurloc: Why do we need drop now?

ShapatinAA: As it gave [this output](https://github.com/GreengageDB/greengage/actions/runs/24715668850/job/72294457079#step:7:2550) and new test suite ideally should not be dependent on any previous results I thought that we should drop language and recreate language here.

KnightMurloc: Then why not `CREATE IF EXISTS ...`?

ShapatinAA: You mean `CREATE OR REPLACE`? Well then we will not have "clear" start.

KnightMurloc: I think I've started to understand. You're using `DROP... CASCADE` to remove all plpython3 functions?

KnightMurloc (dismissed):

bimboterminator1 (dismissed):

src/test/regress/expected/qp_olap_window.out
```diff
@@ -3,7 +3,7 @@
 --
 -- start_ignore
 -- count number of certain operators in a given plan
-create language plpythonu;
+create language plpython3uu;
```
magf:
```suggestion
create language plpython3u;
```

src/test/regress/expected/qp_olap_window.out
```diff
@@ -15,7 +15,7 @@ for i in range(len(rv)):
         result = result+1
 return result
 $$
-language plpythonu;
+language plpython3uu;
```
magf:
```suggestion
language plpython3u;
```

magf (changes requested):
Two `plpython3u` have duplicate `u` at the tail in the `src/test/regress/expected/qp_olap_window.out`

src/test/regress/expected/qp_olap_window.out
ShapatinAA: Thank you for noticing! `ignore` section did its evil thing...
Fixed.

src/test/regress/expected/qp_olap_window.out
ShapatinAA: Fixed.

magf (approved):

bimboterminator1:
> except for
places where import of greengage internal libraries happens. There we need to
address libraries with version of plpython this libraries were build to.

Where are these adjustments?

bimboterminator1 (approved):

silent-observer (approved):

PR #366, opened 2026-04-02
GG-333: Add volatile expression alert for replicated tables to gpcheckcat

VoidZeroNull0:
What happens?
Although 7cd685d protects replicated tables from main ways to
break its policy via usage of volatile expressions, some loopholes still exists
(one is used in violate_replicated_policy.sql‎ for behave test) and we can't do
much about them, except to alert the user.

How can we help the user?
gpcheckcat exists for similar purposes, so adding command to it, which alerts
the user about such data inconsistency should help. 

What command does?
Shows all columns with volatile expressions of all replicated tables among all
non-system schemas. 

Can we resync replicated table among segments?
No, as it is not obvious how to do it. If we choose to sync with one of the
segments, it raises question of what segment to choose as main to resync data
with. Even if we delegate this responsibility to user, gpcheckcat does not
contain many interactive ways to provide arguments to command to begin with.

Tests?
Unit tests for broken connection / cursor reading are provided.  Behave test to
check alert of inconsistent data also provided.

### DIFF
--- a/gpMgmt/bin/gpcheckcat
+++ b/gpMgmt/bin/gpcheckcat
@@ -576,9 +576,7 @@ def checkReplicatedDistribPolicy():
         JOIN pg_attribute a ON c.oid = a.attrelid
         JOIN pg_attrdef ad ON c.oid = ad.adrelid AND a.attnum = ad.adnum
         JOIN gp_distribution_policy d ON c.oid = d.localoid
-        JOIN pg_proc p ON 
-            ad.adbin ~ ':funcid \d+' AND
-            regexp_replace(ad.adbin, '.*:funcid (\d+) .*', '\\1', 'g')::int = p.oid
+        JOIN pg_proc p ON ad.adbin ~ (':funcid ' || p.oid || ' ')
     WHERE c.relkind = 'r'
         AND d.policytype = 'r'
         AND a.attnum > 0
@@ -608,6 +606,8 @@ def checkReplicatedDistribPolicy():
                         'This must be manually resolved. Check the gpcheckcat log for details.'
                     )
                 logger.error('replicated tables has %d issue(s)' % len(err))
+                logger.error('Replace the VOLATILE defaults with a STABLE expression via `ALTER TABLE ... SET DEFAULT ...`, '
+                             'also sync the data among segments.')
                 logger.error('The following columns and their tables must be manually fixed and synced:')
                 for e in err:
                     logger.error(formatErr(e[0], e[1], e[2]))
--- a/gpMgmt/bin/gpcheckcat
+++ b/gpMgmt/bin/gpcheckcat
@@ -601,9 +601,9 @@ def checkReplicatedDistribPolicy():
                 setError(ERROR_NOREPAIR)
                 logger.info('[FAIL] replicated tables')
                 myprint(
-                        '[ERROR]: Some columns, containing volatile expressions as defaults, '
+                        '[ERROR]: %d columns, containing volatile expressions as defaults, '
                         'have violated replicated table\'s distribution policy. '
-                        'This must be manually resolved. Check the gpcheckcat log for details.'
+                        'This must be manually resolved. Check the gpcheckcat log for details.' % len(err)
                     )
                 logger.error('replicated tables has %d issue(s)' % len(err))
                 logger.error('Replace the VOLATILE defaults with a STABLE expression via `ALTER TABLE ... SET DEFAULT ...`, '
--- a/gpMgmt/bin/gpcheckcat
+++ b/gpMgmt/bin/gpcheckcat
@@ -576,7 +576,9 @@ def checkReplicatedDistribPolicy():
         JOIN pg_attribute a ON c.oid = a.attrelid
         JOIN pg_attrdef ad ON c.oid = ad.adrelid AND a.attnum = ad.adnum
         JOIN gp_distribution_policy d ON c.oid = d.localoid
-        JOIN pg_proc p ON pg_get_expr(ad.adbin, ad.adrelid) ~ ('\m' || p.proname || '\s*\(')
+        JOIN pg_proc p ON 
+            ad.adbin ~ ':funcid \d+' AND
+            regexp_replace(ad.adbin, '.*:funcid (\d+) .*', '\\1', 'g')::int = p.oid
     WHERE c.relkind = 'r'
         AND d.policytype = 'r'
         AND a.attnum > 0
--- a/gpMgmt/bin/gpcheckcat
+++ b/gpMgmt/bin/gpcheckcat
@@ -560,6 +560,64 @@ def checkDistribPolicy():
 
     checkConstraintsRepair()
 
+def checkReplicatedDistribPolicy():
+    logger.info('-----------------------------------')
+    logger.info('Checking distribution violations on replicated tables')
+
+    # SELECT all replicated tables with volatile functions in default expressions from non-system schemas
+    qry = '''
+    SELECT
+        n.nspname AS nspname,
+        c.relname AS relname,
+        a.attname AS attname,
+        v.def_expr AS default_expr
+    FROM pg_class c
+        JOIN pg_namespace n ON c.relnamespace = n.oid
+        JOIN pg_attribute a ON c.oid = a.attrelid
+        JOIN pg_attrdef ad ON c.oid = ad.adrelid AND a.attnum = ad.adnum
+        JOIN gp_distribution_policy d ON c.oid = d.localoid
+    CROSS JOIN LATERAL (
+        SELECT pg_get_expr(ad.adbin, ad.adrelid)
+    ) AS v(def_expr)
+    WHERE c.relkind = 'r'
+        AND d.policytype = 'r'
+        AND a.attnum > 0
+        AND NOT a.attisdropped
+        AND n.nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast')
+        AND EXISTS (
+            SELECT 1 FROM pg_proc p
+            WHERE p.provolatile = 'v'
+                AND v.def_expr ~ ('\m' || p.proname || '\s*\(')
+        );
+    '''
+
+    try:
+        conn = connect2(GV.cfg[GV.coordinator_dbid])
+        with conn.cursor(cursor_factory=psycopg2.extras.RealDictCursor) as curs:
+            curs.execute(qry)
+
+            err = []
+            for row in curs.fetchall():
+                err.append([GV.cfg[GV.coordinator_dbid], ('nspname', 'relname', 'attname', 'default_expr'), row])
+
+            if not err:
+                logger.info('[OK] replicated tables')
+            else:
+                GV.checkStatus = False
+                setError(ERROR_NOREPAIR)
+                logger.info('[FAIL] replicated tables')
+                logger.error('replicated tables has %d issue(s)' % len(err))
+                for e in err:
+                    logger.error(formatErr(e[0], e[1], e[2]))
+                if len(err) > 100:
+                    logger.error('...')
+                    logger.error(qry)
+    except Exception as e:
+        GV.checkStatus = False
+        setError(ERROR_NOREPAIR)
+        myprint('[ERROR] executing test: checkReplicatedDistribPolicy')
+        myprint('  Execution error: ' + str(e))
+
 #############
 def checkPartitionIntegrity():
     logger.info('-----------------------------------')
@@ -2346,6 +2404,14 @@ all_checks = {
             "fn": lambda: checkMixDistPolicy(),
             "version": 'main',
             "order": 17,
+            "online": True
+        },
+    "replicated_distribution_policy":
+        {
+            "description": "Check policy violations of replicated tables",
+            "fn": lambda: checkReplicatedDistribPolicy(),
+            "version": 'main',
+            "order": 18,
             "online": True
         }
 
--- a/gpMgmt/bin/gpcheckcat
+++ b/gpMgmt/bin/gpcheckcat
@@ -570,25 +570,19 @@ def checkReplicatedDistribPolicy():
         n.nspname AS nspname,
         c.relname AS relname,
         a.attname AS attname,
-        v.def_expr AS default_expr
+        pg_get_expr(ad.adbin, ad.adrelid) AS default_expr
     FROM pg_class c
         JOIN pg_namespace n ON c.relnamespace = n.oid
         JOIN pg_attribute a ON c.oid = a.attrelid
         JOIN pg_attrdef ad ON c.oid = ad.adrelid AND a.attnum = ad.adnum
         JOIN gp_distribution_policy d ON c.oid = d.localoid
-    CROSS JOIN LATERAL (
-        SELECT pg_get_expr(ad.adbin, ad.adrelid)
-    ) AS v(def_expr)
+        JOIN pg_proc p ON pg_get_expr(ad.adbin, ad.adrelid) ~ ('\m' || p.proname || '\s*\(')
     WHERE c.relkind = 'r'
         AND d.policytype = 'r'
         AND a.attnum > 0
         AND NOT a.attisdropped
         AND n.nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast')
-        AND EXISTS (
-            SELECT 1 FROM pg_proc p
-            WHERE p.provolatile = 'v'
-                AND v.def_expr ~ ('\m' || p.proname || '\s*\(')
-        );
+        AND p.provolatile = 'v';
     '''
 
     try:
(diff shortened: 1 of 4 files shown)


gpMgmt/bin/gpcheckcat
```diff
@@ -560,6 +560,64 @@ def checkDistribPolicy():
...
+    WHERE c.relkind = 'r'
+        AND d.policytype = 'r'
+        AND a.attnum > 0
+        AND NOT a.attisdropped
+        AND n.nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast')
+        AND EXISTS (
+            SELECT 1 FROM pg_proc p
+            WHERE p.provolatile = 'v'
+                AND v.def_expr ~ ('\m' || p.proname || '\s*\(')
+        );
```
RekGRpth:
```suggestion
    SELECT
        n.nspname AS nspname,
        c.relname AS relname,
        a.attname AS attname,
        pg_get_expr(ad.adbin, ad.adrelid) AS default_expr
    FROM pg_class c
        JOIN pg_namespace n ON c.relnamespace = n.oid
        JOIN pg_attribute a ON c.oid = a.attrelid
        JOIN pg_attrdef ad ON c.oid = ad.adrelid AND a.attnum = ad.adnum
        JOIN gp_distribution_policy d ON c.oid = d.localoid
        JOIN pg_proc p ON pg_get_expr(ad.adbin, ad.adrelid) ~ ('\m' || p.proname || '\s*\(')
    WHERE c.relkind = 'r'
        AND d.policytype = 'r'
        AND a.attnum > 0
        AND NOT a.attisdropped
        AND n.nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast')
        AND p.provolatile = 'v'
```

gpMgmt/bin/gpcheckcat
```diff
@@ -560,6 +560,64 @@ def checkDistribPolicy():
...
+    ) AS v(def_expr)
+    WHERE c.relkind = 'r'
+        AND d.policytype = 'r'
+        AND a.attnum > 0
+        AND NOT a.attisdropped
+        AND n.nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast')
+        AND EXISTS (
+            SELECT 1 FROM pg_proc p
+            WHERE p.provolatile = 'v'
+                AND v.def_expr ~ ('\m' || p.proname || '\s*\(')
```
RekGRpth: This condition does not seem very reliable.

gpMgmt/test/behave/mgmt_utils/gpcheckcat.feature
```diff
@@ -235,6 +235,14 @@ Feature: gpcheckcat tests
         Then gpcheckcat should return a return code of 0
         And the user runs "dropdb constraint_db"
 
+    Scenario: gpcheckcat should report replicated tables policy violation via default volatile expressions
+        Given database "policy_violation_db" is dropped and recreated
+        And the user runs "psql policy_violation_db -f test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql"
+        Then psql should return a return code of 0
+        When the user runs "gpcheckcat -R replicated_distribution_policy policy_violation_db"
+        Then gpcheckcat should return a return code of 3
```
RekGRpth, line 243: Maybe it’s worth to fix the error output, tests can do that.

VoidZeroNull0: I don't understand what you mean. Could you, please, suggest some changes or explain a bit more?

RekGRpth: something like `Then gpcheckcat should print "reported here: owner" to stdout`

VoidZeroNull0: Added:
```Python
  myprint(
          '[ERROR]: Some columns, containing volatile expressions as defaults, '
          'have violated replicated table\'s distribution policy. '
          'This must be manually resolved. Check the gpcheckcat log for details.'
      )
 ```
 ```
 Then gpcheckcat should print "have violated replicated table's distribution policy" to stdout
 ```

File: gpMgmt/bin/gpcheckcat (hunk 1 above)
Comment (VoidZeroNull0) [reply gpMgmt/bin/gpcheckcat]: Changed to suggested one.

File: gpMgmt/bin/gpcheckcat [hunk 4]
```diff
@@ -560,6 +560,64 @@ def checkDistribPolicy():
...
+        n.nspname AS nspname,
+        c.relname AS relname,
+        a.attname AS attname,
+        pg_get_expr(ad.adbin, ad.adrelid) AS default_expr
+    FROM pg_class c
+        JOIN pg_namespace n ON c.relnamespace = n.oid
+        JOIN pg_attribute a ON c.oid = a.attrelid
+        JOIN pg_attrdef ad ON c.oid = ad.adrelid AND a.attnum = ad.adnum
+        JOIN gp_distribution_policy d ON c.oid = d.localoid
+        JOIN pg_proc p ON pg_get_expr(ad.adbin, ad.adrelid) ~ ('\m' || p.proname || '\s*\(')
```
RekGRpth: This condition does not seem very reliable.

VoidZeroNull0: Why do you think so? Can you give an example that won't work on this condition?

RekGRpth:
```sql
create table t (a int default random());
select * from pg_attrdef ad join pg_proc p on regexp_replace(adbin, '.*:funcid (\d+) .*', '\1', 'g')::int = p.oid;
  oid  | adrelid | adnum |                                                                                                                                                                   adbin                                                                                                                                                         
           | oid  | proname | pronamespace | proowner | prolang | procost | prorows | provariadic | prosupport | prokind | prosecdef | proleakproof | proisstrict | proretset | provolatile | proparallel | pronargs | pronargdefaults | prorettype | proargtypes | proallargtypes | proargmodes | proargnames | proargdefaults | protrftypes | prosrc  | probin | proconfig | proacl | proexeclocation 
-------+---------+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------+---------+--------------+----------+---------+---------+---------+-------------+------------+---------+-----------+--------------------------------------+-------------+-------------+----------+-----------------+------------+-------------+----------------+-------------+-------------+----------------+-------------+---------+--------+-----------+--------+-----------------
 17092 |   17089 |     1 | {FUNCEXPR :funcid 317 :funcresulttype 23 :funcretset false :funcvariadic false :funcformat 2 :funccollid 0 :inputcollid 0 :args ({FUNCEXPR :funcid 1598 :funcresulttype 701 :funcretset false :funcvariadic false :funcformat 0 :funccollid 0 :inputcollid 0 :args <> :is_tablefunc false :location 30}) :is_tablefunc false :location -1} | 1598 | random  |           11 |       10 |      12 |       1 |       0 |           0 | -          | f       | f         | f            | t           | f         | v           | r           |        0 |               0 |        701 |             |                |             |             |                |             | drandom |        |           |        | a
(1 row)

Comment (VoidZeroNull0) [reply gpMgmt/bin/gpcheckcat]: I agree. Did not think about searching for function's id inside `adbin`  without parsing it. Absolutely better this way.

Also had to check that `adbin` contains some functions, otherwise error would arise:
```SQL
ad.adbin ~ ':funcid \d+'
```

gpMgmt/bin/gpcheckcat
```diff
@@ -560,6 +560,66 @@ def checkDistribPolicy():
...
+        a.attname AS attname,
+        pg_get_expr(ad.adbin, ad.adrelid) AS default_expr
+    FROM pg_class c
+        JOIN pg_namespace n ON c.relnamespace = n.oid
+        JOIN pg_attribute a ON c.oid = a.attrelid
+        JOIN pg_attrdef ad ON c.oid = ad.adrelid AND a.attnum = ad.adnum
+        JOIN gp_distribution_policy d ON c.oid = d.localoid
+        JOIN pg_proc p ON 
+            ad.adbin ~ ':funcid \d+' AND
+            regexp_replace(ad.adbin, '.*:funcid (\d+) .*', '\\1', 'g')::int = p.oid
```
bandetto: The regex `:funcid \d+` matches only one funcid per expression. A default like `fn_vol() + fn_vol2()` contains two funcids, but `regexp_replace(..., 'g')` with a greedy `.*` capture group will only grab one.

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
```diff
@@ -0,0 +1,65 @@
...
+    a_vol INT DEFAULT (42*fn_vol())+42,
+    b_vol INT DEFAULT fn_val() + 42,
+    c_vol FLOAT
+) DISTRIBUTED REPLICATED;
+
+CREATE TABLE dist_replicated_non_vol(
+    a FLOAT,
+    b FLOAT,
+    c INT DEFAULT fn_val()
+) DISTRIBUTED REPLICATED
```
bandetto: How does this file work without semicolons at the end of the statements?

gpMgmt/bin/gpcheckcat
```diff
@@ -560,6 +560,66 @@ def checkDistribPolicy():
...
+            if not err:
+                logger.info('[OK] replicated tables')
+            else:
+                GV.checkStatus = False
+                setError(ERROR_NOREPAIR)
+                logger.info('[FAIL] replicated tables')
+                myprint(
+                        '[ERROR]: Some columns, containing volatile expressions as defaults, '
+                        'have violated replicated table\'s distribution policy. '
+                        'This must be manually resolved. Check the gpcheckcat log for details.'
```
bandetto:
```suggestion
                        'Replace the VOLATILE default with a STABLE expression via `ALTER TABLE ... SET DEFAULT ...`'
```

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
VoidZeroNull0: Good question. Behave tests don't seem to catch output error if test script has errors in syntax. Anyway, fixed.

gpMgmt/bin/gpcheckcat
VoidZeroNull0: Good idea. Yet, I think it is better to put this solution message in `logger.error()`, as in the log file will be a list of corrupted columns. Also added about the need of syncing (however, no straight method for this could be specified) - tables could be corrupted due to usage of VOLATILE expressions.

gpMgmt/bin/gpcheckcat
VoidZeroNull0: Thanks for the notice! Replaced with `ad.adbin ~ (':funcid ' || p.oid || ' ')`. This way, it should work in described case, because we search for exact `funcid` substring in `adbin`.

gpMgmt/bin/gpcheckcat
```diff
@@ -560,6 +560,66 @@ def checkDistribPolicy():
...
+    logger.info('-----------------------------------')
+    logger.info('Checking distribution violations on replicated tables')
+
+    # SELECT all replicated tables with volatile functions in default expressions from non-system schemas
+    qry = '''
+    SELECT
+        n.nspname AS nspname,
+        c.relname AS relname,
+        a.attname AS attname,
+        pg_get_expr(ad.adbin, ad.adrelid) AS default_expr
```
RekGRpth, line 573: Why is this line needed if it is not used anywhere below?

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
```diff
@@ -0,0 +1,65 @@
...
+    b INT,
+    c INT DEFAULT 42
+) DISTRIBUTED BY (a);
+
+CREATE TABLE dist_randomly (
+    a FLOAT,
+    b INT,
+    c INT DEFAULT 42
+) DISTRIBUTED RANDOMLY;
+
```
RekGRpth: Why so many different cases if only one error message is being checked?

gpMgmt/bin/gpcheckcat
VoidZeroNull0: It is used in log outputing for readability (so that it was easier for user to fix):
```Python
for row in curs.fetchall():
    err.append([GV.cfg[GV.coordinator_dbid], ('nspname', 'relname', 'attname', 'default_expr'), row])
```

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
VoidZeroNull0: Originally, I was expecting to check that command detected some exact number of problematic columns, but turned out step in behave for doing that does not exist - we can check stdout, but not the logs. I think, i will add the change to check the count in stdout (an justify number of cases).

VoidZeroNull0: Done in b866a69.

RekGRpth: > Done in [b866a69](https://github.com/GreengageDB/greengage/commit/b866a69c8890d73e73370b2884d2f9ced6c5036d).

But that's just the number of errors. You also have error-free cases. What's the point of them if they're not checked?

VoidZeroNull0: Well, the idea was that they should not contribute to error count. In reality, we should we check that gpcheckcat has not included them - to do so, check every log line. I can see what you mean, gonna leave only replicated ones.

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
```diff
@@ -0,0 +1,53 @@
...
+CREATE FUNCTION fn_val()
+        RETURNS int
+        LANGUAGE plpgsql
+        IMMUTABLE
+AS $$
+BEGIN
+	RETURN 42;
+END;
+$$
+EXECUTE ON ANY;
```
RekGRpth, line 21: may be remove it too?

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
```diff
@@ -0,0 +1,53 @@
...
+    a_vol INT DEFAULT (42*fn_vol())+42,
+    b_vol INT DEFAULT fn_val() + 42,
+    c_vol FLOAT
+) DISTRIBUTED REPLICATED;
+
+CREATE TABLE dist_replicated_non_vol(
+    a FLOAT,
+    b FLOAT,
+    c INT DEFAULT fn_val()
+) DISTRIBUTED REPLICATED;
```
RekGRpth, line 41: may be remove it too?

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
```diff
@@ -0,0 +1,53 @@
...
+CREATE OR REPLACE FUNCTION fn_vol()
+        RETURNS int
+        LANGUAGE plpgsql
+        VOLATILE
+AS $$
+BEGIN
+	RETURN (random() * 1000)::int;
+END;
+$$
+EXECUTE ON ANY;
```
RekGRpth: why not
```sql
alter function fn_vol volatile
```

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
VoidZeroNull0: I think this should stay, so that it won't be grabbed by command accidentally.

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
VoidZeroNull0: Needs to stay, to check that it is not counted towards total error count.

gpMgmt/test/behave/mgmt_utils/steps/data/gpcheckcat/violate_replicated_policy.sql
VoidZeroNull0: Removed RANDOMLY distributed and DISTRIBUTED by key.

bandetto (approved):
owo

RekGRpth (approved):

PR #367, opened 2026-04-02
Protect replicated tables with volatile functions from rewrite (6.x)

VoidZeroNull0:
What happens?
Distribution policy of replicated tables can be violated via usage of ALTER
TABLE (or CREATE TABLE) and a volatile expression as a default
value.

Why it happens?
No check for this case is made anywhere in the code, thus different data may
appear on segment's table.

How do we fix this?
Main check for expression volatility and table's distribution policy
inside AddRelationNewConstraintsn(), additional ones in
ATExecCookedColumnDefault() and ATExecSetDistributedBy() are made, to output
an error before table is altered or created.

Tests?
Two tests were developed. One for volatile case, which is prohibited. Another
one for immutable - it works. These cases applied to both ALTER TABLE (ADD
COLUMN, ALTER COLUMN, SET DISTRIBUTED) and CREATE TABLE.

Changes from original commit:
Logic to prohibit violating policy via usage of LIKE during CREATE TABLE was
moved to DefineRelation(), as ATExecCookedColumnDefault() route does not exists
in 6.x.

Changes in tests: 
minor changes due to functions syntax difference and order of
error detail message, also some changes to ao and aocs tests were removed as
those tests are not present in 6.x.

(cherry picked from commit 7cd685d)

Ticket: GG-203

!!!REBASE-ONLY!!!

### DIFF
--- a/src/backend/catalog/heap.c
+++ b/src/backend/catalog/heap.c
@@ -2863,6 +2863,14 @@ AddRelationNewConstraints(Relation rel,
 			(IsA(expr, Const) &&((Const *) expr)->constisnull))
 			continue;
 
+		/* Protect replicated tables from volatile expressions as DEFAULT value */
+		if (GpPolicyIsReplicated(rel->rd_cdbpolicy) &&
+			contain_volatile_functions_not_nextval(expr))
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
+					 errmsg("volatile expressions are not supported as "
+							"default values for columns in replicated tables")));
+
 		StoreAttrDefault(rel, colDef->attnum, expr, is_internal);
 
 		cooked = (CookedConstraint *) palloc(sizeof(CookedConstraint));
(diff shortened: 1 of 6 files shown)


RekGRpth:
The commit message must indicate the changes compared to the original.

RekGRpth:
It's better to indicate `co-authored-by` at the very end of the message, otherwise GitHub will display it incorrectly.

VoidZeroNull0:
> The commit message must indicate the changes compared to the original.

> It's better to indicate co-authored-by at the very end of the message

Gonna change this.

VoidZeroNull0:
>The commit message must indicate the changes compared to the original.

>It's better to indicate co-authored-by at the very end of the message, otherwise GitHub will display it incorrectly.

Done both.

RekGRpth:
> Changes from original commit:

This should be written NOT INSTEAD of the original message, but AFTER it.
It should also be formatted to 80 characters.

RekGRpth:
> This should be written NOT INSTEAD of the original message, but AFTER it.
> It should also be formatted to 80 characters.

I mean commit descr, not only pr descr

RekGRpth (approved):

red1452 (approved):

PR #371, opened 2026-04-03
GG-360: Prohibit moving from bypassed resource group

RekGRpth:
Prohibit moving from bypassed resource group

Some "simple" queries could be bypassed by the check_and_unassign_from_resgroup
function. After this, attempting to move these bypassed queries to another
resource group would return an incorrect error message stating "process <pid>
is in IDLE state." However, the pg_stat_activity view showed that these queries
were still active. This occurred because, before bypassing the queries, the
UnassignResGroup->sessionResetSlot functions reset the current session's
resGroupSlot field, which the ResGroupGetGroupIdBySessionId function used to
determine the current group when moving the session. If no current group was
specified, the pg_resgroup_move_query move function assumed the session was in
an idle state. Proper error messages should be issued regarding the inability
to move these bypassed queries. Store the group ID in the bypassResGroupId
field of the current session and add a new IsResGroupBypassedBySessionId
function to check for the setting of this field. Add the corresponding test.

Ticket: GG-360


VoidZeroNull0 (approved):
LGTM.

GirgerS (approved):

PR #372, opened 2026-04-06
ADBDEV-9615: Compile backend libpq for regress_gp.so

bandetto:
Compile backend libpq for regress_gp.so

Modified regress.so build to symlink and compile libpq sources without FRONTEND
macro.

The regress_gp.so now uses backend memory management (palloc()/pfree()) instead
of frontend allocators (malloc()/free()).


RekGRpth:
How is this same problem solved on 6X?

bandetto:
Commit b3ad725c12092ec8eec6017bc0c896db52845c17 does not exist on 6x.

RekGRpth:
> Commit [b3ad725](https://github.com/GreengageDB/greengage/commit/b3ad725c12092ec8eec6017bc0c896db52845c17) does not exist on 6x.

Compare symlinks in `src/backend/libpq` and `src/interfaces/libpq` for 6X and 7X.

src/interfaces/libpq/libpq-int.h
```diff
@@ -729,6 +729,7 @@ extern int	pqWaitTimeout(int forRead, int forWrite, PGconn *conn,
 						  time_t finish_time);
 extern int	pqReadReady(PGconn *conn);
 extern int	pqWriteReady(PGconn *conn);
+extern int	pqIsBackend(void);
```
RekGRpth, line 732: Is this necessary in production? Wouldn't it be better to do this only in debug builds?

src/test/regress/GNUmakefile
```diff
@@ -253,6 +271,8 @@ clean maintainer-clean: clean-lib
 	rm -f $(OBJS) refint$(DLSUFFIX) autoinc$(DLSUFFIX)
 	rm -f pg_regress_main.o pg_regress.o pg_regress$(X)
 	rm -f twophase_pqexecparams
+	rm -f regress_gp.so
+	rm -f $(LIBPQ_BE_SRCS)
```
RekGRpth, line 275: Do we need to remove `LIBPQ_BE_OBJS` as well?

src/interfaces/libpq/fe-misc.c
```diff
@@ -1373,6 +1373,20 @@ PQenv2encoding(void)
...
+ * Return 1 if the current library has been compiled for backend usage.
+ * 0 otherwise.
+ */
+int
+pqIsBackend(void)
+{
+#ifdef FRONTEND
+	return 0;
+#else
+	return 1;
```
RekGRpth, line 1386: why not `true` and `false`?

src/test/regress/GNUmakefile
bandetto: They are added to $(OBJS).

src/interfaces/libpq/fe-misc.c
bandetto: libpq does not use bool types.

src/interfaces/libpq/libpq-int.h
bandetto: Yeah, why not? If something needs to link with libpq, they can see whether it's the backend or frontend at runtime too.

VoidZeroNull0 (approved):
LGTM.

RekGRpth (approved):

PR #373, opened 2026-04-07
GG-356: Incorrect Host header when connecting gpfdist via K8s Ingress

VoidZeroNull0:
What happens?
Routing when using K8s Ingress and gpfdist fails with error 404. 

Why it happens?
When using gpfdist or gpfdists protocols, 'Host' field of header is not filled
with URI and instead automatically gets assigned with a resolved IP. So, on the
next hop to host routing fails, because host can have several domains hosted.
RFC 7320 explicitly states this.

How do we fix this?
'Host' field of HTTP request header must be original unresolved URI used in
request (not IP).

Tests?
No tests provided as it seems gpfdist only has regression tests and checking
HTTP header is not supported. Yet, it can be manually checked using utilities
like netcat.


VoidZeroNull0:
[RFC](https://datatracker.ietf.org/doc/html/rfc7230#section-5.4):
```
   The "Host" header field in a request provides the host and port
   information from the target URI, enabling the origin server to
   distinguish among resources while servicing requests for multiple
   host names on a single IP address.
```
```
   A client MUST send a Host header field in all HTTP/1.1 request
   messages.  If the target URI includes an authority component, then a
   client MUST send a field-value for Host that is identical to that
   authority component, excluding any userinfo subcomponent and its "@"
   delimiter (Section 2.7.1).  If the authority component is missing or
   undefined for the target URI, then a client MUST send a Host header
   field with an empty field-value.
```

GirgerS (dismissed):

src/backend/access/external/url_curl.c
```diff
@@ -1281,7 +1281,7 @@ url_curl_fopen(char *url, bool forwrite, extvar_t *ev, CopyState pstate)
 	/*
 	 * support multihomed http use cases. see MPP-11874
```
ShapatinAA: Should we update comment?

VoidZeroNull0: Seems like a good idea!

ShapatinAA (approved):

GirgerS (approved):

PR #374, opened 2026-04-07
GG-355: Changing GUC gp_role during session leads to segfault on a segment

GirgerS:
When changing gp_role to any role other than utility during a session, 
primary segments would segfault because the new value of the guc 
gets dispatched to them.

This functionality and the related code seems dubious and largely 
outdated, and there is no point in fixing it. So let's make a quick fix
of the immediate problem that may affect the users by making sure
that this code path never gets executed.

There is a proper fix of it already, f6297b96eeb4937629bd796333f6118340613197
but it never made it way to 6.x version. Sadly, I wasn't able to port 
it in a timely manner.
Therefore cleanup of the outdated logic remains as a separate task.

### DIFF
--- a/src/backend/cdb/cdbvars.c
+++ b/src/backend/cdb/cdbvars.c
@@ -495,11 +495,11 @@ assign_gp_session_role(const char *newval, void *extra)
 
 
 /*
- * Check and Assign routines for "gp_role" option.  This variable has context
- * PGC_SUSET so that is can only be set by a superuser via the SET command.
- * (It can also be set using an option on postmaster start, but this isn't
- * interesting beccause the derived global CdbRole is always set (along with
- * CdbSessionRole) on backend startup for a new connection.
+ * Check and Assign routines for "gp_role" option. Initially,
+ * this variable had context PGC_SUSET so that is could only be set by
+ * a superuser via the SET command. In greengage, we swapped it
+ * to PGC_BACKEND, as the code responsible for changing Gp_role at runtime
+ * is outdated and doesn't work anymore.
  *
  * See src/backend/util/misc/guc.c for option definition.
  */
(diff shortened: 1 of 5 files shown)


GirgerS:
Error messages when setting this guc during connection are still cryptic, but at least there are no segfaults.

```bash
regrig@fourier:~/greengage/6.x$ PGOPTIONS="-c gp_role=dispatch" psql postgres
psql (9.4.26)
Type "help" for help.

postgres=# \q
regrig@fourier:~/greengage/6.x$ PGOPTIONS="-c gp_role=execute" psql postgres
psql: FATAL:  reader could not find writer proc entry
DETAIL:  lock [0,2964] AccessShareLock 0. Probably because writer gang is gone somehow. Maybe try rerunning.
regrig@fourier:~/greengage/6.x$ PGOPTIONS="-c gp_role=utility" psql postgres
psql: FATAL:  System was started in master-only utility mode - only utility mode connections are allowed
regrig@fourier:~/greengage/6.x$ PGOPTIONS="-c gp_session_role=dispatch" psql postgres
psql (9.4.26)
Type "help" for help.

postgres=# \q
regrig@fourier:~/greengage/6.x$ PGOPTIONS="-c gp_session_role=execute" psql postgres
psql: FATAL:  reader could not find writer proc entry
DETAIL:  lock [0,2964] AccessShareLock 0. Probably because writer gang is gone somehow. Maybe try rerunning.
regrig@fourier:~/greengage/6.x$ PGOPTIONS="-c gp_session_role=utility" psql postgres
psql (9.4.26)
Type "help" for help.

postgres=# \q
```

hilltracer (approved):

KnightMurloc (approved):

PR #375, opened 2026-04-08
Implement progress reporting for ggrebalance

whitehawk:
Implement progress reporting for ggrebalance

This patch adds support for options:
1. '--detailed-progress',
2. '--simple-progress',
3. '--no-progress'.

When detailed or simple progress options are used, a view 'rebalance_progress'
is created in the ggrebalance schema. It allows to track the status of current
ggrebalance operation. The amount of details in the view depends on the
progress option selection.

As now the progress can be tracked via the mentioned view, function
'print_progress()' is totally removed, as it was a temporary solution.

Note: currently progress report is focused only on the shrink operation.

Plus fix test 'test 6.1' from 'ggrebalance_misc_options' suite
('--non-interactive-mode' was missing).

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
+++ b/gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
@@ -22,6 +22,12 @@ class RebalanceSchema:
     STATE_CATEGORY_REBALANCE = 'REBALANCE'
     STATE_CATEGORY_MAIN = 'MAIN'
 
+    class ProgressType(Enum):
+        PROGRESS_NOT_DEFINED = 0
+        PROGRESS_NO = 1
+        PROGRESS_SIMPLE = 2
+        PROGRESS_DETAILED = 3
+
     def __init__(self, conn: dbconn.Connection):
         self.schema_name = 'ggrebalance'
         self.rebalance_status = 'rebalance_status'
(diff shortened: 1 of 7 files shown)


gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
```diff
@@ -54,6 +67,206 @@ def createSchema(self, plan: Plan) -> None:
 
         self.savePlan(plan)
 
+        self.progress_type = progress_type
+
+        if self.progress_type != self.ProgressType.PROGRESS_NO:
+            states_str = ',\n'.join(f"'{s}'" for s in shrink_rollback_states)
+            dbconn.execSQL(self.conn, f'''
+CREATE FUNCTION ggrebalance._is_rollback()
```
KnightMurloc: If we take the schema name from the variable everywhere, should we use the variable here too?

bimboterminator1 (approved):

gpMgmt/bin/ggrebalance_modules/shrink.py
```diff
@@ -713,15 +713,59 @@ def prepare_shrink_schema(self, is_rollback: bool) -> None:
...
+                        BEGIN
+                            FOR rec IN
+                                SELECT
+                                    pg_catalog.pg_relation_size(c.oid) AS size_bytes
+                                FROM gp_dist_random('pg_class') c
+                                WHERE c.oid = p_rel_oid
+                                AND c.gp_segment_id {'<' if is_rollback else '>='} gp_segment_id_limit
+                            LOOP
+                                total := total + rec.size_bytes;
+                            END LOOP;
```
KnightMurloc: It looks like we can omit the loop by little hack:
```sql
with cte as MATERIALIZED (
    select * from gp_dist_random('pg_class') c WHERE c.oid = p_rel_oid and c.gp_segment_id >= gp_segment_id_limit
) select sum(pg_catalog.pg_relation_size(cte.oid)) from cte;
```

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
whitehawk: Yes, we should, Thank you for noticing.

KnightMurloc (approved):

PR #381, opened 2026-04-13
Enable coverage analysis for python code

ShapatinAA:
Introduce Python code coverage analysis for the GreengageDB gpMgmt component.
The changes enable automated collection, aggregation, and reporting of test
coverage data during behave BDD test execution as well as gpMgmt unit tests which
are currently enabled in the CI, helping developers identify untested code paths.

Changes include adding coverage to the Dockerfile image, injecting coverage
startup code into the sitecustomize.py file, running data processing and report
generation.

Running coverage analysis is handled by the COVERAGE_PROCESS_START variable,
which is set now for CI runs.

Reports are saved in the /tmp/coverage-data directory which is expected to be a
mounted volume when running docker container.

Co-authored-by: Denis Kovalev <d.kovalev@arenadata.io>


dkovalev1:
Are the coverage results are available yet? It would be great if the description had a steps how to see coverage results.

ci/scripts/behave_gpdb.bash
```diff
@@ -11,6 +11,17 @@ function gen_env(){
...
 		source /usr/local/greengage-db-devel/greengage_path.sh
 
+		cat > /tmp/sitecustomize.py <<INNER_EOF
+		import coverage
+		coverage.process_startup()
+		INNER_EOF
+
+		# Now copy everything over to the hosts.
+		gpscp -f /tmp/hostfile_all -v /tmp/sitecustomize.py =:/usr/local/greengage-db-devel/lib/python
+		gpssh -f /tmp/hostfile_all -v -e "echo 'export COVERAGE_PROCESS_START=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_behave' >> /usr/local/greengage-db-devel/greengage_path.sh"
```
RekGRpth: replace `/home/gpadmin/gpdb_src/gpMgmt/` with `\${1}/gpdb_src/gpMgmt/`, as below

ci/scripts/init_containers.sh
```diff
@@ -40,3 +40,11 @@ do
     $service bash -c "for HOST in $services; do echo \"\$(host \"\$HOST\" | grep 'has address' | head -n 1 | cut -d ' ' -f 4) \$HOST\" >>/etc/hosts; done" &
 done
 wait
+
+# Add host names of all cluster nodes to /tmp/hostfile_all for coverage collect
+for service in $services
+do
+  docker compose -p $project -f ci/docker-compose.yaml exec -T \
+    $service bash -c "for HOST in $services; do echo \"\$HOST\" >>/tmp/hostfile_all; done" &
+done
```
RekGRpth: I don't quite understand why the `hostfile_all` file is placed on all hosts if it is only used on one?

ci/scripts/behave_gpdb.bash
```diff
@@ -19,6 +30,19 @@ function gen_env(){
...
+		if [ ! -z "\${COVERAGE_PROCESS_START}" ] && [ \$(ls /tmp/pre-coverage-data | wc -l) -gt 0 ]; then
+			LOCK_FILE=/tmp/coverage-data/coverage.lock
+			flock "\$LOCK_FILE" -c "
+				mkdir -p /tmp/coverage-data
+				cp -r /tmp/pre-coverage-data/* /tmp/coverage-data/
+				cd /tmp/coverage-data
+				coverage combine --append --rcfile=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_behave coverage-data*
+				coverage html --rcfile=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_behave --show-contexts -d ./coverage-html
+			"
+		fi
```
RekGRpth: Why not make coverage collection a separate step in CI, like log collection?

ci/scripts/run_behave_tests.bash
```diff
@@ -45,9 +48,10 @@ run_feature() {
   fi
   echo "Started $feature behave tests on cluster $cluster and project $project"
   bash ci/scripts/init_containers.sh $project
-
+
```
RekGRpth: remove this change

ci/scripts/behave_gpdb.bash
RekGRpth: Why is it necessary to copy the export of the `COVERAGE_PROCESS_START` environment variable if it is already set when executing `behave_gpdb.bash` below?

concourse/scripts/ic_gpdb.bash
```diff
@@ -32,6 +32,7 @@ function gen_env(){
 		cd "\${1}/gpdb_src"
 		source gpAux/gpdemo/gpdemo-env.sh
 		export TEST_PGFDW=1
+		export COVERAGE_PROCESS_START="/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_unit"
```
RekGRpth: This change is redundant because we don't use this script anywhere.

ci/scripts/behave_gpdb.bash
```diff
@@ -11,6 +11,17 @@ function gen_env(){
 
 		source /usr/local/greengage-db-devel/greengage_path.sh
 
+		cat > /tmp/sitecustomize.py <<INNER_EOF
+		import coverage
+		coverage.process_startup()
+		INNER_EOF
```
RekGRpth: Why do this if there is a `coverage-setup` step below?

ci/scripts/behave_gpdb.bash
ShapatinAA: It gets set only for cdw and we need it on each host. It appears that previous launch of container with this environmental variable set via `-e` is excessive.

concourse/scripts/ic_gpdb.bash
ShapatinAA: We do use this script for running installcheck-world tests
```bash
docker run -i --name $LOG_NAME \
    -v $LOG_DIR:/logs \
    -e CI \
    -e TEST_OS \
    -e TEST_OS_VERSION \
    -e DUMP_DB \
    -e MAKE_TEST_COMMAND \
    --sysctl 'kernel.sem=500 1024000 200 4096' \
    ${IMAGE,,} \
    /bin/bash << 'END_OF_SCRIPT'
  set -x
  ssh-keygen -A
  /usr/sbin/sshd
  
  cd /home/gpadmin || exit 1
  
  exit_code=0
  bash gpdb_src/concourse/scripts/ic_gpdb.bash || exit_code=1
  echo "Script gpdb_src/concourse/scripts/ic_gpdb.bash finished with $exit_code"
  if [ -n "$DUMP_DB" ]; then
    tar -rf "/logs/${TEST_OS}${TEST_OS_VERSION}_postgres_sqldump.tar" ./sqldump/dump.sql || exit_code=$?
    echo "Dump collection './sqldump/dump.sql' finished with $exit_code"
  fi
  exit $exit_code
  END_OF_SCRIPT
```

ci/scripts/behave_gpdb.bash
ShapatinAA: Because when running behave tests we do not do do this step. Moreover we `make` on only one host, but we need to set up coverage environment on all hosts.

ci/scripts/behave_gpdb.bash
ShapatinAA: It is task for CI isn't it? This lines provide possibility for "local" (on ur own machine) coverage collection via launching behave test with scripts.

ci/scripts/init_containers.sh
ShapatinAA: Well yep, I didn't thought it would hurt anything and I didn't want to bring `cdw` host-name here.

ShapatinAA:
> Are the coverage results are available yet? It would be great if the description had a steps how to see coverage results.

May I clarify what do you mean? Instructions for user in coverages readme how to get results (in html-form)? Or current overall statistic should be placed somewhere?

ci/scripts/behave_gpdb.bash
RekGRpth: Place this code in a separate file that CI can call.

ci/scripts/behave_gpdb.bash
RekGRpth: > we need it on each host

So put it in compose.

dkovalev1:
> > Are the coverage results are available yet? It would be great if the description had a steps how to see coverage results.
> 
> May I clarify what do you mean? Instructions for user in coverages readme how to get results (in html-form)? Or current overall statistic should be placed somewhere?

I think both. For the user the is an instruction in readme, but for the CI we shall clearly state what to expect from the work, what to take and where to place.

> Reports are saved in the /tmp/coverage-data directory which is expected to be a
mounted volume when running docker container.

Docker compose already maps `/tmp/coverage-data` stores some results to the `coverage` folder. Do we have html reports yet? Do we have that results available after the run yet?

ci/scripts/behave_gpdb.bash
RekGRpth: Wouldn't it be easier to mount this file in compose on all hosts?

ci/scripts/behave_gpdb.bash
ShapatinAA: Do you mean we need to put it in `docker-compose.yaml` or `init_containers.sh`?
Both of this files used in resgroups and pollution of resgroups with coverage setup makes them fail.

ci/scripts/behave_gpdb.bash
ShapatinAA: Well maybe. But I haven't seen this kind of move anywhere in our project. Should we add dependency on docker compose like that?

ci/scripts/init_containers.sh
ShapatinAA: So I decided to just make this file on all hosts.

ci/scripts/behave_gpdb.bash
RekGRpth: You can hack compose in `run_behave_tests.bash` only for behave tests.

ci/scripts/behave_gpdb.bash
RekGRpth: You can hack compose in `run_behave_tests.bash` only for behave tests.

ci/scripts/init_containers.sh
RekGRpth: You can hack compose in run_behave_tests.bash only for behave tests.

ci/scripts/behave_gpdb.bash
ShapatinAA: What do you mean? How do I bring environmental variable from `run_behave_tests.bash` in running project?

ci/scripts/behave_gpdb.bash
ShapatinAA: Shouldn't we skip that? I mounted directory (`/tmp/coverage-data`, where all needed artifacts are being placed) to `gpdb_src/coverage` directory. So CI team **need only to collect** what in `gpdb_src/coverage` will be.
But, for `modularity` I can bring this code in different **file** in `ci/scripts` directory.

ci/scripts/behave_gpdb.bash
```diff
@@ -11,6 +11,17 @@ function gen_env(){
...
+
+		# Now copy everything over to the hosts.
+		gpscp -f /tmp/hostfile_all -v /tmp/sitecustomize.py =:/usr/local/greengage-db-devel/lib/python
+		gpssh -f /tmp/hostfile_all -v -e "echo 'export COVERAGE_PROCESS_START=\${1}/gpdb_src/gpMgmt/test/coveragerc_behave' >> /usr/local/greengage-db-devel/greengage_path.sh"
+
+		source /usr/local/greengage-db-devel/greengage_path.sh
```
RekGRpth: This all seems very hacky.

ci/scripts/init_containers.sh
```diff
@@ -40,3 +40,11 @@ do
...
+wait
```
RekGRpth:
```suggestion
for HOST in $services
do
  echo "$HOST\" >>/tmp/hostfile_all
done
```

concourse/scripts/ic_gpdb.bash
RekGRpth:
```suggestion
		export COVERAGE_PROCESS_START="gpMgmt/test/coveragerc_unit"
```
or
```suggestion
		export COVERAGE_PROCESS_START="\${1}/gpdb_src/gpMgmt/test/coveragerc_unit"
```

ci/scripts/run_behave_tests.bash
```diff
@@ -47,7 +50,8 @@ run_feature() {
   bash ci/scripts/init_containers.sh $project
```
RekGRpth, line 49:
```suggestion
  docker compose -p $project -f ci/docker-compose.yaml --env-file ci/behave.env up -d
  bash ci/scripts/init_containers.sh $project
```
and put `COVERAGE_PROCESS_START=...` to `ci/behave.env`

ci/scripts/run_behave_tests.bash
```diff
@@ -28,7 +28,10 @@ fi
 processes=3
 
 rm -rf allure-results
+rm -rf coverage
 mkdir allure-results -pm 777
+mkdir coverage
+chmod 777 coverage
```
RekGRpth:
```suggestion
cat > coverage/sitecustomize.py <<EOF
import coverage
coverage.process_startup()
EOF
```

ci/scripts/behave_gpdb.bash
RekGRpth:
```suggestion
		gpssh -f /tmp/hostfile_all -v -e "cp /tmp/coverage-data/sitecustomize.py /usr/local/greengage-db-devel/lib/python"
```

ci/scripts/behave_gpdb.bash
```diff
@@ -19,6 +30,20 @@ function gen_env(){
 		else
 				flags="\${BEHAVE_FLAGS}" make -f Makefile.behave behave
 		fi
+
+		gpscp -f /tmp/hostfile_all -v =:/tmp/pre-coverage-data/* /tmp/pre-coverage-data/ || true
```
RekGRpth: Why this unnecessary copying if you could have each host store the results in a shared volume in `/tmp/coverage-data`?

ci/scripts/run_behave_tests.bash
ShapatinAA: fixed

ci/scripts/behave_gpdb.bash
ShapatinAA: fixed

ci/scripts/behave_gpdb.bash
ShapatinAA: I used this function from 7.x as example for making it into 6.x
https://github.com/GreengageDB/greengage/blob/f7da5b00fa03aa570045908675b2b36cdcc7ba87/concourse/scripts/common.bash#L224-L256

ci/scripts/behave_gpdb.bash
ShapatinAA: Because behave tests run in parallel and while one test entered critical section, the other might continue to create coverage-associated files (which might be in form not suitable for `combine`)

RekGRpth: So, let your parallel tests write their results to different directories, and then combine everything at the very end.

ci/scripts/behave_gpdb.bash
RekGRpth: This doesn't mean it's not a very hacky.

ci/scripts/behave_gpdb.bash
ShapatinAA: Why would it be preferable, than `gpscp`? Because it would not waste time on copying?

ci/scripts/run_behave_tests.bash
ShapatinAA: As I understand this will not work, as `--env-file` is only for changing environment, where `docker-compose.yaml` takes it's values.

ci/scripts/init_containers.sh
ShapatinAA: This will create `/tmp/hostfile_all` on local machine, we can save it in `coverage/` and mount it within `docker-compose.yaml` but, then, resgroup tests will have it too, and then we need to make variance of `docker-compose.yaml` so...

ci/scripts/run_behave_tests.bash
ShapatinAA: Well, I can make move creation of this file, in run_behave_tests, but in the end we again `gpssh` it (instead of `gpscp`) and then it's just tossing around.

concourse/scripts/ic_gpdb.bash
ShapatinAA: fixed

ci/scripts/behave_gpdb.bash
ShapatinAA: Btw removed it from compose command execution on cdw

ci/scripts/init_containers.sh
RekGRpth: > but, then, resgroup tests will have it too

And in your version, on all hosts too!

ShapatinAA: But coverage directory created **only** in behave testing and as we don't need `hostfile_all` in resgroup we will need to make 2 different `init_containers` which seems to me much worse than having this in all tests.

RekGRpth: So, your version of `init_containers.sh` already affects resource groups! Move your code to `run_behave_tests.bash`

ci/scripts/run_behave_tests.bash
```diff
@@ -47,7 +50,7 @@ run_feature() {
   bash ci/scripts/init_containers.sh $project
 
   docker compose -p $project -f "$docker_compose_path" exec -T \
-    -e FEATURE="$feature" -e BEHAVE_FLAGS="--tags $feature --tags=$cluster \
+      -e FEATURE="$feature" -e BEHAVE_FLAGS="--tags $feature --tags=$cluster \
```
RekGRpth: remove this change

ShapatinAA: fixed

ci/scripts/init_containers.sh
ShapatinAA: fixed

ShapatinAA:
> I think both. For the user the is an instruction in readme, but for the CI we shall clearly state what to expect from the work, what to take and where to place.

Okay. Readme already has all instructions for that. Also I think task for CI team described good enough, so they should see and understand where to take and where to place artifacts after tests (exact names of directories where artifacts will be saved and how they will do it, I think is up to them)

> Docker compose already maps `/tmp/coverage-data` stores some results to the `coverage` folder. Do we have html reports yet? Do we have that results available after the run yet?

If we run them locally, will be having results for each of behave test in `coverage/` directory. But only for behave test, as we provide artifact "download" only for them (via mounting files with `volumes` in `docker-compose.yaml`)

ci/scripts/run_behave_tests.bash
```diff
@@ -46,6 +48,12 @@ run_feature() {
   echo "Started $feature behave tests on cluster $cluster and project $project"
   bash ci/scripts/init_containers.sh $project
 
+  services=$(docker compose -p $project -f ci/docker-compose.yaml config --services | tr '\n' ' ')
+
+  # Add host names of all cluster nodes to /tmp/hostfile_all for coverage collect
+  docker compose -p $project -f ci/docker-compose.yaml exec -T \
+    cdw bash -c "for HOST in $services; do echo \"\$HOST\" >>/tmp/hostfile_all; done"
```
RekGRpth:
```suggestion
for HOST in $services; do echo "$HOST" >>/tmp/hostfile_all; done
```

ci/scripts/behave_gpdb.bash
RekGRpth: How confident are you that none of the behave tests modify the `/tmp/hostfile_all` file (or will modify it in the future)?

ci/scripts/behave_gpdb.bash
```diff
@@ -19,6 +30,20 @@ function gen_env(){
...
+
+		if [ ! -z "\${COVERAGE_PROCESS_START}" ] && [ \$(ls /tmp/pre-coverage-data | wc -l) -gt 0 ]; then
+			LOCK_FILE=/tmp/coverage-data/coverage.lock
+			flock "\$LOCK_FILE" -c "
+				mkdir -p /tmp/coverage-data
+				cp -r /tmp/pre-coverage-data/* /tmp/coverage-data/
+				cd /tmp/coverage-data
+				coverage combine --append --rcfile=\${1}/gpdb_src/gpMgmt/test/coveragerc_behave coverage-data*
+				coverage html --rcfile=\${1}/gpdb_src/gpMgmt/test/coveragerc_combine_report --show-contexts -d ./coverage-html
+				mv /tmp/pre-coverage-data/coverage-data /tmp/coverage-data/
```
RekGRpth: On GitLab, all behave tests are run in parallel in four threads. Won't this code overwrite the results? Perhaps the results should be written to a sub-directory named `$project`?

ci/scripts/behave_gpdb.bash
ShapatinAA: Well, we were not using this file in 6.x, so this is it's first time appearance. I assume that nobody needed this file in past and so will not need it in future. And if someone will want to use file with name like this one, I think this someone will be forced to firstly check this code (as files-search will show `hostfile_all` name here) and no conflicts of modifications will be done.

Also I took this file from 7.x, from Pivotal "leftovers", and it served same purpose - names of all hosts in cluster.

ci/scripts/behave_gpdb.bash
ShapatinAA: This is why I made this section critical with 
```bash
LOCK_FILE=/tmp/coverage-data/coverage.lock
			flock "\$LOCK_FILE" -c "
			...
```

ShapatinAA: The only problem that I faced was with this command
```
coverage combine --append --rcfile=\${1}/gpdb_src/gpMgmt/test/coveragerc_behave coverage-data*
```
If 2 processes will simultaneously enter this section they would try to combine same files and one process will be removing files faster than other. So we will get error.

ci/scripts/run_behave_tests.bash
```diff
@@ -46,6 +48,12 @@ run_feature() {
...
+
```
RekGRpth, line 62:
```suggestion

for service in $services
do
  docker compose -p $project -f ci/docker-compose.yaml exec -T $service bash & -ex <<EOF
echo "import coverage; coverage.process_startup()" >/usr/local/greengage-db-devel/lib/python
echo "export COVERAGE_PROCESS_START=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_unit" >>/usr/local/greengage-db-devel/greengage_path.sh
EOF
done
wait
```

ci/scripts/behave_gpdb.bash
RekGRpth: That's why I suggest storing the results in separate sub-directories, so that additional locking isn't required.

ShapatinAA: If so, we will need to provide new global variable, so it can be referenced in configuration file as prefix/suffix of directory with coverage files. Will it be okay?
With this we can get rid of `hostfile_all`

RekGRpth: > With this we can get rid of `hostfile_all`

That's exactly what I want.

ci/scripts/run_behave_tests.bash
```diff
@@ -46,6 +48,12 @@ run_feature() {
...
   docker compose -p $project -f "$docker_compose_path" exec -T \
     -e FEATURE="$feature" -e BEHAVE_FLAGS="--tags $feature --tags=$cluster \
```
RekGRpth, line 64:
```suggestion
    -e FEATURE="$feature" -e PROJECT="$project" -e BEHAVE_FLAGS="--tags $feature --tags=$cluster \
```
for sub-directories

ci/scripts/behave_gpdb.bash
RekGRpth: > If so, we will need to provide new global variable, so it can be referenced in configuration file as prefix/suffix of directory with coverage files. Will it be okay?

yes, [see](https://github.com/GreengageDB/greengage/pull/381#discussion_r3199596865)

ci/scripts/run_behave_tests.bash
ShapatinAA: done

ci/scripts/run_behave_tests.bash
ShapatinAA: done

ci/scripts/behave_gpdb.bash
ShapatinAA: done

ci/scripts/run_behave_tests.bash
```diff
@@ -46,6 +48,17 @@ run_feature() {
...
+  services=$(docker compose -p $project -f ci/docker-compose.yaml config --services | tr '\n' ' ')
+  
+  # Setup coverage collection
+  for service in $services
+  do
+    docker compose -p $project -f ci/docker-compose.yaml exec -T $service bash -c "
+      echo 'import coverage; coverage.process_startup()' > /usr/local/greengage-db-devel/lib/python/sitecustomize.py &&
+      echo 'export COVERAGE_PROCESS_START=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_behave' >> /usr/local/greengage-db-devel/greengage_path.sh
+      echo 'export PROJECT=$project' >> /usr/local/greengage-db-devel/greengage_path.sh"
+  done
```
RekGRpth:
```suggestion
      echo 'export COVERAGE_PROCESS_START=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_behave' >> /usr/local/greengage-db-devel/greengage_path.sh &&
      echo 'export PROJECT=$project' >> /usr/local/greengage-db-devel/greengage_path.sh" &
  done
  wait
```

ci/scripts/run_behave_tests.bash
```diff
@@ -46,6 +48,17 @@ run_feature() {
   echo "Started $feature behave tests on cluster $cluster and project $project"
   bash ci/scripts/init_containers.sh $project
 
+  services=$(docker compose -p $project -f ci/docker-compose.yaml config --services | tr '\n' ' ')
+  
+  # Setup coverage collection
+  for service in $services
+  do
+    docker compose -p $project -f ci/docker-compose.yaml exec -T $service bash -c "
+      echo 'import coverage; coverage.process_startup()' > /usr/local/greengage-db-devel/lib/python/sitecustomize.py &&
```
RekGRpth, line 57: Can we avoid hardcoding the `/usr/local/greengage-db-devel` path? It could be configured to a different path...

ci/scripts/run_behave_tests.bash
```diff
@@ -56,6 +69,31 @@ run_feature() {
     cdw gpdb_src/ci/scripts/behave_gpdb.bash
   status=$?
 
+  docker compose -p "$project" -f "$docker_compose_path" exec -T \
+    -e FEATURE="$feature" -e PROJECT="$project" \
+    cdw bash -s <<'EOF'
+      set -ex
```
RekGRpth:
```suggestion
    cdw bash -eux <<EOF
```

ci/scripts/run_behave_tests.bash
```diff
@@ -56,6 +69,31 @@ run_feature() {
...
+
+      LOCK_FILE=/tmp/coverage-data/coverage.lock
+			flock "$LOCK_FILE" -c "
+        coverage combine --append \
+          --rcfile=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_combine_report \
+          coverage-data*
+        coverage html \
+          --rcfile=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_combine_report \
+          --show-contexts -d ./coverage-html
+      "
```
RekGRpth, line 95: Why do you need to do this separately for each `$PROJECT`? Why can't you just do it once, after all `$PROJECT`s have been executed?

gpMgmt/test/coveragerc_behave
```diff
@@ -0,0 +1,38 @@
...
+parallel = True
+context = ${FEATURE}
+data_file = /tmp/coverage-data/$PROJECT-coverage-data/coverage-data
+branch = True
+omit =
+        */__init__.py*
+        */sitecustomize.py*
+        */test/*
+        */gpload_test/*
+        /usr/local/greengage-db-devel/lib/python/pygresql/*
```
RekGRpth: Can we use `$GPHOME` instead of `/usr/local/greengage-db-devel` here?

gpMgmt/test/coveragerc_combine_report
```diff
@@ -0,0 +1,40 @@
...
+        /usr/local/greengage-db-devel/lib/python/yaml/*
+        /usr/local/greengage-db-devel/lib/python/sitecustomize.py
+        /usr/local/greengage-db-devel/lib/python/subprocess32.py
+        /usr/local/greengage-db-devel/lib/python/subprocess32.pyc
+        /usr/local/greengage-db-devel/bin/lib/pexpect/*
+        /usr/local/greengage-db-devel/lib/python/psutil/*
+        */dist-packages/*
+        /usr/local/bin/behave
+        /usr/local/bin/coverage
+        /tmp/*
```
RekGRpth: add new line

ci/scripts/run_behave_tests.bash
ShapatinAA: We need to have clusters environment (executable python files), for generating fancy report.

RekGRpth: For this purpose, you can launch a separate container at the very end.

ShapatinAA: I think starting and initializing new container after tearing previous down, so we can finish up the work on conceptual level worse than doing extra-work on each step. 
Beside this combination does reduce size of all artifacts from coverage collection. So the only extra is computation.

ci/scripts/run_behave_tests.bash
ShapatinAA: I wish to generalize this approach, but it seems, that we don't have meanings for it.
Thou to justify it I want to say that this code is being ran after `init_containers.sh`, which does this
`source gpdb_src/concourse/scripts/common.bash && install_gpdb &&`
where `install_gpdb()` installs database at one specific path, so nothing can be "configured" to a different path (unless we mess with this script).
```bash
function install_gpdb() {
	[ ! -d /usr/local/greengage-db-devel ] && mkdir -p /usr/local/greengage-db-devel
	tar -xzf bin_gpdb/bin_gpdb.tar.gz -C /usr/local/greengage-db-devel
}
```

gpMgmt/test/coveragerc_behave
ShapatinAA: fixed

gpMgmt/test/coveragerc_combine_report
ShapatinAA: fixed

ci/scripts/run_behave_tests.bash
ShapatinAA: fixed

ci/scripts/run_behave_tests.bash
ShapatinAA: fixed

ci/scripts/run_behave_tests.bash
```diff
@@ -56,6 +70,31 @@ run_feature() {
...
+      fi
+
+      LOCK_FILE=/tmp/coverage-data/coverage.lock
+			flock "$LOCK_FILE" -c "
+        coverage combine --append \
+          --rcfile=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_combine_report \
+          coverage-data*
+        coverage html \
+          --rcfile=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_combine_report \
+          --show-contexts -d ./coverage-html
```
RekGRpth, line 94: Is it possible to move at least this part to the very end? To avoid a lock file?

ShapatinAA: What do you mean by "end"? 

If we're gonna leave this code inside `run_feature()` we'll be having race condition in any way, as simultaneous 
`coverage combine --append --rcfile=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_combine_report coverage-data*` will fight for `coverage-data*` files.
 
And moving it out would solve this problem, but would require clusters setup again.

RekGRpth (approved):

dkovalev1 (approved):

PR #384, opened 2026-04-13
Eliminate demo cluster creation before behave tests (#193)

red1452:
Eliminate demo cluster creation before behave tests

Currently, behave tests require explicitly creating a demo cluster and setting
the PGPORT and MASTER_DATA_DIRECTORY variables before running. Create a demo
cluster and set the variables before running feature tests.

Previously, the init_cluster fixture was used only for the concourse cluster,
which doesn't yet run on our CI. In this patch, this fixture is now used for
the demo cluster, and will be used for the concourse cluster in a separate PR.

Ticket: GG-155
(cherry picked from commit https://github.com/GreengageDB/greengage/commit/0977773599ee1c855a504c06cebacdda8ea4001d)

Fixed conflicts after porting patch from GG6: master_data_dir was renamed to
coordinator_data_dir.

##

Do not squash

### DIFF
--- a/ci/scripts/behave_gpdb.bash
+++ b/ci/scripts/behave_gpdb.bash
@@ -11,8 +11,6 @@ function gen_env(){
 
 		source /usr/local/greengage-db-devel/greengage_path.sh
 
-		source gpdb_src/gpAux/gpdemo/gpdemo-env.sh
-
 		cd "\${1}/gpdb_src/gpMgmt/"
 		BEHAVE_TAGS="${BEHAVE_TAGS}"
 		BEHAVE_FLAGS="${BEHAVE_FLAGS}"
@@ -32,10 +30,6 @@ function _main() {
 				echo "FATAL: BEHAVE_TAGS or BEHAVE_FLAGS not set"
 				exit 1
 		fi
-
-		# Run inside a subshell so it does not pollute the environment after
-		# sourcing greengage_path
-		time (make_cluster)
 
 		time gen_env
 
--- a/gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
+++ b/gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
@@ -40,9 +40,7 @@ from gppylib import pgconf
 from gppylib.commands.gp import get_coordinatordatadir
 from gppylib.parseutils import canonicalize_address
 
-coordinator_data_dir = gp.get_coordinatordatadir()
-if coordinator_data_dir is None:
-    raise Exception('Please set COORDINATOR_DATA_DIRECTORY in environment')
+coordinator_data_dir = None
 
 def create_local_demo_cluster(context, extra_config='', with_mirrors='true', with_standby='true', num_primaries=None):
     stop_database_if_started(context)
@@ -51,15 +49,24 @@ def create_local_demo_cluster(context, extra_config='', with_mirrors='true', wit
         num_primaries = os.getenv('NUM_PRIMARY_MIRROR_PAIRS', 3)
 
     os.environ['PGPORT'] = '15432'
+    demoDir = os.path.abspath("%s/../gpAux/gpdemo" % os.getcwd())
+    global coordinator_data_dir
+    coordinator_data_dir = "%s/datadirs/qddir/demoDataDir-1" % demoDir
+    os.environ['COORDINATOR_DATA_DIRECTORY'] = coordinator_data_dir
+
     cmd = """
         cd ../gpAux/gpdemo &&
         export DEMO_PORT_BASE={port_base} &&
         export NUM_PRIMARY_MIRROR_PAIRS={num_primary_mirror_pairs} &&
+        export PGPORT={pgport} &&
+        export COORDINATOR_DATA_DIRECTORY={coordinator_data_dir} &&
         export WITH_STANDBY={with_standby} &&
         export WITH_MIRRORS={with_mirrors} &&
         ./demo_cluster.sh -d && ./demo_cluster.sh -c &&
         {extra_config} ./demo_cluster.sh
     """.format(port_base=os.getenv('PORT_BASE', 15432),
+               pgport=os.getenv('PGPORT', 15432),
+               coordinator_data_dir=os.getenv('COORDINATOR_DATA_DIRECTORY', coordinator_data_dir),
                num_primary_mirror_pairs=num_primaries,
                with_mirrors=with_mirrors,
                with_standby=with_standby,
@@ -121,10 +128,6 @@ def impl(context, checksum_toggle):
 
 @given('the cluster is generated with "{num_primaries}" primaries only')
 def impl(context, num_primaries):
-    os.environ['PGPORT'] = '15432'
-    demoDir = os.path.abspath("%s/../gpAux/gpdemo" % os.getcwd())
-    os.environ['COORDINATOR_DATA_DIRECTORY'] = "%s/datadirs/qddir/demoDataDir-1" % demoDir
-
     create_local_demo_cluster(context, with_mirrors='false', with_standby='false', num_primaries=num_primaries)
 
     context.gpexpand_mirrors_enabled = False
@@ -235,19 +238,27 @@ def impl(context, checksum_toggle):
             is_ok = False
 
     if not is_ok:
-        stop_database(context)
+        stop_database_if_started(context)
 
         os.environ['PGPORT'] = '15432'
         port_base = os.getenv('PORT_BASE', 15432)
+        demoDir = os.path.abspath("%s/../gpAux/gpdemo" % os.getcwd())
+        global coordinator_data_dir
+        coordinator_data_dir = "%s/datadirs/qddir/demoDataDir-1" % demoDir
+        os.environ['COORDINATOR_DATA_DIRECTORY'] = coordinator_data_dir
 
         cmd = """
         cd ../gpAux/gpdemo; \
             export DEMO_PORT_BASE={port_base} && \
             export NUM_PRIMARY_MIRROR_PAIRS={num_primary_mirror_pairs} && \
+            export PGPORT={pgport} &&
+            export COORDINATOR_DATA_DIRECTORY={coordinator_data_dir} &&
             export WITH_MIRRORS={with_mirrors} && \
             ./demo_cluster.sh -d && ./demo_cluster.sh -c && \
             env EXTRA_CONFIG="HEAP_CHECKSUM={checksum_toggle}" ./demo_cluster.sh
         """.format(port_base=port_base,
+                   pgport=os.getenv('PGPORT', 15432),
+                   coordinator_data_dir=os.getenv('COORDINATOR_DATA_DIRECTORY', coordinator_data_dir),
                    num_primary_mirror_pairs=os.getenv('NUM_PRIMARY_MIRROR_PAIRS', 3),
                    with_mirrors='true',
                    checksum_toggle=checksum_toggle)
@@ -2864,6 +2875,7 @@ def _create_cluster(context, coordinator_host, segment_host_list, hba_hostnames=
     global coordinator_data_dir
     coordinator_data_dir = os.path.join(context.working_directory, 'data/coordinator/gpseg-1')
     os.environ['COORDINATOR_DATA_DIRECTORY'] = coordinator_data_dir
+    os.environ['PGPORT'] = '10300'
 
     try:
         with closing(dbconn.connect(dbconn.DbURL(dbname='template1'), unsetSearchPath=False)) as conn:
--- a/gpMgmt/test/behave_utils/utils.py
+++ b/gpMgmt/test/behave_utils/utils.py
@@ -22,10 +22,7 @@ from gppylib.utils import escape_string
 PARTITION_START_DATE = '2010-01-01'
 PARTITION_END_DATE = '2013-01-01'
 
-coordinator_data_dir = get_coordinatordatadir()
-if coordinator_data_dir is None:
-    raise Exception('COORDINATOR_DATA_DIRECTORY is not set')
-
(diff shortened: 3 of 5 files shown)


RekGRpth:
Why did the tests fail?

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -51,15 +49,24 @@ def create_local_demo_cluster(context, extra_config='', with_mirrors='true', wit
         num_primaries = os.getenv('NUM_PRIMARY_MIRROR_PAIRS', 3)
 
     os.environ['PGPORT'] = '15432'
+    demoDir = os.path.abspath("%s/../gpAux/gpdemo" % os.getcwd())
+    global coordinator_data_dir
+    coordinator_data_dir = "%s/datadirs/qddir/demoDataDir-1" % demoDir
+    os.environ['COORDINATOR_DATA_DIRECTORY'] = coordinator_data_dir
```
dkovalev1, line 55: Is this environment var cleaned somewhere? Or its not needed?

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -51,15 +49,24 @@ def create_local_demo_cluster(context, extra_config='', with_mirrors='true', wit
...
         export NUM_PRIMARY_MIRROR_PAIRS={num_primary_mirror_pairs} &&
+        export PGPORT={pgport} &&
+        export COORDINATOR_DATA_DIRECTORY={coordinator_data_dir} &&
         export WITH_STANDBY={with_standby} &&
         export WITH_MIRRORS={with_mirrors} &&
         ./demo_cluster.sh -d && ./demo_cluster.sh -c &&
         {extra_config} ./demo_cluster.sh
     """.format(port_base=os.getenv('PORT_BASE', 15432),
+               pgport=os.getenv('PGPORT', 15432),
+               coordinator_data_dir=os.getenv('COORDINATOR_DATA_DIRECTORY', coordinator_data_dir),
```
dkovalev1, line 69: is it ever possible to get a fallback of default value of `coordinator_data_dir` here?

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
dkovalev1, line 104: Is it possible that `coordinator_data_dir` will have initial `None` value? (and other reading places)

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -2864,6 +2875,7 @@ def _create_cluster(context, coordinator_host, segment_host_list, hba_hostnames=
     global coordinator_data_dir
     coordinator_data_dir = os.path.join(context.working_directory, 'data/coordinator/gpseg-1')
     os.environ['COORDINATOR_DATA_DIRECTORY'] = coordinator_data_dir
+    os.environ['PGPORT'] = '15432'
```
RekGRpth: why not 10300 like in e9f1a92?

red1452: Such value is used at 7X: https://github.com/GreengageDB/greengage/blob/d5c3cd1c30a623d7ff41a8f71736e8bbf9d6807a/gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py#L53

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
red1452: No, it isn't. At line 54 we set correct value.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
RekGRpth: Port `15432` is the port for the demo cluster, and for the concourse cluster, the port should be `10300`.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
red1452: Function `check_database_is_running` returns "false", if value of `coordinator_data_dir` is `None`.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
red1452: There are places, where environments are cleaned.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
dkovalev1: then why to use `os.getenv()` here instead of just `coordinator_data_dir` ?

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
dkovalev1: I suggest defining constants like 
```
DEMO_CLUSTER_PORT=`15432`
CONCOURSE_CLUSTER_PORT=`10300`
```
to have less questions in advance

gpMgmt/bin/gppylib/test/unit/test_unit_recoveryinfo.py
```diff
@@ -534,7 +534,7 @@ def test_get_last_reported_error(self):
     def test_get_current_logfile_path(self):
         coordinator_datadir = os.environ.get('COORDINATOR_DATA_DIRECTORY')
         if not coordinator_datadir:
-            coordinator_datadir = os.environ.get('MASTER_DATA_DIRECTORY')
+            coordinator_datadir = os.environ.get('COORDINATOR_DATA_DIRECTORY')
```
dkovalev1: This like looks identical to 535. How will it help?

gpMgmt/bin/gppylib/commands/gp.py
```diff
@@ -1222,7 +1222,7 @@ def get_coordinatordatadir():
     else:
         coordinator_datadir = os.environ.get('COORDINATOR_DATA_DIRECTORY')
         if not coordinator_datadir:
-            coordinator_datadir = os.environ.get('MASTER_DATA_DIRECTORY')
+            coordinator_datadir = os.environ.get('COORDINATOR_DATA_DIRECTORY')
```
dkovalev1: This line is identical to 1223 and contradicts to comment above

ci/scripts/behave_gpdb.bash
```diff
@@ -33,10 +31,6 @@ function _main() {
 				exit 1
 		fi
 
-		# Run inside a subshell so it does not pollute the environment after
-		# sourcing greengage_path
-		time (make_cluster)
```
dkovalev1, line 38: [Here](https://github.com/GreengageDB/greengage/blob/c635992b5d5b6df7394b412210e77cf0e5da73fb/concourse/scripts/behave_gpdb.bash#L42) the code is almost identical. Why do we leave it in concourse scripts?

dkovalev1:
> and set the variables before running feature tests.

The term "feature test" does not appear anywhere in the source code and might confuse a reader (confuses me). Could you explain it in more details?

dkovalev1:
I believe it should be 2 commits in the PR - the original one and the changes made above to to adapt to 7.x.

dkovalev1:
Am I understand correctly that key change in this PR is start demo cluster in the `init_cluster` fixture instead of starting real cluster in docker compose, misleadingly called `concourse cluster` ? And as a consequence, do not start demo cluster in the script.  If so, the PR title "Eliminate demo cluster creation..." does not fully represent the intention.

gpMgmt/test/behave_utils/utils.py
```diff
@@ -185,10 +182,12 @@ def check_return_code(context, ret_code):
 
 
 def check_database_is_running(context):
-    if not 'PGPORT' in os.environ:
-        raise Exception('PGPORT should be set')
+    if not 'PGPORT' in os.environ or not 'COORDINATOR_DATA_DIRECTORY' in os.environ:
```
dkovalev1: I agree with SonarCloud here, pythonic construct is `A not in B`

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
RekGRpth: > I suggest defining constants like
> 
> ```
> DEMO_CLUSTER_PORT=`15432`
> CONCOURSE_CLUSTER_PORT=`10300`
> ```
> 
> to have less questions in advance

This will already be a refactoring, and not just a cherry-pick forward port.

ci/scripts/behave_gpdb.bash
RekGRpth: This is cherry pick and we don't use those scripts anywhere.

gpMgmt/bin/gppylib/commands/gp.py
RekGRpth: I agree, this change shouldn't be made. `MASTER_DATA_DIRECTORY` is used here for backward compatibility and is not a bug.

gpMgmt/bin/gppylib/test/unit/test_unit_recoveryinfo.py
RekGRpth: I agree, this change shouldn't be made. `MASTER_DATA_DIRECTORY` is used here for backward compatibility and is not a bug.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
RekGRpth: This is cherry pick

gpMgmt/test/behave_utils/utils.py
RekGRpth: Before the patch it was `not A in B` and in the original patch it was the same, but I agree, it would be more Pythonic to say `A not in B`, you can correct it with a description of the change relative to the original commit.

red1452:
> The term "feature test" does not appear anywhere in the source code and might confuse a reader (confuses me). Could you explain it in more details?

"feature test" means that all behave tests divides on features and can be run separately.

> I believe it should be 2 commits in the PR - the original one and the changes made above to to adapt to 7.x.

I think, that one commit for all is enough, because at each commit we should get working project, but if we separe patch on two commits, after the first commit we wiil get broken project.

> Am I understand correctly that key change in this PR is start demo cluster in the init_cluster fixture instead of starting real cluster in docker compose, misleadingly called concourse cluster ? And as a consequence, do not start demo cluster in the script. If so, the PR title "Eliminate demo cluster creation..." does not fully represent the intention.

You undestand correctly, but it is cherry-pick, and changing title here is not good idea.

dkovalev1:
> > The term "feature test" does not appear anywhere in the source code and might confuse a reader (confuses me). Could you explain it in more details?
> 
> "feature test" means that all behave tests divides on features and can be run separately.
> 
Then may be rephrase, like "before running behave tests" or "before running behave feature"?

> I think, that one commit for all is enough, because at each commit we should get working project, but if we separe patch on two commits, after the first commit we wiil get broken project.
> 
We had such a practice - PR commit contains to smaller ones - initial and delta. IMHO its much cleaner and easy to read when see what is original and what is the part of the patch. Project does not gets broken in this way. Example: https://github.com/arenadata/gpdb/pull/1546

> You undestand correctly, but it is cherry-pick, and changing title here is not good idea.

okay

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
RekGRpth: > dkovalev1 marked this comment as resolved

???

RekGRpth: My initial question is still actual.

red1452:
> > I think, that one commit for all is enough, because at each commit we should get working project, but if we separe patch on two commits, after the first commit we wiil get broken project.
> 
> We had such a practice - PR commit contains to smaller ones - initial and delta. IMHO its much cleaner and easy to read when see what is original and what is the part of the patch. Project does not gets broken in this way. Example: [arenadata/gpdb#1546](https://github.com/arenadata/gpdb/pull/1546)

@RekGRpth, do you agree? If yes, I will split the patch into two commits.

RekGRpth:
> > > I think, that one commit for all is enough, because at each commit we should get working project, but if we separe patch on two commits, after the first commit we wiil get broken project.
> > 
> > 
> > We had such a practice - PR commit contains to smaller ones - initial and delta. IMHO its much cleaner and easy to read when see what is original and what is the part of the patch. Project does not gets broken in this way. Example: [arenadata/gpdb#1546](https://github.com/arenadata/gpdb/pull/1546)
> 
> @RekGRpth, do you agree? If yes, I will split the patch into two commits.

No, I think that a back/forward-port should be one commit - a cherry-pick indicating the changes.

RekGRpth (approved):

dkovalev1:
> > > > I think, that one commit for all is enough, because at each commit we should get working project, but if we separe patch on two commits, after the first commit we wiil get broken project.
> > > 
> > > 
> > > We had such a practice - PR commit contains to smaller ones - initial and delta. IMHO its much cleaner and easy to read when see what is original and what is the part of the patch. Project does not gets broken in this way. Example: [arenadata/gpdb#1546](https://github.com/arenadata/gpdb/pull/1546)
> > 
> > 
> > @RekGRpth, do you agree? If yes, I will split the patch into two commits.
> 
> No, I think that a back/forward-port should be one commit - a cherry-pick indicating the changes.

okay

dkovalev1 (approved):

PR #387, opened 2026-04-15
GG-389: Optimize container start

dkovalev1:
In the docker container gpadmin user is created when all sources already copied 
and permissions are set after that with the chown -R command.
This involves modifying all files in the docker FS layer, which is copy-on-write,
and leads to rather long container start and extra disk space usage.

This patch sets proper permissions when copying files during image creation.

GG-389


ci/Dockerfile.ubuntu
```diff
@@ -56,9 +56,13 @@ RUN rm -rf gpdb_src/.git/
 
 FROM base AS test
 
-COPY --from=code  /home/gpadmin/gpdb_src gpdb_src
-COPY --from=build /home/gpadmin/bin_gpdb bin_gpdb
-COPY --from=build /home/gpadmin/gpdb_src/VERSION gpdb_src
+COPY --from=build /home/gpadmin/gpdb_src/concourse/scripts gpdb_src/concourse/scripts
+
+RUN /home/gpadmin/gpdb_src/concourse/scripts/setup_gpadmin_user.bash ubuntu
```
RekGRpth: I don't think generating keys for ssh when building an image is a good idea.

concourse/scripts/setup_gpadmin_user.bash
```diff
@@ -41,7 +41,7 @@ transfer_ownership() {
     [ -f gpdb_src/gpMgmt/bin/gpload_test/gpload2/data_file.csv ] && chown gpadmin:gpadmin gpdb_src/gpMgmt/bin/gpload_test/gpload2/data_file.csv
     [ -d /usr/local/gpdb ] && chown -R gpadmin:gpadmin /usr/local/gpdb
     [ -d /usr/local/greengage-db-devel ] && chown -R gpadmin:gpadmin /usr/local/greengage-db-devel
-    chown -R gpadmin:gpadmin /home/gpadmin
+    [ -d /home/gpadmin ] && chown -R gpadmin:gpadmin /home/gpadmin
```
RekGRpth: Why set the permissions again if they were already set when building the image?

dkovalev1: Normally at container start these lines are not executed. They executed when building image.
At the container start it becomes a defensive code, for the case if permissions were not set properly for some reason.

RekGRpth: > at container start these lines are not executed

why? Isn't this what is done when calling `setup_gpadmin_user.bash`?

dkovalev1: [this was the idea of the patch](https://github.com/GreengageDB/greengage/blob/0f17ecbea6d437b0ed9b4f1c212221fcf72d1a90/concourse/scripts/setup_gpadmin_user.bash#L68)

RekGRpth:
Why did the tests fail?

ci/Dockerfile.ubuntu
RekGRpth: And how is the argument `ubuntu` used?

RekGRpth: Is it possible to source this file and call only the add user function from it? Or at least pass the user setup request as an argument?

RekGRpth:
> Why did the tests fail?

```
20260415:07:26:28:000399 gpinitsystem:cdw:gpadmin-[WARN]:-File permission mismatch.  The gpad owns the Greengage Database installation directory.
20260415:07:26:28:000399 gpinitsystem:cdw:gpadmin-[WARN]:-You are currently logged in as root and may not have sufficient
20260415:07:26:28:000399 gpinitsystem:cdw:gpadmin-[WARN]:-permissions to run the Greengage binaries and management utilities.
```

concourse/scripts/setup_gpadmin_user.bash
```diff
@@ -58,7 +58,16 @@ set_limits() {
...
 setup_gpadmin_user() {
-  groupadd supergroup
+  # grep -q "^${supergroup}:" /etc/group
+  if ! getent group supergroup > /dev/null; then
+    groupadd supergroup
+  fi
+
+  if id -u gpadmin > /dev/null; then
+    echo "gpadmin user already exists"
+    return
```
RekGRpth: So, it turns out that directories mounted during container startup will have incorrect permissions?

RekGRpth: In a similar situation, I use this [hack](https://github.com/RekGRpth/postgres.docker/blob/12aa8161a8d78f363e09cf9ed73c93a1cbc9564a/bin/docker_entrypoint.sh#L21).

concourse/scripts/common.bash
```diff
@@ -55,8 +55,9 @@ function build_arch() {
 ## ----------------------------------------------------------------------
 
 function install_gpdb() {
-	[ ! -d /usr/local/greengage-db-devel ] && mkdir -p /usr/local/greengage-db-devel
-	tar -xzf bin_gpdb/bin_gpdb.tar.gz -C /usr/local/greengage-db-devel
+	gpdb_bin_dir=/usr/local/greengage-db-devel
+	[ ! -d ${gpdb_bin_dir} ] && mkdir -p ${gpdb_bin_dir} && chown gpadmin:gpadmin ${gpdb_bin_dir}
+	su gpadmin -c "tar -xzf bin_gpdb/bin_gpdb.tar.gz -C ${gpdb_bin_dir}"
```
RekGRpth:
```suggestion
	[ ! -d $gpdb_bin_dir ] && mkdir -p $gpdb_bin_dir && chown gpadmin:gpadmin $gpdb_bin_dir
	su gpadmin -c "tar -xzf bin_gpdb/bin_gpdb.tar.gz -C $gpdb_bin_dir"
```

concourse/scripts/setup_gpadmin_user.bash
```diff
@@ -57,8 +57,13 @@ set_limits() {
...
 }
 
-setup_gpadmin_user() {
-  groupadd supergroup
+add_gpadmin_user() {
+  if ! getent group supergroup > /dev/null; then
+    groupadd supergroup
+  fi
+
+  TEST_OS=$1
```
RekGRpth:
```suggestion
  TEST_OS="$(determine_os)"
```

concourse/scripts/setup_gpadmin_user.bash
```diff
@@ -73,13 +78,17 @@ setup_gpadmin_user() {
...
+
+}
+
+setup_gpadmin_user() {
+
   echo -e "password\npassword" | passwd gpadmin
   # Add user to sudoers list required for gpinitsystem test
   echo "gpadmin ALL = NOPASSWD : ALL" >> /etc/sudoers
   setup_ssh_for_user gpadmin
-  transfer_ownership
```
RekGRpth, line 82: This `transfer_ownership` function is no longer used anywhere? Why, then, was it changed above?

concourse/scripts/common.bash
dkovalev1: please explain why to change

RekGRpth: Remove redundant braces, they are only needed in cases like these: `${gpdb_bin_dir}_6`

dkovalev1: both styles are used in the script, braces more often

concourse/scripts/setup_gpadmin_user.bash
```diff
@@ -152,10 +151,12 @@ workaround_before_concourse_stops_stripping_suid_bits() {
 }
 
 _main() {
+  set -e
   TEST_OS=$(determine_os)
   setup_gpadmin_user
```
RekGRpth, line 189: why not
```suggestion
  setup_ssh_for_user gpadmin
```

dkovalev1: Just to keep the place where things performed at the container start. I considered it is more convenient.

concourse/scripts/setup_gpadmin_user.bash
```diff
@@ -73,16 +68,20 @@ setup_gpadmin_user() {
...
+
   echo -e "password\npassword" | passwd gpadmin
   # Add user to sudoers list required for gpinitsystem test
   echo "gpadmin ALL = NOPASSWD : ALL" >> /etc/sudoers
-  setup_ssh_for_user gpadmin
-  transfer_ownership
   set_limits
 }
 
+setup_gpadmin_user() {
```
RekGRpth, line 111: Is this function needed anywhere else?

dkovalev1: This function was a place where user related actions performed at container start performed. Now only ssh keys are created here, but  I left it to keep the overall structure.

RekGRpth: Are there any examples where this function is called separately, and not when calling this file?

dkovalev1: not yet

RekGRpth:
How about moving the `install_and_configure_gpdb` (or at least `install_gpdb`) call from startup to the image? But in a way that would allow it to be called at startup (in which case it wouldn't do anything). Once we've eliminated the startup call, we can remove the unused `bin_gpdb.tar.gz` archives from the image.

RekGRpth:
Replace the word `adb` in the title!

ci/Dockerfile.ubuntu
```diff
@@ -56,10 +56,17 @@ RUN rm -rf gpdb_src/.git/
...
+    
+COPY --chown=gpadmin:gpadmin --from=code /home/gpadmin/gpdb_src gpdb_src
+COPY --chown=gpadmin:gpadmin --from=build /home/gpadmin/bin_gpdb bin_gpdb
+COPY --chown=gpadmin:gpadmin --from=build /home/gpadmin/gpdb_src/VERSION gpdb_src
+
+RUN source /home/gpadmin/gpdb_src/concourse/scripts/common.bash; install_and_configure_gpdb; \
 # Install entab used by pgindent utility.
 # This should be done using gpdb sources.
-RUN make -C gpdb_src/src/tools/entab install clean
+    make -C gpdb_src/src/tools/entab install clean
```
RekGRpth: Either use `&&` between commands, or add `set -eux` at the beginning, otherwise if there is an error in the first commands, the rest will still be executed.

RekGRpth: still actual for first command, I prefer `set -eux`

dkovalev1: Construct like
```
[ "${BASH_SOURCE[0]}" = "$0" ] && _main "$@"
```
from `setup_gpadmin_user.bash` returns 1 if script sourced and  _main not executed, `set -eux` fails too

RekGRpth: So call the `setup_gpadmin_user.bash` file with the `add_gpadmin_user` argument, not `source` it.

dkovalev1: fixed

ci/Dockerfile.ubuntu
```diff
@@ -56,10 +56,19 @@ RUN rm -rf gpdb_src/.git/
 
 FROM base AS test
 
-COPY --from=code  /home/gpadmin/gpdb_src gpdb_src
-COPY --from=build /home/gpadmin/bin_gpdb bin_gpdb
-COPY --from=build /home/gpadmin/gpdb_src/VERSION gpdb_src
+COPY --from=build /home/gpadmin/gpdb_src/concourse/scripts gpdb_src/concourse/scripts
 
+RUN source /home/gpadmin/gpdb_src/concourse/scripts/setup_gpadmin_user.bash; \
```
RekGRpth:
```suggestion
RUN source gpdb_src/concourse/scripts/setup_gpadmin_user.bash; \
```

ci/Dockerfile.ubuntu
```diff
@@ -56,10 +56,19 @@ RUN rm -rf gpdb_src/.git/
...
+    add_gpadmin_user && \
+    chown -R gpadmin:gpadmin /home/gpadmin
```
RekGRpth:
```suggestion
    chown -R gpadmin:gpadmin .
```

ci/Dockerfile.ubuntu
```diff
@@ -56,10 +56,19 @@ RUN rm -rf gpdb_src/.git/
...
+    
+COPY --chown=gpadmin:gpadmin --from=code /home/gpadmin/gpdb_src gpdb_src
+COPY --chown=gpadmin:gpadmin --from=build /home/gpadmin/bin_gpdb bin_gpdb
+COPY --chown=gpadmin:gpadmin --from=build /home/gpadmin/gpdb_src/VERSION gpdb_src
+
+RUN set -eux; \
+    source /home/gpadmin/gpdb_src/concourse/scripts/common.bash; \
```
RekGRpth:
```suggestion
    source gpdb_src/concourse/scripts/common.bash; \
```

concourse/scripts/common.bash
```diff
@@ -72,12 +79,17 @@ function configure() {
 	# The full set of configure options which were used for building the
 	# tree must be used here as well since the toplevel Makefile depends
 	# on these options for deciding what to test. Since we don't ship
-	./configure --prefix=/usr/local/greengage-db-devel --with-perl --with-python --with-libxml --with-uuid=e2fs --enable-mapreduce --enable-orafce --enable-tap-tests --disable-orca --with-openssl ${CONFIGURE_FLAGS}
+	su gpadmin -c "./configure --prefix=/usr/local/greengage-db-devel --with-perl --with-python --with-libxml --with-uuid=e2fs --enable-mapreduce --enable-orafce --enable-tap-tests --disable-orca --with-openssl ${CONFIGURE_FLAGS}"
```
RekGRpth:
```suggestion
	su gpadmin -c "./configure --prefix=${gpdb_bin_dir} --with-perl --with-python --with-libxml --with-uuid=e2fs --enable-mapreduce --enable-orafce --enable-tap-tests --disable-orca --with-openssl ${CONFIGURE_FLAGS}"
```

RekGRpth: and in other all places in this file

ci/Dockerfile.ubuntu
```diff
@@ -56,10 +56,19 @@ RUN rm -rf gpdb_src/.git/
 
 FROM base AS test
 
-COPY --from=code  /home/gpadmin/gpdb_src gpdb_src
-COPY --from=build /home/gpadmin/bin_gpdb bin_gpdb
-COPY --from=build /home/gpadmin/gpdb_src/VERSION gpdb_src
+COPY --from=build /home/gpadmin/gpdb_src/concourse/scripts gpdb_src/concourse/scripts
 
+RUN source /home/gpadmin/gpdb_src/concourse/scripts/setup_gpadmin_user.bash; \
+    add_gpadmin_user && \
```
magf, line 72: May be parametrize `gpadmin` name for files in Docker?
Or use `1000:1000`.

magf:
I agree that avoiding `chown -R` is a valid optimization, but I don't think this change addresses the actual problem.

The root cause is that the image is still built and run around a privileged `root` user.
We should instead design the image and startup flow to run as the unprivileged `gpadmin` user from the beginning.

A non-root container model would also eliminate host-side ownership issues when bind-mounting directories for logs, build artifacts, test results, or source code, and would reduce the impact of incorrect volume mappings.

 ---

This patch improves one consequence of the current design, but I would prefer to address the underlying issue and move toward a fully non-root container model.

magf (dismissed):

ci/Dockerfile.ubuntu
dkovalev1: `1000:1000` is not a good practice anymore - ubuntu 24.04 has a user `ubuntu` with id 1000 which interfere. 

Also I do not see a need to make a `gpadmin` user parametrized at this moment - with current scripts its useless, while changing user in scripts is a big change and not related to this optimization.

PR #391, opened 2026-04-15
GG-390 Bump resgroup workflow version (6.x)

uaqq:
### Problem

Previously, when resgroup tests were cancelled, logs were not always collected and uploaded as artifacts. This made debugging failures difficult.

#### Start the QEMU VM

Initialize and boot the QEMU virtual machine.

#### Run tests via SSH

Connect to the QEMU VM via SSH and run tests.

#### Collect logs (separate step)

This step runs regardless of test outcome (`always()`), ensuring logs are preserved.

### Implementation details

Log collection is decoupled from test execution.

SSH configuration is prepared to enable interaction with Docker running inside the QEMU VM:

- `DOCKER_HOST=ssh://qemu-vm` variable is used to run Docker commands directly on the VM.

Log collection removed from `run_resgroup.sh` script.
CI condition was added to run_resgroup.sh script.

Task: GG-180

### DIFF
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -103,7 +103,7 @@ jobs:
       contents: read  # Explicit for default behavior
       packages: read  # Explicit for GHCR access clarity
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-resgroup.yml@v28
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-resgroup.yml@v29
     with:
       version: 6
       target_os: ${{ matrix.target_os }}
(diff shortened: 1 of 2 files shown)


silent-observer:
In this PR no logs were collected ([here](https://github.com/GreengageDB/greengage/actions/runs/24532035844/job/71755568949?pr=391)), is that expected? Maybe we should test 6.x as well?

uaqq:
> In this PR no logs were collected ([here](https://github.com/GreengageDB/greengage/actions/runs/24532035844/job/71755568949?pr=391)), is that expected? Maybe we should test 6.x as well?

my bad. ci condition was added to run_resgroup_test.bash script and logs are collected succesfully

silent-observer (approved):

GirgerS (approved):

PR #392, opened 2026-04-16
GG-380: Fix gp_sparse_vector build issues

VoidZeroNull0:
What happens?
Override CFLAGS += -std=gnu99 is placed before contrib-global.mk include, which
resets CFLAGS, making the override ineffective. Also, warnings about
unitialized variables inside compword_to_int8() function in SparseData.h
outputed.

How do we fix this?
Place override after include to correctly append, also cherry-pick 3950340 to
both solve warnings  and provide additional fix for ARM.

Ticket: GG-380

!!!REBASE ONLY!!!

### DIFF
--- a/gpcontrib/gp_sparse_vector/SparseData.h
+++ b/gpcontrib/gp_sparse_vector/SparseData.h
@@ -46,7 +46,7 @@
  * Note that if the ptr is NULL, a zero size is returned
  */
 #define	int8compstoragesize(ptr) \
-	(((ptr) == NULL) ? 0 : (((*((char *)(ptr)) < 0) ? 1 : (1 + *((char *)(ptr))))) )
+	(((ptr) == NULL) ? 0 : (((*((int8 *)(ptr)) < 0) ? 1 : (1 + *((int8 *)(ptr))))) )
 
 /*------------------------------------------------------------------------------
  * SparseData holds information about a sparse array of values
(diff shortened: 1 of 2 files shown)


RekGRpth:
Why don't we want to make it compatible with the ARM platform, as in the commit https://github.com/GreengageDB/greengage/commit/3950340e689e8a534b9a41e70bf91909dfd0584b?

VoidZeroNull0:
> Why don't we want to make it compatible with the ARM platform, as in the commit 3950340?

Seems like a great idea. Was going to write about it.

gpcontrib/gp_sparse_vector/SparseData.h
```diff
@@ -248,11 +248,11 @@ static inline void int8_to_compword(int64 num, char entry[9])
...
 	char size = int8compstoragesize(entry);
-	int16_t num_2;
+	int16_t num_2 = 0;
 	char *numptr2 = (char *)(&num_2);
-	int32_t num_4;
+	int32_t num_4 = 0;
 	char *numptr4 = (char *)(&num_4);
-	int64 num;
+	int64 num = 0;
 	char *numptr8 = (char *)(&num);
```
RekGRpth: Maybe we should remove the changes to this file in the first commit?

RekGRpth:
> CFLAGS append gets overwritten by including contrib-global.mk

Could you provide more information about moving the flag? Something like

> override CFLAGS += -std=gnu99 is placed before contrib-global.mk include, which resets CFLAGS, making the override ineffective.

VoidZeroNull0:
> Could you provide more information about moving the flag?

Added to PR description.

gpcontrib/gp_sparse_vector/SparseData.h
VoidZeroNull0: Seems to be outdated:
https://github.com/GreengageDB/greengage/blob/b460c7175dcf29480c87427b83151ef6e549170c/gpcontrib/gp_sparse_vector/SparseData.h#L235-L243

Or do you mean revert and use just cherry-pick and modify Makefile?

RekGRpth: > do you mean revert and use just cherry-pick and modify Makefile?

Yes, in two different commits.

VoidZeroNull0: But what is the point, if they are going to get squashed? Or should I rebase?

RekGRpth: > rebase

only

VoidZeroNull0: Job's done.

RekGRpth:
> Ticket: GG-380

After this line in the first commit, you can add co-authored by...

RekGRpth:
I think we need to cherry-pick commit dd2d4bd here too.

RekGRpth (approved):

KnightMurloc (approved):

PR #394, opened 2026-04-16
GG-401: Fix isolation2 dependencies to avoid rebuilds

silent-observer:
Commit 44e96ec948f9caa7c8c667716a6d6c2d71e29716 added submake dependencies to
isolation2 Makefile. Unfortunately this caused isolation2_regress.so to be
rebuilt every time it was used, so on every installcheck. This caused
permission issues in diskquota CI. This patch switches to using order-only
dependencies in Makefile, as already done in many other Makefiles using
submake targets.

### DIFF
--- a/src/test/isolation2/Makefile
+++ b/src/test/isolation2/Makefile
@@ -53,13 +53,13 @@ explain.pm:
 data:
 	rm -f $@ && $(LN_S) $(top_builddir)/src/test/regress/data
 
-isolation2_regress.o: submake-libpgport
+isolation2_regress.o: | submake-libpgport
 
 .PHONY: scan_flaky_fault_injectors
 scan_flaky_fault_injectors:
 	$(top_builddir)/src/test/regress/scan_flaky_fault_injectors.sh
 
-pg_isolation2_regress$(X): isolation2_main.o pg_regress.o submake-libpgport submake-libpq
+pg_isolation2_regress$(X): isolation2_main.o pg_regress.o | submake-libpgport submake-libpq
 	$(CC) $(CFLAGS) $(filter %.o,$^) $(libpq_pgport) $(LDFLAGS) $(LDFLAGS_EX) $(LIBS) -o $@
 
 clean distclean:


red1452 (approved):

src/test/isolation2/Makefile
```diff
@@ -53,13 +53,13 @@ explain.pm:
 data:
 	rm -f $@ && $(LN_S) $(top_builddir)/src/test/regress/data
 
-isolation2_regress.o: submake-libpgport
+isolation2_regress.o: | submake-libpgport
```
hilltracer, line 56: How building of isolation2_regress.o related with isolation2_regress.so? Maybe describe it in more detail?

hilltracer, line 56: What if libpgport ever changes? Should we rebuild isolation2_regress.o in this case?

src/test/isolation2/Makefile
```diff
@@ -53,13 +53,13 @@ explain.pm:
...
 
 .PHONY: scan_flaky_fault_injectors
 scan_flaky_fault_injectors:
 	$(top_builddir)/src/test/regress/scan_flaky_fault_injectors.sh
 
-pg_isolation2_regress$(X): isolation2_main.o pg_regress.o submake-libpgport submake-libpq
+pg_isolation2_regress$(X): isolation2_main.o pg_regress.o | submake-libpgport submake-libpq
```
hilltracer, line 62: How pg_isolation2_regress is related with described isolation2_regress.so problem? Please explain in patch description why done this change?

src/test/isolation2/Makefile
silent-observer: isolation2_regress.so is built from isolation2_regress.o with rules in Makefile.shlib. To build isolation2_regress.o correctly we need generated imports from libpgport, so this dependency was added here

silent-observer: Technically yes, but this is the approach taken by most submake rules everywhere else, even in regress. This doesn't affect CI or release builds since those are rebuilt from scratch anyway, so it can only affect local development, where we can always call `make clean`. Also, note that libpgport is only used there for auto-generated header files, so I would assume those rarely change significantly enough to break this.

src/test/isolation2/Makefile
silent-observer: It is not directly related, and I think the build is still successful regardless. However, the order-only dependencies are used everywhere else for submake targets, so I think it's better to keep this consistent with all other cases.

hilltracer (approved):

PR #396, opened 2026-04-17
Add final summary to ggrebalance

whitehawk:
Add final summary to ggrebalance

This patch adds final summary output to ggrebalance tool. Final summary consists
from shrink and rebalance statistics plus warnings section. Shrink and rebalance
statistics are not output in case '--no-progress' options is used. Warnings
sections is output always (if any warnings exist).

Some details about the change:
 - 'configurationInterface' is added back, as now it is used when checking for
warnings.
 - In order to show correct shrink statistics in case of shrink rollback,
we save 'rebalance_progress' view into the 'rebalance_progress_history' table.
And add the information about shrink duration into the table, as it will be
lost as well during rollback. We do it in 'backupShrinkProgress()'.
 - Distribution of rebalance internal 'table_rebalance_status_detail' is
changed from 'replicated' to hash-distributed. There are 2 reasons: A) On a 
real cluster the table can be pretty large, so no need to make it replicated.
B) There is some strange issue that hits with ~50% reproduction rate (and only
via psycopg; not reproduced in psql) when performing CTAS from a view over the
replicated table - the result table has wrong data (as if the source table was
empty). It is a separate issue, out of scope of this PR. So far it is not
reproduced with non-replicated table.

New tests are not added. Some old tests are modified instead:
 - rebalance_test 8.1.3 - check final summary with '--simple-progress' plus 
 warnings due to cancelled moves that caused segments down;
 - rebalance_test 8.1.2 - check final summary with '--simple-progress' plus 
 warnings due to rolled back moves that caused unbalanced cluster;
 - rebalance_test 8.2.2 - check final summary with '--no-progress' plus 
 warnings due to rolled back moves that caused unbalanced cluster;
 - rebalance_test 8.2.4 - check final summary with '--simple-progress' plus 
 warnings due to cancelled back moves that caused unbalanced cluster;
 - rebalance_test 8.3.1 - check final summary with '--simple-progress' plus 
 warnings due to cancelled back moves that caused unbalanced cluster
 and segments down;
 - rebalance_test 2 - check final summary with '--simple-progress' with 
 shrink and rebalance;
 - misc_options_test 21 - check final summary with '--simple-progress' 
 with shrink and shrink rollback;
 - rebalance_test 3 - check final summary with '--simple-progress' with 
 no shrink and rebalance;
 - rebalance_test 10 - check final summary with '--simple-progress' with 
 no shrink and rebalance and full rebalance rollback;
 - misc_options_test 21 - check final summary with '--no-progress' with 
 shrink and shrink rollback;
 - misc_options_test 21 - check final summary with '--detailed-progress' 
 with shrink and shrink rollback;

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
+++ b/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
@@ -20,6 +20,8 @@ except ImportError as e:
 
 class GGRebalanceMainSM:
 
+    summary_separator_str = '================================================================================'
+
     states_not_logged = [
         'STATE_OPTIONS_VALIDATION',
         'STATE_CLEANUP',
@@ -194,9 +196,9 @@ class GGRebalanceMainSM:
             self.logger.info('Skip final shrink summary report (specify "--simple-progress" or "--detailed-progress" to enable it).')
             return
 
-        self.logger.info('================================================================================')
+        self.logger.info(self.summary_separator_str)
         self.logger.info('                                   SHRINK                                   ')
-        self.logger.info('================================================================================')
+        self.logger.info(self.summary_separator_str)
         self.logger.info(f'Tables shrunk:\t\t{summary.tables_shrunk}')
         if self.rebalance_schema.getProgressType() == self.rebalance_schema.ProgressType.PROGRESS_DETAILED:
             self.logger.info(f'Bytes processed:\t{summary.bytes_processed}')
@@ -232,9 +234,9 @@ class GGRebalanceMainSM:
         if self.rebalance_schema.getProgressType() == self.rebalance_schema.ProgressType.PROGRESS_NO:
             self.logger.info('Skip final rebalance summary report (specify "--simple-progress" or "--detailed-progress" to enable it).')
         else:
-            self.logger.info('================================================================================')
+            self.logger.info(self.summary_separator_str)
             self.logger.info('                                   REBALANCE                                   ')
-            self.logger.info('================================================================================')
+            self.logger.info(self.summary_separator_str)
             self.logger.info(f'Segments moved:\t\t{segments_moved}')
             self.logger.info(f'Rolled back moves:\t\t{len(rolled_back_moves)}')
             self.logger.info(f'Cancelled moves:\t\t{len(cancelled_moves)}')
@@ -250,9 +252,9 @@ class GGRebalanceMainSM:
         # we show warnings regardless the progress type set by user
         show_warnings = len(segments_down) > 0 or len(cancelled_moves) > 0 or not balanced
         if show_warnings:
-            self.logger.warning('================================================================================')
+            self.logger.warning(self.summary_separator_str)
             self.logger.warning('                                   WARNINGS                                    ')
-            self.logger.warning('================================================================================')
+            self.logger.warning(self.summary_separator_str)
 
             if len(cancelled_moves) > 0:
                 self.logger.warning('------------------------------- Cancelled moves  -------------------------------')
--- a/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
+++ b/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
@@ -7,18 +7,20 @@ try:
     from gppylib.commands.gp import *
     from gppylib.gplog import *
     from gppylib.system.environment import *
+    from gppylib.system import configurationInterface
     from ggrebalance_modules.planner import *
-    from ggrebalance_modules.rebalance_schema import RebalanceSchema
+    from ggrebalance_modules.rebalance_schema import RebalanceSchema, RebalanceSummaryInfo
     from gppylib.fault_injection import *
     from ggrebalance_modules.shrink import GGShrink
     from ggrebalance_modules.ggrebalance_sm import RebalanceSM
+    from ggrebalance_modules.rebalance_step import RebalanceStep
+    from ggrebalance_modules.rebalance_commons import is_gparray_balanced
 except ImportError as e:
     sys.exit('ERROR: Cannot import modules.  Please check that you have sourced greenplum_path.sh.  Detail: ' + str(e))
 
 class GGRebalanceMainSM:
 
     states_not_logged = [
-        'STATE_START',
         'STATE_OPTIONS_VALIDATION',
         'STATE_CLEANUP',
         'STATE_ROLLBACK',
@@ -30,6 +32,7 @@ class GGRebalanceMainSM:
     ]
 
     states_logged = [
+        'STATE_START',
         'STATE_SETUP_SCHEMA_STARTED',
         'STATE_SETUP_SCHEMA_DONE',
         'STATE_EXECUTOR_STARTED',
@@ -149,7 +152,7 @@ class GGRebalanceMainSM:
         self.shrink_state_from_prev_run = self.rebalance_schema.getShrinkStateFromPreviousRun()
         self.is_shrink_rollback_in_progress = self.gg_shrink.state_is_from_rollback_flow(self.shrink_state_from_prev_run)
         self.prev_shrink_run_was_complete = self.gg_shrink.state_is_final(self.shrink_state_from_prev_run)
-
+        self.is_summary_output_required = False
 
     def on_every_state(self) -> None:
         if self.state in self.states_logged:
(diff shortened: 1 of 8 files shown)


KnightMurloc:
> Distribution of rebalance internal 'table_rebalance_status_detail' is
changed from 'replicated' to hash-distributed.

Is it safe? Could the table become unavailable in some scenario because of this?

gpMgmt/bin/ggrebalance_modules/rebalance_commons.py
```diff
@@ -744,6 +745,25 @@ def check_batch_available_space(self,
...
+        if pair.mirrorDB.getSegmentHostName() == pair.primaryDB.getSegmentHostName():
+            return False
+    load = next(iter(primaries_by_host.values()))
+    for n in primaries_by_host.values():
+        if n != load:
+            return False
+    for n in mirrors_by_host.values():
+        if n != load:
+            return False
+    return True
```
KnightMurloc, line 765: We already have `already_balanced` in the planner. Is it possible to create a unified function?

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
```diff
@@ -54,7 +70,7 @@ def createSchema(self, plan: Plan, progress_type: ProgressType, shrink_rollback_
                        rebalance_finished TIMESTAMP WITH TIME ZONE,
                        source_bytes NUMERIC,
                        CONSTRAINT unique_fqn UNIQUE (db_name, schema_name, rel_name))
-                       DISTRIBUTED REPLICATED''')
+                       DISTRIBUTED BY (rel_name)''')
```
KnightMurloc, line 87: Why not distributed randomly?

whitehawk:
> > Distribution of rebalance internal 'table_rebalance_status_detail' is
> > changed from 'replicated' to hash-distributed.
> 
> Is it safe? Could the table become unavailable in some scenario because of this?

It should be safe - we do redistribution of this table along with others when performing shrink. The 'rebalance_shrink' test suite passes Ok with this change.

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
```diff
@@ -440,3 +470,100 @@ def getProgressType(self) -> ProgressType:
             else:
                 self.progress_type = self.ProgressType.PROGRESS_DETAILED
         return self.progress_type
+
+    def getShrinkTotalTime(self, is_rollback: bool = False) -> str:
+        target_status = 'done' if not is_rollback else 'none'
+        if not is_rollback:
+            filter_condition = "cte.state != 'STATE_SHRINK_TABLES_STARTED'"
+        else:
+            filter_condition = "cte.state != 'STATE_SHRINK_ROLLBACK_SHRINKED_TABLES_START' AND cte.state LIKE 'STATE_SHRINK_ROLLBACK_%'"
```
KnightMurloc, line 493: Why are these types of events separated?

KnightMurloc:
I found an edge case when there are no tables in the cluster: `[INFO]:-Shrink total time:	None`. Is it acceptable?

gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
```diff
@@ -180,6 +183,104 @@ def shutdown(self) -> None:
...
+            return
+
+        self.logger.info('================================================================================')
+        self.logger.info('                                   SHRINK                                   ')
+        self.logger.info('================================================================================')
+        self.logger.info(f'Tables shrunk:\t\t{summary.tables_shrunk}')
+        if self.rebalance_schema.getProgressType() == self.rebalance_schema.ProgressType.PROGRESS_DETAILED:
+            self.logger.info(f'Bytes processed:\t{summary.bytes_processed}')
+            self.logger.info(f'Shrink rate:\t\t{summary.shrink_rate}')
+        self.logger.info(f'Shrink total time:\t\t{summary.shrink_total_time}')
```
KnightMurloc: For correct alignment, one tab is needed here.
```suggestion
        self.logger.info(f'Shrink total time:\t{summary.shrink_total_time}')
```

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
```diff
@@ -425,6 +444,17 @@ def getExecutionSteps(self, status_filter: List[RebalanceStep.Status]) -> List[R
 
         return result
 
+    def backupShrinkProgress(self, logger) -> None:
```
KnightMurloc: `logger` is unused

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
```diff
@@ -440,3 +470,100 @@ def getProgressType(self) -> ProgressType:
...
         return self.progress_type
+
+    def getShrinkTotalTime(self, is_rollback: bool = False) -> str:
+        target_status = 'done' if not is_rollback else 'none'
+        if not is_rollback:
+            filter_condition = "cte.state != 'STATE_SHRINK_TABLES_STARTED'"
+        else:
+            filter_condition = "cte.state != 'STATE_SHRINK_ROLLBACK_SHRINKED_TABLES_START' AND cte.state LIKE 'STATE_SHRINK_ROLLBACK_%'"
+
+        return dbconn.querySingleton(self.conn, f"""
```
KnightMurloc, line 503: Add comments about the logic of this query.

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
whitehawk: because:
```
ERROR:  UNIQUE and DISTRIBUTED RANDOMLY are incompatible
```

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
whitehawk: If we calculate the rollback time - we need to consider only duration of rollback states (which start with `STATE_SHRINK_ROLLBACK_`).
And we exclude the state where we do actual table redistribution ('STATE_SHRINK_TABLES_STARTED' and 'STATE_SHRINK_ROLLBACK_SHRINKED_TABLES_START') - details about the reason are in the newly added comments for the query below.

gpMgmt/bin/ggrebalance_modules/rebalance_commons.py
whitehawk: `already_balanced()` expects `load` as parameter, while `is_gparray_balanced()` calculates it by itself. And there are a couple of other differences in logic. All in all, I think that it should be possible to unify the code, but it will require some rework and re-testing of the planner code, and it looks more a separate refactoring. So I would like to avoid it now, as we need to speed up development at this point of time.

whitehawk:
> I found an edge case when there are no tables in the cluster: `[INFO]:-Shrink total time: None`. Is it acceptable?

Fixed.

KnightMurloc (approved):

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
```diff
@@ -440,3 +470,108 @@ def getProgressType(self) -> ProgressType:
...
         return self.progress_type
+
+    def getShrinkTotalTime(self, is_rollback: bool = False) -> str:
+        target_status = 'done' if not is_rollback else 'none'
+        if not is_rollback:
+            filter_condition = "cte.state != 'STATE_SHRINK_TABLES_STARTED'"
+        else:
+            filter_condition = "cte.state != 'STATE_SHRINK_ROLLBACK_SHRINKED_TABLES_START' AND cte.state LIKE 'STATE_SHRINK_ROLLBACK_%'"
+
+        # In order to get total shrink (or shrink rollback time), we need to consider several items:
```
bandetto:
```suggestion
        # In order to get total shrink time (or shrink rollback time), we need to consider several items:
```

gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
```diff
@@ -180,6 +183,104 @@ def shutdown(self) -> None:
...
+
+        segments_down = []
+        gp_array = configurationInterface.getConfigurationProvider().loadSystemConfig(useUtilityMode=False, verbose=False)
+        for segment in gp_array.getSegDbList():
+            if segment.isSegmentDown():
+                segments_down.append(segment)
+
+        balanced = is_gparray_balanced(gp_array)
+
+        # we show warnings regardless the progress type set by user
```
bandetto, line 252: Shrink without rebalance skips warnings?
https://github.com/GreengageDB/greengage/blob/fb206ecec9b47682e8f23fad2e1f6364356b2be4/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py#L213-L214

gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
```diff
@@ -180,6 +183,104 @@ def shutdown(self) -> None:
...
+            status = step.getStatus()
+            if status == RebalanceStep.Status.CANCELLED:
+                cancelled_moves.append(step.getMove())
+            elif status == RebalanceStep.Status.DONE:
+                if step.isRollback():
+                    rolled_back_moves.append(step.getMove())
+                else:
+                    segments_moved += 1
+            else:
+                raise Exception(f'Unexpected status for executed step: {str(step)}')
```
bandetto, line 232: I think this is too harsh. How `executed_rebalance_steps` is built? Is there a chance any steps get into the array with a different status?

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
```diff
@@ -440,3 +470,108 @@ def getProgressType(self) -> ProgressType:
...
+                                f"SELECT stat_name, stat_value FROM {self.schema_name}.{self.rebalance_progress_view}")
+            if not result.is_rollback:
+                for stat_name, stat_value in cursor:
+                    match stat_name:
+                        case '1.1. Tables shrunk':
+                            result.tables_shrunk = stat_value
+                        case '2.1. Bytes processed':
+                            result.bytes_processed = stat_value
+                        case '3.1. Estimated shrink rate':
+                            result.shrink_rate = stat_value
```
bandetto:
```
     109:        '1.1. Tables shrunk' AS stat_name,
     127:        '1.1. Tables rolled back' AS stat_name,
     174:        '1.1. Tables shrunk' AS stat_name,
     189:        '2.1. Bytes processed' AS stat_name,
     204:        '3.1. Estimated shrink rate' AS stat_name,
     239:        '1.1. Tables rolled back' AS stat_name,
     254:        '2.1. Bytes processed' AS stat_name,
     269:        '3.1. Estimated shrink rate' AS stat_name,
     545:                        case '1.1. Tables shrunk':
     547:                        case '2.1. Bytes processed':
     549:                        case '3.1. Estimated shrink rate':
     556:                        case '1.1. Tables rolled back':
     558:                        case '2.1. Bytes processed':
     560:                        case '3.1. Estimated shrink rate':
     567:                        case '1.1. Tables shrunk':
     569:                        case '3.1. Estimated shrink rate':
```

The same 4 labels appear 9 times in the view DDL and 8 times in the match arms. Do we want to assign the strings to constants?

gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
whitehawk: Yes, for now alost all the warnings are consequences of the rebalance operation.

Without rebalance, after shrink the cluster most likely will be left in unbalanced state, but that is a natural result of the tool operation without rebalance, so it is should be Ok not to show warning in this case.

gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
whitehawk: We show summary only at the end of the normal or rollback operation. By that time, all steps should be either DONE or CANCELLED. Any other result should point out that we have some issue with the rebalance executor logic - so I think throwing an exception in this case is a proper think to do.

gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
```diff
@@ -180,6 +183,104 @@ def shutdown(self) -> None:
...
+        if not self.is_shrink_planned():
+            return
+
+        if self.rebalance_schema.getProgressType() == self.rebalance_schema.ProgressType.PROGRESS_NO:
+            self.logger.info('Skip final shrink summary report (specify "--simple-progress" or "--detailed-progress" to enable it).')
+            return
+
+        self.logger.info('================================================================================')
+        self.logger.info('                                   SHRINK                                   ')
+        self.logger.info('================================================================================')
```
bandetto: Couldn't grep this equals separator string anywhere else. I propose it to be a constant for future use too.

bandetto (approved):

KnightMurloc (approved):

PR #397, opened 2026-04-20
Support interruption of table redistribution by the SIGINT

whitehawk:
Support interruption of table redistribution by the SIGINT

Problem description:
Workers that execute table redistribution ('ALTER TABLE ... REBALANCE') were
not interrupted by the SIGINT. If during the shrink of a large table, SIGINT was
issued, ggrebalance waited till the table processing had been complete, and only
after the completion it terminated the execution.
It also affected the interruption by the '--end' and '--duration' flags.

Root cause:
Worker for the table redistribution (TableRebalanceTask) is inherited from the
SQLCommand class. Child classes of SQLCommand should set 'self.cancel_conn' to
the psycopg connection they wish to cancel and check 'self.cancel_flag' in
order to properly handle cancellation of the worker's current operation. It was
not done.

Fix:
Set 'self.cancel_conn' to the connection used for table redistribution. It will
invoke 'cancel()' method on the psycopg connection (refer to
'SQLCommand.cancel()' for details). But the 'cancel()' works only if there is
an active query being executed, otherwise is does nothing. So, we also check
the 'self.cancel_flag' flag in the beginning and before every potentially long
operation (which is 'ALTER TABLE ... REBALANCE' and 'ANALYZE'), and bail out if
it is set. For other queries, we allow them to finish without interruption,
as they should not take much time.

Note: there is a very small chance that 'SQLCommand.cancel()' is issued in
a tiny time window when we've already checked the flag, but the query hasn't
started to execute, so connection.cancel() will fire out without any effect.
To handle it, TableRebalanceTask defines its own 'cancel()' method
implementation, which calls the superclass's method() and then retries to
cancel the query till the flag 'long_operation_in_progress' (which signals that
a long query may still be running) is on.

Plus, this patch fixes logger instance usage in TableRebalanceTask.

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/shrink.py
+++ b/gpMgmt/bin/ggrebalance_modules/shrink.py
@@ -622,6 +622,18 @@ class GGShrink:
                 fun(self, attempt)
             return func_with_faults
 
+        # decorator to help with interruption of potentially long queries
+        def long_object_method(func):
+            def long_func(*args, **kwargs):
+                try:
+                    self = args[0]
+                    self.long_operation_in_progress = True
+                    self.check_cancel()
+                    return func(*args, **kwargs)
+                finally:
+                    self.long_operation_in_progress = False
+            return long_func
+
         def table_exists(self, conn: dbconn.Connection) -> bool:
             if dbconn.querySingleton(conn, f"""
                 SELECT count(1)
@@ -641,10 +653,16 @@ class GGShrink:
                 return False
             return True
 
+        @long_object_method
         def rebalance_table(self, conn: dbconn.Connection) -> None:
             dbconn.execSQL(conn,
                            f'''ALTER TABLE {escapeDoubleQuoteInSQLString(self.schema_name, True)}.{escapeDoubleQuoteInSQLString(self.rel_name, True)}
                            REBALANCE {self.target_segment_count}''')
+
+        @long_object_method
+        def analyze_table(self, conn: dbconn.Connection) -> None:
+            dbconn.execSQL(conn,
+                           f'''ANALYZE {escapeDoubleQuoteInSQLString(self.schema_name, True)}.{escapeDoubleQuoteInSQLString(self.rel_name, True)}''')
 
         def check_cancel(self):
             if self.cancel_flag:
@@ -653,7 +671,7 @@ class GGShrink:
         @wrap_table_rebalance_with_faults
         def process_table(self, attempt: int) -> None:
             self.shrink.logger.info(f'Start table rebalance for "{self.db_name}"."{self.schema_name}"."{self.rel_name}" to {self.target_segment_count} segments (attempt {attempt})')
-            # check for cancel_flag at the beginning and before each long operation
+            # check for cancel_flag at the beginning and before each long operation (refer to long_object_method decorator)
             self.check_cancel()
             self.shrink.rebalance_schema.setTableRebalanceStartTime(self.db_name, self.schema_name, self.rel_name)
             if self.db_exists(self.shrink.rebalance_schema.conn):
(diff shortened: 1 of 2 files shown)


gpMgmt/bin/ggrebalance_modules/shrink.py
```diff
@@ -645,23 +646,37 @@ def rebalance_table(self, conn: dbconn.Connection) -> None:
...
                 dburl = dbconn.DbURL(dbname=self.db_name, port=self.shrink.gpEnv.getCoordinatorPort())
                 with closing(dbconn.connect(dburl, encoding='UTF8')) as conn:
+                    self.cancel_conn = conn
+
                     dbconn.execSQL(conn, 'BEGIN')
 
                     if self.table_exists(conn):
                         if not self.table_is_rebalanced(conn):
+                            self.long_operation_in_progress = True
+                            self.check_cancel()
```
dkovalev1: is there any particular reason to do `check_cancel()` while `long_operation_in_progress` is set?

gpMgmt/bin/ggrebalance_modules/shrink.py
```diff
@@ -645,23 +646,37 @@ def rebalance_table(self, conn: dbconn.Connection) -> None:
...
             if self.db_exists(self.shrink.rebalance_schema.conn):
                 dburl = dbconn.DbURL(dbname=self.db_name, port=self.shrink.gpEnv.getCoordinatorPort())
                 with closing(dbconn.connect(dburl, encoding='UTF8')) as conn:
+                    self.cancel_conn = conn
+
                     dbconn.execSQL(conn, 'BEGIN')
 
                     if self.table_exists(conn):
                         if not self.table_is_rebalanced(conn):
+                            self.long_operation_in_progress = True
```
dkovalev1: may I suggest the following trick:
1. Put long_operation_in_progress under a decorator
2. Mark long function as a 
```
@long_operation
def rebalance_table(self):
...
```
3. In the decorator set and clean long_operation_in_progress flag in a try...finally block

This will make sure that the flag is reset in any cases while keeping the code clean.

gpMgmt/test/behave/mgmt_utils/ggrebalance_basics.feature
```diff
@@ -120,16 +120,22 @@ Feature: ggrebalance behave tests
          And a working directory of the test as '/data/gpdata/ggrebalance'
          And a cluster is created with mirrors on "cdw" and "sdw1, sdw2, sdw3"
          And all files in gpAdminLogs directory are deleted
-         And set fault inject "on_enter_STATE_REBALANCE_DONE_begin"
          And database "test_db_1" exists
          And schema "test_schema_1" exists in "test_db_1"
-         And there is a "heap" table "test_schema_1.test_table_1" in "test_db_1" with "100" rows
+         And there is a "heap" table "test_schema_1.test_table_1" in "test_db_1" with "10000000" rows
```
KnightMurloc: Remembering the experience with gpexpand, it seems to me that it is a bad idea to create a long operation by creating a large amount of data. Maybe there is some kind of suitable fault injector that will allow us to stably simulate a delay?

KnightMurloc:
Is it normal that we will print 5 messages per table when canceling?
```
ggrebalance:cdw:gpadmin-[INFO]:-Start table rebalance for "postgres"."public"."t11" to 4 segments (attempt 1)
ggrebalance:cdw:gpadmin-[WARNING]:-Cancelled table rebalance for "postgres"."public"."t11"
ggrebalance:cdw:gpadmin-[INFO]:-Start table rebalance for "postgres"."public"."t11" to 4 segments (attempt 2)
ggrebalance:cdw:gpadmin-[ERROR]:-Cancelled table rebalance for "postgres"."public"."t11"
ggrebalance:cdw:gpadmin-[ERROR]:-Failed to process the db object "postgres"."public"."t11" for 2 attempts
```

gpMgmt/bin/ggrebalance_modules/shrink.py
whitehawk: Good idea!

gpMgmt/bin/ggrebalance_modules/shrink.py
whitehawk: If we do `check_cancel()` before `long_operation_in_progress` is set, there is still some chance to issue `cancel()` in between these events,  and the query will not be interrupted.

whitehawk:
> Is it normal that we will print 5 messages per table when canceling?
> 
> ```
> ggrebalance:cdw:gpadmin-[INFO]:-Start table rebalance for "postgres"."public"."t11" to 4 segments (attempt 1)
> ggrebalance:cdw:gpadmin-[WARNING]:-Cancelled table rebalance for "postgres"."public"."t11"
> ggrebalance:cdw:gpadmin-[INFO]:-Start table rebalance for "postgres"."public"."t11" to 4 segments (attempt 2)
> ggrebalance:cdw:gpadmin-[ERROR]:-Cancelled table rebalance for "postgres"."public"."t11"
> ggrebalance:cdw:gpadmin-[ERROR]:-Failed to process the db object "postgres"."public"."t11" for 2 attempts
> ```

I think it is Ok - we see all key events here. For the first release having more logs I think would be usefull. We can reduce the amount of logs later, if there are any complains.

gpMgmt/test/behave/mgmt_utils/ggrebalance_basics.feature
whitehawk: What exactly do you think can go wrong?

KnightMurloc: The operation will be completed before we cancel it.

whitehawk: Used explicit locking from a separate session to block ALTER TABLE operation of ggrebalance

dkovalev1 (approved):

KnightMurloc (approved):

PR #398, opened 2026-04-21
Quit ggrebalance if '--show-plan' is used

whitehawk:
Quit ggrebalance if '--show-plan' is used

Problem description:
ggrebalance didn't quit if '--show-plan' option was specified. According to the
current requirements, expected behaviour is following: show plan always, and
don't perform any other actions (shrink/rebalance) if '--show-plan' is used.

This patch aligns the tool behaviour with the requirements.

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
+++ b/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
@@ -113,7 +113,7 @@ class GGRebalanceMainSM:
         },
         {
             'trigger': 'move_to_STATE_END',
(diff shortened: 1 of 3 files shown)


bimboterminator1 (approved):

KnightMurloc:
It seems to me that this change is a bit counter-intuitive. I think it's worth at least correcting the `--help` and pointing out that the `--show-plan` parameter prevent any real actions.

whitehawk:
> It seems to me that this change is a bit counter-intuitive. I think it's worth at least correcting the `--help` and pointing out that the `--show-plan` parameter prevent any real actions.

Updated help file.

KnightMurloc (approved):

PR #399, opened 2026-04-22
Check for running ggrebalance in the gpexpand

whitehawk:
Check for running ggrebalance in the gpexpand

This patch adds checking into the gpexpand that will throw an error: 
 - if there is a running instance of ggrebalance detected via pid file check;
 - or if there is a ggrebalance schema existing and
latest rebalance state isn't equal to STATE_EXECUTOR_DONE.

Schema check are encapsulated into the RebalanceSchema class and are available
with the help of the 'checkOperationInProgress()' class method.

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
+++ b/gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
@@ -2,10 +2,13 @@
 
 from psycopg2.extensions import cursor
 from gppylib.db import dbconn
+from gppylib.system.environment import GpCoordinatorEnvironment
 from gppylib.utils import escape_string
 from typing import List
 from ggrebalance_modules.planner import Plan, deserializePlan
 from ggrebalance_modules.rebalance_step import *
+
+DBNAME = 'postgres'
 
 STATE_NOT_DEFINED = 'not defined'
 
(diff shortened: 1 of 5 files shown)


gpMgmt/bin/gpexpand
```diff
@@ -242,11 +243,10 @@ def remove_pid_file(coordinator_data_directory):
         pass
 
 
-def is_gpexpand_running(coordinator_data_directory):
-    """Checks if there is another instance of gpexpand running"""
+def check_pid_file(coordinator_data_directory, pid_filename):
     is_running = False
     try:
-        fp = open(coordinator_data_directory + '/gpexpand.pid', 'r')
+        fp = open(coordinator_data_directory + f'/{pid_filename}', 'r')
```
dkovalev1:
```suggestion
        fp = open(os.path.join(coordinator_data_directory, pid_filename), 'r')
```

gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
```diff
@@ -22,19 +25,20 @@ class RebalanceSchema:
     STATE_CATEGORY_REBALANCE = 'REBALANCE'
     STATE_CATEGORY_MAIN = 'MAIN'
 
+    schema_name = 'ggrebalance'
+    rebalance_status = 'rebalance_status'
+    table_rebalance_status_detail = 'table_rebalance_status_detail'
+    rebalance_progress_view = 'rebalance_progress'
+    saved_plan = 'saved_plan'
+    segment_move_steps = 'segment_move_steps'
+
```
dkovalev1, line 63: why is this change needed?

whitehawk: schema and table names were moved from the object fields to class fields to make them accessible from the class method.

dkovalev1 (approved):

VoidZeroNull0 (approved):

PR #403, opened 2026-04-24
GG-403: Fix pg_dump for partitioned tables with droped column during pg_upgrade [7X]

GirgerS:
If there is a partitioned table in the database containing 
deleted columns, pg_dump in 7X generates sql that contains 
catalog tables no longer present in 7X, as it uses 
PostgreSQL partitioned tables.

To fix this issue, use a separate query when upgrading 
from 6X.

### DIFF
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -17539,7 +17539,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 					appendStringLiteralAH(q, tbinfo->attnames[j], fout);
 
 					/* GPDB partitioning */
-					if (fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
+					if (dopt->binary_upgrade
+						&& GPDB6_MAJOR_PGVERSION <= fout->remoteVersion
+						&& fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
+					{
+						// binary-upgrade edge-case. Target version (i.e. Greengage 7.x)
+						// uses PostgreSQL partitions, and its catalog doesn't have
+						// 'pg_partition' and 'pg_partition_rule',
+						// but because we are dumping from Greengage 6.x,
+						// the rest of the code expects that we will deal
+						// with partitions here (in 7.x each partition
+						// updates its attributes seperately).
+						//
+						// Gladly, there is a builtin function to traverse partition
+						// hierarchy.
+						//
+						// Also, because non-partitioned tables can also get here,
+						// we can't rely on 'pg_partition_tree' to always report
+						// the table. Therefore, it is hardcoded into the query.
+
+						appendPQExpBufferStr(q, "\n  AND attrelid IN (SELECT ");
+						appendStringLiteralAH(q, qualrelname, fout);
+						appendPQExpBufferStr(q, "::pg_catalog.regclass ");
+						appendPQExpBufferStr(q, "UNION SELECT relid FROM pg_partition_tree(");
+						appendStringLiteralAH(q, qualrelname, fout);
+						appendPQExpBufferStr(q, "::pg_catalog.regclass));\n");
+					}
+					else if (fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
 					{
 						/*
 						 * Do for all descendants of a partition table.
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -17567,14 +17567,11 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 						 * but because we are dumping from Greengage 6.x,
 						 * the rest of the code expects that we will deal
 						 * with partitions here (in 7.x each partition
-						 * updates its attributes seperately).
+						 * updates its attributes separately).
 						 *
-						 * Gladly, there is a builtin function to traverse partition
-						 * hierarchy.
-						 *
-						 * Also, because non-partitioned tables can also get here,
-						 * we can't rely on 'pg_partition_tree' to always report
-						 * the table. Therefore, it is hardcoded into the query.
+						 * Use pg_partition_tree() to find partition children.
+						 * For non-partitioned tables it returns no rows,
+						 * so include the root relation explicitly.
 						 */
 						appendPQExpBufferStr(q, "\n  AND attrelid IN (SELECT ");
 						appendStringLiteralAH(q, qualrelname, fout);
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -17539,33 +17539,7 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 					appendStringLiteralAH(q, tbinfo->attnames[j], fout);
 
 					/* GPDB partitioning */
-					if (dopt->binary_upgrade
-						&& GPDB6_MAJOR_PGVERSION <= fout->remoteVersion
-						&& fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
-					{
-						// binary-upgrade edge-case. Target version (i.e. Greengage 7.x)
-						// uses PostgreSQL partitions, and its catalog doesn't have
-						// 'pg_partition' and 'pg_partition_rule',
-						// but because we are dumping from Greengage 6.x,
-						// the rest of the code expects that we will deal
-						// with partitions here (in 7.x each partition
-						// updates its attributes seperately).
-						//
-						// Gladly, there is a builtin function to traverse partition
-						// hierarchy.
-						//
-						// Also, because non-partitioned tables can also get here,
-						// we can't rely on 'pg_partition_tree' to always report
-						// the table. Therefore, it is hardcoded into the query.
-
-						appendPQExpBufferStr(q, "\n  AND attrelid IN (SELECT ");
-						appendStringLiteralAH(q, qualrelname, fout);
-						appendPQExpBufferStr(q, "::pg_catalog.regclass ");
-						appendPQExpBufferStr(q, "UNION SELECT relid FROM pg_partition_tree(");
-						appendStringLiteralAH(q, qualrelname, fout);
-						appendPQExpBufferStr(q, "::pg_catalog.regclass));\n");
-					}
-					else if (fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
+					if (fout->remoteVersion < GPDB6_MAJOR_PGVERSION)
 					{
 						/*
 						 * Do for all descendants of a partition table.
@@ -17581,6 +17555,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 										  "pr.paroid = p.oid AND p.parrelid = ");
 						appendStringLiteralAH(q, qualrelname, fout);
 						appendPQExpBufferStr(q, "::pg_catalog.regclass);\n");
+					}
+					else if (fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
+					{
+						/*
+						 * Case for 6.x -> 7.x migration
+						 *
+						 * Target version (i.e. Greengage 7.x)
+						 * uses PostgreSQL partitions, and its catalog doesn't have
+						 * 'pg_partition' and 'pg_partition_rule',
+						 * but because we are dumping from Greengage 6.x,
+						 * the rest of the code expects that we will deal
+						 * with partitions here (in 7.x each partition
+						 * updates its attributes seperately).
+						 *
+						 * Gladly, there is a builtin function to traverse partition
+						 * hierarchy.
+						 *
+						 * Also, because non-partitioned tables can also get here,
+						 * we can't rely on 'pg_partition_tree' to always report
+						 * the table. Therefore, it is hardcoded into the query.
+						 */
+						appendPQExpBufferStr(q, "\n  AND attrelid IN (SELECT ");
+						appendStringLiteralAH(q, qualrelname, fout);
+						appendPQExpBufferStr(q, "::pg_catalog.regclass ");
+						appendPQExpBufferStr(q, "UNION SELECT relid FROM pg_partition_tree(");
+						appendStringLiteralAH(q, qualrelname, fout);
+						appendPQExpBufferStr(q, "::pg_catalog.regclass));\n");
 					}
 					else
 					{


GirgerS:
It doesn't seem that we run `pg_dump` tests, so we have to verify these changes manually.

Please try to run `pg_dump --binary-upgrade` on 6.x, and then run the resulting query manually on 7.x, as well as running `ggupgrade` (when the source cluster contains the table described in the ticket). Everything work as expected for me. 

Also, `ggupgrade` doesn't work right now. To test this parch, you have to apply both https://github.com/GreengageDB/ggupgrade/pull/5 and https://github.com/GreengageDB/ggupgrade/pull/6 to the main branch.

src/bin/pg_dump/pg_dump.c
```diff
@@ -17539,7 +17539,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
...
+						// Also, because non-partitioned tables can also get here,
+						// we can't rely on 'pg_partition_tree' to always report
+						// the table. Therefore, it is hardcoded into the query.
+
+						appendPQExpBufferStr(q, "\n  AND attrelid IN (SELECT ");
+						appendStringLiteralAH(q, qualrelname, fout);
+						appendPQExpBufferStr(q, "::pg_catalog.regclass ");
+						appendPQExpBufferStr(q, "UNION SELECT relid FROM pg_partition_tree(");
+						appendStringLiteralAH(q, qualrelname, fout);
+						appendPQExpBufferStr(q, "::pg_catalog.regclass));\n");
```
KnightMurloc: Why can't we just leave only the next branch?

GirgerS: Do we have `pg_partition_tree` in 6.x?

```
postgres=# select * from pg_proc where proname = 'pg_partition_tree';
(0 rows)
```

KnightMurloc: Sorry, I got it a little mixed up, I was referring to the last branch.
```c
						appendPQExpBufferStr(q, "\n  AND attrelid = ");
						appendStringLiteralAH(q, qualrelname, fout);
						appendPQExpBufferStr(q, "::pg_catalog.regclass;\n");
```

GirgerS: ~~Ah, you probably meant changing only the `UNION SELECT` part of the query. I believe this case it worth is own branch. And also, this way each branch represent it own case, so it is immediately clear to the reader what we have to deal with.~~

GirgerS: Thanks, github, for not updating the conversation.

> Sorry, I got it a little mixed up, I was referring to the last branch.

The [rest of the code](https://github.com/GreengageDB/greengage/blob/ee51f6cd0105c22631108d0472c984b2153aa034/src/bin/pg_dump/pg_dump.c#L6978-L6989) assumes this to be the case, because it updates each partition seperately. Try running `pg_dump --binary upgrade` on 7.x cluster with the same table that is present in the ticket.

src/bin/pg_dump/pg_dump.c
```diff
@@ -17539,7 +17539,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 					appendStringLiteralAH(q, tbinfo->attnames[j], fout);
 
 					/* GPDB partitioning */
-					if (fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
+					if (dopt->binary_upgrade
```
hilltracer: We already have this condition here:
https://github.com/GreengageDB/greengage/blob/96b35bed93037522c1ae7ef0fff3c2a7e55852c1/src/bin/pg_dump/pg_dump.c#L17517

src/bin/pg_dump/pg_dump.c
```diff
@@ -17539,7 +17539,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
...
+						&& GPDB6_MAJOR_PGVERSION <= fout->remoteVersion
+						&& fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
+					{
+						// binary-upgrade edge-case. Target version (i.e. Greengage 7.x)
```
hilltracer: The previous code uses
```
/*
 *
*/
```
comment style

hilltracer:
> Also, ggupgrade doesn't work right now. To test this parch, you have to apply both https://github.com/GreengageDB/ggupgrade/pull/5 and https://github.com/GreengageDB/ggupgrade/pull/6 to the main branch.

Please explain in more detail how can I test it with help ggupgrade. I've never run ggupgrade before

hilltracer:
Is it ok to use new 7.x pg_partition_tree(..) function? Are we sure that resulting SQL won't run on 6.x?

GirgerS:
> Is it ok to use new 7.x pg_partition_tree(..) function? Are we sure that resulting SQL won't run on 6.x?

I believe that `pg_dump --binary-upgrade` is used only for `pg_upgrade` migration. Also, we modify only Greengage 7.x.

GirgerS:
> > Also, ggupgrade doesn't work right now. To test this parch, you have to apply both [GreengageDB/ggupgrade#5](https://github.com/GreengageDB/ggupgrade/pull/5) and [GreengageDB/ggupgrade#6](https://github.com/GreengageDB/ggupgrade/pull/6) to the main branch.
> 
> Please explain in more detail how can I test it with help ggupgrade. I've never run ggupgrade before

Well, it is gonna take some time (because it is far from being in a working state). I propose to postpone this for tomorrow (i.e next week).

hilltracer:
> Well, it is gonna take some time (because it is far from being in a working state). I propose to postpone this for tomorrow (i.e next week).

Certainly. I'm also ok if you do it yourself in the future. And just describe the result.

src/bin/pg_dump/pg_dump.c
```diff
@@ -17539,7 +17539,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 					appendStringLiteralAH(q, tbinfo->attnames[j], fout);
 
 					/* GPDB partitioning */
-					if (fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
+					if (dopt->binary_upgrade
+						&& GPDB6_MAJOR_PGVERSION <= fout->remoteVersion
```
dkovalev1:
```suggestion
						&& fout->remoteVersion >= GPDB6_MAJOR_PGVERSION
```

dkovalev1: so we check that fout->remoteVersion == GPDB6_MAJOR_PGVERSION here?

GirgerS: > so we check that fout->remoteVersion == GPDB6_MAJOR_PGVERSION here?

Logically, yes. But in the practice, `GPDB6_MAJOR_PGVERSION` is defined as the first PostgreSQL version that Greengage/Greenplum 6 was based on. The same thing for  `GPDB7_MAJOR_PGVERSION'. So there is a range of values between them.

https://github.com/GreengageDB/greengage/blob/96b35bed93037522c1ae7ef0fff3c2a7e55852c1/src/bin/pg_dump/pg_backup.h#L29-L32

src/bin/pg_dump/pg_dump.c
GirgerS: Thank you, pretty stupid by me to overlook this. I updated the condition.

src/bin/pg_dump/pg_dump.c
GirgerS: Ty.

src/bin/pg_dump/pg_dump.c
```diff
@@ -17556,6 +17556,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
...
+						/*
+						 * Case for 6.x -> 7.x migration
+						 *
+						 * Target version (i.e. Greengage 7.x)
+						 * uses PostgreSQL partitions, and its catalog doesn't have
+						 * 'pg_partition' and 'pg_partition_rule',
+						 * but because we are dumping from Greengage 6.x,
+						 * the rest of the code expects that we will deal
+						 * with partitions here (in 7.x each partition
+						 * updates its attributes seperately).
```
hilltracer:
```suggestion
						 * updates its attributes separately).
```

src/bin/pg_dump/pg_dump.c
```diff
@@ -17556,6 +17556,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
...
+						 *
+						 * Gladly, there is a builtin function to traverse partition
+						 * hierarchy.
+						 *
+						 * Also, because non-partitioned tables can also get here,
+						 * we can't rely on 'pg_partition_tree' to always report
+						 * the table. Therefore, it is hardcoded into the query.
```
hilltracer: I think better to use somenthing more businesslike here:
Use pg_partition_tree() on the target to find partition children.
For non-partitioned tables it returns no rows, so include the root
relation explicitly.

src/bin/pg_dump/pg_dump.c
```diff
@@ -17556,6 +17556,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
...
+						 *
+						 * Target version (i.e. Greengage 7.x)
+						 * uses PostgreSQL partitions, and its catalog doesn't have
+						 * 'pg_partition' and 'pg_partition_rule',
+						 * but because we are dumping from Greengage 6.x,
+						 * the rest of the code expects that we will deal
+						 * with partitions here (in 7.x each partition
+						 * updates its attributes seperately).
+						 *
+						 * Gladly, there is a builtin function to traverse partition
```
dkovalev1: Do you mean `pg_partition_tree` mentioned below here?

dkovalev1:
How can we check it?

src/bin/pg_dump/pg_dump.c
```diff
@@ -17556,6 +17556,33 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 						appendStringLiteralAH(q, qualrelname, fout);
 						appendPQExpBufferStr(q, "::pg_catalog.regclass);\n");
 					}
+					else if (fout->remoteVersion < GPDB7_MAJOR_PGVERSION)
+					{
+						/*
+						 * Case for 6.x -> 7.x migration
+						 *
+						 * Target version (i.e. Greengage 7.x)
+						 * uses PostgreSQL partitions, and its catalog doesn't have
```
dkovalev1, line 17565: What is "Target" in this context? How can we know while dumping schema about target version and its abilities?

GirgerS: `pg_dump --binary-upgrade` is [marked](https://www.postgresql.org/docs/current/app-pgdump.html) as 'for internal usage', and judging from grep, it is used only from [this](https://github.com/GreengageDB/greengage/blob/e30fe1d13cbbd92ccba3c5a397754ab029949562/src/bin/pg_upgrade/dump.c#L19-L77) function, when we upgrading a cluster with `pg_upgrade`.

There is also the same flag for [pg_resetval](https://github.com/GreengageDB/greengage/blob/e30fe1d13cbbd92ccba3c5a397754ab029949562/src/bin/pg_resetwal/pg_resetwal.c#L386-L389) and [pg_restore](https://github.com/GreengageDB/greengage/blob/e30fe1d13cbbd92ccba3c5a397754ab029949562/src/bin/pg_dump/pg_restore.c#L130-L132), but they are not related to pg_dump.

And we [run](https://github.com/GreengageDB/ggupgrade/blob/878e671d039012ac07d93090dfa5d78cb7f1bc93/hub/upgrade_coordinator.go#L61-L62) `pg_upgrade` directly to 7.x. Meaning that we should never see anything below 7.x as the target version. There is also [this](https://github.com/GreengageDB/greengage/blob/e30fe1d13cbbd92ccba3c5a397754ab029949562/src/bin/pg_upgrade/check.c#L332-L336) check.

(there is a subtlety that there are several postgres versions between Greengage/Greenplum 6.x and 7.x, and I don't know what would happen if we try to upgrade to any of those (i.e. PostgreSQL 11), but I doubt that this can happen).

GirgerS:
> How can we check it?

Manually? The easiest way is to `pg_dump --binary-upgrade` a 6.x cluster with the table specified in the ticket, and then manually run generated sql in a separate 7.x cluster. Running `pg_upgrade` manually is also an option, but I've not tried it.

And if you want to run ggupgrate, let me know, I'll give more detailed explanation. 

----- 

Also, there are some automatic pg_dump tests, but I doubt that we run them.

src/bin/pg_dump/pg_dump.c
GirgerS: Nice catch.

src/bin/pg_dump/pg_dump.c
GirgerS: Agreed, I like how concise your description is.

src/bin/pg_dump/pg_dump.c
GirgerS: Yes

hilltracer:
> The easiest way is to `pg_dump --binary-upgrade` a 6.x cluster with the table specified in the ticket, and then manually run generated sql in a separate 7.x cluster

I've tested it in that way. The command works fine
<details>
<summary>spoiler</summary>

```sql
CREATE TABLE public.aoco_dropped_col_and_encoding_partitioned (
    a integer,
    "........pg.dropped.2........" INTEGER /* dummy */ ENCODING (compresstype=zlib,compresslevel=1,blocksize=32768),
    c integer,
    COLUMN a ENCODING (compresstype=none,blocksize=32768,compresslevel=0),
    COLUMN c ENCODING (compresstype=none,blocksize=32768,compresslevel=0)
)
WITH (appendonly='true', orientation='column') DISTRIBUTED BY (a) PARTITION BY RANGE(c) 
          (
          PARTITION p1 START (1) END (4) WITH (tablename='aoco_dropped_col_and_encoding_partitioned_1_prt_p1', appendonly='true', orientation='column' ) 
                    COLUMN a ENCODING (compresstype=none, blocksize=32768, compresslevel=0) 
                    COLUMN c ENCODING (compresstype=none, blocksize=32768, compresslevel=0)
          );
CREATE TABLE
 ;
SET allow_system_table_mods = true;
SET
-- For binary upgrade, recreate dropped column.
UPDATE pg_catalog.pg_attribute
SET attlen = 4, attalign = 'i', attbyval = false
WHERE attname = '........pg.dropped.2........'
  AND attrelid IN (SELECT 'public.aoco_dropped_col_and_encoding_partitioned'::pg_catalog.regclass UNION SELECT relid FROM pg_partition_tree('public.aoco_dropped_col_and_encoding_partitioned'::pg_catalog.regclass));
UPDATE 2
ALTER TABLE public.aoco_dropped_col_and_encoding_partitioned DROP COLUMN "........pg.dropped.2........";
ALTER TABLE
RESET allow_system_table_mods;
RESET
SET allow_system_table_mods = true;
SET
```

</details>

And the result for pg_catalog.pg_attribute is identical with situation, when I droped column from table manually as specified in the ticket.
So LGTM

hilltracer (approved):

src/bin/pg_dump/pg_dump.c
dkovalev1: may be you rephrase it to make it unambiguous then?

GirgerS: We did it already: https://github.com/GreengageDB/greengage/pull/403/changes/da21360a87ebc9c0a4a79223e42dbca15126fd20

dkovalev1 (approved):

PR #416, opened 2026-04-28
GG-409: ALTER RESOURCE GROUP inside transaction

bandetto:
ALTER RESOURCE GROUP inside transaction

ALTER RESOURCE GROUP did not support execution inside transaction blocks 
because resource group changes are hard to roll back safely: they affect both
catalog state and in-memory/shared-memory state, and uncommitted changes must
block other resource group modifications.

Introduce an opt-in postmaster GUC that lifts the restriction. With it on,
several ALTERs run in one transaction and roll back together if any of them
fails or if the user chooses to abort. The catalog write happens during the
ALTER itself, while the in-memory cgroup and shared memory apply still run at
top-level commit.

Check callbacks at PRE_COMMIT against the final pg_resgroupcapability state, so
ALTERs rolled back by subtransactions or overwritten by later ALTERs are not
applied.

CREATE and DROP keep the transaction block restriction. Reject moving a query
to another resource group while resource group capabilities are being edited by
an uncommitted transaction.

Co-authored-by: bandetto <v.sarmin@arenadata.io>


gpcontrib/gp_internal_tools/gp_resource_group.c
```diff
@@ -84,6 +86,21 @@ pg_resgroup_move_query(PG_FUNCTION_ARGS)
...
+		/*
+		 * Refuse the move if any backend is currently editing a resource group.
+		 * ALTER RESOURCE GROUP holds ExclusiveLock on pg_resgroupcapability 
+		 * until end of its transaction, so a conditional acquisition of
+		 * RowExclusiveLock fails fast in that case.
+		 */
+		if (!ConditionalLockRelationOid(ResGroupCapabilityRelationId,
+										RowExclusiveLock))
+		{
+			ereport(ERROR,
```
RekGRpth, line 102: Why can't we just wait for that process to release the lock here?

src/backend/commands/resgroupcmds.c
```diff
@@ -531,6 +561,27 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+
+		/*
+		 * Track the context for the PRE_COMMIT validation hook. The list
+		 * lives across the transaction, so allocate the cell in
+		 * TopMemoryContext. It is freed at the terminal event.
+		 */
+		if (gp_resource_group_enable_alter_in_transaction)
+		{
+			MemoryContext oldcxt = MemoryContextSwitchTo(TopMemoryContext);
+			pending_alter_callbacks = lappend(pending_alter_callbacks,
```
RekGRpth: Why do you need an extra list in `TopMemoryContext`? `RegisterXactCallback` already adds it to its list in `TopMemoryContext`; just specify `callbackCtx` as the `arg` argument.

src/backend/commands/resgroupcmds.c
```diff
@@ -1144,12 +1195,76 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+						   MaxCpuSetLength) == 0;
+		case RESGROUP_LIMIT_TYPE_MEMORY:
+			return recorded->memLimit == current->memLimit;
+		case RESGROUP_LIMIT_TYPE_MEMORY_SHARED_QUOTA:
+			return recorded->memSharedQuota == current->memSharedQuota;
+		case RESGROUP_LIMIT_TYPE_MEMORY_SPILL_RATIO:
+			return recorded->memSpillRatio == current->memSpillRatio;
+		case RESGROUP_LIMIT_TYPE_MEMORY_AUDITOR:
+			return recorded->memAuditor == current->memAuditor;
+		default:
```
RekGRpth:
```suggestion
		case RESGROUP_LIMIT_TYPE_UNKNOWN:
```

src/backend/commands/resgroupcmds.c
```diff
@@ -531,6 +561,27 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+			MemoryContext oldcxt = MemoryContextSwitchTo(TopMemoryContext);
+			pending_alter_callbacks = lappend(pending_alter_callbacks,
+											  callbackCtx);
+			MemoryContextSwitchTo(oldcxt);
+
+			if (!alter_pre_commit_callback_registered)
+			{
+				RegisterXactCallback(alterResgroupPreCommitCallback, NULL);
+				alter_pre_commit_callback_registered = true;
+			}
```
RekGRpth: Why not use
```suggestion
			RegisterXactCallbackOnce(alterResgroupPreCommitCallback, callbackCtx); 
```
here and simplify `alterResgroupPreCommitCallback`?

src/backend/commands/resgroupcmds.c
```diff
@@ -1144,12 +1195,76 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+	}
+}
+
+static void
+alterResgroupPreCommitCallback(XactEvent event, void *arg)
+{
+	ListCell   *lc;
+
+	if (event == XACT_EVENT_PRE_COMMIT &&
+		gp_resource_group_enable_alter_in_transaction &&
```
RekGRpth: Is it possible to call this function when `gp_resource_group_enable_alter_in_transaction` is disabled?

src/backend/utils/misc/guc_gp.c
```diff
@@ -3169,6 +3170,15 @@ struct config_bool ConfigureNamesBool_gp[] =
 		false, NULL, NULL
 	},
 
+	{
+		{"gp_resource_group_enable_alter_in_transaction", PGC_POSTMASTER, RESOURCES,
```
RekGRpth, line 3174: Why was this specific `PGC_POSTMASTER` context chosen? Apparently, enabling this GUC doesn't allocate any shared resources.

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
```diff
@@ -0,0 +1,221 @@
+-- ------------------------------------------------------
+-- Test: ALTER RESOURCE GROUP inside a transaction block.
+-- ------------------------------------------------------
+
+-- start_ignore
+! gpconfig -c gp_resource_group_enable_alter_in_transaction -v on;
+
+! gpstop -rai;
+-- end_ignore
```
RekGRpth: remove ignore blocks from output here and below

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
```diff
@@ -0,0 +1,110 @@
+-- ------------------------------------------------------
+-- Test: ALTER RESOURCE GROUP inside a transaction block.
+-- ------------------------------------------------------
+
+-- start_ignore
+! gpconfig -c gp_resource_group_enable_alter_in_transaction -v on;
+! gpstop -rai;
```
hilltracer: Why did you used hard immediately `-rai` here? Why not -arf for example?

src/backend/commands/resgroupcmds.c
```diff
@@ -1144,12 +1195,76 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+			return recorded->memSpillRatio == current->memSpillRatio;
+		case RESGROUP_LIMIT_TYPE_MEMORY_AUDITOR:
+			return recorded->memAuditor == current->memAuditor;
+		default:
+			return true;
+	}
+}
+
+static void
+alterResgroupPreCommitCallback(XactEvent event, void *arg)
```
hilltracer: How about some text explanation of purpose `alterResgroupPreCommitCallback` in comment before the function?

src/backend/commands/resgroupcmds.c
```diff
@@ -416,6 +430,22 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+	/*
+	 * The session cannot alter its own resource group inside a transaction
+	 * block, the change would not be visible to this session before COMMIT.
+	 */
+	if (Gp_role == GP_ROLE_DISPATCH &&
+		gp_resource_group_enable_alter_in_transaction &&
+		IsTransactionBlock() &&
+		groupid == GetMyResGroupId())
+	{
+		ereport(ERROR,
```
hilltracer: see new requirements. This should be warning: https://gitlab.adsw.io/arenadata/development/adb-requirements/-/merge_requests/13

src/backend/commands/resgroupcmds.c
```diff
@@ -1144,12 +1195,76 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+		Relation rel = heap_open(ResGroupCapabilityRelationId, AccessShareLock);
+
+		foreach (lc, pending_alter_callbacks)
+		{
+			ResourceGroupCallbackContext *ctx = lfirst(lc);
+			ResGroupCaps currentCaps;
+
+			GetResGroupCapabilities(rel, ctx->groupid, &currentCaps);
+
+			if (!resgroupCapFieldMatches(ctx->limittype, &ctx->caps,
```
hilltracer: hm.. we detect rollback using matching `old value == new value`. How about this edge case?

```sql
BEGIN;

SAVEPOINT s;
ALTER RESOURCE GROUP rg SET CONCURRENCY 10;
ROLLBACK TO SAVEPOINT s;

ALTER RESOURCE GROUP rg SET CONCURRENCY 10;

COMMIT;
```

gpcontrib/gp_internal_tools/gp_resource_group.c
```diff
@@ -84,6 +86,21 @@ pg_resgroup_move_query(PG_FUNCTION_ARGS)
 		pid_t pid = PG_GETARG_INT32(0);
 		groupName = text_to_cstring(PG_GETARG_TEXT_PP(1));
 
+		/*
+		 * Refuse the move if any backend is currently editing a resource group.
+		 * ALTER RESOURCE GROUP holds ExclusiveLock on pg_resgroupcapability 
+		 * until end of its transaction, so a conditional acquisition of
+		 * RowExclusiveLock fails fast in that case.
+		 */
+		if (!ConditionalLockRelationOid(ResGroupCapabilityRelationId,
```
hilltracer, line 99: We take the lock here but don't release it after moving. So in the long living transaction it will block ALTER RESOURCE GROUP until the end of current transaction. Is it ok? Can we release lock after the move?

gpcontrib/gp_internal_tools/gp_resource_group.c
hilltracer: According SRS we don't want to wait for pg_move command

> This possible problems can resolve with waiting, but if its are resolved with waiting, administrator will wait. If performing `pg_resgroup_move_query` returns error, administrator will receive a response immediately and may be able to resolve the long-term editing of resource groups. Administrator will have a choice:: wait (for example, repeatable call with `BEGIN EXCEPTION`) or immediately exit.

src/backend/commands/resgroupcmds.c
hilltracer: We register the xact callback only once per backend lifetime, while the pending context is transaction-scoped and may be deleted on COMMIT/ABORT. Passing that pending context as arg would leave a dangling pointer in the callback registry.

src/backend/commands/resgroupcmds.c
hilltracer: Use a regular callback because once callbacks are not fired at PRE_COMMIT, where we need to validate final catalog state.
https://github.com/GreengageDB/greengage/blob/c4e1085d26bae201ef29191373ecf7b6b92c534f/src/backend/access/transam/xact.c#L2657

src/backend/commands/resgroupcmds.c
hilltracer: Added

src/backend/utils/misc/guc_gp.c
hilltracer: According new SRS:

> Setting gp_resource_group_alter_tran is postmaster, because it changes the way resource groups are modified.
If users decide to use this approach, they will have functions and scripts that use ALTER RESOURCE GROUP in transactions, which will not work with disable gp_resource_group_alter_tran.
So they will not need to change this setting often.

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: reworked

src/backend/commands/resgroupcmds.c
hilltracer: I think the old check is more suitable, since it takes into account other values, besides, all the code around is written in a similar way

src/backend/commands/resgroupcmds.c
hilltracer: reworked

src/backend/commands/resgroupcmds.c
hilltracer: reworked

src/backend/commands/resgroupcmds.c
hilltracer: reworked

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
hilltracer: removed

src/backend/commands/resgroupcmds.c
```diff
@@ -39,6 +39,8 @@
 #include "utils/resowner.h"
 #include "utils/syscache.h"
 #include "utils/faultinjector.h"
+#include "utils/memutils.h"
+#include "nodes/pg_list.h"
```
RekGRpth: sort headers

gpcontrib/gp_internal_tools/gp_resource_group.c
```diff
@@ -84,6 +86,19 @@ pg_resgroup_move_query(PG_FUNCTION_ARGS)
 		pid_t pid = PG_GETARG_INT32(0);
 		groupName = text_to_cstring(PG_GETARG_TEXT_PP(1));
 
+		/*
+		* Fail if ALTER RESOURCE GROUP is in progress.
+		* ALTER holds ExclusiveLock until end of transaction.
+		* RowShareLock allows concurrent moves but conflicts with it.
+		*/
```
RekGRpth:
```suggestion
		 * Fail if ALTER RESOURCE GROUP is in progress.
		 * ALTER holds ExclusiveLock until end of transaction.
		 * RowShareLock allows concurrent moves but conflicts with it.
		 */
```

src/backend/commands/resgroupcmds.c
```diff
@@ -360,11 +371,23 @@ DropResourceGroup(DropResourceGroupStmt *stmt)
...
 
+void
+AlterResourceGroup(AlterResourceGroupStmt *stmt)
+{
+	/*
+	* Legacy ABI entry point has no ProcessUtilityContext.
+	* Treat it as top-level so IsInTransactionChain() can still detect
+	* explicit transaction blocks, subtransactions and non-standalone
+	* transaction states, but does not assume function/multi-command context.
+	*/
```
RekGRpth:
```suggestion
	 * Legacy ABI entry point has no ProcessUtilityContext.
	 * Treat it as top-level so IsInTransactionChain() can still detect
	 * explicit transaction blocks, subtransactions and non-standalone
	 * transaction states, but does not assume function/multi-command context.
	 */
```

src/backend/commands/resgroupcmds.c
```diff
@@ -416,6 +439,20 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
 				 errmsg("resource group \"%s\" does not exist",
 						stmt->name)));
 
+	/*
+	* The session alters its own resource group inside a transaction block.
+	* The change will not be visible to this session before COMMIT.
+	*/
```
RekGRpth:
```suggestion
	 * The session alters its own resource group inside a transaction block.
	 * The change will not be visible to this session before COMMIT.
	 */
```

src/backend/commands/resgroupcmds.c
```diff
@@ -531,6 +568,31 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
 		callbackCtx->limittype = limitType;
 		callbackCtx->caps = caps;
 		callbackCtx->oldCaps = oldCaps;
+		callbackCtx->should_apply = true;
+
+		/*
+		* Track callbacks only for ALTER inside a transaction chain. Top-level
+		* ALTER keeps the old path. The list cell must live until COMMIT/ABORT.
+		*/
```
RekGRpth:
```suggestion
		 * Track callbacks only for ALTER inside a transaction chain. Top-level
		 * ALTER keeps the old path. The list cell must live until COMMIT/ABORT.
		 */
```

src/backend/commands/resgroupcmds.c
```diff
@@ -531,6 +568,31 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+			pending_alter_callbacks = lappend(pending_alter_callbacks,
+											  callbackCtx);
+			MemoryContextSwitchTo(oldcxt);
+
+			if (!alter_pre_commit_callback_registered)
+			{
+				/*
+				* Use a regular callback because once callbacks are not fired at
+				* PRE_COMMIT, where we need to validate final catalog state.
+				*/
```
RekGRpth:
```suggestion
				 * Use a regular callback because once callbacks are not fired at
				 * PRE_COMMIT, where we need to validate final catalog state.
				 */
```

src/backend/commands/resgroupcmds.c
RekGRpth: ok, why not
```suggestion
			RegisterXactCallback(alterResgroupPreCommitCallback, callbackCtx); 
```

src/backend/commands/resgroupcmds.c
```diff
@@ -1144,12 +1206,65 @@ alterResgroupCallback(XactEvent event, void *arg)
 {
 	ResourceGroupCallbackContext *callbackCtx = arg;
 
-	if (event == XACT_EVENT_COMMIT)
+	if (event == XACT_EVENT_COMMIT && callbackCtx->should_apply)
+	{
 		ResGroupAlterOnCommit(callbackCtx);
+	}
```
RekGRpth, line 1252:
```suggestion
		ResGroupAlterOnCommit(callbackCtx);
```

src/backend/commands/resgroupcmds.c
```diff
@@ -1144,12 +1206,65 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+	Assert(gp_resource_group_enable_alter_in_transaction);
+
+	if (event == XACT_EVENT_COMMIT || event == XACT_EVENT_ABORT)
+	{
+		list_free(pending_alter_callbacks);
+		pending_alter_callbacks = NIL;
+		return;
+	}
+	else if (event == XACT_EVENT_PRE_COMMIT)
+	{
```
RekGRpth, line 1295:
```suggestion
	}
	
	if (event != XACT_EVENT_PRE_COMMIT)
		return;
		

```

src/backend/commands/resgroupcmds.c
RekGRpth: But this might lead us to forget to add specific handling for cases where the enumeration grows.

src/backend/utils/resgroup/resgroup.c
```diff
@@ -888,14 +888,84 @@ ResGroupCreateOnAbort(const ResourceGroupCallbackContext *callbackCtx)
...
+	{
+		case RESGROUP_LIMIT_TYPE_CONCURRENCY:
+			dst->concurrency = src->concurrency;
+			break;
+		case RESGROUP_LIMIT_TYPE_CPU:
+		case RESGROUP_LIMIT_TYPE_CPUSET:
+			dst->cpuRateLimit = src->cpuRateLimit;
+			StrNCpy(dst->cpuset,
+					src->cpuset,
+					sizeof(dst->cpuset));
```
RekGRpth:
```suggestion
					MaxCpuSetLength);
```

src/backend/utils/resgroup/resgroup.c
```diff
@@ -904,54 +974,90 @@ ResGroupAlterOnCommit(const ResourceGroupCallbackContext *callbackCtx)
...
 		{
-			if (gp_resource_group_enable_cgroup_cpuset)
+			/*
+			 * More than one callback may survive for the same final value, for
+			 * example after savepoint rollback and a later matching ALTER. If
+			 * this value is already in shared memory, skip this callback.
+			 */
+			if (ResGroupCapFieldMatches(callbackCtx->limittype,
+										 &callbackCtx->caps,
+										 &group->caps))
```
RekGRpth:
```suggestion
										&callbackCtx->caps,
										&group->caps))
```

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
```diff
@@ -0,0 +1,256 @@
+-- ------------------------------------------------------
+-- Test: ALTER RESOURCE GROUP inside a transaction block.
+-- ------------------------------------------------------
+
+-- start_ignore
+! gpconfig -c gp_resource_group_enable_alter_in_transaction -v on;
+! gpstop -arf;
+
```
RekGRpth, line 8: in isolation2 you may use following
```suggestion
!\retcode gpconfig -c gp_resource_group_enable_alter_in_transaction -v on;
!\retcode gpstop -rai;
```

RekGRpth: still actual

src/backend/commands/resgroupcmds.c
hilltracer: Because it's a regular callback and it should be registered only once (after that in will leave inside backend instance), so the flag alter_pre_commit_callback_registered prevents several running of RegisterXactCallback().
RegisterXactCallbackOnce should be registered each time since it self destroyed after first running.

RekGRpth: What's the problem with registering the required number of callbacks? Why do they have an argument then?

src/backend/utils/resgroup/resgroup.c
hilltracer: I would prefer to safe old code as it was done in other places, e.g.
https://github.com/GreengageDB/greengage/blob/8588ef72701c41553587541567a242dc262b6cc7/src/backend/commands/resgroupcmds.c#L507

RekGRpth: so fix [compare](https://github.com/GreengageDB/greengage/pull/416/changes#diff-14f7111e45936f7f6b0fbccf732d91ab7957e461b8a3e641381e1b003f113a18R907-R908) above

src/backend/commands/resgroupcmds.c
hilltracer: reworked

src/backend/commands/resgroupcmds.c
hilltracer: done

gpcontrib/gp_internal_tools/gp_resource_group.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/utils/resgroup/resgroup.c
hilltracer: done

src/backend/utils/resgroup/resgroup.c
hilltracer: done

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: done

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
RekGRpth: still actual

src/backend/commands/resgroupcmds.c
```diff
@@ -1144,12 +1206,64 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+
+	if (event == XACT_EVENT_COMMIT || event == XACT_EVENT_ABORT)
+	{
+		list_free(pending_alter_callbacks);
+		pending_alter_callbacks = NIL;
+		return;
+	}
+
+	if (event == XACT_EVENT_PRE_COMMIT)
+	{
```
RekGRpth:
```suggestion
	if (event != XACT_EVENT_PRE_COMMIT)
		return;
```

hilltracer: done

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: Because many callbacks will remain in the backend and will be called each time a commit is performed.

hilltracer:
I've also manually checked these 2 scenatio from SRS:
2. Use kill -9 terminates all processes and reinitializes shared memory, as well as reconfigures all resource groups
3. Using pg_terminate_backend during COMMIT correctly terminates the request by calling ResGroupAlterOnCommit for all modified resource groups

<details>
<summary>spoiler</summary>

- first session:
```sql
SELECT * FROM rg_alter_tran_status;
CREATE VIEW
    groupname    | concurrency | cpu_rate_limit | memory_limit 
-----------------+-------------+----------------+--------------
 rg_alter_tran   | 2           | 10             | 10
 rg_alter_tran_b | 2           | 10             | 10
(2 rows)

gpadmin=# SELECT * FROM rg_alter_tran_runtime_status;
    groupname    | concurrency | cpu_rate_limit | memory_limit 
-----------------+-------------+----------------+--------------
 rg_alter_tran   |           2 |             10 |           10
 rg_alter_tran_b |           2 |             10 |           10
(2 rows)
BEGIN;
ALTER RESOURCE GROUP rg_alter_tran   SET CONCURRENCY 6;
ALTER RESOURCE GROUP rg_alter_tran   SET MEMORY_LIMIT 7;
COMMIT; -- stop after (or during) first callback using debug break point
```

- second session:
```sql
killall -9 postgres
```
or
```sql
select pg_terminate_backend(<pid>);
```

- first session:
```sql
-- restart backend again
gpadmin=# SELECT * FROM rg_alter_tran_status;
SELECT * FROM rg_alter_tran_runtime_status;
    groupname    | concurrency | cpu_rate_limit | memory_limit 
-----------------+-------------+----------------+--------------
 rg_alter_tran   | 6           | 10             | 7
 rg_alter_tran_b | 2           | 10             | 10
(2 rows)

    groupname    | concurrency | cpu_rate_limit | memory_limit 
-----------------+-------------+----------------+--------------
 rg_alter_tran   |           6 |             10 |            7
 rg_alter_tran_b |           2 |             10 |           10
(2 rows)
```
so all new values applying fine after cluster failures

</details>

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
```diff
@@ -0,0 +1,256 @@
+-- ------------------------------------------------------
+-- Test: ALTER RESOURCE GROUP inside a transaction block.
+-- ------------------------------------------------------
+
+!\retcode gpconfig -c gp_resource_group_enable_alter_in_transaction -v on;
+
+!\retcode gpstop -arf;
```
whitehawk: Why is '-f' flag used? It is marked as `<deprecated>` in `gpstop`.

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
```diff
@@ -0,0 +1,256 @@
...
+ALTER RESOURCE GROUP rg_alter_tran_b SET CONCURRENCY 6;
+COMMIT;
+SELECT * FROM rg_alter_tran_status;
+
+-- 2 Not applying settings if there is a ROLLBACK.
+-- Multiple resource groups are changed to verify atomic rollback.
+BEGIN;
+ALTER RESOURCE GROUP rg_alter_tran   SET CONCURRENCY 7;
+ALTER RESOURCE GROUP rg_alter_tran_b SET CONCURRENCY 8;
+ROLLBACK;
```
whitehawk, line 126: Please add test cases with full crash recovery and WAL replay, for ex:
 - segfault happens before the commit, therefore the changes should not be applied after crash recovery;
 - segfault happens right after the commit, and the changes should be replayed Ok;
 - also we need to check such scenarios in both coordinator and segments.

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
```diff
@@ -0,0 +1,458 @@
...
+-- only one apply should matter.
+BEGIN;
+BEGIN
+SAVEPOINT s1;
+SAVEPOINT
+ALTER RESOURCE GROUP rg_alter_tran SET CONCURRENCY 11;
+ALTER
+ROLLBACK TO SAVEPOINT s1;
+ROLLBACK
+ALTER RESOURCE GROUP rg_alter_tran SET CONCURRENCY 11;
```
whitehawk, line 286: We apply the same value before the rollback to savepoint, and after. How do we know which of them was applied?

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
```diff
@@ -0,0 +1,458 @@
...
+ groupname       | concurrency | cpu_rate_limit | memory_limit 
+-----------------+-------------+----------------+--------------
+ rg_alter_tran   | 15          | 10             | 9            
+ rg_alter_tran_b | 21          | 10             | 8            
+(2 rows)
+
+
+-- Concurrent ALTERs
+
+-- 18 Error during pg_resgroup_move_query when ALTER is uncommitted.
```
whitehawk: Can we add a test case when ALTER is applied to a group which is not the same as used in `pg_resgroup_move_query`? I suppose there should be no error in this case, correct?

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
```diff
@@ -0,0 +1,458 @@
...
+return None $$ LANGUAGE plpython3u;
+CREATE
+
+CREATE OR REPLACE VIEW rg_alter_tran_runtime_status AS SELECT groupname, rg_alter_tran_runtime_cap(groupname, 1) AS concurrency, rg_alter_tran_runtime_cap(groupname, 2) AS cpu_rate_limit, rg_alter_tran_runtime_cap(groupname, 3) AS memory_limit FROM (VALUES ('rg_alter_tran'), ('rg_alter_tran_b')) AS v(groupname) ORDER BY groupname;
+CREATE
+
+SELECT * FROM rg_alter_tran_status;
+ groupname       | concurrency | cpu_rate_limit | memory_limit 
+-----------------+-------------+----------------+--------------
+ rg_alter_tran   | 2           | 10             | 10
```
whitehawk, line 40: Are these values from coordinator only? Why are not we checking the value on segments?

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
```diff
@@ -0,0 +1,256 @@
...
+--   3 - memory_limit
+--   4 - memory_shared_quota
+--   5 - memory_spill_ratio
+--   6 - memory_auditor
+CREATE OR REPLACE FUNCTION rg_alter_tran_runtime_cap(group_name text, cap_id int)
+RETURNS int
+AS $$
+import json
+
+res = plpy.execute("""
```
whitehawk: Do we really need to use a python function? Maybe there are ways to get status easier?

hilltracer:
I've decided to remove skipping duplicate no-op operations. Since don't want to change old behavior, so ALTER's with value that has not changed any will be applied in any case. Commit: 09d721d66

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: Alternative is to use of `gpstop -ar` but in this case we can wait any open sessions. Many tests already run with -arf or -rai (but I think immediate mode is too hard for this case). I think it's better to use it also to prevent influence of previous tests.

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
hilltracer: done

> I suppose there should be no error in this case, correct?

no, any ALTER for any resource group will block pg_move, CREATE resourse group and DROP.

it's expected by SRS:

> Add a check of editing resource groups to the pg_resgroup_move_query function without an exclusive lock (!ConditionalLockRelationOid(ResGroupCapabilityRelationId, ExclusiveLock)) without ExclusiveLock). Otherwise, there may be the following problems, when transferring to another resource group that is currently being edited

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
hilltracer: thanks, added total checks

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: reworked

.abi-check/6.30.1/postgres.types.ignore
whitehawk, line 1: Do we need this file?

src/backend/commands/resgroupcmds.c
```diff
@@ -531,7 +567,35 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+			 * In transaction, collect ALTER data for later synchronous apply.
+			 */
+			Assert(gp_resource_group_enable_alter_in_transaction);
+			MemoryContext oldcxt = MemoryContextSwitchTo(TopMemoryContext);
+			resgroup_alter_tran_callbacks = lappend(
+												resgroup_alter_tran_callbacks,
+												callbackCtx);
+			MemoryContextSwitchTo(oldcxt);
+
+			if (!alter_tran_callback_registered)
```
whitehawk: `alter_tran_callback_registered` looks like is used only inside this function. Can we make it `static` local var? Why should it be global?

src/backend/commands/resgroupcmds.c
```diff
@@ -531,7 +567,35 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+			{
+				/*
+				 * Use a regular callback because once callbacks are not fired
+				 * at PRE_COMMIT, where we need to validate final catalog state.
+				 */
+				RegisterXactCallback(alterResgroupTranCallback, NULL);
```
whitehawk: Can't we use callback argument instead of global `resgroup_alter_tran_callbacks`?

src/backend/commands/resgroupcmds.c
```diff
@@ -1150,6 +1214,154 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+	{
+		foreach (lc, resgroup_alter_tran_callbacks)
+			pfree(lfirst(lc));
+
+		list_free(resgroup_alter_tran_callbacks);
+		resgroup_alter_tran_callbacks = NIL;
+		return;
+	}
+
+	if (event == XACT_EVENT_COMMIT)
```
whitehawk: In the [CommitTransaction](https://github.com/GreengageDB/greengage/blob/dd8cc071f96dc3e00f72956adf6f8a1dfbc4f42d/src/backend/access/transam/xact.c#L2780) the callback with `XACT_EVENT_COMMIT` will be called in a section marked with following comment: 
```
	/*
	 * This is all post-commit cleanup.  Note that if an error is raised here,
	 * it's too late to abort the transaction.  This should be just
	 * noncritical resource releasing.
	 *
```
I suppose that resgroup update is a bit beyond 'noncritical resource releasing'. Aren't we updating the resgroups too late?

Btw, `RegisterXactCallback()` has a comment:
```
 * These functions are intended for use by dynamically loaded modules.
 * For built-in modules we generally just hardwire the appropriate calls
 * (mainly because it's easier to control the order that way, where needed).
```
Aren't we misusing this api?

src/backend/commands/resgroupcmds.c
```diff
@@ -1150,6 +1214,154 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+			return left->memLimit == right->memLimit;
+		case RESGROUP_LIMIT_TYPE_MEMORY_SHARED_QUOTA:
+			return left->memSharedQuota == right->memSharedQuota;
+		case RESGROUP_LIMIT_TYPE_MEMORY_SPILL_RATIO:
+			return left->memSpillRatio == right->memSpillRatio;
+		case RESGROUP_LIMIT_TYPE_MEMORY_AUDITOR:
+			return left->memAuditor == right->memAuditor;
+		case RESGROUP_LIMIT_TYPE_UNKNOWN:
+		case RESGROUP_LIMIT_TYPE_COUNT:
+			break;
```
whitehawk, line 1221:
```suggestion
		default:
			break;
```

src/backend/commands/resgroupcmds.c
```diff
@@ -1150,6 +1214,154 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+
+		foreach (lc, resgroup_alter_tran_callbacks)
+		{
+			ResourceGroupCallbackContext *ctx = lfirst(lc);
+			ResGroupCaps finalCaps;
+
+			GetResGroupCapabilities(rel, ctx->groupid, &finalCaps);
+
+			/* Skip changes rolled back or overwritten before commit. */
+			if (!resGroupCapFieldMatches(ctx->limittype,
```
whitehawk: will it skip the changes applied at a later point? For, ex:
```
ALTER RESOURCE GROUP rg_alter_tran   SET CONCURRENCY 20;
ALTER RESOURCE GROUP rg_alter_tran   SET CONCURRENCY 21;
ALTER RESOURCE GROUP rg_alter_tran   SET CONCURRENCY 22;
```
does it mean that only `CONCURRENCY 20` will be applied?

src/backend/utils/resgroup/resgroup.c
whitehawk, line 1019: Not directly related to this patch, but it appears that we do not restore old memory context from `ErrorContext` when handling errors here, while we should do it, as we do not re-throw error.

src/backend/commands/resgroupcmds.c
```diff
@@ -531,7 +567,35 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
 		callbackCtx->caps = caps;
 		callbackCtx->oldCaps = oldCaps;
-		RegisterXactCallbackOnce(alterResgroupCallback, callbackCtx);
+		if (IsInTransactionChain(isTopLevel))
+		{
+			/*
+			 * In transaction, collect ALTER data for later synchronous apply.
+			 */
+			Assert(gp_resource_group_enable_alter_in_transaction);
+			MemoryContext oldcxt = MemoryContextSwitchTo(TopMemoryContext);
```
whitehawk: Why is `TopMemoryContext` used? Maybe better to use `CurTransactionContext` or `TopTransactionContext`?

src/backend/utils/resgroup/resgroup.c
```diff
@@ -904,18 +951,25 @@ ResGroupAlterOnCommit(const ResourceGroupCallbackContext *callbackCtx)
 		savedInterruptHoldoffCount = InterruptHoldoffCount;
 		group = groupHashFind(callbackCtx->groupid, true);
 
-		group->caps = callbackCtx->caps;
+		oldCaps = group->caps;
```
whitehawk, line 959: `ResourceGroupCallbackContext.oldCaps` seems to be not used anymore, should it be removed?

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
whitehawk: `-M fast` is the non-deprecated way of setting the shutdown mode.

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
whitehawk: Hmm, in the .out file I still do not see per-segment info in the output of `SELECT * FROM rg_alter_tran_status;`. How does the per-segment check work?

src/backend/commands/resgroupcmds.c
RekGRpth: If we do this, we might forget to handle a new option specifically when adding it to the enumeration. In the current implementation, we'll get a compiler warning.

src/backend/commands/resgroupcmds.c
RekGRpth: The `resgroup_alter_tran_callbacks` list can be used in another transaction.

src/backend/commands/resgroupcmds.c
```diff
@@ -531,7 +567,35 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+			resgroup_alter_tran_callbacks = lappend(
+												resgroup_alter_tran_callbacks,
+												callbackCtx);
```
RekGRpth: wrong indent

src/backend/commands/resgroupcmds.c
```diff
@@ -531,7 +567,35 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
 		callbackCtx->limittype = limitType;
 		callbackCtx->caps = caps;
 		callbackCtx->oldCaps = oldCaps;
-		RegisterXactCallbackOnce(alterResgroupCallback, callbackCtx);
+		if (IsInTransactionChain(isTopLevel))
```
RekGRpth: add empty line before

src/backend/commands/resgroupcmds.c
```diff
@@ -1150,6 +1214,154 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+
+	if (resgroup_alter_tran_callbacks == NIL)
+		return;
+
+	if (event == XACT_EVENT_ABORT)
+	{
+		foreach (lc, resgroup_alter_tran_callbacks)
+			pfree(lfirst(lc));
+
+		list_free(resgroup_alter_tran_callbacks);
```
RekGRpth: `list_free_deep`?

src/backend/utils/resgroup/resgroup.c
```diff
@@ -941,9 +995,9 @@ ResGroupAlterOnCommit(const ResourceGroupCallbackContext *callbackCtx)
 								  defaultCpusetGroup,
 								  MaxCpuSetLength);
 			/* Add old value to default group
-			 * sub new value from default group */
-			char *cpuset= getCpuSetByRole(callbackCtx->caps.cpuset);
-			char *oldcpuset = getCpuSetByRole(callbackCtx->oldCaps.cpuset);
+			* sub new value from default group */
```
RekGRpth: wrong indent

src/backend/commands/resgroupcmds.c
RekGRpth: To handle multiple options it is better to use `if-else if-else if`, see for example `plpgsql_xact_cb`.

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: no. The finalCaps downloaded from heap table which will contain only final transaction value 22. So here we will skip 20 and 21 setting.

The 20 will be in this case:
```sql
ALTER RESOURCE GROUP rg_alter_tran   SET CONCURRENCY 20;
SAVEPOINT s1;
ALTER RESOURCE GROUP rg_alter_tran   SET CONCURRENCY 21;
ALTER RESOURCE GROUP rg_alter_tran   SET CONCURRENCY 22;
ROLLBACK TO SAVEPOINT s1;
```

src/backend/utils/resgroup/resgroup.c
hilltracer: I though about removing ResourceGroupCallbackContext.oldCaps. But safe it as it is done to not break ABI compatibility.
Also keep oldCaps initialization to not have garbage data inside it and minimize diff in AlterResourceGroup().

src/backend/commands/resgroupcmds.c
hilltracer: Agree, it looks bad. But
ResGroupAlterOnCommit() is already called from an XACT_EVENT_COMMIT callback in the existing code path. So the current patch does not move runtime apply later; it keeps the same model.

The reason why it's not on PRE_COMMIT is probably that runtime state must only be changed if the catalog transaction commits. If it were applied before the commit point and the transaction later aborted, we would need rollback logic for shared memory and cgroup state, which is exactly the hard part.

Here the upstream removed their own AtEOXact_ResGroup() / HandleResGroupDDLCallbacks() mechanism and changed it to xact callback mechanism:
https://github.com/GreengageDB/greengage/commit/c1cdb99d6f58f8d4a0ba21b6ab48fc1e437254ff

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
hilltracer: Data from all segments is grouped into one row to minimize output. If there is any inconsistency, the check will determine this. Added a comment about this.

.abi-check/6.30.1/postgres.types.ignore
hilltracer: removed

src/backend/commands/resgroupcmds.c
hilltracer: converted to local var. Thx.

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: ok, reworked

src/backend/commands/resgroupcmds.c
hilltracer: no. See this discussion

Image

src/backend/commands/resgroupcmds.c
whitehawk: Ok, lets leave it as is for now.

src/backend/utils/resgroup/resgroup.c
hilltracer: Added restoring to oldMemoryContext. Thanks.

src/backend/commands/resgroupcmds.c
hilltracer: done. Thx.

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/utils/resgroup/resgroup.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: > The resgroup_alter_tran_callbacks list can be used in another transaction.

No, since we take exclusive lock. But on segments TopTransactionContext [is released](https://github.com/GreengageDB/greengage/blob/c4e1085d26bae201ef29191373ecf7b6b92c534f/src/backend/access/transam/xact.c#L3190) before XACT_EVENT_COMMIT during PrepareTransaction(). So we have to use TopMemoryContext here.

src/test/isolation2/expected/resgroup/resgroup_alter_in_transaction.out
hilltracer: reworked with fault injector

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: done

gpcontrib/gp_internal_tools/gp_resource_group.c
```diff
@@ -87,29 +89,51 @@ pg_resgroup_move_query(PG_FUNCTION_ARGS)
...
+			if (currentGroupId == InvalidOid)
+				ereport(ERROR,
+						(errcode(ERRCODE_UNDEFINED_OBJECT),
+						(errmsg("process %d is in IDLE state", pid))));
+			if (currentGroupId != groupId)
+				ResGroupMoveQuery(sessionId, groupId, groupName);
+		}
+		PG_CATCH();
+		{
+			UnlockRelationOid(ResGroupCapabilityRelationId, RowShareLock);
```
RekGRpth: Doesn't the lock get removed when the transaction is aborted?

src/backend/commands/resgroupcmds.c
```diff
@@ -531,7 +565,27 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+		/*
+		 * Collect ALTERs and apply them together at COMMIT after PRE_COMMIT
+		 * validates final catalog state. Outside an explicit transaction this
+		 * list contains only one ALTER.
+		 */
+		static bool alter_tran_callback_registered = false;
+		MemoryContext oldcxt = MemoryContextSwitchTo(TopMemoryContext);
+		resgroup_alter_callbacks = lappend(resgroup_alter_callbacks,
+										   callbackCtx);
+		MemoryContextSwitchTo(oldcxt);
```
RekGRpth, line 577: Why do we need to accumulate all the `callbackCtx` if we can only apply one? Wouldn't it be easier to save only the changed options in one `callbackCtx` and apply them later?

src/backend/utils/resgroup/resgroup.c
```diff
@@ -888,34 +891,90 @@ ResGroupCreateOnAbort(const ResourceGroupCallbackContext *callbackCtx)
...
 	PG_TRY();
 	{
 		savedInterruptHoldoffCount = InterruptHoldoffCount;
 		group = groupHashFind(callbackCtx->groupid, true);
 
-		group->caps = callbackCtx->caps;
+		oldCaps = group->caps;
+		newCaps = oldCaps;
+
+		resGroupCapFieldApply(callbackCtx->limittype,
```
whitehawk: Maybe just pass the entire `callbackCtx` into `resGroupCapFieldApply()`? Therefore we can reduce number of parameters.

src/backend/utils/resgroup/resgroup.c
```diff
@@ -957,6 +1016,7 @@ ResGroupAlterOnCommit(const ResourceGroupCallbackContext *callbackCtx)
 	PG_CATCH();
 	{
 		InterruptHoldoffCount = savedInterruptHoldoffCount;
+		MemoryContextSwitchTo(oldMemoryContext);
```
whitehawk: We have the same issue in `ResGroupDropFinish()` and in `ResGroupCreateOnAbort()`. And, as it is not related directly to the scope of this PR, I suggest to create a separate ticket, and introduce this fix for memory context restore in a separate patch.

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
```diff
@@ -0,0 +1,307 @@
...
+SELECT * FROM rg_alter_tran_status;
+
+1q:
+2q:
+
+
+-- Crash recovery
+
+-- 22 Coordinator segfault before COMMIT must not apply ALTERs.
+1: SELECT gp_inject_fault('exec_simple_query_start', 'reset', dbid) FROM gp_segment_configuration WHERE role = 'p' AND content = -1;
```
whitehawk, line 395: Do we need this fault 'reset' here at the 1st usage?

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
```diff
@@ -0,0 +1,307 @@
...
+1: SELECT gp_inject_fault('exec_simple_query_start', 'segv', dbid) FROM gp_segment_configuration WHERE role = 'p' AND content = -1;
+-- The backend dies at the very start of COMMIT.
+1: @post_run 'echo ""' : COMMIT;
+-- Wait until coordinator accepts connections again.
+! while [ `psql -tc "SELECT 1;" postgres 2>/dev/null | wc -l` != '2' ]; do sleep 1; done;
+1q:
+1: SELECT * FROM rg_alter_tran_status;
+
+-- 23 Segment segfault before COMMIT must not apply ALTERs.
+-- Crash before COMMIT: during COMMIT, coordinator changes may already be
```
whitehawk: Doesn't that sound like an issue?

src/backend/commands/resgroupcmds.c
hilltracer: We must to store all ALTERs even for one limittype because some of them may be rolled back. I modified test 17 to show this.

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: I think yes. To eliminate the influence of previous unfinished runs.

gpcontrib/gp_internal_tools/gp_resource_group.c
hilltracer: Yes. So we don't need this PG_CATCH. So I reverted it and added test (see test 20)

src/backend/utils/resgroup/resgroup.c
hilltracer: done

src/backend/utils/resgroup/resgroup.c
hilltracer: done. I reverted this fix and created new ticket GG-499

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: Yes. I reworked previous test and removed this comment.

whitehawk (dismissed):

src/backend/commands/resgroupcmds.c
```diff
@@ -1133,21 +1187,148 @@ dropResgroupCallback(XactEvent event, void *arg)
...
 }
 
+/*
+ * Check whether the field changed by limittype has the same value in both
+ * capability snapshots.
+ */
+static bool
+resGroupCapFieldMatches(ResGroupLimitType limittype,
+						const ResGroupCaps *left,
+						const ResGroupCaps *right)
```
RekGRpth:
```suggestion
resGroupCapFieldMatches(ResourceGroupCallbackContext *ctx,
						const ResGroupCaps *caps)
```

src/backend/commands/resgroupcmds.c
```diff
@@ -1133,21 +1187,148 @@ dropResgroupCallback(XactEvent event, void *arg)
...
+				rejected = lappend(rejected, ctx);
+				continue;
+			}
+			/* Drop duplicate target callbacks. */
+			if (resgroupAlterCallbackAlreadyPrepared(prepared, ctx))
+			{
+				rejected = lappend(rejected, ctx);
+				continue;
+			}
+			prepared = lappend(prepared, ctx);
```
RekGRpth:
```suggestion
			if (resGroupCapFieldMatches(ctx->limittype, &ctx->caps, &finalCaps) &&
				!resgroupAlterCallbackAlreadyPrepared(prepared, ctx))
				prepared = lappend(prepared, ctx);
			else
				rejected = lappend(rejected, ctx);
```

src/backend/utils/resgroup/resgroup.c
```diff
@@ -888,34 +890,88 @@ ResGroupCreateOnAbort(const ResourceGroupCallbackContext *callbackCtx)
...
+			(errcode(ERRCODE_UNDEFINED_OBJECT),
+			(errmsg("invalid resource group limit type: %d",
+					 callbackCtx->limittype))));
+}
+
 /*
  * Apply the new resgroup caps.
+ *
+ * Only callbackCtx->limittype is applied. The previous state is read
+ * from shared memory.
```
RekGRpth: Could you please explain this and other changes in this function in more detail?

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
```diff
@@ -0,0 +1,316 @@
...
+     json_array_elements((d.value::json)->'info') AS seg,
+     json_array_elements(seg->'groups') AS grp,
+     pg_resgroup r
+WHERE r.rsgname IN ('rg_alter_tran', 'rg_alter_tran_b')
+  AND (grp->>'group_id')::oid = r.oid
+ORDER BY groupname, gp_segment_id;
+
+-- Use PL/pgSQL to run heap and runtime checks as separate statements;
+-- one SQL query may fail with "multiple segworker groups is not supported".
+CREATE OR REPLACE FUNCTION rg_alter_tran_all_status() RETURNS TABLE(gp_segment_id int, groupname text, concurrency int, cpu_rate_limit int, memory_limit int) AS $$ BEGIN RETURN QUERY EXECUTE 'SELECT gp_segment_id, groupname, concurrency, cpu_rate_limit, memory_limit FROM rg_alter_tran_heap_status'; RETURN QUERY EXECUTE 'SELECT gp_segment_id, groupname, concurrency, cpu_rate_limit, memory_limit FROM rg_alter_tran_runtime_status'; END; $$ LANGUAGE plpgsql EXECUTE ON MASTER;
```
RekGRpth: Could you please format all the long lines in this file pretty?

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/utils/resgroup/resgroup.c
hilltracer: done

src/test/isolation2/sql/resgroup/resgroup_alter_in_transaction.sql
hilltracer: done

bandetto (commented):
@hilltracer GitHub does not techincally allow to approve PRs created by myself. This message can be considered as an approval.

src/backend/utils/resgroup/resgroup.c
```diff
@@ -889,33 +891,88 @@ ResGroupCreateOnAbort(const ResourceGroupCallbackContext *callbackCtx)
...
+
+/*
+ * Apply the new resgroup caps and update dependent runtime state.
+ *
+ * Each callback keeps a full capability snapshot, but only its limittype is
+ * installed over the current shared-memory state. This lets multiple callbacks
+ * from one transaction update different fields without overwriting each other.
  */
 void
 ResGroupAlterOnCommit(const ResourceGroupCallbackContext *callbackCtx)
```
RekGRpth, line 943: Before the patch, the `ResGroupAlterOnCommit` function was called only once, but now it can be called several times. How good is it that now internal functions like `ResGroupOps_SetCpuRateLimit`, `ResGroupOps_SetCpuSet` will now be called several times?

RekGRpth: Why is it not possible to collect one resulting `ResourceGroupCallbackContext` in the loop above, and then call this function `ResGroupAlterOnCommit` for it once?

hilltracer: ResGroupOps_SetCpuRateLimit() and ResGroupOps_SetCpuSet() will not be
called multiple times for the same final target.  During PRE_COMMIT we keep only
final, non-duplicate callbacks: if several callbacks target the same group,
limit type, and resulting value, only one of them is left in the prepared list.

CPU rate limit and CPUSET are handled as one runtime target intentionally.  They
are mutually exclusive in resource group capabilities: setting CPU rate clears
CPUSET, while setting CPUSET disables CPU rate limit.  Because of that,
resGroupCapFieldMatches() treats `RESGROUP_LIMIT_TYPE_CPU` and
`RESGROUP_LIMIT_TYPE_CPUSET` as the same effective CPU configuration by comparing
both `cpuRateLimit` and `cpuset`.

So if one transaction contains several CPU/CPUSET ALTERs, only the final catalog
state survives PRE_COMMIT validation, and duplicate callbacks for the same final
CPU configuration are dropped before COMMIT apply.

hilltracer: > Why is it not possible to collect one resulting ResourceGroupCallbackContext in the loop above, and then call this function ResGroupAlterOnCommit for it once?

A transaction may change several independent limit types.  In that case we need
one final callback per changed limit type, so memory, concurrency, CPU and
cpuset-specific side effects are all executed.

RekGRpth (dismissed):

RekGRpth (approved):

whitehawk (approved):

PR #418, opened 2026-04-29
Add copyright headers to ggrebalance

whitehawk:
Add copyright headers to ggrebalance

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/test/config.py
+++ b/gpMgmt/bin/ggrebalance_modules/test/config.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2026 Greengage Community
+# Copyright (c) 2025-Present, Greengage Community
 #
 
 import os
--- a/gpMgmt/bin/ggrebalance_modules/test/config.py
+++ b/gpMgmt/bin/ggrebalance_modules/test/config.py
@@ -1,3 +1,7 @@
+#
+# Copyright (c) 2026 Greengage Community
+#
+
 import os
 
 from gppylib.gparray import Segment, GpArray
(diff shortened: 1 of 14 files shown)


gpMgmt/bin/ggrebalance_modules/test/config.py
```diff
@@ -1,3 +1,7 @@
+#
+# Copyright (c) 2026 Greengage Community
```
RekGRpth: why not
```suggestion
# Copyright (c) 2025-Present Greengage Community
```

whitehawk: Because in the existing files ([example](https://github.com/GreengageDB/greengage/blob/f7099c8d918cdd64bd160c7e381a419c295fcf21/src/backend/catalog/storage_pending_deletes.c#L6)] we already have copyright in a format \<the year of adding\> + 'Greengage Community' .

Generally speaking, the information about proper copyright header format is not an engineering question and it should be defined in the code-style/code-review guidelines, but so far we do not have it (though I've requested it several times). If we had it, that could eliminate any such questions...

RekGRpth: > we already have copyright in a format <the year of adding> + 'Greengage Community' .

https://github.com/GreengageDB/greengage/blob/f7099c8d918cdd64bd160c7e381a419c295fcf21/src/include/nodes/pathnodes.h#L9

RekGRpth: Work on the `rebalance` began in `2025`, I think. Are you planning to update the copyrights every year?

RekGRpth (approved):

red1452 (approved):

PR #421, opened 2026-04-30
GG-426: Port a check from upstream repo [7X]

GirgerS:
Views that use functions with changed signatures will cause upgrade to fail. This happens during metadata restore on the target cluster because pg_restore will error trying to create a view using functions with changed arguments and/or return types. This is not ideal as we could be many hours into upgrade before a view using a function with changed signature causes pg_upgrade to fail This check calls a support function on the source cluster to check if views using such functions with changed signature exist.

(cherry picked from commit 3c0ffffc651fbbe9febd005eee4cd4022fff2bb2)

Changes compared to the original commit:
- Use fully qualified names, and remove queries that change schemas.
- Properly format error message.
- Close created file.
- Unhardcode FirstNormalObjectId.

---

MERGE WITH REBASE

### DIFF
--- a/src/bin/pg_upgrade/greengage/check_gp.c
+++ b/src/bin/pg_upgrade/greengage/check_gp.c
@@ -15,6 +15,7 @@
  */
 
 #include "pg_upgrade_greengage.h"
+#include "catalog/pg_magic_oid.h"
 #include "catalog/pg_class_d.h"
 
 #define RELSTORAGE_EXTERNAL	'x'
@@ -31,6 +32,7 @@ static void check_views_with_removed_operators(void);
 static void check_views_with_removed_functions(void);
 static void check_views_with_removed_types(void);
 static void check_for_disallowed_pg_operator(void);
+static void check_views_with_changed_function_signatures(void);
 
 /*
  *	check_greengage
@@ -55,6 +57,7 @@ check_greengage(void)
 	check_views_with_removed_functions();
 	check_views_with_removed_types();
 	check_for_disallowed_pg_operator();
+	check_views_with_changed_function_signatures();
 }
 
 /*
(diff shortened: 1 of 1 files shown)


GirgerS:
The [function](https://github.com/GreengageDB/greengage/blob/6.x/contrib/pg_upgrade_support/pg_upgrade_support.c#L719-L741) used in the query. I didn't check the [hardcoded](https://github.com/GreengageDB/greengage/blob/6.x/contrib/pg_upgrade_support/pg_upgrade_support.c#L751-L813) oids though, because it is not a part of this patch, and the ticket it to fix a test anyway. 

To test the changes, create a similar view in 6.x cluster:

```sql 
create view badview_2 as select 'public' || (select '.' ||pg_catalog.to_regclass('aoco_dropped_col_and_encoding_partitioned'));
```

And then run pg_upgrade (changing paths to those from you local system):
```
../7.x/bin/pg_upgrade --old-bindir='/home/regrig/greengage/6.x/bin' --new-bindir='/home/regrig/greengage/7.x/bin' --old-datadir="/home/regrig/greengage/6.x-data/data/gpseg-1" --new-datadir="/home/regrig/greengage/7.x-data/data/gpseg-1" --check
```

whitehawk (dismissed):

red1452:
Why is original commit hash different from hash in the commit message?
`11973217973ec7decfdf6d411fce70cedb160135`
`3c0ffffc651fbbe9febd005eee4cd4022fff2bb2`

red1452 (approved):

GirgerS:
> Why is original commit hash different from hash in the commit message?

Thanks for noticing! I somehow cherry-picked a local commit:

```
git name-rev 3c0ffffc651fbbe9febd005eee4cd4022fff2bb2
3c0ffffc651fbbe9febd005eee4cd4022fff2bb2 stash~2
```

I remember redoing cherry-pick with `-x`, but not sure how I ended up there.

whitehawk (approved):

PR #422, opened 2026-05-04
ADBDEV-9727: Fix loging the Exceptions in gplog

KnightMurloc:
In Python 2, if an Exception contains unicode characters, an attempt to convert
it to unicode will result in an error.

This patch restores the old behavior. If we log an Exception, we first cast it
to the str. After that, if it is a byte string, then decode it to the unicode.
If the Exception contains unicode, then catch it and log a fixed message.

### DIFF
--- a/gpMgmt/bin/gppylib/gplog.py
+++ b/gpMgmt/bin/gppylib/gplog.py
@@ -320,10 +320,13 @@ class EncodingFileHandler(logging.FileHandler):
         logging.FileHandler.__init__(self, filename, mode, encoding, delay)
 
     def emit(self, record):
+        if not isinstance(record.msg, (string_types, binary_type)):
+            try:
+                record.msg = str(record.msg)
+            except UnicodeEncodeError:
+                record.msg = "Can't decode exception message"
         if isinstance(record.msg, binary_type):
-            record.msg = record.msg.decode('utf-8')
-        elif not isinstance(record.msg, text_type):
-            record.msg = text_type(record.msg)
+            record.msg = record.msg.decode('utf-8', errors='replace')
         logging.FileHandler.emit(self, record)
 
 
@@ -336,8 +339,11 @@ class EncodingStreamHandler(logging.StreamHandler):
         logging.StreamHandler.__init__(self, strm)
 
     def emit(self, record):
+        if not isinstance(record.msg, (string_types, binary_type)):
+            try:
+                record.msg = str(record.msg)
+            except UnicodeEncodeError:
+                record.msg = "Can't decode exception message"
         if isinstance(record.msg, binary_type):
-            record.msg = record.msg.decode('utf-8')
-        elif not isinstance(record.msg, text_type):
-            record.msg = text_type(record.msg)
+            record.msg = record.msg.decode('utf-8', errors='replace')
         logging.StreamHandler.emit(self, record)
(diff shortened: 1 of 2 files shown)


gpMgmt/bin/gppylib/gplog.py
```diff
@@ -320,10 +320,13 @@ def __init__(self, filename, mode='a', encoding=None, delay=0):
         logging.FileHandler.__init__(self, filename, mode, encoding, delay)
 
     def emit(self, record):
+        if not isinstance(record.msg, (string_types, binary_type)):
+            try:
+                record.msg = str(record.msg)
```
dkovalev1, line 325: Which exact types can be here?

KnightMurloc: Technically, any type passed to `logger.log` except strings. But more likely Exception.

dkovalev1, line 325: with python2 I getting:
```
>>> isinstance (u"пр", basestring)
True
```
It goes to the logging.FileHandler.emit(self, record) unchanged, is it expected?

gpMgmt/bin/gppylib/gplog.py
```diff
@@ -320,10 +320,13 @@ def __init__(self, filename, mode='a', encoding=None, delay=0):
         logging.FileHandler.__init__(self, filename, mode, encoding, delay)
 
     def emit(self, record):
+        if not isinstance(record.msg, (string_types, binary_type)):
```
dkovalev1, line 323: will the logic like 
```
if isinstance(record.msg, binary_type):
  ...
elif not if isinstance(record.msg, string_types):
  ...
else:
   # string_types, do nothing?
   ...
```
be more clear?

KnightMurloc: As I understand it, we always want to output unicode. But we can't get unicode from Exception directly in Python 2. Therefore, if it is not a string, we first convert it to a byte string. Then we convert it to unicode.

dkovalev1: okay with passing unicode, but what you think about if conditions?

KnightMurloc: With this approach, it will look something like this:
```python
if isinstance(record.msg, binary_type):
    record.msg = record.msg.decode('utf-8', errors='replace')
elif not if isinstance(record.msg, string_types):
    try:
        record.msg = str(record.msg).decode('utf-8', errors='replace')
    except UnicodeEncodeError:
        record.msg = "Can't decode exception message"
```
Do you think it's better?

gpMgmt/bin/gppylib/gplog.py
KnightMurloc: Yes, since it's unicode.

gpMgmt/bin/gppylib/gplog.py
dkovalev1: From my point of view yes, but it is up to you

KnightMurloc: This version does not work in Python 3 because `str` is already unicode. Adding branches for the version seems to me a bad idea. The current version works fine on both Pythons.

dkovalev1 (approved):

ShapatinAA (approved):

PR #423, opened 2026-05-05
GG-396: Fix assert trigger when vacuum AO table after ggupgrade

GirgerS:
Vacuuming AO table after pg_upgrade from 6.x to 7.x triggered 
an assertion inside AppendOnlyCompact function. 
 
It was caused by the difference in the handling of the pg_aoseg 
tables on the coordinator:

- in 6.x, these tables contain the totals of the whole distributed 
  table
- in 7.x, the code expects them to remain empty. Otherwise it 
  is possible to reach executor-only code.
 
Note that there is an exception to this rule in 7.x:
`ALTER TABLE ... REPACK BY COLUMNS (...)` and `CLUSTER ...`
queries do create an oddly specific entry for RESERVED_SEGNO
segfile of the AO table with zero tuples. But these commands are 
absent in 6.x, so pg_upgrade code can ignore this edge case.
 
Therefore the most simple fix is to skip copying these tables on 
the coordinator during upgrade. Also, don't copy pg_aoblkdir, 
pg_aovisimap and gp_fastsequence tables.

This patch is done as a part of gguprade migration.
Specifically, it fixes vacuum_freeze_tables test.

Ticket: GG-396

### DIFF
--- a/src/bin/pg_upgrade/pg_upgrade.c
+++ b/src/bin/pg_upgrade/pg_upgrade.c
@@ -241,26 +241,15 @@ main(int argc, char **argv)
 		 * however. So we still need to restore those separately on each
 		 * server.
 		 *
-		 * Because we are concerned with 6.x -> 7.x migration here, this
-		 * step should be done only for segments, because starting from
+		 * Because we are concerned with 6.x -> 7.x (or further) migration here,
+		 * this step should be done only for segments, because starting from
 		 * Greengage 7, pg_aoseg table entries are almost never created
-		 * on the coordinator. With a notable exception:
-		 *
-		 * 'ALTER TABLE ... REPACK BY COLUMNS (...)' and 'CLUSTER ...'
-		 * queries do create an oddly specific entry for RESERVED_SEGNO
-		 * segfile of the AO table with zero tuples. It looks like a flaw
-		 * in the logic, moreover, these commands are absent in 6.x.
-		 *
-		 * If any other pg_aoseg entry (that is, not similar to the one
-		 * created by the aforementioned queries) is present in the target
-		 * cluser, it would cause a failure to 'VACUUM' corresponding
-		 * AO table (see vacuum_ao.c , line 378).
+		 * on the coordinator.
 		 *
 		 * 'restore_aosegment_tables' function also deals with other tables,
 		 * so, in total, when upgrading the coordinator:
 		 *
-		 * Don't copy pg_aoseg tables. Coordinator can always recreate
-		 * necessary entries in the above edge case.
+		 * Don't copy pg_aoseg tables.
 		 *
 		 * Don't copy pg_aoblkdir and pg_aovisimap tables, because they should
 		 * be empty too.
--- a/src/bin/pg_upgrade/pg_upgrade.c
+++ b/src/bin/pg_upgrade/pg_upgrade.c
@@ -243,9 +243,33 @@ main(int argc, char **argv)
 		 *
 		 * Because we are concerned with 6.x -> 7.x migration here, this
 		 * step should be done only for segments, because starting from
-		 * Greengage 7, pg_aoseg tables on the coordinator are expected
-		 * to remain empty. Don't copy pg_aoblkdir and pg_aovisimap tables
-		 * either, they should be empty too.
+		 * Greengage 7, pg_aoseg table entries are almost never created
+		 * on the coordinator. With a notable exception:
+		 *
+		 * 'ALTER TABLE ... REPACK BY COLUMNS (...)' and 'CLUSTER ...'
+		 * queries do create an oddly specific entry for RESERVED_SEGNO
+		 * segfile of the AO table with zero tuples. It looks like a flaw
+		 * in the logic, moreover, these commands are absent in 6.x.
+		 *
+		 * If any other pg_aoseg entry (that is, not similar to the one
+		 * created by the aforementioned queries) is present in the target
+		 * cluser, it would cause a failure to 'VACUUM' corresponding
+		 * AO table (see vacuum_ao.c , line 378).
+		 *
+		 * 'restore_aosegment_tables' function also deals with other tables,
+		 * so, in total, when upgrading the coordinator:
+		 *
+		 * Don't copy pg_aoseg tables. Coordinator can always recreate
+		 * necessary entries in the above edge-case.
+		 *
+		 * Don't copy pg_aoblkdir and pg_aovisimap tabes, because they should
+		 * be empty too.
+		 *
+		 * Don't copy gp_fastsequence table, it will be filled when recreating
+		 * AO tables.
+		 *
+		 * (the difference between versions comes mainly from
+		 *  the commit 5778f31124d5737e18d76f8d06f07f96c31c8be7)
 		 */
 		restore_aosegment_tables();
 	}
--- a/src/bin/pg_upgrade/pg_upgrade.c
+++ b/src/bin/pg_upgrade/pg_upgrade.c
@@ -260,9 +260,9 @@ main(int argc, char **argv)
 		 * so, in total, when upgrading the coordinator:
 		 *
 		 * Don't copy pg_aoseg tables. Coordinator can always recreate
-		 * necessary entries in the above edge-case.
+		 * necessary entries in the above edge case.
 		 *
-		 * Don't copy pg_aoblkdir and pg_aovisimap tabes, because they should
+		 * Don't copy pg_aoblkdir and pg_aovisimap tables, because they should
 		 * be empty too.
 		 *
 		 * Don't copy gp_fastsequence table, it will be filled when recreating
--- a/src/bin/pg_upgrade/pg_upgrade.c
+++ b/src/bin/pg_upgrade/pg_upgrade.c
@@ -231,15 +231,24 @@ main(int argc, char **argv)
 		create_new_objects();
 	}
 
-	/*
-	 * In a segment, the data directory already contains all the objects,
-	 * because the segment is initialized by taking a physical copy of the
-	 * upgraded QD data directory. The auxiliary AO tables - containing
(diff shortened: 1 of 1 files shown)


src/bin/pg_upgrade/pg_upgrade.c
```diff
@@ -231,15 +231,24 @@ main(int argc, char **argv)
...
-	/*
-	 * In a segment, the data directory already contains all the objects,
-	 * because the segment is initialized by taking a physical copy of the
-	 * upgraded QD data directory. The auxiliary AO tables - containing
-	 * information about the segment files, are different in each server,
-	 * however. So we still need to restore those separately on each
-	 * server.
-	 */
-	restore_aosegment_tables();
+	if (!is_greengage_dispatcher_mode())
```
red1452, line 234: According to comment below, is it correct check? As I understand correctly, it is check for type of connection, which does not separate primary end mirror.

GirgerS: I see, I slightly confused the terminology. I meant primary segments or mirrors, that is, simply segments.

red1452 (commented):
What about tests?

GirgerS:
> What about tests?

The main purpose of this patch is to fix the test. 

https://github.com/GreengageDB/ggupgrade/blob/main/test/acceptance/pg_upgrade/6-to-7/upgradeable_tests/target_cluster_regress/sql/vacuum_freeze_tables.sql#L1-L18

It is not really intuitive that we (would) test pg_upgrade in the separate repository, but it is what it is. We do run [this](https://github.com/GreengageDB/greengage/blob//7.x/src/bin/pg_upgrade/test_gpdb.sh) test though, but it is not that helpful.

red1452:
> > What about tests?
> 
> The main purpose of this patch is to fix the test.
> 
> https://github.com/GreengageDB/ggupgrade/blob/main/test/acceptance/pg_upgrade/6-to-7/upgradeable_tests/target_cluster_regress/sql/vacuum_freeze_tables.sql#L1-L18
> 
> It is not really intuitive that we (would) test pg_upgrade in the separate repository, but it is what it is. We do run [this](https://github.com/GreengageDB/greengage/blob//7.x/src/bin/pg_upgrade/test_gpdb.sh) test though, but it is not that helpful.

Add note to the description.

whitehawk:
In the description:
'It was caused be the difference' -> 'It was caused by the difference' ?

src/bin/pg_upgrade/pg_upgrade.c
```diff
@@ -231,15 +231,24 @@ main(int argc, char **argv)
...
+		 * however. So we still need to restore those separately on each
+		 * server.
+		 *
+		 * Because we are concerned with 6.x -> 7.x migration here, this
+		 * step should be done only for segments, because starting from
+		 * Greengage 7, pg_aoseg tables on the coordinator are expected
+		 * to remain empty. Don't copy pg_aoblkdir and pg_aovisimap tables
+		 * either, they should be empty too.
+		 */
+		restore_aosegment_tables();
```
whitehawk, line 263: `restore_aosegment_tables()` also do some updates of `gp_fastsequence` table. Is it Ok to skip this table handling on the coordinator as well?

GirgerS: > restore_aosegment_tables() also do some updates of gp_fastsequence table. Is it Ok to skip this table handling on the coordinator as well?

I think so.

I left it without much attention, because it directly references `pg_aoseg` entries, which should are absent on the coordinator in 7.x. But I doubt that it count as a complete answer, so:

`gp_fastsequence` DOES have entries on the coordinator, but they are left with [default](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/catalog/gp_fastsequence.c#L77-L81) values.

They reason for that being that `gp_fastsequence` sees to be updated only via [these](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/catalog/gp_fastsequence.c#L89-L283) functions. Call tree for them looks like: 

---

... -> [AOCSSegmentFileFullCompaction](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/access/aocs/aocs_compaction.c#L238-L380) ([has](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/access/aocs/aocs_compaction.c#L409) an assertion that we are on the executor) -> AOCSMoveTuple-> aocs_insert_values -> GetFastSequences  -> insert_or_update_fastsequence

... -> [aoco_relation_cluster_internals](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/access/aocs/aocsam_handler.c#L1428-L1613) (call chain [continues](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/access/aocs/aocsam_handler.c#L1516-L1606) only if there are tuples in the table. It it not the case on the cooridnator) -> aocs_insert_values -> GetFastSequences -> insert_or_update_fastsequence

... -> [ao_column_methods](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/access/aocs/aocsam_handler.c#L2750-L2828) (access method. I assume that there are should be no inserts to the AO table on the coordinator, because data is stored on the segments) -> aoco_tuple_insert/aoco_multi_insert/aoco_tuple_update  -> aocs_insert -> aocs_insert_values -> GetFastSequences -> insert_or_update_fastsequence

... -> [AOCSCompact](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/access/aocs/aocs_compaction.c#L395-L455) (asserts that we are on the executor ) -> aocs_insert_init -> GetFastSequences -> insert_or_update_fastsequence

... -> [ao_column_methods](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/access/aocs/aocsam_handler.c#L2750-L2828) (the same as above) -> aoco_tuple_insert/aoco_multi_insert/aoco_tuple_update -> aocs_insert_init -> GetFastSequences -> insert_or_update_fastsequence

... -> [aoco_relation_cluster_internals](https://github.com/GreengageDB/greengage/blob/b192a8e7da980e1b858f33487fce309505459b61/src/backend/access/aocs/aocsam_handler.c#L1428-L1613) (we have already seen this function. But it IS executed on the coordinator (see the following comment)) -> aocs_insert_init -> GetFastSequences -> insert_or_update_fastsequence

... and similar calls for row-oriented AO tables. they should mirror the calls above ...

GirgerS: Turns out, contrary to what I've told, the last call is the exception to the `no pg_aoseg entries on the coordinator` statement. When running `CLUSTER ...` or `ALTER TABLE ... REPACK BY COLUMNS (...);` , we DO get an entry in the pg_aoseg table.

From what I've seen, this edge-case feels unnecessary, more like a flaw in the existing code. But I don't think that the patch needs to be changed (besides documenting everything, of course). 

Also, there is no vacuum error in such case, because the created `pg_aoseg` entry doesn't have any rows. `GetFastSequences` eventually gets called from there, but miraculously enough, it doesn't change `gp_fastsequence`, because there should be no values in the AO table on the coordinator.

GirgerS:
> From what I've seen, this edge-case feels unnecessary, more like a flaw in the existing code. But I don't think that the patch needs to be changed (besides documenting everything, of course).

I've looked into it a bit more, in 6.x `ALTER TABLE ... REPACK BY COLUMNS (...);` command doesn't exist, and `CLUSTER ...` gives an [error](https://github.com/GreengageDB/greengage/blob/ab73e971e1a15890dddf695fb534e7ba6ab8c27c/src/backend/commands/cluster.c#L331-L336). So as far as the migration is concerned, we shouldn't see these edge cases.

[test_repack_and_cluster.txt](https://github.com/user-attachments/files/27469207/test_repack_and_cluster.txt)

Also, there is an additional information about the difference between versions in the commit 5778f31124d5737e18d76f8d06f07f96c31c8be7.

whitehawk (dismissed):

src/bin/pg_upgrade/pg_upgrade.c
```diff
@@ -231,15 +231,48 @@ main(int argc, char **argv)
...
+		 *
+		 * 'ALTER TABLE ... REPACK BY COLUMNS (...)' and 'CLUSTER ...'
+		 * queries do create an oddly specific entry for RESERVED_SEGNO
+		 * segfile of the AO table with zero tuples. It looks like a flaw
+		 * in the logic, moreover, these commands are absent in 6.x.
+		 *
+		 * If any other pg_aoseg entry (that is, not similar to the one
+		 * created by the aforementioned queries) is present in the target
+		 * cluser, it would cause a failure to 'VACUUM' corresponding
+		 * AO table (see vacuum_ao.c , line 378).
```
whitehawk: I think its better not to mention exact line - if some code is added to the referenced file, the line number will be no more relevant. And I think it is better to make the entire comment a bit more compact - it pretty large now.

GirgerS: Maybe it is better move it to the PR description then?

whitehawk: I think instead of file:line you can refer to specific function.
And yes, maybe some part of the comment move to the PR description.

GirgerS: Should be done. Is is always possible to make the comment even smaller by not mentioning other tables (e.g. `gp_fastsequence`), but I think that it is worth explaining.

whitehawk (approved):

red1452 (approved):

PR #424, opened 2026-05-05
GG-336: [7.x] Adjust GP_HLL_ERROR_MARGIN according to the reasearch

GirgerS:
Adjust  GP_HLL_ERROR_MARGIN (#424)

In the ANALYZE command, GP_HLL_ERROR_MARGIN is used as a threshold
to determine if all values are distinct. It was set too low, leading to incorrect
statistics calculation and the possible generation of a suboptimal plan.

The problem has a probabilistic nature and does not appear on every run.
Therefore, a regression test was not provided, as it would be a flaky test.
However, even with the fix, due to the probabilistic nature, we cannot guarantee
100% accuracy.
Checks were performed by using the introduced unit test and running ANALYZE
several times to inspect n_distinct.

GG-336

(cherry picked from commit 462a17fdc5e1abf2bf2490a3d12588140bed5e50)

---

MERGE WITH REBASE

### DIFF
--- a/src/backend/commands/analyze.c
+++ b/src/backend/commands/analyze.c
@@ -138,11 +138,11 @@
 #include "tcop/utility.h"
 
 /*
- * For Hyperloglog, we define an error margin of 0.3%. If the number of
- * distinct values estimated by hyperloglog is within an error of 0.3%,
+ * For Hyperloglog, we define an error margin as dependency of requested standard 
+ * error. If the number of distinct values estimated by hyperloglog is within an error,
  * we consider everything as distinct.
  */
-#define GP_HLL_ERROR_MARGIN  0.003
+#define GP_HLL_ERROR_MARGIN  (DEFAULT_ERROR * 2)
 
 /* Fix attr number of return record of function gp_acquire_sample_rows */
 #define FIX_ATTR_NUM  3
--- a/src/backend/utils/hyperloglog/test/Makefile
+++ b/src/backend/utils/hyperloglog/test/Makefile
@@ -1,0 +1,7 @@
+subdir=src/backend/utils/hyperloglog
+top_builddir=../../../../..
+include $(top_builddir)/src/Makefile.global
+
+TARGETS=gp_hyperloglog
+
+include $(top_builddir)/src/backend/mock.mk
--- a/src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
+++ b/src/backend/utils/hyperloglog/test/gp_hyperloglog_test.c
@@ -1,0 +1,276 @@
+/*-------------------------------------------------------------------------
+ *
+ * gp_hyperloglog_test.c
+ *	  HyperLogLog algorithm test program
+ *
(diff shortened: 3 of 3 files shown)


GirgerS:
To see the difference before and after the patch, run the following test program from the parent ticket. It was necessary to add `setof` to the function return to make it work in 7.x.

```sql
set optimizer_analyze_root_partition=on;
set optimizer_analyze_midlevel_partition=on;

create extension pgcrypto;

drop table if exists test_financial_account_ada;
drop table if exists sampledata;

create table test_financial_account_ada (
    fin_account_cft_hk uuid not null,
    balance_flg bool null,
    account_no text null
)
with (
    appendonly=true,
    orientation=column,
    compresstype=zstd,
    compresslevel=1
)
distributed by (account_no)
partition by list(balance_flg)
          (
          partition bal values(true) with (appendonly='true', orientation='column', compresstype=zstd, compresslevel='1')
                    column fin_account_cft_hk encoding (compresstype=zstd, compresslevel=1, blocksize=32768)
                    column balance_flg encoding (compresstype=zstd, compresslevel=1, blocksize=32768)
                    column account_no encoding (compresstype=zstd, compresslevel=1, blocksize=32768),
          partition nobal values(false) with (appendonly='true', orientation='column', compresstype=zstd, compresslevel='1')
                    column fin_account_cft_hk encoding (compresstype=zstd, compresslevel=1, blocksize=32768)
                    column balance_flg encoding (compresstype=zstd, compresslevel=1, blocksize=32768)
                    column account_no encoding (compresstype=zstd, compresslevel=1, blocksize=32768)
          );
alter table test_financial_account_ada alter column account_no set statistics 1000;

--- prepare unique values in account_no
insert into test_financial_account_ada
select
    gen_random_uuid(),
    true,
    '913178106000063' || lpad(s.account_id::text, 9, '0') as account_number
from
    generate_series(1, 3000000) as s(account_id);

insert into test_financial_account_ada
select
    gen_random_uuid(),
    false,
    '913178106000063' || lpad(s.account_id::text, 9, '0') as account_number
from
    generate_series(3000001, 6000000) as s(account_id);

--- make sure value are unique

select count(distinct account_no) from test_financial_account_ada;

create table sampledata(step int, tablename text, n_distinct float);

create or replace function analyze_test(count int) returns setof void
as $$
declare
    i int;
begin
    for i in 1..count loop
      analyze test_financial_account_ada; -- this will fail

      insert into sampledata
        select i, t.tablename, t.n_distinct from pg_stats  t
          where t.tablename in ('test_financial_account_ada',
                          'test_financial_account_ada_1_prt_bal',
                          'test_financial_account_ada_1_prt_nobal')
          and t.attname = 'account_no';
      end loop;
    return;
end;
$$ language plpgsql
execute on master;

select analyze_test(25);
select * from sampledata order by 1,2;


drop table test_financial_account_ada;
drop function analyze_test(count int);
```

red1452 (approved):

ShapatinAA (approved):

PR #428, opened 2026-05-08
CI-5581: Port CI infrastructure from 7.x to 8.x

magf:
### GitHub Actions

Add `.github/workflows/greengage-ci.yml` with a full pipeline:
- Stages: build -> test (behave, regression, orca, resgroup, jit) -> upload
  - Tests run on PRs only
  - Upload runs on push to `next` or version tags (`8.*`)
- Add workflow README with configuration and backward-compatibility notes
  for `target_os_version`

### Directory layout

- Rename `arenadata/` -> `ci/` to reflect generic CI purpose
- Remove `arenadata/Dockerfile.ubuntu` and `arenadata/docker-compose.yaml`;
  replace with the new versions described below

### Dockerfile

Rewrite `ci/Dockerfile.ubuntu`:
- Add `ARG OS_VERSION` (default `22.04`) to support multiple Ubuntu versions
- Mirror substitution now uses `$UBUNTU_CODENAME` from `/etc/os-release`
  instead of hardcoded `jammy`
- Python pip flags adjusted automatically for Ubuntu 24.04
  (`--break-system-packages --ignore-installed`)
- Remove the separate `apt install` call for test packages - all packages
  are now installed in a single `README.Ubuntu.bash` invocation
- Add `COPY ... VERSION` to the `test` stage

### README.Ubuntu.bash

- Add `export DEBIAN_FRONTEND=noninteractive` at the top
- Consolidate all packages into one `apt-get install` call, adding:
  `fakeroot`, `iproute2`, `libipc-run-perl`, `libossp-uuid-dev`,
  `libpq-dev`, `lsof`, `protobuf-compiler`, `rsync`, `sudo`
- Sort the package list alphabetically

### docker-compose

Rewrite `ci/docker-compose.yaml`:
- Use YAML anchor `x-common-service` to eliminate repetition
- Add `ulimits: nofile: 65535`
- Add per-service log volumes (`/logs`) and `sqldump` volume on `cdw`

### Scripts

- `run_behave_tests.bash`:
  - Switch `docker-compose` (v1) -> `docker compose` (v2 plugin)
  - Exclude `gpexpand` until sqldump is available in CI
  - Add log collection via `behave_collect_logs.bash` when `$CI` is set
  - Fix `[[ $status > 0 ]]` -> `[[ $status -gt 0 ]]`
  - Add `-f pretty` formatter
- `behave_gpdb.bash`:
  - Remove `make_cluster` call, `gpdemo-env.sh` sourcing, and sqldump
    download for `gpexpand` - cluster lifecycle is now managed externally
- Add `ci/scripts/behave_collect_logs.bash`:
  - Collects `gpAdminLogs`, `log`, and `pg_log` directories into tarballs
    under `/logs` after each feature run
- Add `ci/scripts/run_resgroup_test.bash`:
  - Standalone script to run resource group regression tests inside a
    single Docker container

### Behave test infrastructure

- `environment.py`:
  - Replace per-scenario `concourse_cluster` fixture with a single
    per-feature `init_cluster` fixture guarded by `context.cluster_created`
  - Remove `gpconfig`, `gpstop`, `gpmovemirrors` from `tags_to_skip`
- `behave_utils/arenadata/` -> `behave_utils/ci/`; update all imports
- `ci/fixtures.py`:
  - Simplify cluster init - use `a standard local demo cluster is created`
    instead of a custom multi-host `gpinitsystem` invocation
- `mgmt_utils/steps/mgmt_utils.py`:
  - Make `coordinator_data_dir` lazy (set to `None` at module level,
    populated in `create_local_demo_cluster` and `_create_cluster`)
  - Export `PGPORT` and `COORDINATOR_DATA_DIRECTORY` inside the
    demo-cluster shell command
- `behave_utils/utils.py`:
  - Make `coordinator_data_dir` lazy
  - `check_database_is_running` returns `False` instead of raising when
    env vars are absent
- Mark `cross_subnet`, `gpssh`, and `gpexpand_icw_db_concourse` with `@skip`

### compile_gpdb.bash

Handle repos with no previous git tag:
- `git describe` now falls back to an empty string via `2>/dev/null || echo ''`
- Log range uses `${PREV_TAG:+$PREV_TAG..}HEAD` so a full `HEAD` log is
  produced when no prior tag exists

### compile_gpdb_remote_windows.bash

- Remove unused `GIT_TAG` variable and its `set GPDB_VERSION` injection
  into the remote build script
- Fix trailing whitespace

### dumpdb.bash / ic_gpdb.bash

- `dumpdb.bash`:
  - Create `sqldump/` with `mkdir -pm 777` before dumping
  - Skip `xz` compression when running in CI (`$CI` is set) to avoid
    unnecessary processing and simplify log collection
- `ic_gpdb.bash`:
  - Guard `chmod 777 sqldump` with a directory existence check to avoid
    failures on a clean workspace

### postmaster / fts_segment_reset

- Add `FAULT_INJECTOR` hook `backend_abort_handling` in `CleanupBackend`:
  - Guarded by `#ifdef FAULT_INJECTOR`, only active in `PM_RUN` state
  - Makes crash-recovery timing deterministic in tests
- Update `fts_segment_reset.sql` and its expected output:
  - Inject and wait on `backend_abort_handling` before triggering a
    segment PANIC
  - Eliminates a race condition that caused intermittent test failures

### SSL tests

- `001_ssltests.pl`:
  - Push `client_wrongperms` into `@keys` so the world-readable temporary
    key file is removed in the final cleanup phase
- `002_scram.pl`:
  - Rename the temporary client key to `client_scram_tmp.key` to avoid
    conflicts with parallel tests
  - Add `sslrootcert=invalid` to the connection string
  - Unlink the key explicitly after use

Task: [CI-5581](https://tracker.yandex.ru/CI-5581)


arenadata/scripts/behave_gpdb.bash
```diff
@@ -40,10 +31,6 @@ function _main() {
 				exit 1
 		fi
 
-		# Run inside a subshell so it does not pollute the environment after
-		# sourcing greenplum_path
-		time (make_cluster)
-
```
RekGRpth, line 46: I think this change should be a separate PR.

ci/readme.md
```diff
@@ -101,30 +101,30 @@ Feature files are located in `gpMgmt/test/behave/mgmt_utils`
...
 
 for Ubuntu:
 ```bash
 # Run all tests
-IMAGE=hub.adsw.io/library/gpdb8_u22:${BRANCH_NAME} bash arenadata/scripts/run_behave_tests.bash
+IMAGE=hub.adsw.io/library/gpdb8_u22:${BRANCH_NAME} bash ci/scripts/run_behave_tests.bash
 
 # Run specific features
-IMAGE=hub.adsw.io/library/gpdb8_u22:${BRANCH_NAME} bash arenadata/scripts/run_behave_tests.bash gpstart gpstop
+IMAGE=hub.adsw.io/library/gpdb8_u22:${BRANCH_NAME} bash ci/scripts/run_behave_tests.bash gpstart gpstop
```
RekGRpth, line 115: I don't think the `hub.adsw.io` host should be included in this PR.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -38,9 +38,7 @@
 from gppylib.operations.package import linux_distribution_id, linux_distribution_version
 from gppylib.commands.gp import get_coordinatordatadir
 
-coordinator_data_dir = gp.get_coordinatordatadir()
-if coordinator_data_dir is None:
-    raise Exception('Please set COORDINATOR_DATA_DIRECTORY in environment')
+coordinator_data_dir = None
```
RekGRpth, line 41: I think these changes to this file should be in a separate PR.

gpMgmt/test/behave/mgmt_utils/environment.py
```diff
@@ -19,11 +19,16 @@ def before_all(context):
...
 def before_feature(context, feature):
     # we should be able to run gpexpand without having a cluster initialized
-    tags_to_skip = ['gpexpand', 'gpaddmirrors', 'gpstate', 'gpmovemirrors',
-                    'gpconfig', 'gpssh-exkeys', 'gpstop', 'gpinitsystem', 'cross_subnet']
+    tags_to_skip = ['gpexpand', 'gpaddmirrors', 'gpstate',
+                    'gpssh-exkeys', 'gpinitsystem', 'cross_subnet']
     if set(context.feature.tags).intersection(tags_to_skip):
         return
 
+    if not hasattr(context, "cluster_created"):
```
RekGRpth, line 27: I think the changes to this file should be in a separate PR.

gpMgmt/test/behave_utils/ci/fixtures.py
```diff
@@ -0,0 +1,10 @@
+from behave import fixture
+
+
+@fixture
+def init_cluster(context):
+    context.execute_steps(u"""
+        Given the database is not running
+        And the user runs command "rm -rf ~/gpAdminLogs/gpinitsystem*"
+        And a standard local demo cluster is created
+    """)
```
RekGRpth, line 10: I think the contents of this file should be moved unchanged from `gpMgmt/test/behave_utils/arenadata/fixtures.py`.

gpMgmt/test/behave_utils/utils.py
```diff
@@ -22,10 +22,7 @@
 PARTITION_START_DATE = '2010-01-01'
 PARTITION_END_DATE = '2013-01-01'
 
-coordinator_data_dir = get_coordinatordatadir()
-if coordinator_data_dir is None:
-    raise Exception('COORDINATOR_DATA_DIRECTORY is not set')
-
+coordinator_data_dir = None
```
RekGRpth, line 25: I think the changes to this file should be in a separate PR.

src/backend/postmaster/postmaster.c
```diff
@@ -3920,6 +3920,21 @@ CleanupBackend(int pid,
...
+			 * Code below is for test purposes only. If a fault is set, it
+			 * will formally be completed without doing anything.
+			 *
+			 * Additional note: This check of pmState == PM_RUN is just extra
+			 * safety measurments. If we try to access this code not from
+			 * PM_RUN the behaviour can be unpredictable.
+			 */
+			SIMPLE_FAULT_INJECTOR("backend_abort_handling");
+		}
+#endif
```
RekGRpth, line 3939: I think the changes to this file should be in a separate PR.

src/test/isolation2/sql/fts_segment_reset.sql
```diff
@@ -26,12 +26,34 @@ from gp_segment_configuration where role = 'p' and content = 0;
...
+-- We need to set fault at the handler of abort signal from backend
+-- so then we can wait until it is triggered. We need to do this because
+-- sometimes process of calling abort() (which is issued by the injection
+-- of 'panic' fault in the next lines of this test) and abortion itself can
+-- take a long time to complete. So, sometimes this test can give unexpected
+-- results, when you think that commands sended after issuing panic fault
+-- should fail, but they are being executed without any errors.
+select gp_inject_fault('backend_abort_handling', 'skip', dbid)
+from gp_segment_configuration where role = 'p' and content = 0;
+
```
RekGRpth, line 38: I think the changes to this file should be in a separate PR.

src/test/ssl/t/001_ssltests.pl
```diff
@@ -48,9 +48,11 @@
 
 # Also make a copy of that explicitly world-readable.  We can't
 # necessarily rely on the file in the source tree having those
-# permissions.
+# permissions.  Add it to @keys to include it in the final clean
+# up phase.
 copy("ssl/client.key", "ssl/client_wrongperms_tmp.key");
 chmod 0644, "ssl/client_wrongperms_tmp.key";
+push @keys, 'client_wrongperms';
```
RekGRpth, line 59: I think the changes to this file should be in a separate PR.

src/test/ssl/t/002_scram.pl
```diff
@@ -89,14 +89,20 @@
...
 test_connect_fails(
-	"sslcert=ssl/client.crt sslkey=ssl/client_tmp.key hostaddr=$SERVERHOSTADDR",
+	"sslcert=ssl/client.crt sslkey=$client_tmp_key sslrootcert=invalid hostaddr=$SERVERHOSTADDR",
 	"dbname=certdb user=ssltestuser channel_binding=require",
 	qr/channel binding required, but server authenticated client without channel binding/,
 	"Cert authentication and channel_binding=require");
 
+# clean up
+unlink($client_tmp_key);
+
```
RekGRpth, line 107: I think the changes to this file should be in a separate PR.

.github/workflows/greengage-ci.yml
```diff
@@ -0,0 +1,160 @@
...
+      fail-fast: false
+      matrix:
+        target_os: [ubuntu]
+    permissions:
+      contents: read  # Explicit for default behavior
+      packages: write # Required for GHCR access
+      actions: write  # Required for artifact upload
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-build.yml@v33
+    with:
+      version: 8
```
bandetto, line 28: I think the README in this directory describes a different workflow than what's actually implemented here. It documents an `env:` block with `VERSION`/`OS_MATRIX`, a `generate-matrix` job, a `package` job, and a separate `Greengage release` workflow. In this file the version and matrix are hardcoded, `package` is commented out, and `generate-matrix`/release are absent. Maybe we should either align the workflow with the README or update the README, since they diverge quite a bit right now.

ci/Dockerfile.ubuntu
```diff
@@ -0,0 +1,75 @@
...
+COPY README.Ubuntu.bash ./
+
+RUN set -eux; \
+    . /etc/os-release; \
+    sed -i \
+      -e "s|http://archive.ubuntu.com/ubuntu/  ${UBUNTU_CODENAME} |${UBUNTU_MAIN_MIRROR}/ ${UBUNTU_CODENAME} |g" \
+      -e "s|http://archive.ubuntu.com/ubuntu/  ${UBUNTU_CODENAME}-updates |${UBUNTU_UPDATES_MIRROR}/ ${UBUNTU_CODENAME}-updates |g" \
+      -e "s|http://archive.ubuntu.com/ubuntu/  ${UBUNTU_CODENAME}-backports |${UBUNTU_BACKPORTS_MIRROR}/ ${UBUNTU_CODENAME}-backports |g" \
+      -e "s|http://security.ubuntu.com/ubuntu/ ${UBUNTU_CODENAME}-security |${UBUNTU_SECURITY_MIRROR}/ ${UBUNTU_CODENAME}-security |g" \
+      /etc/apt/sources.list; \
```
bandetto, line 21: I think the mirror substitution may not be taking effect here owo. The `sed` patterns use a double space (`ubuntu/  ${UBUNTU_CODENAME}`), while the stock `sources.list` uses a single space, so the expression likely won't match anything. Also, on Ubuntu 24.04 the sources moved to the deb822 format (`/etc/apt/sources.list.d/ubuntu.sources`) and `/etc/apt/sources.list` is effectively empty, so the mirror override becomes a no-op on 24.04. Since multi-version support is the goal of this PR, it might be worth handling this case.

ci/scripts/run_resgroup_test.bash
```diff
@@ -0,0 +1,39 @@
...
+exitcode=1
+
+cd /home/gpadmin/
+ssh-keygen -A
+/usr/sbin/sshd
+source gpdb_src/concourse/scripts/common.bash
+install_and_configure_gpdb
+gpdb_src/concourse/scripts/setup_gpadmin_user.bash
+make_cluster
+chmod -R 777 /sys/fs/cgroup/{memory,cpu,cpuset}
```
bandetto, line 24: These paths assume cgroup v1 (`/sys/fs/cgroup/{memory,cpu,cpuset}`). Ubuntu 24.04 defaults to cgroup v2 with a unified hierarchy, so these paths won't exist and the resgroup test would likely fail there. This is the same multi-version scenario this PR targets, so I think we should account for it.

README.Ubuntu.bash
```diff
@@ -1,48 +1,57 @@
...
 	pkg-config \
+	protobuf-compiler \
 	python3-dev \
 	python3-pip \
-	python3-psycopg2 \
 	python3-psutil \
+	python3-psycopg2 \
 	python3-yaml \
+	rsync \
+	sudo \
```
bandetto, line 54: I think `openjdk-17-jdk` (and `python3.11`) was dropped here owo - it used to be installed in the test stage of the Dockerfile. If ORCA or the test suites need Java at runtime, this could be a regression. Maybe we should double-check that the orca-tests job still has a JDK available.

src/backend/postmaster/postmaster.c
```diff
@@ -3920,6 +3920,21 @@ CleanupBackend(int pid,
...
+		if (pmState == PM_RUN)
+		{
+			/*
+			 * Code below is for test purposes only. If a fault is set, it
+			 * will formally be completed without doing anything.
+			 *
+			 * Additional note: This check of pmState == PM_RUN is just extra
+			 * safety measurments. If we try to access this code not from
+			 * PM_RUN the behaviour can be unpredictable.
+			 */
```
bandetto, line 3936: This is more of a scope question: this fault injector, along with the `fts_segment_reset` and SSL test changes (`001_ssltests.pl`, `002_scram.pl`), are PostgreSQL-side patches rather than test-infrastructure changes. I'm not sure why they're bundled into a CI port PR. I think it might be cleaner to revert them here and handle them in a separate PR so this one stays focused on infrastructure. What do you think?

magf:
@Stolb27, branch `next` not ready for CI implementation.
Make it works locally before implement automation.

PR #429, opened 2026-05-12
Forbid table inheritance from a partitioned table

whitehawk:
Forbid table inheritance from a partitioned table

Problem description:
If some table was inherited from a leaf of a partitioned table, the leaf table
wasn't properly redistributed by the gpexpand.

Root cause:
gpexpand applies 'ALTER TABLE ... EXPAND PARTITION PREPARE' for all partitioned
tables at its 1st phase, and applies 'ALTER TABLE ... SET WITH (REORGANIZE=true)
DISTRIBUTED ...' for all leaf tables at its 2nd stage. 'ALTER TABLE ...
EXPAND PARTITION PREPARE' for all non-leaf tables only changes numsegments
of the current policy to expanded cluster size. The check for non-leaf condition
is done via 'relhassubclass' flag of 'pg_class' record (refer to
'ATExecExpandPartitionTablePrepare()' function).
Partitioned tables in 6x are implemented with the use of table inheritance.
When some table was inherited from a leaf table, the 'relhassubclass' flag for
the leaf table was set to 'true'. Therefore, it was treated as non-leaf table.
If it was hash-distributed, the distribution policy didn't change, only the
segment count was updated (while, for all hash-distributed leaves the
distribution is expected to be changed to 'randomly' distributed).
And, as the 'relhassubclass' flag was true, gpexpand skipped 'ALTER TABLE ...
SET WITH (REORGANIZE=true) DISTRIBUTED ...' operation for the leaf table. This
operation is supposed to perform proper redistribution of the table's data
according to the new segment configuration.
As result, after the gpexpand finished its work, the leaf table had updated
distribution policy, but the data remained to reside on the old segments.
As a consequence, if we tried to insert data from the partitioned table to some
other table with the same distribution policy, the planner created a plan
without motions, and the data was inserted into the wrong segments, spoiling the
other table as well.

Fix:
Mixing table partitioning and table inheritance is error prone, especially in
Greengage 6x, where partitioning uses much of the inheritance logic. In
Greengage 7x and in vanilla Postgres the mixing is already forbidden.
Therefore, we follow the same way in 6x - this patch forbids table inheritance
from a partitioned table.


src/test/regress/sql/inherit.sql
```diff
@@ -631,3 +631,56 @@ CREATE TABLE child2(t text) DISTRIBUTED BY (t);
 ALTER TABLE child2 INHERIT parent;
 
 DROP TABLE child, child2, parent;
+
+-- Test that partitioned table can't inherit a parent table
+DROP TABLE IF EXISTS test_table, test_table_partitioned;
```
red1452, line 637: It is better to use ignore block for commands with "IF EXISTS".

src/backend/commands/tablecmds.c
```diff
@@ -2253,6 +2253,21 @@ MergeAttributes(List *schema, List *supers, char relpersistence, bool isPartitio
...
 
+		/*
+		 * We do not allow partitioned tables and partitions to participate in
+		 * regular inheritance.
+		 */
+		if (!isPartitioned && rel_is_partitioned(RelationGetRelid(relation)))
+			ereport(ERROR,
+					(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+					 errmsg("cannot inherit from partitioned table \"%s\"",
+							RelationGetRelationName(relation))));
```
KnightMurloc, line 2264: Why it's not safe to inherit from partitioned table (root)?

whitehawk: Because it raises many uncertainties - should this inherited table be involved in partition pruning, should the inherited table be scaned when the partitioned table is scaned, should insert operation consider the inherited table, etc.... So, in Postgres and in 7x inheritance from root is forbiden, and we do in this patch the same.

src/test/regress/sql/inherit.sql
whitehawk: added

KnightMurloc (approved):

red1452 (approved):

PR #432, opened 2026-05-12
GG-352: Fix external web table get negative cost estimates

riogor:
Postgres query optimizer set negative costs for plans with non-distributed
(ON MASTER) external web tables.
Optimizer didn't set proper values from pg_class and used default -1 value for
tuple count.

Make optimizer use default values from pg_class.
Fix affected tests.


dkovalev1:
It could be useful if we had a test which demonstrates original issue and the fact it has been fixed.

riogor:
Hmm
Maybe something simple, like this:
```sql
CREATE EXTERNAL WEB TABLE ext_t1(c1 int) EXECUTE 'seq 1 10' ON MASTER FORMAT 'text';

EXPLAIN SELECT * FROM ext_t1;
                             QUERY PLAN                              
---------------------------------------------------------------------
 External Scan on ext_t1  (cost=0.00..11000.00 rows=1000000 width=4)
 Optimizer: Postgres query optimizer
(2 rows)
```
Without fix it produces a negative cost plan, as expected:
```sql
CREATE EXTERNAL WEB TABLE ext_t1(c1 int) EXECUTE'seq 1 10' ON MASTER FORMAT 'text';

EXPLAIN SELECT * FROM ext_t1;
                         QUERY PLAN                         
------------------------------------------------------------
 External Scan on ext_t1  (cost=0.00..-0.01 rows=1 width=4)
 Optimizer: Postgres query optimizer
(2 rows)
```

Not sure how to show the fact that there was an original issue.

dkovalev1:
> ```sql
> CREATE EXTERNAL WEB TABLE ext_t1(c1 int) EXECUTE 'seq 1 10' ON MASTER FORMAT 'text';
> 
> EXPLAIN SELECT * FROM ext_t1;
>                              QUERY PLAN                              
> ---------------------------------------------------------------------
>  External Scan on ext_t1  (cost=0.00..11000.00 rows=1000000 width=4)
>  Optimizer: Postgres query optimizer
> (2 rows)
> ```

Thanks! I suggest to add this to the test suite, to some sutable existing test. IMO, external_table is a good candidate.

src/test/regress/input/external_table.source
```diff
@@ -3643,3 +3643,14 @@ FROM ext_subplan_t1;
...
 DROP EXTERNAL TABLE ext_subplan_t2;
+
+-- Test non-negative plan costs on non-distributed external web tables
+
+CREATE EXTERNAL WEB TABLE ext_t1(c1 int) EXECUTE 'seq 1 10' ON MASTER FORMAT 'text';
+
+EXPLAIN
+SELECT * FROM ext_t1;
+
+SELECT * FROM ext_t1;
```
dkovalev1: IMHO this statement is useless for the test and can be safely removed.

riogor: Removed in [b54fd32](https://github.com/GreengageDB/greengage/pull/432/commits/b54fd32391e89e099d7c991845bf8add6f06b10a)

dkovalev1 (approved):

red1452 (commented):
Why is cost smaller at 7x than at your patch? Your value: `cost=29296.95..42592.86 rows=1000001` and at 7x:  `cost=2.08..2.10 rows=2`. Is it correct difference?

riogor:
Because 7.x converts all external tables to foreign tables and uses gp_exttable_fdw to work with them. Broadcom has a [doc page](https://techdocs.broadcom.com/us/en/vmware-tanzu/data-solutions/tanzu-greenplum/7/greenplum-database/admin_guide-external-map_ext_to_foreign.html) about it.

And so, plan in 7.x uses foreign scan instead of external scan, with different cost calculation functions and different default values for tuple and page count in pg_class.

In 6.x default values for external tables are being set [here](https://github.com/GreengageDB/greengage/blob/c20b995fe4c22173b2566ecd44c997cd902dc0b1/src/backend/catalog/heap.c#L1220-L1225).

red1452 (approved):

PR #435, opened 2026-05-14
GG-465: Fix issue with SET LOCAL inside PL functions (7.x)

VoidZeroNull0:
What happened?
Local GUCs (SET LOCAL) get lost inside PLPG functions (see test case for more
details), thus writing commands may have unexpected results or fail.

Why it happens?
It happens because SET LOCAL is executed in a separate transaction, because no
DTX (2PC) is set up, so its effect is popped.

How do we fix this?
Let's make all SET LOCAL commands on segments be just SET, so that they will
last until master synchronizes them using previously saved value
(gp_guc_restore_list). Synchronization happens on next master's transaction or
when transaction control statement is met (COMMIT, ROLLBACK). Specifically,
syncing is handled inside AtEOXact_SPI(). Note - on master, SET LOCAL is
still executed as SET LOCAL.

Alternate solutions?
DTX transaction context could be set up on the SET step, so that all subsequent
writing commands get inside this new transaction and thus, get required GUC.

Tests?
Test case to check consistency and correct passage of local GUCs inside DO was
developed.

Ticket: GG-465

### DIFF
--- a/src/backend/executor/spi.c
+++ b/src/backend/executor/spi.c
@@ -308,6 +308,9 @@ _SPI_commit(bool chain)
 				DispatchSyncPGVariable(gconfig);
 			}
 			CommitTransactionCommand();
+			
+			list_free(gp_guc_restore_list);
+			gp_guc_restore_list = NIL;
 		}
 
 		/* Immediately start a new transaction */
@@ -417,6 +420,9 @@ _SPI_rollback(bool chain)
 				DispatchSyncPGVariable(gconfig);
 			}
 			CommitTransactionCommand();
+
+			list_free(gp_guc_restore_list);
+			gp_guc_restore_list = NIL;
 		}
 
 		/* Immediately start a new transaction */
--- a/src/backend/executor/spi.c
+++ b/src/backend/executor/spi.c
@@ -412,7 +412,7 @@ _SPI_rollback(bool chain)
 
 			foreach(lc, gp_guc_restore_list)
 			{
-				struct config_generic *gconfig = (struct config_generic *) lfirst(lc);
+				struct config_generic *gconfig = lfirst(lc);
 
 				DispatchSyncPGVariable(gconfig);
 			}
--- a/src/backend/executor/spi.c
+++ b/src/backend/executor/spi.c
@@ -293,6 +293,23 @@ _SPI_commit(bool chain)
 		/* Do the deed */
 		CommitTransactionCommand();
 
+		/* Resync GUCs */
+		if (Gp_role == GP_ROLE_DISPATCH && gp_guc_restore_list)
+		{
+			StartTransactionCommand();
+			if (chain)
+				RestoreTransactionCharacteristics();
+			ListCell   *lc;
+
+			foreach(lc, gp_guc_restore_list)
+			{
+				struct config_generic *gconfig = (struct config_generic *) lfirst(lc);
+
+				DispatchSyncPGVariable(gconfig);
+			}
+			CommitTransactionCommand();
+		}
+
 		/* Immediately start a new transaction */
 		StartTransactionCommand();
 		if (chain)
@@ -384,6 +401,23 @@ _SPI_rollback(bool chain)
 
 		/* Do the deed */
 		AbortCurrentTransaction();
+
+		/* Resync GUCs */
+		if (Gp_role == GP_ROLE_DISPATCH && gp_guc_restore_list)
+		{
+			StartTransactionCommand();
+			if (chain)
+				RestoreTransactionCharacteristics();
+			ListCell   *lc;
+
+			foreach(lc, gp_guc_restore_list)
+			{
+				struct config_generic *gconfig = (struct config_generic *) lfirst(lc);
+
+				DispatchSyncPGVariable(gconfig);
+			}
+			CommitTransactionCommand();
+		}
 
 		/* Immediately start a new transaction */
 		StartTransactionCommand();
--- a/src/backend/executor/spi.c
+++ b/src/backend/executor/spi.c
@@ -303,7 +303,7 @@ _SPI_commit(bool chain)
 
 			foreach(lc, gp_guc_restore_list)
 			{
-				struct config_generic *gconfig = (struct config_generic *) lfirst(lc);
+				struct config_generic *gconfig = lfirst(lc);
 
 				DispatchSyncPGVariable(gconfig);
 			}
--- a/src/backend/executor/spi.c
+++ b/src/backend/executor/spi.c
@@ -293,26 +293,6 @@ _SPI_commit(bool chain)
 		/* Do the deed */
 		CommitTransactionCommand();
 
-		/* Resync GUCs */
-		if (Gp_role == GP_ROLE_DISPATCH && gp_guc_restore_list)
-		{
-			StartTransactionCommand();
-			if (chain)
-				RestoreTransactionCharacteristics();
-			ListCell   *lc;
-
-			foreach(lc, gp_guc_restore_list)
-			{
-				struct config_generic *gconfig = lfirst(lc);
-
-				DispatchSyncPGVariable(gconfig);
-			}
-			CommitTransactionCommand();
-			
-			list_free(gp_guc_restore_list);
-			gp_guc_restore_list = NIL;
-		}
-
 		/* Immediately start a new transaction */
 		StartTransactionCommand();
 		if (chain)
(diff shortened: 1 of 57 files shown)


RekGRpth:
Why did tests fail?

src/backend/utils/misc/guc.c
```diff
@@ -7089,6 +7089,16 @@ set_config_option(const char *name, const char *value,
...
 
+	/*
+	 * Make SET LOCAL just SET when executing on segments.
+	 * It is needed for correct passage of local GUCs as they might be discarded.
+	 * Anyway, master executes SET as LOCAL, so on next transaction
+	 * GUC will resynchronize and everything will be back in place.
+	 */
+	if (Gp_role == GP_ROLE_EXECUTE &&
+		action == GUC_ACTION_LOCAL)
+		action = GUC_ACTION_SET;
```
RekGRpth, line 7098: pgindent your code

src/test/regress/expected/plpgsql_optimizer.out
```diff
@@ -5329,3 +5329,38 @@ end;
 $$;
 ERROR:  value for domain plpgsql_arr_domain violates check constraint "plpgsql_arr_domain_check"
 CONTEXT:  PL/pgSQL function inline_code_block line 4 at assignment
+-- Test consistency and passage of SET LOCAL GUC to writing commands
+--start_ignore
+drop schema if exists s cascade;
+--end_ignore
+create schema s;
```
RekGRpth, line 5333: Remove ignore block from output

src/test/regress/sql/plpgsql.sql
```diff
@@ -4150,3 +4150,36 @@ begin
...
+select c.gp_segment_id
+from gp_dist_random('pg_class') c
+join pg_namespace n on n.oid = c.relnamespace
+where c.relname = 'test_table'
+  and c.relkind = 'r'
+  and n.nspname = 's'
+group by c.gp_segment_id
+order by c.gp_segment_id;
+
+drop schema s cascade;
```
RekGRpth: add empty line

src/test/regress/sql/plpgsql.sql
```diff
@@ -4150,3 +4150,36 @@ begin
...
+
+--check existence on segments
+select c.gp_segment_id
+from gp_dist_random('pg_class') c
+join pg_namespace n on n.oid = c.relnamespace
+where c.relname = 'test_table'
+  and c.relkind = 'r'
+  and n.nspname = 's'
+group by c.gp_segment_id
+order by c.gp_segment_id;
```
RekGRpth: use union all to combine two query into one

src/backend/utils/misc/guc.c
VoidZeroNull0: Done.

src/test/regress/expected/plpgsql_optimizer.out
VoidZeroNull0: Done.

src/test/regress/sql/plpgsql.sql
VoidZeroNull0: Done.

src/test/regress/sql/plpgsql.sql
VoidZeroNull0: Done.

src/test/regress/sql/plpgsql.sql
```diff
@@ -4150,3 +4150,36 @@ begin
...
+
+union all
+
+--check existence on segments
+select c.gp_segment_id from gp_dist_random('pg_class') c
+join pg_namespace n on n.oid = c.relnamespace
+where c.relname = 'test_table'
+  and c.relkind = 'r'
+  and n.nspname = 's'
+order by gp_segment_id;
```
RekGRpth: I mean something like
```suggestion
with c as (
    select gp_segment_id, relname, relkind, relnamespace from pg_class
    union all
    select gp_segment_id, relname, relkind, relnamespace from gp_dist_random('pg_class')
) select gp_segment_id from c
join pg_namespace n on n.oid = c.relnamespace
where c.relname = 'test_table' and c.relkind = 'r' and n.nspname = 's'
```

VoidZeroNull0: Ok!

KnightMurloc:
Value is failed to restore if we commit transaction in do block:
This PR:
```
DO $$
DECLARE
    _result text;
BEGIN
    select string_agg(setting, '|') into _result from gp_dist_random('pg_settings') where name = 'search_path';
    RAISE NOTICE 'before %', _result;
    set local search_path = '"$user",public,test';
    select string_agg(setting, '|') into _result from gp_dist_random('pg_settings') where name = 'search_path';
    RAISE NOTICE 'after  %', _result;
    commit;
    select string_agg(setting, '|') into _result from gp_dist_random('pg_settings') where name = 'search_path';
    RAISE NOTICE 'after  %', _result;
END
$$ LANGUAGE plpgsql;

NOTICE:  before "$user", public|"$user", public|"$user", public
NOTICE:  after  """$user"",public,test"|"""$user"",public,test"|"""$user"",public,test"
NOTICE:  after  """$user"",public,test"|"""$user"",public,test"|"""$user"",public,test"
DO
```
Postgres 12:
```
NOTICE:  before "$user", public
NOTICE:  after  """$user"",public,test"
NOTICE:  after  "$user", public
DO
```

VoidZeroNull0:
> Value is failed to restore if we commit transaction in do block

This also seems to be the issue pre-patch judging by debugging. During `DO` block execution program does not seem to enter `send_guc_to_QE()` (which syncs GUCs among QD and QEs) or react in any way to `COMMIT` inside. I guess, it is because `DO` block logically treated as a separate transaction with no control statements that could come. Anyway, going to research it further.

KnightMurloc:
> I guess, it is because DO block logically treated as a separate transaction with no control statements that could come. Anyway, going to research it further.

By the way, Postgres allow transaction control (commit/rollback) inside DO block since Postgres 11.

src/test/regress/sql/plpgsql.sql
VoidZeroNull0: Done!

VoidZeroNull0:
>> Value is failed to restore if we commit transaction in do block

> Anyway, going to research it further.

Have implemented GUCs recovery in 74c6bea. Had to modify relevant SPI commit and rollback commands.

RekGRpth:
> > > Value is failed to restore if we commit transaction in do block
> 
> > Anyway, going to research it further.
> 
> Have implemented GUCs recovery in [74c6bea](https://github.com/GreengageDB/greengage/commit/74c6bea18b6cec701f90a55603fa7d2faa0b688a). Had to modify relevant SPI commit and rollback commands.

add tests for these cases

VoidZeroNull0:
> add tests for these cases

They were added:

https://github.com/GreengageDB/greengage/blob/0da35f3669e670992e40af7ea5108358f2a0508b/src/test/regress/expected/plpgsql.out#L5362-L5391

src/backend/executor/spi.c
```diff
@@ -293,6 +293,23 @@ _SPI_commit(bool chain)
...
+		if (Gp_role == GP_ROLE_DISPATCH && gp_guc_restore_list)
+		{
+			StartTransactionCommand();
+			if (chain)
+				RestoreTransactionCharacteristics();
+			ListCell   *lc;
+
+			foreach(lc, gp_guc_restore_list)
+			{
+				struct config_generic *gconfig = (struct config_generic *) lfirst(lc);
```
RekGRpth:
```suggestion
				struct config_generic *gconfig = lfirst(lc);
```

src/backend/executor/spi.c
```diff
@@ -385,6 +402,23 @@ _SPI_rollback(bool chain)
...
```
RekGRpth:
```suggestion
				struct config_generic *gconfig = lfirst(lc);
```

src/backend/executor/spi.c
VoidZeroNull0: Isn't implicit casting kind of unsafe?

RekGRpth: A pointer to void can be cast to any pointer. By explicitly casting, you hide potential errors where a pointer to void is converted to a pointer to a specific type. Therefore, you should not explicitly cast pointers to void.

VoidZeroNull0: Understood, thanks for noticing!

src/backend/executor/spi.c
```diff
@@ -293,6 +293,26 @@ _SPI_commit(bool chain)
...
+				RestoreTransactionCharacteristics();
+			ListCell   *lc;
+
+			foreach(lc, gp_guc_restore_list)
+			{
+				struct config_generic *gconfig = lfirst(lc);
+
+				DispatchSyncPGVariable(gconfig);
+			}
+			CommitTransactionCommand();
```
KnightMurloc: Why do we need additional transaction?

src/backend/executor/spi.c
```diff
@@ -385,6 +405,26 @@ _SPI_rollback(bool chain)
...
+
+			list_free(gp_guc_restore_list);
+			gp_guc_restore_list = NIL;
+		}
```
KnightMurloc: Can we avoid duplication by place it in `AtEOXact_SPI`?

VoidZeroNull0: Good idea! Done.

src/backend/executor/spi.c
VoidZeroNull0: I thought it was needed cause we are inbetween transactions, but turns out that no. Thanks for noticing!

RekGRpth:
actualize pr descr about SPI

RekGRpth (approved):

KnightMurloc (approved):

PR #438, opened 2026-05-18
Add checks into gpcheckcat for usual inheritance from partitioned tables

whitehawk:
Add checks into gpcheckcat for usual inheritance from partitioned tables

Inheriting a table from a partitioned table can cause problems - for example,
data can be not redistributed in a partitioned table after gpexpand.
This patch adds checks into gpcheckcat that allow to detect if such cases of
inheritance have already happened.

Test are not added as the capability to inherit a table from a partitioned
table is forbidden in a separate patch.


whitehawk:
SQL command that can be used for test (Note: need to test without db987c3ae04fcfe3bf0d5cbc0e3634ed38802e9e, as it forbids such inheritance):
```
CREATE TABLE test_table (
    id INT,
    data_date DATE
)
DISTRIBUTED BY (id)
PARTITION BY RANGE (data_date)
SUBPARTITION BY LIST (id)
(
    PARTITION p_2023 START ('2023-01-01') END ('2024-01-01')
    (
        SUBPARTITION p_2023_ids VALUES (1,2,3,4,5)
    )
);

create table inh_det_root (like test_table)
inherits (test_table);
create table inh_det_interior (like test_table_1_prt_p_2023)
inherits (test_table_1_prt_p_2023);
create table inh_det_leaf (like test_table_1_prt_p_2023_2_prt_p_2023_ids)
inherits (test_table_1_prt_p_2023_2_prt_p_2023_ids);
```

red1452 (approved):

KnightMurloc (approved):

PR #440, opened 2026-05-18
ADBDEV-9854: metadata exchange communication channel

dkovalev1:
This patch is to add possibility of sending some arbitrary binary data
(metadata) from backends to the frontend. In out particular case it could
be segment and coordinator.

The data is sent using direct libpq messages with M as a message type.
Format of the message is pretty simple - just message length and message
body.

To send messages there is a new API function pq_metadatasend which can
be called from an extension.

The messages are processed on the frontend (coordinator in our case) side,
and collected into a linked list ggMetadataList in the cdbconn.c.

The code for message processing is inspired by the notice processing
mechanism which works in a similar way. To lower the number of allocations, 
flexible array is used.

To test the new API there is a test extension and a code for retrieving
messages implemented as well. This code uses methods PQMetadataWalk,
PQgetNextMetadata, PQgetMetadata and PQCleanMetadata. These methods
can be called from an extension as well.

New metadata API is available only at the backend (no FRONTEND macro defined).

ADBDEV-9854


RekGRpth:
> To test the new API `the is` a test extension

there is?

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,167 @@ forwardQENotices(void)
...
+
+/* List of ggMetadataQueue */
+static List *ggMetadataQueues = NIL;
+
+static ggMetadataQueue *PQMetadataFindQueue(ggMetadataQueueId queue_id)
+{
+	ListCell *lc;
+	foreach(lc, ggMetadataQueues)
+	{
+		ggMetadataQueue *queue = (ggMetadataQueue *) lfirst(lc);
```
RekGRpth:
```suggestion
		ggMetadataQueue *queue = lfirst(lc);
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,167 @@ forwardQENotices(void)
...
+		if (queue->id == queue_id)
+			return queue;
+	}
+	return NULL;
+}
+
+static void
+MPPmetadataReceiver(void *arg, ggMetadataChunk *metadata_chunk, ggMetadataQueueId queue_id)
+{
+	SegmentDatabaseDescriptor *segdbDesc = (SegmentDatabaseDescriptor *)arg;
```
RekGRpth:
```suggestion
	SegmentDatabaseDescriptor *segdbDesc = arg;
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,167 @@ forwardQENotices(void)
...
+extern void 		AtCommit_MetadataQueues();
+
+/* Callback executed at transaction end */
+static void
+PQDeleteMetadataQueues()
+{
+	while (ggMetadataQueues)
+	{
+		ListCell *lc = list_head(ggMetadataQueues);
+		ggMetadataQueue *queue = (ggMetadataQueue *) lfirst(lc);
```
RekGRpth:
```suggestion
		ggMetadataQueue *queue = lfirst(lc);
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,167 @@ forwardQENotices(void)
...
+
+ggMetadataQueueId
+PQMetadataNextQueueId()
+{
+	static ggMetadataQueueId queueId = 0;
+	return queueId++;
+}
+
+extern void 		AtAbort_MetadataQueues();
+extern void 		AtCommit_MetadataQueues();
```
RekGRpth: why not in some header?

src/include/libpq/libpq.h
```diff
@@ -75,4 +75,31 @@ extern void secure_close(Port *port);
...
+ */
+extern void pq_metadatasend(const void *data, size_t len, int32 queue_id);
+
+typedef void *ggMetadataChunkIterator;
+
+typedef struct ggMetadataDescriptor
+{
+	int metadataLen;
+	int    segindex;        /* source segment */
+	void *data;
```
RekGRpth: pgindent

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2365,3 +2372,48 @@ build_startup_packet(const PGconn *conn, char *packet,
 
 	return packet_len;
 }
+
+static int
+pgGetMetadataMessage(PGconn *conn, int length)
```
RekGRpth, line 2386: add comment before function like `pqGetErrorNotice3`

src/interfaces/libpq/libpq-fe.h
```diff
@@ -618,6 +618,23 @@ extern int	pg_char_to_encoding(const char *name);
 extern const char *pg_encoding_to_char(int encoding);
 extern int	pg_valid_server_encoding_id(int encoding);
 
+typedef struct ggMetadataChunk
+{
+	struct ggMetadataChunk *next;
```
RekGRpth: why not use `slist_node`?

src/interfaces/libpq/libpq-fe.h
```diff
@@ -618,6 +618,23 @@ extern int	pg_char_to_encoding(const char *name);
...
+	int    segindex;      /* source segment */
+	int    metadataLen;    /* Length of metadata buffer */
+	char   payload[];
```
RekGRpth: why not use `PQExpBufferData`?

src/interfaces/libpq/libpq-fe.h
dkovalev1: I just have not seen advantages of using slist_node. None of pros in the slist_node commit message looks convincing and does not improve simple  embedding a pointer to the next element.

But from the other side - we loose compiler type checking and making data structure more complex. And hiding rather trivial implementation under macros.

src/interfaces/libpq/libpq-fe.h
dkovalev1: PQExpBufferData holds information about an extensible string. It implies separate allocation for the data.
This structure is a single piece of a memory, and also embeds additional data like segindex. It is simple, transparent and readable.

src/interfaces/libpq/libpq-fe.h
RekGRpth: `slist_node` already has a standard API for adding, removing, and iterating over a list, and it is already used in many places in the code.

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,167 @@ forwardQENotices(void)
...
+
+	PQCleanMetadataInternal(queue);
+}
+
+
+ggMetadataQueueId
+PQMetadataNextQueueId()
+{
+	static ggMetadataQueueId queueId = 0;
+	return queueId++;
```
RekGRpth, line 1017: Will this type be enough until it overflows?

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,167 @@ forwardQENotices(void)
 	if (hasNotices)
 		pq_flush();
 }
+
+typedef struct ggMetadataQueue
+{
+	ggMetadataQueueId id;
+	int count;
+	ggMetadataChunk *chunks;
```
RekGRpth, line 897: pgindent

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -877,31 +877,34 @@ forwardQENotices(void)
...
+	int			count;
 	ggMetadataChunk *chunks;
-} ggMetadataQueue;
+}	ggMetadataQueue;
 
 /* List of ggMetadataQueue */
 static List *ggMetadataQueues = NIL;
 
-static ggMetadataQueue *PQMetadataFindQueue(ggMetadataQueueId queue_id)
+static ggMetadataQueue * PQMetadataFindQueue(ggMetadataQueueId queue_id)
```
RekGRpth:
```suggestion
static ggMetadataQueue *
PQMetadataFindQueue(ggMetadataQueueId queue_id)
```

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2373,6 +2373,12 @@ build_startup_packet(const PGconn *conn, char *packet,
 	return packet_len;
 }
 
+/*
+ * Read a Metadata response message.
+ * Entry: 'M' message type has already been consumed.
+ * Exit: returns 0 if successfully consumed message.
+ *		 returns 1 otherwise.
+*/
```
RekGRpth:
```suggestion
 */
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,178 @@ forwardQENotices(void)
...
+		ggMetadataQueue *queue = lfirst(lc);
+
+		if (queue->id == queue_id)
+			return queue;
+	}
+	return NULL;
+}
+
+static void
+MPPmetadataReceiver(void *arg, ggMetadataChunk * metadata_chunk, ggMetadataQueueId queue_id)
```
RekGRpth:
```suggestion
MPPmetadataReceiver(void *arg, ggMetadataChunk *metadata_chunk, ggMetadataQueueId queue_id)
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,178 @@ forwardQENotices(void)
...
+	if (!it)
+		return NULL;
+
+	ggMetadataChunk *chunk = (ggMetadataChunk *) it;
+
+	return chunk->next;
+}
+
+void
+PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor * out_desc)
```
RekGRpth:
```suggestion
PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor *out_desc)
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,178 @@ forwardQENotices(void)
...
+{
+	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
+
+	if (!queue)
+		elog(ERROR, "No metadata queue with id %u", queue_id);
+	return queue->count;
+}
+
+static void
+PQCleanMetadataInternal(ggMetadataQueue * queue)
```
RekGRpth:
```suggestion
PQCleanMetadataInternal(ggMetadataQueue *queue)
```

src/interfaces/libpq/libpq-int.h
```diff
@@ -277,6 +286,7 @@ struct pg_result
 	PQaoRelTupCount *aotupcounts;
 	int		nWaits;
 	int		*waitGxids;
+
```
RekGRpth: remove this change

src/interfaces/libpq/libpq-fe.h
dkovalev1: Many but not all, even some new code is not using it.

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: We going to use with a rather big and long distributed transactions. 4G is a big number, I don't think its a big chance for an overflow. And even in a case of overflow, it will not be noticed, it just go from 0 again.

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2365,3 +2372,54 @@ build_startup_packet(const PGconn *conn, char *packet,
...
+	{
+		free(chunk);
+		return 1;
+	}
+
+	if (pqGetnchar(chunk->payload, payload_len, conn))
+	{
+		free(chunk);
+		return 1;
+	}
```
KnightMurloc: Why not do all checks before we do malloc?

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,179 @@ forwardQENotices(void)
...
+static void
+MPPmetadataReceiver(void *arg, ggMetadataChunk *metadata_chunk, ggMetadataQueueId queue_id)
+{
+	SegmentDatabaseDescriptor *segdbDesc = arg;
+
+	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
+
+	if (queue)
+	{
+		metadata_chunk->next = queue->chunks;
```
RekGRpth:
```suggestion
	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);

	if (queue)
	{
		SegmentDatabaseDescriptor *segdbDesc = arg;

		metadata_chunk->next = queue->chunks;
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,179 @@ forwardQENotices(void)
...
+	if (!it)
+		return NULL;
+
+	ggMetadataChunk *chunk = (ggMetadataChunk *) it;
+
+	return chunk->next;
+}
+
+void
+PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor *out_desc)
```
RekGRpth, line 963:
```suggestion
PQsetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor *out_desc)
```

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,179 @@ forwardQENotices(void)
...
+	{
+		metadata_chunk->next = queue->chunks;
+		metadata_chunk->segindex = segdbDesc->segindex;
+		queue->chunks = metadata_chunk;
+		queue->count++;
+	}
+}
+
+ggMetadataChunkIterator
+PQMetadataWalk(ggMetadataQueueId queue_id)
```
RekGRpth, line 942: Why not `PQMetadataChunks` like `PQgetMetadataCount`?

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,179 @@ forwardQENotices(void)
...
+void
+PQCreateMetadataQueue(ggMetadataQueueId queue_id)
+{
+	MemoryContext oldcontext = MemoryContextSwitchTo(TopTransactionContext);
+
+	ggMetadataQueue *queue = palloc0(sizeof(ggMetadataQueue));
+
+	queue->id = queue_id;
+	queue->chunks = NULL;
+	queue->count = 0;
```
RekGRpth:
```suggestion
```
(it is done by `palloc0`)

RekGRpth:
remove `[DRAFT]` from pr title

RekGRpth:
How about naming types with two capital letters `GG`?

src/interfaces/libpq/libpq-fe.h
RekGRpth:
```suggestion
	char   payload[1];	/* VARIABLE LENGTH ARRAY */
```

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2365,3 +2372,54 @@ build_startup_packet(const PGconn *conn, char *packet,
...
+
+	/*
+	 * Since the metadata might be pretty long, we create an own buffer
+	 * rather than using conn->workBuffer.  workBuffer is intended
+	 * for stuff that is expected to be short.
+	 *
+	 * we can not use palloc/pqAlloc here as this is a handler.
+	 */
+
+	ggMetadataChunk *chunk = malloc(sizeof(ggMetadataChunk) + payload_len);
```
RekGRpth:
```suggestion
	ggMetadataChunk *chunk = malloc(offsetof(ggMetadataChunk, payload) + payload_len);
```

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2365,3 +2372,54 @@ build_startup_packet(const PGconn *conn, char *packet,
...
+	}
+
+	if (pqGetnchar(chunk->payload, payload_len, conn))
+	{
+		free(chunk);
+		return 1;
+	}
+
+	/* pass ownership of the link too the handler */
+	conn->metadataHooks.metadataRec(conn->metadataHooks.metadataRecArg, chunk, queue_id);
```
RekGRpth, line 2422: I'm a little confused: is this `pgGetMetadataMessage` function called on the `FRONTEND`? And is `metadataRec` = `MPPmetadataReceiver` defined in the `BACKEND`?

src/interfaces/libpq/libpq-int.h
```diff
@@ -197,6 +197,15 @@ typedef struct
 	void	   *noticeProcArg;
 } PGNoticeHooks;
 
+/* Fields needed for metadata handling */
+typedef struct
+{
+	PQmetadataReceiver metadataRec; /* notice message receiver */
+	void	   *metadataRecArg;
+	PQmetadataProcessor metadataProc;
+	void       *metadataProcArg;
```
RekGRpth: Where are these fields used?

src/interfaces/libpq/libpq-fe.h
dkovalev1: This is not VLA idiom, this is flexible array member idiom, and its written with a [].  C99 section §6.7.2.1, item 16, page 103, https://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf

src/interfaces/libpq/fe-protocol3.c
dkovalev1: C99 says about using sizeof in this case. What are pros of your suggestion?

RekGRpth: It looks like they are the same thing: `sizeof` is used with `[]` and `offsetof` is used with `[1]`.

src/interfaces/libpq/fe-protocol3.c
dkovalev1: `pgGetMetadataMessage` function called when QD recieve "M" message from the QE, and it acts as a frontend side for the QE at that moment. 
`MPPmetadataReceiver` is a callback which works in the QD in its backend context.

src/interfaces/libpq/libpq-int.h
dkovalev1: removed. They were intended to keep it consistent with a notice receiver, not really needed actually

dkovalev1:
> How about naming types with two capital letters `GG`?

This is according spec, do you suggest to change the spec as well? Or you talking about some specific type(s)?

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: Sorry, don't get the idea. What exactly should be `PQMetadataChunks` here?

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: The function is intended to get metadata to the caller from the core. Why should it be called `set`?

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: This `segdbDesc` is about processing arguments and preparing them to be the proper type. IMHO it's much more readable when in the very beginning. It's not about making some logic and the compiler does not generate any extra instructions here. Also your change is not saving stack memory.
So I'd leave it here.

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: okay, removed

src/interfaces/libpq/fe-protocol3.c
RekGRpth: It looks like you've added public functions to the client side, but they can't be used on the regular client.

src/interfaces/libpq/fe-protocol3.c
dkovalev1: Its not the same - sizeof is directed by the standard. And even if they were the same - why change it?

src/backend/cdb/dispatcher/cdbconn.c
RekGRpth: `PQgetMetadataCount` returns `queue->count`
this function returns `queue->chunks`

src/backend/cdb/dispatcher/cdbconn.c
RekGRpth: because it sets `out_desc` and returns nothing

src/interfaces/libpq/fe-protocol3.c
dkovalev1: Can you suggest something?

RekGRpth: hide your new functions under macro `#ifndef FRONTEND`

RekGRpth: or move your new functions in separate file, compiled only on backend

RekGRpth:
> do you suggest to change the spec as well?

Yes, if it is a valid change to be accepted.

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: Got your question finally. The idea is to hide implementation of chunks storage. This makes possible to use different data structures under the hood, as all communication hidden under abstract `ggMetadataChunkIterator` on the client side.  It happens to be equal to queue->chunks at this moment, but its not necessary.

src/interfaces/libpq/libpq-fe.h
RekGRpth: > under macros

This is the advantage: the code is standard and understandable, and there is no loss of performance.

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: it returns metadata by filling out `out_desc`. Error handling is not expected on the caller side, so it returns nothing as a function return code.

src/interfaces/libpq/libpq-fe.h
dkovalev1: It is not a standard, and its understandability is questionable. IDEs and tools does not properly understand it, debugging has an extra level of indirection. We shall read not the well known C-level basix syntax but instead some (ugly?) macros pseudocode. Honestly, this is not an advantage from my point of view.

src/interfaces/libpq/fe-protocol3.c
dkovalev1: will moving them to `libpq-int.h` work for you?

RekGRpth: I can successfully include it in regular client

dkovalev1: > I can successfully include it in regular client

is it good or bad? The file itself is:

```
 *	  This file contains internal definitions meant to be used only by
 *	  the frontend libpq library, not by applications that call it.
```

RekGRpth: I mean it

>will moving them to libpq-int.h work for you?

won't help.

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +873,177 @@ forwardQENotices(void)
...
+	ggMetadataQueueId id;
+	int			count;
+	ggMetadataChunk *chunks;
+}	ggMetadataQueue;
+
+/* List of ggMetadataQueue */
+static List *ggMetadataQueues = NIL;
+
+static ggMetadataQueue *
+PQMetadataFindQueue(ggMetadataQueueId queue_id)
```
RekGRpth, line 904: why not
```suggestion
MPPmetadataFindQueue(ggMetadataQueueId queue_id)
```
because it is not `FRONTEND` (`PQ`)

src/interfaces/libpq/fe-protocol3.c
dkovalev1: I still don't understand the nature of you suggestion. Shall [PQsetNoticeReceiver](https://github.com/GreengageDB/greengage/blob/a6b03cc934598eb1a3013b3d11eec87a6c0fc853/src/interfaces/libpq/libpq-fe.h#L376) be refsactored out as well? Semantically it is very close.
And what could prevent a normal frontend client from using metadata? Okay, we can hide it and do not promise that interface is stable by moving to `libpq-int.h`.

RekGRpth:
There is still something wrong with the naming of functions:

1) for some reason the `PQMetadataWalk` function returns `ggMetadataChunkIterator`, which is not an iterator at all

2) for some reason the `PQgetNextMetadata` function also returns the same `ggMetadataChunkIterator`

3) the `PQgetMetadata` function does not return anything at all

It seems that it is not clear at all what this `metadata` is.

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: because we agreed on the naming on the MVP stage and spec providing

src/interfaces/libpq/fe-protocol3.c
RekGRpth: I can quite easily call `PQsetNoticeReceiver` in regular client.

src/backend/cdb/dispatcher/cdbconn.c
RekGRpth: Spec can't change?

dkovalev1:
> There is still something wrong with the naming of functions:
> 
>     1. for some reason the `PQMetadataWalk` function returns `ggMetadataChunkIterator`, which is not an iterator at all
> 
it IS iterator. Refer an [answer](https://github.com/GreengageDB/greengage/pull/440#discussion_r3279567437) and see usage example in the test.

>     2. for some reason the `PQgetNextMetadata` function also returns the same `ggMetadataChunkIterator`
> 
`PQgetNextMetadata` advances iterator and returns a new version of it, which we can check and use for getting data.

>     3. the `PQgetMetadata` function does not return anything at all
> 
It does return metadata in the second argument as an out parameter. Error code is not supposed  to be checked at the caller so it's void.
> 
> It seems that it is not clear at all what this `metadata` is.

Metadata is an opaque binary data which can be sent from QE as a part of a transaction to be later used on a QD.

RekGRpth:
> Metadata is an opaque binary data which can be sent from QE as a part of a transaction to be later used on a QD.

So this has nothing to do with being a regular client?

dkovalev1:
> > Metadata is an opaque binary data which can be sent from QE as a part of a transaction to be later used on a QD.
> 
> So this has nothing to do with being a regular client?

It is not related to a kind of a client. Metadata sent via libpq protocol and a QE client is to receive it. One can send metadata from the backend (f.e. using custom extension module) and receive it on a client side.
To make this interface less visible to the users I suggest to move it to libpq-int.h.

RekGRpth:
> > > Metadata is an opaque binary data which can be sent from QE as a part of a transaction to be later used on a QD.
> > 
> > 
> > So this has nothing to do with being a regular client?
> 
> It is not related to a kind of a client. Metadata sent via libpq protocol and a QE client is to receive it. One can send metadata from the backend (f.e. using custom extension module) and receive it on a client side. To make this interface less visible to the users I suggest to move it to libpq-int.h.

Try to write tests not as for the backend, but as for the frontend.

dkovalev1:
> Try to write tests not as for the backend, but as for the frontend.

Why? it is out of scope of the task.

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -15,6 +15,7 @@
  */
 #include "postgres.h"
 
+#include "access/xact.h"
```
KnightMurloc: Why do we need this?

src/interfaces/libpq/libpq-fe.h
```diff
@@ -618,6 +618,23 @@ extern int	pg_char_to_encoding(const char *name);
...
+} ggMetadataChunk;
+
+typedef unsigned int ggMetadataQueueId;
+typedef void (*PQmetadataReceiver) (void *arg, ggMetadataChunk *, ggMetadataQueueId);
+typedef void (*PQmetadataProcessor) (void *arg, ggMetadataChunk *);
```
RekGRpth: Where is this type used?

RekGRpth:
> > Try to write tests not as for the backend, but as for the frontend.
> 
> Why? it is out of scope of the task.

You've added new functionality to both the `backend` and the `frontend`, but you're only testing it through the `backend`. The `frontend` isn't even linked:
```bash
main.c:10:(.text+0x61): undefined reference to `PQsetMetadataReceiver'
collect2: error: ld returned 1 exit status
```
where `main.c`
```c
#include <postgres.h>

#include <libpq-fe.h>

static void MPPmetadataReceiver(void *arg, ggMetadataChunk *metadata_chunk, ggMetadataQueueId queue_id) {}

int main(int argc, char *argv[]) {
    PGconn *conn;
    if (!(conn = PQconnectdb("user=gpadmin"))) return 1;
    PQsetMetadataReceiver(conn, &MPPmetadataReceiver, NULL);
    // ...
}
```
and `Makefile`
```Makefile
$(OBJS): Makefile
OBJS = main.o
PG_CONFIG = pg_config
PG_CPPFLAGS += -I$(libpq_srcdir)
PG_LIBS += $(libpq)
PG_LIBS += $(libpq_pgport)
PGXS = $(shell $(PG_CONFIG) --pgxs)
PROGRAM = test
include $(PGXS)
```

dkovalev1:
> > > Try to write tests not as for the backend, but as for the frontend.
> > 
> > 
> > Why? it is out of scope of the task.
> 
> You've added new functionality to both the `backend` and the `frontend`, but you're only testing it through the `backend`. The `frontend` isn't even linked:

This is out of scope of the problem I am solving. That's why I suggest to hide this in libpq-int.h

RekGRpth:
> > > > Try to write tests not as for the backend, but as for the frontend.
> > > 
> > > 
> > > Why? it is out of scope of the task.
> > 
> > 
> > You've added new functionality to both the `backend` and the `frontend`, but you're only testing it through the `backend`. The `frontend` isn't even linked:
> 
> This is out of scope of the problem I am solving. That's why I suggest to hide this in libpq-int.h

This won't help; I'll include `libpq-int.h` and the same thing will happen.
You have a fundamental error: public functions are in the `frontend`, but the implementation is in the `backend`.

src/include/libpq/libpq.h
```diff
@@ -75,4 +75,31 @@ extern void secure_close(Port *port);
...
+
+extern ggMetadataChunkIterator PQMetadataWalk(ggMetadataQueueId queue_id);
+extern ggMetadataChunkIterator PQgetNextMetadata(ggMetadataChunkIterator it);
+extern void PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor * out_desc);
+extern int	PQgetMetadataCount(ggMetadataQueueId queue_id);
+extern void PQCleanMetadata(ggMetadataQueueId queue_id);
+
+extern ggMetadataQueueId PQMetadataNextQueueId(void);
+extern void PQCreateMetadataQueue(ggMetadataQueueId queue_id);
+extern void PQDeleteMetadataQueue(ggMetadataQueueId queue_id);
```
RekGRpth: These functions are still not linked in the regular `backend`.

dkovalev1: I'd appreciate a bit more detail and do you have any suggestions?

RekGRpth:
```bash
main.c:7:(.text+0x19): undefined reference to `PQMetadataWalk'
/usr/bin/ld: /home/gpadmin/gpdb_src/test/main.c:9:(.text+0x2b): undefined reference to `PQgetNextMetadata'
/usr/bin/ld: /home/gpadmin/gpdb_src/test/main.c:14:(.text+0x40): undefined reference to `PQgetMetadataCount'
/usr/bin/ld: /home/gpadmin/gpdb_src/test/main.c:15:(.text+0x4a): undefined reference to `PQCleanMetadata'
/usr/bin/ld: /home/gpadmin/gpdb_src/test/main.c:16:(.text+0x4f): undefined reference to `PQMetadataNextQueueId'
/usr/bin/ld: /home/gpadmin/gpdb_src/test/main.c:17:(.text+0x59): undefined reference to `PQCreateMetadataQueue'
/usr/bin/ld: /home/gpadmin/gpdb_src/test/main.c:18:(.text+0x63): undefined reference to `PQDeleteMetadataQueue'
collect2: error: ld returned 1 exit status
```
main.c
```c
#include <postgres.h>

#include <libpq-fe.h>
#include <libpq/libpq.h>

int main(int argc, char *argv[]) {
    ggMetadataChunkIterator it = PQMetadataWalk(1);

    for (; it; it = PQgetNextMetadata(it))
    {
        // ...
    }

    PQgetMetadataCount(1);
    PQCleanMetadata(1);
    PQMetadataNextQueueId();
    PQCreateMetadataQueue(1);
    PQDeleteMetadataQueue(1);
}
```

RekGRpth: > do you have any suggestions?

Either move the declarations to the `backend`, or move the implementation to the `frontend`. Not like now: one here, one there.

src/include/libpq/libpq.h
```diff
@@ -75,4 +75,5 @@ extern void secure_close(Port *port);
 extern ssize_t secure_read(Port *port, void *ptr, size_t len);
 extern ssize_t secure_write(Port *port, void *ptr, size_t len);
 
+
```
RekGRpth: remove this change

src/interfaces/libpq/libpq-int.h
```diff
@@ -197,6 +197,24 @@ typedef struct
 	void	   *noticeProcArg;
 } PGNoticeHooks;
 
+/* Fields needed for metadata handling */
+typedef struct ggMetadataChunk
+{
+	struct ggMetadataChunk *next;
+	int    segindex;      /* source segment */
+	int    metadataLen;    /* Length of metadata buffer */
+	char   payload[];
```
RekGRpth, line 207: why not
```suggestion
	char		payload[1];	/* ... (VARIABLE LENGTH) */
```
like in `PGMessageField`

dkovalev1: because C standard says about char[] for flexible struct members.

src/interfaces/libpq/libpq-int.h
```diff
@@ -197,6 +197,24 @@ typedef struct
...
+} ggMetadataChunk;
+
+typedef unsigned int ggMetadataQueueId;
+typedef void (*PQmetadataReceiver) (void *arg, ggMetadataChunk *, ggMetadataQueueId);
```
RekGRpth, line 211:
```suggestion
typedef void (*PQmetadataReceiver) (void *arg, void *metadata_chunk, unsigned int queue_id);
```
and remove `typedef struct ggMetadataChunk` from this file

src/backend/access/transam/xact.c
```diff
@@ -84,6 +84,8 @@
 #include "utils/workfile_mgr.h"
 #include "utils/vmem_tracker.h"
 #include "cdb/cdbdisp.h"
+#include "cdb/cdbconn.h"
+
```
RekGRpth, line 88: remove this line?

src/interfaces/libpq/libpq-int.h
dkovalev1: please explain the reasons for it

src/backend/access/transam/xact.c
dkovalev1: removed

src/include/cdb/cdbconn.h
```diff
@@ -106,4 +110,32 @@ bool cdbconn_signalQE(SegmentDatabaseDescriptor *segdbDesc, char *errbuf, bool i
...
+}	ggMetadataDescriptor;
+
+extern ggMetadataChunkIterator PQMetadataWalk(ggMetadataQueueId queue_id);
+extern ggMetadataChunkIterator PQgetNextMetadata(ggMetadataChunkIterator it);
+extern void PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor * out_desc);
+extern int	PQgetMetadataCount(ggMetadataQueueId queue_id);
+extern void PQCleanMetadata(ggMetadataQueueId queue_id);
+
+extern ggMetadataQueueId PQMetadataNextQueueId(void);
+extern void PQCreateMetadataQueue(ggMetadataQueueId queue_id);
```
RekGRpth, line 138: Are these functions used anywhere in this PR, except for testing. Or are they planned for future PRs?

src/interfaces/libpq/libpq-int.h
RekGRpth: To provide a simple public interface and hide the implementation details (structure).

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -286,6 +288,11 @@ pqParseInput3(PGconn *conn)
 						return;
 					conn->asyncStatus = PGASYNC_READY;
 					break;
+				case 'M':       /* CDB: backend sends metadata 'M' message */
+					if (pgGetMetadataMessage(conn, msgLength))
+						return;
+					break;
+				break;
```
KnightMurloc: two breaks?

src/interfaces/libpq/libpq-int.h
```diff
@@ -197,6 +197,24 @@ typedef struct
...
+
+typedef struct
+{
+	PQmetadataReceiver metadataRec; /* notice message receiver */
```
KnightMurloc: > notice message receiver

Not updated comment?

src/interfaces/libpq/fe-protocol3.c
KnightMurloc: Can we hide it in `#ifndef FRONTEND`?

src/test/modules/test_metadata/sql/test_metadata.sql
```diff
@@ -0,0 +1,61 @@
+
```
KnightMurloc: Add test that metadata is cleared in case of error during transaction.

src/interfaces/libpq/libpq-int.h
dkovalev1: Can not agree. I prefer to have a safety which comes with a type checking.

src/include/cdb/cdbconn.h
dkovalev1: yes, this is going to be used from PXF (and perhaps other extensions as well)

src/backend/libpq/pqformat.c
```diff
@@ -681,3 +681,44 @@ pq_getmsgend(StringInfo msg)
...
+void
+pq_metadatasend(const void *data, size_t len, int32 queue_id)
+{
+	StringInfoData buf;
+
+	/*
+	 * Safety check: The protocol message length field is a 32-bit signed int.
+	 * We must ensure the payload + header doesn't overflow it.
+	 */
+	if (len > (size_t)(INT_MAX - 4))
```
KnightMurloc: Message length (4 bytes) is also included in the total message length. So we should take 8 bytes into account.

src/backend/libpq/pqformat.c
```diff
@@ -681,3 +681,44 @@ pq_getmsgend(StringInfo msg)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_PROGRAM_LIMIT_EXCEEDED),
+				 errmsg("custom metadata message too large")));
+
+	/*
+	 * Initialize the message buffer. 'M' is the message type tag.
+	 */
+	pq_beginmessage(&buf, 'M');
+
+	pq_sendint(&buf, queue_id, 4);
```
KnightMurloc: Why not:

```suggestion
	pq_sendint(&buf, queue_id, sizeof(queue_id));
```

src/interfaces/libpq/fe-protocol3.c
dkovalev1: moved

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: not needed anymore, deleted

src/interfaces/libpq/fe-protocol3.c
dkovalev1: deleted

src/interfaces/libpq/libpq-int.h
dkovalev1: updated now

src/test/modules/test_metadata/sql/test_metadata.sql
dkovalev1: added

src/backend/libpq/pqformat.c
dkovalev1: taken

src/backend/libpq/pqformat.c
dkovalev1: okay

src/interfaces/libpq/fe-protocol3.c
dkovalev1: okay

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2365,3 +2373,53 @@ build_startup_packet(const PGconn *conn, char *packet,
...
 }
+
+/*
+ * Read a Metadata response message.
+ * Entry: 'M' message type has already been consumed.
+ * Exit: returns 0 if successfully consumed message.
+ *		 returns 1 otherwise.
+ */
+static int
+pgGetMetadataMessage(PGconn *conn, int length)
```
RekGRpth, line 2386: Is this function needed on the `FRONTEND`?

src/interfaces/libpq/libpq-int.h
```diff
@@ -197,6 +197,24 @@ typedef struct
...
+} ggMetadataChunk;
+
+typedef unsigned int ggMetadataQueueId;
+typedef void (*PQmetadataReceiver) (void *arg, ggMetadataChunk *, ggMetadataQueueId);
+
+typedef struct
+{
+	PQmetadataReceiver metadataRec; /* metadata message receiver */
+	void	   *metadataRecArg;
+} PGMetadataHooks;
```
RekGRpth, line 217: Are these types needed on the `FRONTEND`?

src/interfaces/libpq/libpq-int.h
```diff
@@ -538,6 +556,9 @@ struct pg_conn
 	/* Buffer for receiving various parts of messages */
 	PQExpBufferData workBuffer; /* expansible string */
 	char       *diffoptions;  /* MPP: transfer changed GUCs(require sync) from QD to QEs */
+
+	/* Callback procedures for metadata message processing */
+	PGMetadataHooks metadataHooks;
```
RekGRpth, line 564: Is this field needed on the `FRONTEND`?

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -70,6 +70,8 @@ static void reportErrorPosition(PQExpBuffer msg, const char *query,
 					int loc, int encoding);
 static int build_startup_packet(const PGconn *conn, char *packet,
 					 const PQEnvironmentOption *options);
+static int pgGetMetadataMessage(PGconn *conn, int length);
+void pgGetMetaData(PGconn *conn);
```
RekGRpth: What is this line for?

src/include/cdb/cdbconn.h
```diff
@@ -106,4 +110,32 @@ bool cdbconn_signalQE(SegmentDatabaseDescriptor *segdbDesc, char *errbuf, bool i
...
+typedef struct ggMetadataDescriptor
+{
+	int			metadataLen;
+	int			segindex;		/* source segment */
+	void	   *data;
+}	ggMetadataDescriptor;
+
+extern ggMetadataChunkIterator PQMetadataWalk(ggMetadataQueueId queue_id);
+extern ggMetadataChunkIterator PQgetNextMetadata(ggMetadataChunkIterator it);
+extern void PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor * out_desc);
```
RekGRpth:
```suggestion
extern void PQgetMetadata(ggMetadataChunkIterator it, ggMetadataDescriptor *out_desc);
```

RekGRpth:
Perhaps the PR description should also be updated to indicate that the metadata is only available on the `backend`.

src/test/modules/test_metadata/test_metadata.c
```diff
@@ -0,0 +1,149 @@
...
+test_check_metadata(PG_FUNCTION_ARGS)
+{
+	Assert(Gp_role == GP_ROLE_DISPATCH);
+	ggMetadataQueueId queue_id = PG_GETARG_INT32(0);
+
+	/*
+	 * Metadata records are unordered for the each query, so to make
+	 * predictable
+	 */
+	/* records, make them in few passes */
```
RekGRpth: is this single comment?

dkovalev1: fixed

RekGRpth (dismissed):

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -870,3 +889,186 @@ forwardQENotices(void)
...
+
+	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
+
+	/* 
+	 * queue is expected to be created by the extension
+	 */
+	Assert(queue != NULL);
+
+	if (queue)
+	{
```
RekGRpth: This assertion and this condition are incompatible; please keep one or the other.

dkovalev1: Assertion works for the debug build, so extension developer won't miss programming error with miss of queue creation.
call of `free` intended for the release build - to prevent memory leak if queue creation was missed eventually.

RekGRpth: This shouldn't be done! To prevent leaks, you can use a callback of release memory context or resource owner.

dkovalev1: In a message handler we can not use memory contexts and resource owners.

RekGRpth: so, rework code to avoid leak another way

KnightMurloc: What if we wrap `if` and `else` in `#ifndef USE_ASSERT_CHECKING`?

RekGRpth: Are you suggesting we test one code and release another? An assertion is a statement of what should always be the case, but the release version doesn't include this extra check.

RekGRpth: > In a message handler we can not use memory contexts and resource owners.

Link memory release to connection closing.

dkovalev1: > Link memory release to connection closing.

I don't think making infrastructure parallel to memory contexts is a good idea here.

RekGRpth: see `PQfinish`

RekGRpth:
new flaky?
```diff
--- \/home\/gpadmin\/gpdb_src\/src\/test\/isolation2\/expected\/workfile_mgr_test\.out	2026-05-26 21:28:41.314630486 +0000
+++ \/home\/gpadmin\/gpdb_src\/src\/test\/isolation2\/results\/workfile_mgr_test\.out	2026-05-26 21:28:41.319630607 +0000
@@ -84,9 +88,9 @@
 1: select gp_workfile_mgr_test('fd_large_file_test', 0);
  gp_workfile_mgr_test 
 ----------------------
- t                    
- t                    
- t                    
+ f                    
+ f                    
+ f                    
  t                    
 (4 rows)
```

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2365,3 +2373,56 @@ build_startup_packet(const PGconn *conn, char *packet,
...
+	}
+
+	int payload_len = length - sizeof(ggMetadataQueueId);
+
+	/*
+	 * Since the metadata might be pretty long, we create an own buffer
+	 * rather than using conn->workBuffer.  workBuffer is intended
+	 * for stuff that is expected to be short.
+	 *
+	 * we can not use palloc/pqAlloc here as this is a handler.
```
RekGRpth:
```suggestion
```

src/test/modules/test_metadata/Makefile
```diff
@@ -0,0 +1,29 @@
...
+subdir = src/test/modules/test_metadata
+top_builddir = ../../../..
+include $(top_builddir)/src/Makefile.global
+include $(top_srcdir)/contrib/contrib-global.mk
+endif
+
+
+installcheck: install
+test: clean all install
+	psql postgres -f sql/test_extensions.sql 2>&1
```
RekGRpth: Why is this line needed?

dkovalev1:
> new flaky?
> 
> ```diff
> --- \/home\/gpadmin\/gpdb_src\/src\/test\/isolation2\/expected\/workfile_mgr_test\.out	2026-05-26 21:28:41.314630486 +0000
> +++ \/home\/gpadmin\/gpdb_src\/src\/test\/isolation2\/results\/workfile_mgr_test\.out	2026-05-26 21:28:41.319630607 +0000
> @@ -84,9 +88,9 @@
>  1: select gp_workfile_mgr_test('fd_large_file_test', 0);
>   gp_workfile_mgr_test 
>  ----------------------
> - t                    
> - t                    
> - t                    
> + f                    
> + f                    
> + f                    
>   t                    
>  (4 rows)
> ```

Perhaps an old one but not discovered yet. Failed a test for a huge file read/write in an environment which we can not control. Does not look like a result of this PR.

RekGRpth:
> > new flaky?
> > ```diff
> > --- \/home\/gpadmin\/gpdb_src\/src\/test\/isolation2\/expected\/workfile_mgr_test\.out	2026-05-26 21:28:41.314630486 +0000
> > +++ \/home\/gpadmin\/gpdb_src\/src\/test\/isolation2\/results\/workfile_mgr_test\.out	2026-05-26 21:28:41.319630607 +0000
> > @@ -84,9 +88,9 @@
> >  1: select gp_workfile_mgr_test('fd_large_file_test', 0);
> >   gp_workfile_mgr_test 
> >  ----------------------
> > - t                    
> > - t                    
> > - t                    
> > + f                    
> > + f                    
> > + f                    
> >   t                    
> >  (4 rows)
> > ```
> 
> Perhaps an old one but not discovered yet. Failed a test for a huge file read/write in an environment which we can not control. Does not look like a result of this PR.

create ticket for that

src/test/modules/test_metadata/Makefile
dkovalev1: To be consistent with other Makefiles in the test/modules.

RekGRpth: there is no `sql/test_extensions.sql` in your new extension

dkovalev1:
> create ticket for that

Yes Sir!

KnightMurloc:
Description look's outdated:
```
The code for message processing is inspired by the notice processing
mechanism which works in a similar way. This method has a restriction for
not making PG_THROW during the handler, as it can break the protocol
state. That's we good old malloc is used for memory allocation for
metadata receiver. To lower the number of allocations, flexible array is
used.
```

dkovalev1:
> Description look's outdated:
> 
> ```
> The code for message processing is inspired by the notice processing
> mechanism which works in a similar way. This method has a restriction for
> not making PG_THROW during the handler, as it can break the protocol
> state. That's we good old malloc is used for memory allocation for
> metadata receiver. To lower the number of allocations, flexible array is
> used.
> ```

updated

KnightMurloc (dismissed):

RekGRpth (dismissed):

KnightMurloc (approved):

RekGRpth (approved):

PR #441, opened 2026-05-19
GG-481: Add scripts to test pg_upgrade migration in a demo cluster

GirgerS:
This patch adds infrastructure to run src/bin/pg_upgrade/test_gpdb.sh
test between Greengage 6 and Greengage 7.

The list of changes is as follows:
- src/bin/pg_upgrade/test_gpdb.sh is slightly modified to make
  flags a bit more consistent. This change also affects
  src/bin/pg_upgrade/Makefile.
- src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh is added.
  It is the main script to run the test between different versions.
- ci/Dockerfile.pg_upgrade file is added to run the new test on the
  CI. The CI action itself should be added in the future.
- src/bin/pg_upgrade/README.test.md file is added. It explains how
  to run the newly added test.
- src/bin/pg_upgrade/README.gpdb is updated to contain more 
  information about pg_upgrade testing.

### DIFF
--- a/src/bin/pg_upgrade/README.gpdb
+++ b/src/bin/pg_upgrade/README.gpdb
@@ -25,8 +25,8 @@ Extensions to upstream pg_upgrade
 Functionality for restoring the append-only catalog relations is in
 aotable.c.
 
-Upgrading from 5.0 to 6.0
--------------------------
+Upgrading from 5.0 to 6.0 (Largely outdated)
+--------------------------------------------
 
 The name_pattern_ops and pattern equality operators were removed in 6.0 and
 text equality is treated like bitwise equality. This will cause internal sort
@@ -50,6 +50,18 @@ intention is for this to be a quick test that all objects are handled by
 pg_upgrade; upgrade testing still requires the full run across all
 segments/mirrors. The smoketest can be invoked by "make check" in the
 contrib/pg_upgrade directory.
+
+Also, it is worth mentioning that most of the pg_upgrade tests are performed
+in the ggupgrade repository. This decision was made by the Greenplum team to
+reduce amount of code duplication between ggupgrade and test scripts.
+
+In spite of this, we have added a 6.x to 7.x migration test to the core
+Greengage repository. The reason being that is it simply more convenient
+to test pg_upgrade from the same repository where it is located.
+
+For a more detailed overview of the existing tests, see README.test.md
+and TESTING files
+
 
 GPDB Merge notice:
 ------------------
--- a/src/bin/pg_upgrade/README.gpdb
+++ b/src/bin/pg_upgrade/README.gpdb
@@ -52,8 +52,8 @@ segments/mirrors. The smoketest can be invoked by "make check" in the
 contrib/pg_upgrade directory.
 
 Also, it is worth mentioning that most of the pg_upgrade tests are performed
-in the ggupgrade repository. This decision was made by the Greenplum team to
-reduce amount of code duplication between ggupgrade and test scripts.
+in the ggupgrade repository. This decision was made initially to reduce
+amount of code duplication between ggupgrade and test scripts.
 
 In spite of this, we have added a 6.x to 7.x migration test to the core
 Greengage repository. The reason being that is it simply more convenient
--- a/src/bin/pg_upgrade/README.test.md
+++ b/src/bin/pg_upgrade/README.test.md
@@ -35,8 +35,22 @@ bash   ${GREENGAGE7_SRC}/src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh
 Docker image already has all the necessary environment variables set to run the test WITHOUT a schema. The only thing left is to setup gpadmin user:
 ```bash
 docker build -f ci/Dockerfile.pg_upgrade -t gpdb7_pgupgrade:latest .
-docker run --rm gpdb7_pgupgrade bash -c \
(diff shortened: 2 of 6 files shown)


src/bin/pg_upgrade/README.test.md
```diff
@@ -0,0 +1,67 @@
...
+bash   ${GREENGAGE7_SRC}/src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh
+```
+
+## Running tests from the image
+
+Docker image already has all the necessary environment variables set to run the test WITHOUT a schema. The only thing left is to setup gpadmin user:
+```bash
+docker build -f ci/Dockerfile.pg_upgrade -t gpdb7_pgupgrade:latest .
+docker run --rm gpdb7_pgupgrade bash -c \
+    "gpdb_src/concourse/scripts/setup_gpadmin_user.bash && su gpadmin /home/gpadmin/gpdb_src/src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh"
```
whitehawk: I'm getting 
```
bash: /home/gpadmin/gpdb_src/src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh: No such file or directory
```
if I follow the instruction. Is it expected?

GirgerS: > if I follow the instruction. Is it expected?

Yes, because by default the pg_upgrade image pulls Greengage images from this repository, and I believe that they don't have this file yet (it should be there after this PR is merged).

I updated the description so it tells how to use local images.

src/bin/pg_upgrade/README.gpdb
```diff
@@ -51,6 +51,18 @@ pg_upgrade; upgrade testing still requires the full run across all
 segments/mirrors. The smoketest can be invoked by "make check" in the
 contrib/pg_upgrade directory.
 
+Also, it is worth mentioning that most of the pg_upgrade tests are performed
+in the ggupgrade repository. This decision was made by the Greenplum team to
```
whitehawk: Can we not mention 'Greenplum' here?

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -211,7 +211,7 @@ usage()
 {
 	local appname=`basename $0`
 	echo "usage: $appname -o <dir> -b <dir> [options]"
-	echo " -o <dir>     old cluster data directory"
+	echo " -O <dir>     old cluster data directory"
```
whitehawk, line 215: > modified to make flags a bit more consistent

Do we really need this change? How do you evaluate the consistency?

And, the usage message right above should be updated as well if we leave this change.

src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# We could accept command line arguments instead of relying on environment variables,
+# but it seems too much for a 50-line script. In any case, perform some redundant
+# checks before running the script.
+
+if [ -z "${GREENGAGE7_INSTALLATION:-}" ] || \
+   [ -z "${GREENGAGE6_INSTALLATION:-}" ] || \
+   [ -z "${GREENGAGE7_SRC:-}" ] || \
+   [ -z "${GREENGAGE6_SRC:-}" ]; then
+	echo "This scrips expects the following environment variables to be set:"
```
whitehawk: Can we give here a hint which var exactly was not set?

src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh
```diff
@@ -0,0 +1,57 @@
...
+if [ -z "${SQL_SCHEMA}" ]; then
+    echo "SQL_SCHEMA environment variable is not set. This test will be performed on an empty cluster."
+else
+    echo "SQL_SCHEMA is set. This test will load \"${SQL_SCHEMA}\" before performing upgrade."
+fi
+
+set -x
+
+# Create demo cluster
+export PORT_BASE=7000
```
whitehawk, line 41: Is it for greengage6? For 6x base is usually 6000. Why do we set 7000?

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: `old cluster executable directory` is denoted by the capital B, `new cluster executable directory` - by the lowercase one. 

This change was necessary because initially I added a flag for `new cluster data directory`. It turn out to be unnecessary, but I still left `-O -> -o ` change, it feels right for me.

But it is a purely arbitrary decision, so I'm ok with rolling it back.

src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh
GirgerS: Done.

src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh
GirgerS: Yeah, `gpstart` `gpstop` and similar commands in the `test_gpdb.sh` rely on the PGPORT variable, so we need to have consistent port numbers.

src/bin/pg_upgrade/README.gpdb
GirgerS: Done.

whitehawk (approved):

VoidZeroNull0 (approved):

PR #442, opened 2026-05-19
Optimize ALTER TABLE REBALANCE

whitehawk:
Optimize ALTER TABLE REBALANCE command

ALTER TABLE REBALANCE, in order to shrink a table, uses the approach with
selecting data from the shrunk segments and inserting the data into the
remaining segments.
Before this patch, the plan for the insert query had 'redistribute' motion from
all segments into all segments. That means that the root slice was executed on
all segments, either the scan slice. Scan execution on not-shrunk segments led
to performance overhead since tuples were retrieved first, and only then the
filter (which checks gp_segment_id) was applied.

This patch optimizes the ALTER TABLE REBALANCE command in a way that the insert
plan now has 'redistribute' motion from only shrunk segments to the reduced set
of segments that are left after shrinking. That is achieved by
1. enabling direct dispatch by reworking the condition from
'where gp_segment_id >= N' to 'where gp_segment_id in (N, N+1, ...)';
2. removing 'gp_dist_random' (as forcing random distribution disables
direct dispatch - refer to code in 'GetContentIdsFromPlanForSingleRelation()'
function);
3. limiting the 'numsegments' in the target result relation of the insert.

As 'gp_dist_random' is removed, now we need to disable Orca explicitly during
planning of the insert query.

New tests are not added; passing the current tests set is enough.

Co-authored-by: Alexander Kondakov <a.kondakov@arenadata.io>

### DIFF
--- a/src/backend/cdb/cdbllize.c
+++ b/src/backend/cdb/cdbllize.c
@@ -404,6 +404,10 @@ cdbllize_adjust_top_path(PlannerInfo *root, Path *best_path,
 		Assert(rte->rtekind == RTE_RELATION);
 
 		targetPolicy = GpPolicyFetch(rte->relid);
+		if (targetPolicy->ptype == POLICYTYPE_PARTITIONED &&
+			gp_segment_number_for_table_shrink > 0)
+			targetPolicy->numsegments =
+					Min(targetPolicy->numsegments, gp_segment_number_for_table_shrink);
 	}
 
 	if (query->commandType == CMD_SELECT && query->parentStmtType == PARENTSTMTTYPE_CTAS)
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -17753,16 +17753,10 @@ ATExecShrinkTable(Relation rel, GpPolicy *policy)
 		StringInfoData shrunk_segments_condition;
 		initStringInfo(&shrunk_segments_condition);
 		appendStringInfo(&shrunk_segments_condition, "gp_segment_id in (");
-		for (int shrunk_segment = policy->numsegments;
-			shrunk_segment < getgpsegmentCount();
-			shrunk_segment++)
-			if (shrunk_segment == getgpsegmentCount() - 1)
-				appendStringInfo(&shrunk_segments_condition,
-								 "%d", shrunk_segment);
-			else
-				appendStringInfo(&shrunk_segments_condition,
-								 "%d, ", shrunk_segment);
-		appendStringInfo(&shrunk_segments_condition, ")");
+		int shrunk_segment = policy->numsegments;
+		for (;shrunk_segment < getgpsegmentCount() - 1; shrunk_segment++)
+			appendStringInfo(&shrunk_segments_condition, "%d, ", shrunk_segment);
+		appendStringInfo(&shrunk_segments_condition, "%d)", shrunk_segment);
 
 		appendStringInfo(&sqlstmtInsert,
 						 "insert into %s select * from %s where %s",
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -17738,6 +17738,7 @@ ATExecShrinkTable(Relation rel, GpPolicy *policy)
 	if (Gp_role == GP_ROLE_DISPATCH && GpPolicyIsPartitioned(policy))
 	{
 		volatile bool connected = false;
+		bool saveOptimizerGucValue;
 		StringInfoData sqlstmtInsert;
 		initStringInfo(&sqlstmtInsert);
 		const char *nsp = quote_identifier(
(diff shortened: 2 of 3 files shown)


whitehawk:
To check the effect of the patch, one can use:
```
SET debug_print_plan = on;
SET client_min_messages = log;
```
before `ALTER TABLE REBALANCE` command.

bimboterminator1 (approved):

src/backend/commands/tablecmds.c
```diff
@@ -17749,22 +17750,32 @@ ATExecShrinkTable(Relation rel, GpPolicy *policy)
...
-		 * 'gp_dist_random' will cause fallback to Postgres planner,
-		 * so no need to tweak 'optimizer' value.
-		 */
+		StringInfoData shrunk_segments_condition;
+		initStringInfo(&shrunk_segments_condition);
+		appendStringInfo(&shrunk_segments_condition, "gp_segment_id in (");
+		int shrunk_segment = policy->numsegments;
+		for (;shrunk_segment < getgpsegmentCount() - 1; shrunk_segment++)
+			appendStringInfo(&shrunk_segments_condition, "%d, ", shrunk_segment);
+		appendStringInfo(&shrunk_segments_condition, "%d)", shrunk_segment);
```
hilltracer, line 17759: In case if table is already shrank I see this query in debug
```sql
drop table table_distr_hashed;
create table table_distr_hashed(a int) distributed by (a);
insert into table_distr_hashed select generate_series(1, 20);

begin;
select gp_expand_lock_catalog();
select gp_toolkit.gp_set_rebalance_numsegments(2);
end;

alter table table_distr_hashed rebalance;

"insert into public.table_distr_hashed select * from public.table_distr_hashed where gp_segment_id in (2)"


begin;
select gp_expand_lock_catalog();
select gp_toolkit.gp_set_rebalance_numsegments(1);
end;
alter table table_distr_hashed rebalance;

"insert into public.table_distr_hashed select * from public.table_distr_hashed where gp_segment_id in (1, 2)"
```

So, here we still scan seg 2 but we can avoid it. Since table is don't have data in (2). Does this make sense to use `rel->rd_cdbpolicy->numsegments` in logic?

src/backend/cdb/cdbllize.c
```diff
@@ -404,6 +404,10 @@ cdbllize_adjust_top_path(PlannerInfo *root, Path *best_path,
 		Assert(rte->rtekind == RTE_RELATION);
 
 		targetPolicy = GpPolicyFetch(rte->relid);
+		if (targetPolicy->ptype == POLICYTYPE_PARTITIONED &&
+			gp_segment_number_for_table_shrink > 0)
+			targetPolicy->numsegments =
+					Min(targetPolicy->numsegments, gp_segment_number_for_table_shrink);
```
hilltracer, line 410: Maybe gp_segment_number_for_table_shrink should be tied to a specific relation? Currently it seems to affect any partitioned target relation. Is it guaranteed that during ALTER TABLE REBALANCE we only ever plan DML for the table being rebalanced?

src/backend/commands/tablecmds.c
```diff
@@ -17803,6 +17815,7 @@ ATExecShrinkTable(Relation rel, GpPolicy *policy)
 		pfree(sqlstmtInsert.data);
 
 		gp_segment_number_for_table_shrink = 0;
+		optimizer = saveOptimizerGucValue;
```
hilltracer, line 17818: Maybe add test that GUC is restored after table rebalance? And if it is not too difficult after error processing in PG_CATCH.

src/backend/optimizer/plan/createplan.c
```diff
@@ -3003,7 +3003,8 @@ create_modifytable_plan(PlannerInfo *root, ModifyTablePath *best_path)
 				if (isfirst)
 				{
 					root->curSlice->gangType = GANGTYPE_PRIMARY_WRITER;
-					root->curSlice->numsegments = policy->numsegments;
+					if (gp_segment_number_for_table_shrink == 0)
+						root->curSlice->numsegments = policy->numsegments;
```
hilltracer, line 3013: Maybe add some comment description why we don't touch numsegments here?

src/backend/cdb/cdbllize.c
whitehawk: `gp_segment_number_for_table_shrink` is backend-local, and it is set only during ALTER TABLE REBALANCE, where we plan only the DML for the target relation, and should be reset in both OK and NG scenarios (when some error happened). Coupling it with the relation will require us to move it to some relation struct, where it will be not used in all usual scenarious. At this point I consider this to be unjustified.

src/backend/commands/tablecmds.c
whitehawk: The logic of `optimizer` disabling during internal planning is pretty simple, and is used in a handfull of other places in a similar manner (for ex. in `analyze_rel()`). I don't think that we need to add some specific check for the `optimizer` into the test.

src/backend/optimizer/plan/createplan.c
whitehawk: Updated

src/backend/commands/tablecmds.c
whitehawk: Updated

hilltracer (approved):

bimboterminator1 (approved):

PR #446, opened 2026-05-20
GG-472: Fix pg_dump inconsistencies between 6.x and 7.x

GirgerS:
This is the first round of fixing pg_dump inconsistencies
between 6.x and 7.x versions.

This patch makes sure that src/bin/gp_upgrade/test_gpdb.sh
test doesn't produce any errors on an empty cluster.

There might be more similar errors, but to find most of
them we need to make pg_upgrade being able to handle
SQL dump from the regression test suite.
And since it can take quite a while, let's fix these
simple errors right away.

The list of changes is as follows:
- Mask source cluster versions during cross-version testing.
- Explicitly default to NOBYPASSRLS when dumping from 6.x.
  This reverts 1a33ccf96ea199eabcb891ae6e66259d0913ff7c
- Make sure that gp_enable_statement_trigger GUC doesn't
  appear in the dumps. It is set for the restore connection anyway.

### DIFF
--- a/src/bin/pg_dump/pg_backup_archiver.c
+++ b/src/bin/pg_dump/pg_backup_archiver.c
@@ -3248,8 +3248,13 @@ _doSetFixedOutputState(ArchiveHandle *AH)
 	/*
 	 * Ensure that the setting is on, so that we don't error out while trying
 	 * to restore TRIGGERS from GP6 to GP7.
+	 *
+	 * This is done for the pg_restore connection itself, meaning that this
+	 * setting is set implicitly and it wouldn't appear in sql dumps
+	 * or archives. This is not ideal, but changing this code means
+	 * breaking already existing dumps/archives from GP6.
 	 */
-	if (AH->public.remoteVersion >= GPDB7_MAJOR_PGVERSION)
+	if (RestoringToDB(AH) && AH->public.remoteVersion >= GPDB7_MAJOR_PGVERSION)
 		ahprintf(AH, "SET gp_enable_statement_trigger = on;\n");
 
 	ahprintf(AH, "\n");
--- a/src/bin/pg_dump/pg_dumpall.c
+++ b/src/bin/pg_dump/pg_dumpall.c
@@ -1320,6 +1320,14 @@ dumpRoles(PGconn *conn)
 			else
 				appendPQExpBufferStr(buf, " NOBYPASSRLS");
 		}
+		else
+		{
+			/*
+			 * Default to NOBYPASSRLS for dumps from Greengage 6 to make
+			 * them consistent with dumps from Greenage 7.
+			 */
+			appendPQExpBufferStr(buf, " NOBYPASSRLS");
+		}
 
 		if (strcmp(PQgetvalue(res, i, i_rolconnlimit), "-1") != 0)
 			appendPQExpBuffer(buf, " CONNECTION LIMIT %s",
--- a/src/bin/pg_dump/pg_dumpall.c
+++ b/src/bin/pg_dump/pg_dumpall.c
@@ -1313,21 +1313,10 @@ dumpRoles(PGconn *conn)
 		else
 			appendPQExpBufferStr(buf, " NOREPLICATION");
 
-		if (server_version >= 90600)
-		{
-			if (strcmp(PQgetvalue(res, i, i_rolbypassrls), "t") == 0)
-				appendPQExpBufferStr(buf, " BYPASSRLS");
-			else
-				appendPQExpBufferStr(buf, " NOBYPASSRLS");
-		}
+		if (strcmp(PQgetvalue(res, i, i_rolbypassrls), "t") == 0)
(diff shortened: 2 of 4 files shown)


red1452 (commented):
Typo in the description: "This IS"

red1452 (dismissed):

src/bin/pg_dump/pg_dumpall.c
```diff
@@ -1320,6 +1320,14 @@ dumpRoles(PGconn *conn)
...
 				appendPQExpBufferStr(buf, " NOBYPASSRLS");
 		}
+		else
+		{
+			/*
+			 * Default to NOBYPASSRLS for dumps from Greengage 6 to make
+			 * them consistent with dumps from Greenage 7.
+			 */
+			appendPQExpBufferStr(buf, " NOBYPASSRLS");
+		}
```
hilltracer:
```suggestion
if (strcmp(PQgetvalue(res, i, i_rolbypassrls), "t") == 0)
appendPQExpBufferStr(buf, " BYPASSRLS");
else
appendPQExpBufferStr(buf, " NOBYPASSRLS");
```
Looks like we already have right rolbypassrls column for all database version,
e.g. `false as rolbypassrls` for old versions

src/bin/pg_dump/pg_backup_archiver.c
```diff
@@ -3248,8 +3248,13 @@ _doSetFixedOutputState(ArchiveHandle *AH)
 	/*
 	 * Ensure that the setting is on, so that we don't error out while trying
 	 * to restore TRIGGERS from GP6 to GP7.
+	 *
+	 * This is done for the pg_restore connection itself, meaning that this
+	 * setting is set implicitly and it wouldn't appear in sql dumps
+	 * or archives. This is not ideal, but changing this code means
+	 * breaking already existing dumps/archives from GP6.
```
hilltracer, line 3255: Something like this:
```suggestion
	 *
	 * Enable statement triggers only for a live restore connection.  This
	 * is needed when restoring GP6 archives into GPDB7+, but the setting is
	 * restore-session state and should not be emitted into plain SQL dumps.
 ```

File: src/bin/pg_upgrade/test_gpdb.sh [hunk 3]
```diff
@@ -271,6 +271,13 @@ diff_and_exit() {
 	${NEW_BINDIR}/gpstop -a ${args}
 	COORDINATOR_DATA_DIRECTORY=""; unset COORDINATOR_DATA_DIRECTORY
 	PGPORT=""; unset PGPORT
+
+	# Mask database versions, as they will certainly not match during cross-version testing.
+	# There is already a vanilla PostgreSQL script that does this:
+	# src/test/perl/PostgreSQL/Test/AdjustUpgrade.pm
+	# But it performs many other subsititions, and because cross-version testing is currently
+	# broken for other reasons, it is hard to judge whether we need all of them.
```
hilltracer, line 280: Something like more oficial:

```suggestion
	# The source database version is expected to differ in cross-version
	# upgrade tests. Normalize it before comparing dumps.
```

src/bin/pg_dump/pg_backup_archiver.c
GirgerS: It it a hint that this code is not perfect. For example, it would fail 
for plan SQL dumps:

```sql
-- Source cluster setup
-- Assuming that the following trigger is created: 
-- https://www.postgresql.org/docs/9.4/trigger-example.html

CREATE TABLE ttest (
    x integer
);

CREATE TRIGGER tbefore BEFORE INSERT OR UPDATE OR DELETE ON ttest
    FOR EACH STATEMENT EXECUTE PROCEDURE trigf();
```

```bash
# Dump it, the path is local to my system
../7.x/bin/pg_dumpall > dump.sql
```

```sql
-- Target cluster
-- The trigger from the source cluster should be recreated beforehand
\i ../6.x/dump.sql
```

Then we should see an error: 
`psql:../6.x/dump.sql:155: ERROR:  Triggers for statements are not yet supported`



It doesn't affect pg_upgrade because it uses pg_restore under the hood, but still.

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: This script is still wip, because we were not able to run it on the full regression schema. So I think that is justified to have this *TODO*.

src/bin/pg_dump/pg_dumpall.c
GirgerS: Thanks. I don't know why the upstream patch decided to change this though.

src/bin/pg_upgrade/test_gpdb.sh
hilltracer: ok, just fix mistake in `subsititions`.

hilltracer (approved):

red1452 (approved):

PR #447, opened 2026-05-20
GG-425: Make temp_tablespaces test stable

ShapatinAA:
Make temp_tablespaces test stable

Previously we used get_shareinput_fileset() for getting shared fileset for shareinput scans 
except we shouldn't have been. As this function have been creating 
dsm_segments that lived until postmaster, holding segment, goes down. 

That behavior have been creating problems, when this postmasters have been reused,
because shareinput fileset are never updated. 

To keep filesets consistent this system was reworked so from now we use hashtable for
sharefilesets where handle for according fileset is placed and kept until query execution
ends (either successfully or with abort).

Co-authored-by: Maksim Michkov <m.michkov@arenadata.io>

### DIFF
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -150,7 +150,6 @@ init_tuplestore_state(ShareInputScanState *node, bool skip_waiting)
 
 				shareinput_create_bufname_prefix(rwfile_prefix, sizeof(rwfile_prefix), sisc->share_id);
 				tuplestore_make_shared_many(ts,
-											get_shareinput_fileset(),
 											rwfile_prefix,
 											sisc->nconsumers + 1);
 
@@ -231,7 +230,7 @@ init_tuplestore_state(ShareInputScanState *node, bool skip_waiting)
 			Assert(sisc->cross_slice);
 
 			shareinput_create_bufname_prefix(rwfile_prefix, sizeof(rwfile_prefix), sisc->share_id);
-			ts = tuplestore_open_shared_extended(get_shareinput_fileset(), rwfile_prefix, skip_waiting);
+			ts = tuplestore_open_shared_extended(rwfile_prefix, skip_waiting);
 
 			MemoryContextSwitchTo(old_context);
 		}
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -83,7 +83,7 @@
  * individual ShareInputScan whenever its reference count reaches zero.
  */
 static dsm_handle *shareinput_Xslice_dsm_handle_ptr;
-static SharedFileSet *shareinput_Xslice_fileset;
+/* static SharedFileSet *shareinput_Xslice_fileset; */
 
 /*
  * For local (i.e. intra-slice) variants, we use a 'shareinput_local_state'
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -81,6 +81,7 @@
  * individual ShareInputScan whenever its reference count reaches zero.
  */
 static dsm_handle *shareinput_Xslice_dsm_handle_ptr;
+static SharedFileSet *shareinput_Xslice_fileset;
 
 /*
  * For local (i.e. intra-slice) variants, we use a 'shareinput_local_state'
@@ -613,6 +614,55 @@ ShareInputShmemInit(void)
 
 	shareinput_Xslice_dsm_handle_ptr =
 		ShmemInitStruct("ShareInputScan DSM handle", sizeof(dsm_handle), &found);
+}
+
+/*
+ * Get reference to the SharedFileSet used to hold all the tuplestore files.
+ *
+ * This is exported so that it can also be used by the INITPLAN function
+ * tuplestores.
+ */
+SharedFileSet *
+get_shareinput_fileset(void)
+{
+	dsm_handle		handle;
+
+	if (shareinput_Xslice_fileset == NULL)
+	{
+		dsm_segment *seg;
+
+		LWLockAcquire(ShareInputScanLock, LW_EXCLUSIVE);
+
+		handle = *shareinput_Xslice_dsm_handle_ptr;
+
+		if (handle)
+		{
+			seg = dsm_attach(handle);
+			if (seg == NULL)
+				elog(ERROR, "could not attach to ShareInputScan DSM segment");
+			dsm_pin_mapping(seg);
+
+			shareinput_Xslice_fileset = dsm_segment_address(seg);
+		}
+		else
+		{
+			seg = dsm_create(sizeof(SharedFileSet), 0);
+			dsm_pin_segment(seg);
+			*shareinput_Xslice_dsm_handle_ptr = dsm_segment_handle(seg);
+			dsm_pin_mapping(seg);
+
+			shareinput_Xslice_fileset = dsm_segment_address(seg);
+		}
+
+		if (shareinput_Xslice_fileset->refcnt == 0)
+			SharedFileSetInit(shareinput_Xslice_fileset, seg);
+		else
+			SharedFileSetAttach(shareinput_Xslice_fileset, seg);
+
+		LWLockRelease(ShareInputScanLock);
+	}
+
+	return shareinput_Xslice_fileset;
 }
 
 void
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -66,26 +66,6 @@
 #include "port/atomics.h"
 
 /*
- * UNUSED VARIABLES. WE KEEP TO REDUCE CONFLICTS ON FUTURE POSTGRES UPDASTE.
- *
- * The tuplestore files for all share input scans are held in one SharedFileSet.
- * The SharedFileSet is attached to a single DSM segment that persists until
- * postmaster shutdown. When the reference count of the SharedFileSet reaches
- * zero, the SharedFileSet is automatically destroyed, but it is re-initialized
- * the next time it's needed.
- *
- * The SharedFileSet deletes any remaining files when the reference count
- * reaches zero, but we don't rely on that mechanism. All the files are
- * held in the same SharedFileSet, so it cannot be recycled until all
- * ShareInputScans in the system have finished, which might never happen if
- * new queries are started continuously. The shared tuplestore entries
- * are reference counted separately, and we clean up the files backing each
- * individual ShareInputScan whenever its reference count reaches zero.
- */
-static dsm_handle *shareinput_Xslice_dsm_handle_ptr;
-/* static SharedFileSet *shareinput_Xslice_fileset; */
-
-/*
  * For local (i.e. intra-slice) variants, we use a 'shareinput_local_state'
  * to track the status. It is analogous to 'shareinput_share_state' used for
  * cross-slice scans, but we don't need to keep it in shared memory. These
@@ -607,31 +587,30 @@ shareinput_create_bufname_prefix(char* p, int size, int share_id)
 /*
  * Legacy ABI entry point. Is not supported.
  *
- * Kept to reduce conflicts on future postgres updaste.
+ * Kept for ABI support.
  */
 Size
 ShareInputShmemSize(void)
 {
-	return sizeof(dsm_handle);
+	ereport(ERROR, errmsg("Unsupported sharefileset shared memory initializer."));
 }
 
 /*
  * Legacy ABI entry point. Is not supported.
  *
- * Kept to reduce conflicts on future postgres updaste.
+ * Kept for ABI support.
  */
 void
 ShareInputShmemInit(void)
 {
-	bool		found;
-
-	shareinput_Xslice_dsm_handle_ptr =
-		ShmemInitStruct("ShareInputScan DSM handle", sizeof(dsm_handle), &found);
+	ereport(ERROR, errmsg("Unsupported sharefileset shared memory initializer."));
 }
 
 /*
  * Legacy ABI entry point. Is not supported.
  *
+ * Kept for ABI support.
+ * 
  * Get reference to the SharedFileSet used to hold all the tuplestore files.
  *
  * This is exported so that it can also be used by the INITPLAN function
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -81,7 +81,6 @@
  * individual ShareInputScan whenever its reference count reaches zero.
  */
 static dsm_handle *shareinput_Xslice_dsm_handle_ptr;
-static SharedFileSet *shareinput_Xslice_fileset;
 
 /*
  * For local (i.e. intra-slice) variants, we use a 'shareinput_local_state'
@@ -614,55 +613,6 @@ ShareInputShmemInit(void)
 
 	shareinput_Xslice_dsm_handle_ptr =
 		ShmemInitStruct("ShareInputScan DSM handle", sizeof(dsm_handle), &found);
-}
-
-/*
- * Get reference to the SharedFileSet used to hold all the tuplestore files.
- *
- * This is exported so that it can also be used by the INITPLAN function
- * tuplestores.
- */
-SharedFileSet *
-get_shareinput_fileset(void)
-{
-	dsm_handle		handle;
-
-	if (shareinput_Xslice_fileset == NULL)
-	{
-		dsm_segment *seg;
-
-		LWLockAcquire(ShareInputScanLock, LW_EXCLUSIVE);
-
-		handle = *shareinput_Xslice_dsm_handle_ptr;
-
-		if (handle)
-		{
-			seg = dsm_attach(handle);
-			if (seg == NULL)
-				elog(ERROR, "could not attach to ShareInputScan DSM segment");
-			dsm_pin_mapping(seg);
-
-			shareinput_Xslice_fileset = dsm_segment_address(seg);
-		}
-		else
-		{
-			seg = dsm_create(sizeof(SharedFileSet), 0);
-			dsm_pin_segment(seg);
-			*shareinput_Xslice_dsm_handle_ptr = dsm_segment_handle(seg);
-			dsm_pin_mapping(seg);
-
-			shareinput_Xslice_fileset = dsm_segment_address(seg);
-		}
-
-		if (shareinput_Xslice_fileset->refcnt == 0)
-			SharedFileSetInit(shareinput_Xslice_fileset, seg);
-		else
-			SharedFileSetAttach(shareinput_Xslice_fileset, seg);
-
-		LWLockRelease(ShareInputScanLock);
-	}
-
-	return shareinput_Xslice_fileset;
 }
 
 void
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -150,6 +150,7 @@ init_tuplestore_state(ShareInputScanState *node, bool skip_waiting)
 
 				shareinput_create_bufname_prefix(rwfile_prefix, sizeof(rwfile_prefix), sisc->share_id);
 				tuplestore_make_shared_many(ts,
+											NULL,
 											rwfile_prefix,
 											sisc->nconsumers + 1);
 
@@ -230,7 +231,7 @@ init_tuplestore_state(ShareInputScanState *node, bool skip_waiting)
 			Assert(sisc->cross_slice);
 
 			shareinput_create_bufname_prefix(rwfile_prefix, sizeof(rwfile_prefix), sisc->share_id);
-			ts = tuplestore_open_shared_extended(rwfile_prefix, skip_waiting);
+			ts = tuplestore_open_shared_extended(NULL, rwfile_prefix, skip_waiting);
 
 			MemoryContextSwitchTo(old_context);
 		}
@@ -617,6 +618,8 @@ ShareInputShmemInit(void)
 }
 
 /*
+ * Legacy ABI entry point. Is not supported.
+ *
  * Get reference to the SharedFileSet used to hold all the tuplestore files.
  *
  * This is exported so that it can also be used by the INITPLAN function
@@ -625,44 +628,7 @@ ShareInputShmemInit(void)
 SharedFileSet *
 get_shareinput_fileset(void)
 {
-	dsm_handle		handle;
-
-	if (shareinput_Xslice_fileset == NULL)
-	{
-		dsm_segment *seg;
-
-		LWLockAcquire(ShareInputScanLock, LW_EXCLUSIVE);
-
-		handle = *shareinput_Xslice_dsm_handle_ptr;
-
-		if (handle)
-		{
-			seg = dsm_attach(handle);
-			if (seg == NULL)
-				elog(ERROR, "could not attach to ShareInputScan DSM segment");
-			dsm_pin_mapping(seg);
-
-			shareinput_Xslice_fileset = dsm_segment_address(seg);
-		}
-		else
-		{
-			seg = dsm_create(sizeof(SharedFileSet), 0);
-			dsm_pin_segment(seg);
-			*shareinput_Xslice_dsm_handle_ptr = dsm_segment_handle(seg);
-			dsm_pin_mapping(seg);
-
-			shareinput_Xslice_fileset = dsm_segment_address(seg);
-		}
-
-		if (shareinput_Xslice_fileset->refcnt == 0)
-			SharedFileSetInit(shareinput_Xslice_fileset, seg);
-		else
-			SharedFileSetAttach(shareinput_Xslice_fileset, seg);
-
-		LWLockRelease(ShareInputScanLock);
-	}
-
-	return shareinput_Xslice_fileset;
+	ereport(ERROR, errmsg("Unsupported sharefileset getter."));
 }
 
 void
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -66,6 +66,8 @@
 #include "port/atomics.h"
 
 /*
+ * UNUSED VARIABLES. WE KEEP TO REDUCE CONFLICTS ON FUTURE POSTGRES UPDASTE.
+ *
  * The tuplestore files for all share input scans are held in one SharedFileSet.
  * The SharedFileSet is attached to a single DSM segment that persists until
  * postmaster shutdown. When the reference count of the SharedFileSet reaches
@@ -602,12 +604,22 @@ shareinput_create_bufname_prefix(char* p, int size, int share_id)
 			 gp_session_id, MyProc->queryCommandId, share_id);
 }
 
+/*
+ * Legacy ABI entry point. Is not supported.
+ *
+ * Kept to reduce conflicts on future postgres updaste.
+ */
 Size
 ShareInputShmemSize(void)
 {
 	return sizeof(dsm_handle);
 }
 
+/*
+ * Legacy ABI entry point. Is not supported.
+ *
+ * Kept to reduce conflicts on future postgres updaste.
+ */
 void
 ShareInputShmemInit(void)
 {
--- a/src/backend/storage/file/sharedfileset.c
+++ b/src/backend/storage/file/sharedfileset.c
@@ -260,3 +260,43 @@ SharedFilePath(char *path, SharedFileSet *fileset, const char *name)
 	SharedFileSetPath(dirpath, fileset, ChooseTablespace(fileset, name));
 	snprintf(path, MAXPGPATH, "%s/%s", dirpath, name);
 }
+
+void
+SharedFileSetUnpin(SharedFileSet *fileset)
+{
+	bool delete_all = false;
+
+	SpinLockAcquire(&fileset->mutex);
+
+	if (fileset->refcnt == 0)
+	{
+		SpinLockRelease(&fileset->mutex);
+		elog(ERROR, "could not unpin a SharedFileSet that is already destroyed");
+	}
+
+	fileset->refcnt--;
+
+	if (fileset->refcnt == 0)
+		delete_all = true;
+
+	SpinLockRelease(&fileset->mutex);
+
+	if (delete_all)
+		SharedFileSetDeleteAll(fileset);
+}
+
+void
+SharedFileSetPin(SharedFileSet *fileset)
+{
+	SpinLockAcquire(&fileset->mutex);
+
+	if (fileset->refcnt == 0)
+	{
+		SpinLockRelease(&fileset->mutex);
+		elog(ERROR, "could not pin a SharedFileSet that is already destroyed");
+	}
+
+	fileset->refcnt++;
+
+	SpinLockRelease(&fileset->mutex);
+}
--- a/src/backend/storage/file/sharedfileset.c
+++ b/src/backend/storage/file/sharedfileset.c
@@ -260,7 +260,10 @@ SharedFilePath(char *path, SharedFileSet *fileset, const char *name)
 	SharedFileSetPath(dirpath, fileset, ChooseTablespace(fileset, name));
 	snprintf(path, MAXPGPATH, "%s/%s", dirpath, name);
 }
-
+/*
+ * Unpins shareinput fileset from HTAB.
+ * See comments on SharedFileSetPin() for more information. 
+ */
 void
 SharedFileSetUnpin(SharedFileSet *fileset)
 {
@@ -285,6 +288,18 @@ SharedFileSetUnpin(SharedFileSet *fileset)
 		SharedFileSetDeleteAll(fileset);
 }
 
+/*
+ * Beside pinning SharedFileSet with standard dsm commands
+ * exclusively for shareinput_filesets we need to make extra 
+ * pinning - as HTAB should be also considered as holder.
+ * Otherwise we might get into situation, when, for instance,
+ * writer will complete it's part of the job, detach and destroy
+ * set with on_dsm_detach(). 
+ * 
+ * When all jobs will be done, this fileset will be unpinned from HTAB's
+ * owning in tuplestore_cleanup() or AtAbort_SharedTuplestores() and
+ * cleared with according callback.
+ */
 void
 SharedFileSetPin(SharedFileSet *fileset)
 {
--- a/src/backend/storage/file/sharedfileset.c
+++ b/src/backend/storage/file/sharedfileset.c
@@ -293,7 +293,7 @@ SharedFileSetUnpin(SharedFileSet *fileset)
  * exclusively for shareinput_filesets we need to make extra 
  * pinning - as HTAB should be also considered as holder.
  * Otherwise we might get into situation, when, for instance,
- * writer will complete it's part of the job, detach and destroy
+ * writer will complete its part of the job, detach and destroy
  * set with on_dsm_detach(). 
  * 
  * When all jobs will be done, this fileset will be unpinned from HTAB's
--- a/src/backend/storage/ipc/ipci.c
+++ b/src/backend/storage/ipc/ipci.c
@@ -198,6 +198,7 @@ CreateSharedMemoryAndSemaphores(int port)
 		size = add_size(size, CheckpointerShmemSize());
 		size = add_size(size, CancelBackendMsgShmemSize());
 		size = add_size(size, WorkFileShmemSize());
+		/* size = add_size(size, ShareInputShmemSize()); */
 		size = add_size(size, SharedTuplestoreShmemSize());
 
 #ifdef FAULT_INJECTOR
@@ -379,6 +380,7 @@ CreateSharedMemoryAndSemaphores(int port)
 	AsyncShmemInit();
 	BackendCancelShmemInit();
 	WorkFileShmemInit();
+	/* ShareInputShmemInit(); */
 	SharedTuplestoreShmemInit();
 
 	/*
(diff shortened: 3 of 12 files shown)


src/backend/executor/nodeShareInputScan.c
```diff
@@ -618,6 +630,8 @@ ShareInputShmemInit(void)
 }
```
VoidZeroNull0, line 607: Should not we make it like `get_shareinput_fileset()` - only output an error because it is no longer supported?

src/backend/executor/nodeShareInputScan.c
```diff
@@ -602,12 +604,22 @@ shareinput_create_bufname_prefix(char* p, int size, int share_id)
...
+/*
+ * Legacy ABI entry point. Is not supported.
+ *
+ * Kept to reduce conflicts on future postgres updaste.
+ */
 Size
 ShareInputShmemSize(void)
 {
 	return sizeof(dsm_handle);
 }
```
VoidZeroNull0, line 596:
```suggestion
Size
ShareInputShmemSize(void)
{
	ereport(ERROR, errmsg("Unsupported share input function."));
}
```
Should not we add `ereport()` about this function being discontinued?

src/backend/executor/nodeShareInputScan.c
```diff
@@ -66,6 +66,8 @@
 #include "port/atomics.h"
 
 /*
+ * UNUSED VARIABLES. WE KEEP TO REDUCE CONFLICTS ON FUTURE POSTGRES UPDASTE.
```
VoidZeroNull0: To my knowledge, `ShareInputScan` is a node specific to Greenplum, so it does not exist in vanilla Postgres. Comments below and unused variables should be safe to remove.

src/backend/utils/sort/tuplestore.c
```diff
@@ -1949,6 +2036,9 @@ void
 tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, const char *filename, uint32 ntotal)
```
VoidZeroNull0: Seems like a good idea, to add a comment here, saying that `fileset`'s input has no effect and kept for ABI only.

src/backend/storage/ipc/ipci.c
```diff
@@ -198,7 +198,7 @@ CreateSharedMemoryAndSemaphores(int port)
 		size = add_size(size, CheckpointerShmemSize());
 		size = add_size(size, CancelBackendMsgShmemSize());
 		size = add_size(size, WorkFileShmemSize());
-		size = add_size(size, ShareInputShmemSize());
+		/* size = add_size(size, ShareInputShmemSize()); */
```
VoidZeroNull0: Should we keep at as comment and not remove it wholly?

VoidZeroNull0 (changes requested):

src/backend/utils/sort/tuplestore.c
```diff
@@ -673,23 +696,87 @@ tuplestore_cleanup(Tuplestorestate *state, bool should_abort)
...
 				 */
 				Assert(strlen(state->shared_filename) < NAMEDATALEN);
-				if (sstate->created)
-					BufFileDeleteShared(state->fileset, state->shared_filename);
+
+				strlcpy(delete_tag,
+						state->shared_filename,
+						sizeof(delete_tag));
+
+				delete_handle = sstate->sfs_handle;
```
silent-observer: Why is `delete_handle` needed? It is exactly the same as `detach_handle` in case when `should_remove_entry` is true, so can't we use `detach_handle` instead?

src/backend/utils/sort/tuplestore.c
```diff
@@ -673,23 +696,87 @@ tuplestore_cleanup(Tuplestorestate *state, bool should_abort)
...
+				delete_creator_pid = sstate->sfs_creator_pid;
+				delete_sfs_number = sstate->sfs_number;
+
+				should_delete_file = sstate->created;
+				should_release_fileset = delete_handle != DSM_HANDLE_INVALID;
```
silent-observer: Why is this a separate variable when the handle is available too?

src/backend/utils/sort/tuplestore.c
```diff
@@ -673,23 +696,87 @@ tuplestore_cleanup(Tuplestorestate *state, bool should_abort)
...
+				dsm_unpin_segment(delete_handle);
+			}
+		}
 	}
 
 	if (state->myfile)
 		BufFileClose(state->myfile);
+
+	/* Detach manually so we don't get warning */
+	if (local_detach_handle != DSM_HANDLE_INVALID)
```
silent-observer: Is it okay if we try to detach the same dsm after we already deleted it earlier?

src/backend/utils/sort/tuplestore.c
silent-observer: Is the fileset argument even still needed, if we do not initialize it here?

src/backend/utils/sort/tuplestore.c
silent-observer: I suggest also renaming `fileset` argument to something else, since names do not affect ABI. Then we can avoid calling it `real_fileset` too.

src/backend/utils/sort/tuplestore.c
```diff
@@ -2249,21 +2386,40 @@ AtAbort_SharedTuplestores()
...
+			* Also delete the shared memory entry. We do not need to delete
+			* files if the SharedFileSet was already cleaned up.
+			*/
+			if (sstate->created &&
+				fileset != NULL &&
+				sstate->sfs_creator_pid == fileset->creator_pid &&
+				sstate->sfs_number == fileset->number)
+			{
+				BufFileDeleteShared(fileset, sstate->tag);
+				sstate->created = false;
```
silent-observer: Why? Aren't we going to delete the shared entry soon anyway?

src/backend/utils/sort/tuplestore.c
```diff
@@ -1966,35 +2056,62 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
...
 
 	/*
-	 * Switch to tape-based operation, like in tuplestore_puttuple_common().
-	 * We could delay this until tuplestore_freeze(), but we know we'll have
-	 * to write everything to the file anyway, so let's not waste memory
-	 * buffering the tuples in the meanwhile.
+	 * From this point, AtAbort_SharedTuplestores() can find this state
+	 * and tuplestore_cleanup() owns num_current.
 	 */
-	PrepareTempTablespaces();
```
silent-observer, line 1979: Why is this removed?

src/backend/executor/nodeShareInputScan.c
ShapatinAA: Yes, thank you for mentioning that! I completely forgot to check it.

src/backend/utils/sort/tuplestore.c
```diff
@@ -1966,35 +2056,62 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
...
-	state->myfile = BufFileCreateShared(fileset, filename, state->work_set);
-	CurrentResourceOwner = oldowner;
-
 	/* Make sure the file only exists if and only if shared state was successfully created */
 	PG_TRY();
 	{
-		state->shared_state = get_shared_state(fileset, filename);
-		state->shared_state->num_total = ntotal;
-		state->shared_state->created = true;
+		state->myfile = BufFileCreateShared(real_fileset, filename, state->work_set);
```
silent-observer: Why do we actually need a whole try-catch for this? Originally it was needed because there could be an error in shared state creation, and the file would still not be accounted for in the abort. However, now the shared state is created first, so it should not be an issue

src/backend/utils/sort/tuplestore.c
ShapatinAA: Yes, thank you for noticing! I added `local_detach_handle` lately so, forgot to remove this one.

src/backend/utils/sort/tuplestore.c
silent-observer: Actually, do we even need to delete files one by one now? Unpinning the fileset should already automatically delete all of them. This couldn't be done before because there was only one global fileset, but now I think we don't have to keep track of individual files at all.

src/backend/utils/sort/tuplestore.c
ShapatinAA: As this function server no purpose in this context. Via it we mess with variable, which we don't use anyway here `tempTableSpaces`. The only place where it used correctly in our task is `SharedFileSetInit()`, because
```c
/* Capture the tablespace OIDs so that all backends agree on them. */
	PrepareTempTablespaces();
	fileset->ntablespaces =
		GetTempTablespaces(&fileset->tablespaces[0],
						   lengthof(fileset->tablespaces));
```
And `GetTempTablespaces()` relies on `tempTableSpaces`.

src/backend/utils/sort/tuplestore.c
ShapatinAA: Hmm, abort really can be switch to true at `AtAbort_SharedTuplestores()`, but shouldn't we bring context back?

silent-observer: What context though? PG_TRY does not attach any context to the error, the stacktrace is collected the moment an error is thrown.

src/backend/utils/sort/tuplestore.c
```diff
@@ -2006,13 +2123,18 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
...
-	state->backward = true;
-	state->status = TSS_WRITEFILE;
 
-	/* Remember this tuplestore to clear it in case of error */
-	local_shared_tuplestores = lappend(local_shared_tuplestores, state);
+	CurrentResourceOwner = oldowner;
+
+	LWLockAcquire(ShareInputScanLock, LW_EXCLUSIVE);
+	
+	state->shared_state->num_total = ntotal;
```
silent-observer: Why not set `num_total` immediately when the shared state is created? And `created` can be turned into an atomic instead of taking a lock (if `created` is still needed at all)

ShapatinAA: As I understand we can not move it in `get_shared_state()` as if we're going to change `num_total` **before** creation of `myfile`, then someone in case of abort or simple cleanup will try to delete it from `local_shared_tuplestores` and try to remove non-existing file here:
```c
if (fileset != NULL &&
	fileset->creator_pid == delete_creator_pid &&
	fileset->number == delete_sfs_number)
{
	BufFileDeleteShared(fileset, delete_tag);
}
```
Same with `created`. As we set `fileset` before creation of file and with this kind of deletion we will face problems. 
```c
if (sstate->created &&
	fileset != NULL &&
	sstate->sfs_creator_pid == fileset->creator_pid &&
	sstate->sfs_number == fileset->number)
{
	BufFileDeleteShared(fileset, sstate->tag);
	elog((Debug_shareinput_xslice ? LOG : DEBUG1), "SISC (file=%s, slice=%d): file deleted on abort",
		sstate->tag, currentSliceId);
}
```

src/backend/utils/sort/tuplestore.c
ShapatinAA: Thank you for noting! Yes, we can skip created=false part now. I used to have some weird bug with it in previous versions.
But I don't think we should remove deletion from here as this function throws an error and `DeleteAll...` does not.

src/backend/utils/sort/tuplestore.c
ShapatinAA: Didn't noticed it's static. Thank you!

src/backend/utils/sort/tuplestore.c
silent-observer: You mean the error that file doesn't exist? I don't think that's an issue, if the file somehow goes missing then even reading it will already throw an error a lot earlier. Here we just need to clean up everything, to ensure nothing is left. I think replacing all `BufFileDeleteShared` with `DeleteAll...` would simplify code by a lot.

src/backend/utils/sort/tuplestore.c
ShapatinAA: Yes, thank you! Fixed.

src/backend/executor/nodeShareInputScan.c
ShapatinAA: Thank you! Fixed

src/backend/executor/nodeShareInputScan.c
ShapatinAA: Thank you! Fixed

src/backend/executor/nodeShareInputScan.c
ShapatinAA: Fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: Thank you! Fixed.

src/backend/storage/ipc/ipci.c
ShapatinAA: Yep, fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: Fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: Fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: Fixed

src/backend/utils/sort/tuplestore.c
silent-observer: This is still relevant. It also applies to regular cleanup as well, so maybe we can remove `created` entirely if we fully switch to using shared fileset

ShapatinAA: Yes, I'm looking into it.

src/backend/utils/sort/tuplestore.c
ShapatinAA: And now to move `ntotal` to `get_shared_state()` we should make it's signature larger - as it is used in `tuplestore_open_shared_extended` also (and in this method it doesn't have `ntotal` at all). Should we consider it?

silent-observer: As far as I remember, there are cases where `ntotal` just isn't available to the reader slices at all, and `get_shared_state()` might first get opened by a reader slice. So instead we set `ntotal` to maximum value (so that no readers accidentally try to delete it too early), and then the writer sets it (since readers can't actually proceed before the writer fills the tuplestore). `ntotal` doesn't need synchronization because it is only written by a writer process during creation and read at cleanup, which is either an abort (and `ntotal` doesn't matter), or a successful cleanup (and `ntotal` must have been already set earlier, with many locking operations before that)

src/backend/storage/file/sharedfileset.c
```diff
@@ -260,3 +260,58 @@ SharedFilePath(char *path, SharedFileSet *fileset, const char *name)
...
+	if (delete_all)
+		SharedFileSetDeleteAll(fileset);
+}
+
+/*
+ * Beside pinning SharedFileSet with standard dsm commands
+ * exclusively for shareinput_filesets we need to make extra 
+ * pinning - as HTAB should be also considered as holder.
+ * Otherwise we might get into situation, when, for instance,
+ * writer will complete it's part of the job, detach and destroy
```
silent-observer:
```suggestion
 * writer will complete its part of the job, detach and destroy
```

src/backend/utils/sort/tuplestore.c
```diff
@@ -647,14 +655,21 @@ tuplestore_cleanup(Tuplestorestate *state, bool should_abort)
...
 		/*
 		 * If we are aborting, set the variable in shared memory.
 		 * This must be done before everything else, since if we can't
 		 * delete the tuplestore right now, we expect other backends to
 		 * do so after they're done using it.
 		 */
 		if (should_abort)
+		{
 			sstate->aborting = true;
+			ConditionVariableBroadcast(&sstate->ready_done_cv);
```
silent-observer: Why do we need this?

src/backend/utils/sort/tuplestore.c
silent-observer: Now that I think about it, I should have called it `ready_cv`, since it's only used with `sstate->ready` right now, and it is not used anymore for waking up the producer. Could you please rename it and fix the comment above, since we're modifying this code anyway?

src/backend/utils/sort/tuplestore.c
```diff
@@ -1966,53 +2019,69 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
...
-	local_shared_tuplestores = lappend(local_shared_tuplestores, state);
+	CurrentResourceOwner = oldowner;
+
+	LWLockAcquire(ShareInputScanLock, LW_EXCLUSIVE);
+	
+	state->shared_state->num_total = ntotal;
 
 	LWLockRelease(ShareInputScanLock);
+
+	state->backward = true;
```
silent-observer, line 2045: We can probably move these two lines back to their original place now

src/backend/utils/sort/tuplestore.c
```diff
@@ -2134,21 +2211,18 @@ tuplestore_open_shared_extended(SharedFileSet *fileset, const char *filename, bo
...
-	if (sstate->aborting) {
+	if (sstate->aborting) 
+	{
 		sstate->num_current--;
-		/*
-		 * This couldn't be a freshly created shared state,
-	     * so someone else must be holding it right now.
-		 */
-		Assert(sstate->num_current > 0);
+		LWLockRelease(ShareInputScanLock);
```
silent-observer: Why remove the assert? It still couldn't be a fresh state, because if it was, it can't be `aborting`. We never released the lock, and no one could have set it for us.

Also, no need to explicitly release LWLock here, `ereport(ERROR, ...)` automatically releases all LWLocks

src/backend/utils/sort/tuplestore.c
```diff
@@ -2171,23 +2245,38 @@ tuplestore_open_shared_extended(SharedFileSet *fileset, const char *filename, bo
 
 	state->share_status = TSHARE_READER;
 	state->frozen = false;
-	state->fileset = fileset;
 	state->shared_filename = pstrdup(filename);
 
+	/* Remember this tuplestore in case we have to close it while aborting */
+	local_shared_tuplestores = lappend(local_shared_tuplestores, state);
```
silent-observer: Now, this is an issue. `state->fileset` is not initialized here, and if an error happens before it is set (even the error you added), `tuplestore_cleanup` will try to work with not initialized memory. I think this line should remain where it was.

src/backend/utils/sort/tuplestore.c
```diff
@@ -2171,23 +2245,38 @@ tuplestore_open_shared_extended(SharedFileSet *fileset, const char *filename, bo
...
+
 	/* Only open file if requested */
 	if (!skip_open) {
 		/* Wait until writer is ready */
 		tuplestore_reader_waitready(sstate);
 
+		LWLockAcquire(ShareInputScanLock, LW_SHARED);
+
+		if (sstate->aborting)
```
silent-observer: `aborting` was used above to prevent readers from waiting on an already aborted state (because no one is going to wake them up). Why do we need an explicit cancel here?

src/test/regress/input/temp_tablespaces.source
```diff
@@ -27,6 +27,103 @@ drop tablespace some_default_tablespace;
...
+\gset
+-- end_ignore
+
+-- start_ignore
+create temp table sisc_set_pids as
+select logpid
+from gp_toolkit.gp_log_system
+where logseverity = 'LOG'
+  and logtime between :query_start_set and :query_end_set
+  and logmessage like $$%fault triggered, fault name:'sisc_xslice_temp_files' fault type:'skip'%$$;
```
silent-observer, line 95: What's the purpose of this complicated test? Weren't we just trying to check that the share input scan properly uses `temp_tablespaces`?

src/backend/utils/sort/tuplestore.c
ShapatinAA: I mean if we would to move `num_total` real initialization to `get_shared_state()`. This suggestion was discussed in the beginning of this thread. After that we can get rid of this
```c
LWLockAcquire(ShareInputScanLock, LW_EXCLUSIVE);

state->shared_state->num_total = ntotal;

LWLockRelease(ShareInputScanLock);
``` 

I thought that ntotal can be negative somehow, and that we need to have `get_shared_state(const char *filename, uint32 num_total, bool set_num_total)` signature, but in reader we can simply pass 0, and it will be a sign, that `num_total` should not be set.

src/backend/utils/sort/tuplestore.c
ShapatinAA: Thank you for noticing this! Test failed, and I think it can be the reason.

src/test/regress/input/temp_tablespaces.source
ShapatinAA: Yep. We should use correct `temp_tablespaces` in case of reusage of backends. And to make sure that we're reusing same backends I haven't found anything batter than to scan log-file on fault triggered lines, as pids of processes that used to set this variable also go with them.

silent-observer: Can you just use `pg_backend_pid()` inside the query using SISC? Then we can at least check where the query runs on. Also, I'm not sure this is even necessary, since previously there was a single global fileset, so even different PIDs will see the same problem.

silent-observer: P.S. You can also use something like `select gp_segment_id, pid from gp_stat_activity where sess_id = current_setting('gp_session_id')::int;` to obtain the list of currently active PIDs of this session. It seems like a fair assumption that if the list didn't change between before both queries and after, then backends didn't change

ShapatinAA: > `pg_backend_pid()`

What will `pg_backend_pid()` return? I didn't used it because as I understand it will not return pids, which execute slice2 of test query.

And I want to say that this is necessary because while it was global fileset, the problems with old behavior of `shareinput_fileset` was **only** and **always** connected to **reusage** of backends. So to test it exactly this - is crucial thing.

>P.S. You can also use something like select gp_segment_id, pid from gp_stat_activity where sess_id = current_setting('gp_session_id')::int; to obtain the list of currently active PIDs of this session. It seems like a fair assumption that if the list didn't change between before both queries and after, then backends didn't change

Thank you! Didn't know that. Will use.

silent-observer: > What will pg_backend_pid() return? I didn't used it because as I understand it will not return pids, which execute slice2 of test query.

It returns the PID of the process where this function is executed, it should be possible to come up with a query that executes it in the same slice as ShareInputScan. But yes, `gp_stat_activity` seems more useful in this case.

ShapatinAA: Unfortunately `gp_stat_activity` didn't worked for me. I came with this query
```sql
create temp table sisc_set_pids as
select pid, query
from gp_stat_activity 
where sess_id = current_setting('gp_session_id')::int
and query like '%with a1 as materialized%'
and query not like '%create temp table sisc_unset_pids as%';
``` 
And it gives me nothing. As if processes isn't reused. TBH I'm okay with querying logs. It's not like it is uncommon thing.

silent-observer: As far as I can tell, `query` shows the current query, so for this to work it would have to run in parallel with the SISC query.

> It's not like it is uncommon thing.

Is it? As far as I can tell, there are only 3 tests using `gp_log_system` currently (outside of gp_toolkit itself): `external_table` (which tests the log output itself), `gp_fastsequence` (which enables custom debug logging GUC to collect internal info), and `subtrx_overflow` (which also uses a custom GUC). However, using `gp_stat_activity` and `pg_stat_activity` seems to be extremely common. I think parsing logs is an overkill for this situation, since we just want to get PIDs of all currently running backends in this session.

ShapatinAA: >As far as I can tell, query shows the current query

While I was testing things out in psql I was getting this results.
```
postgres=# with a1 as materialized (select * from tts_temp)
select count(*) from a1 t1, a1 t2;
  count  
---------
 1000000
(1 row)

postgres=# select pid, query
from gp_stat_activity 
where sess_id = current_setting('gp_session_id')::int
and query like '%with a1 as materialized%'
and query not like '%create temp table sisc_unset_pids as%';
  pid   |                      query                       
--------+--------------------------------------------------
 400321 | with a1 as materialized (select * from tts_temp)+
        | select count(*) from a1 t1, a1 t2;
 400320 | with a1 as materialized (select * from tts_temp)+
        | select count(*) from a1 t1, a1 t2;
 400322 | with a1 as materialized (select * from tts_temp)+
        | select count(*) from a1 t1, a1 t2;
(3 rows)

postgres=# with a1 as materialized (select * from tts_temp)
select count(*) from a1 t1, a1 t2;
  count  
---------
 1000000
(1 row)

postgres=# select pid, query
from gp_stat_activity 
where sess_id = current_setting('gp_session_id')::int
and query like '%with a1 as materialized%'
and query not like '%create temp table sisc_unset_pids as%';
  pid   |                      query                       
--------+--------------------------------------------------
 400321 | with a1 as materialized (select * from tts_temp)+
        | select count(*) from a1 t1, a1 t2;
 400320 | with a1 as materialized (select * from tts_temp)+
        | select count(*) from a1 t1, a1 t2;
 400322 | with a1 as materialized (select * from tts_temp)+
        | select count(*) from a1 t1, a1 t2;
(3 rows)
```

Unfortunately I still couldn't managed to get processes to be reused in regression testing. 

With querying gp_stat_activity 
```sql
with a1 as materialized (select * from tts_temp)
select count(*) from a1 t1, a1 t2;
```

does not reuse workers anymore (`NOTICE:  sisc_xslice` messages show it).

I switched to isolation2 testing framework for this purposes, but it seems, that it doesn't print `NOTICE` level messages to client.

>As far as I can tell, there are only 3 tests using gp_log_system

Yea, I might exaggerated a little bit this fact... It's just I already have used it for writing some tests previously and got used to it)

src/backend/utils/sort/tuplestore.c
ShapatinAA: I thought that we shouldn't do anything below
```c
tuplestore_reader_waitready(sstate);

LWLockAcquire(ShareInputScanLock, LW_SHARED);

if (sstate->aborting)  
{
	LWLockRelease(ShareInputScanLock);
	ereport(ERROR,
			(errcode(ERRCODE_GP_OPERATION_CANCELED),
			errmsg("tuplestore operation aborted, canceling")));
}
handle = sstate->sfs_handle;

LWLockRelease(ShareInputScanLock);

fileset = attach_shareinput_fileset(handle);

state->fileset = fileset;
state->myfile = BufFileOpenShared(fileset, filename);
state->readptrs[0].file = 0;
state->readptrs[0].offset = 0L;
state->status = TSS_READFILE;
```

 in case of abortion on writer slice after waking up. But we're not doing it all in lock anyway, so can just skip that and hopefully wait for abort command.

silent-observer: Yes, it's not necessary in this case. Even if we held a lock it would have been fine, the only problem was waiting on a condition that will never happen.

src/test/regress/input/temp_tablespaces.source
silent-observer: Does the check not work without `and query like '%with a1 as materialized%'
and query not like '%create temp table sisc_unset_pids as%';`? Again, all backends are supposed to be reused for queries in one session, unless something happens in between. If this doesn't work, then yeah, I guess your log-based approach is the only viable option

ShapatinAA: Yep, without `and query like '%with a1 as materialized%' and query not like '%create temp table sisc_unset_pids as%';` it shows only current queries (which is currently running `select pid, query from gp_stat_activity...`), and this is not what we're interested in.

src/backend/utils/sort/tuplestore.c
ShapatinAA: fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: Done.

src/backend/utils/sort/tuplestore.c
ShapatinAA: yep, fixed

src/backend/storage/file/sharedfileset.c
ShapatinAA: fixed

src/backend/utils/sort/tuplestore.c
ShapatinAA: Nvm, I tested what I wanted to do and it looked bad. So I just moved `state->shared_state->num_total = ntotal;` a little bit higher.

src/backend/utils/sort/tuplestore.c
```diff
@@ -182,21 +182,20 @@ typedef struct
...
 	/*
-	 * ready_done_cv is used for signaling when the scan becomes "ready", and
-	 * when it becomes "done". The producer wakes up everyone waiting on this
-	 * condition variable when it sets ready = true. Also, when the last
-	 * consumer finishes the scan (ndone reaches nconsumers), it wakes up the
-	 * producer using this same condition variable.
+	 * ready_cv is used for signaling when the scan becomes "ready". 
+	 * The producer wakes up everyone waiting on this condition variable when
+	 * it sets ready = true. Also, when the last consumer finishes the scan
+	 * (ndone reaches nconsumers), it wakes up the producer using this same
```
silent-observer: It doesn't actually do that, this comment was outdated. Only the part until "ready = true" is still relevant

src/backend/utils/sort/tuplestore.c
```diff
@@ -1966,39 +2016,55 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
...
+		fileset = create_shareinput_fileset(&handle);
+
+		sstate->sfs_handle = handle;
+		sstate->sfs_creator_pid = fileset->creator_pid;
+		sstate->sfs_number = fileset->number;
+	}
+	else
+	{
+		handle = sstate->sfs_handle;
+		fileset = attach_shareinput_fileset(handle);
```
silent-observer: Do we need this? Who can theoretically create the fileset other than the writer? This seems to be the only usage of `create_shareinput_fileset`

src/backend/utils/sort/tuplestore.c
```diff
@@ -1966,39 +2016,55 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
 
 	LWLockAcquire(ShareInputScanLock, LW_EXCLUSIVE);
 
+	sstate = get_shared_state(filename);
+	Assert(sstate->session_id == gp_session_id);
+
+	if (sstate->aborting) {
```
silent-observer: Do we need this? The writer doesn't wait for the condition variable, so there is no chance of deadlocks even if we create the fileset (and delete it after)

src/backend/utils/sort/tuplestore.c
```diff
@@ -1966,39 +2016,55 @@ tuplestore_make_shared_many(Tuplestorestate *state, SharedFileSet *fileset, cons
...
+	local_shared_tuplestores = lappend(local_shared_tuplestores, state);
+
+	LWLockRelease(ShareInputScanLock);
 
 	/* associate the file with the store's resource owner */
 	oldowner = CurrentResourceOwner;
 	CurrentResourceOwner = state->resowner;
 
 	state->myfile = BufFileCreateShared(fileset, filename, state->work_set);
-	CurrentResourceOwner = oldowner;
```
silent-observer, line 1986: Why not just leave this line here?

src/backend/utils/sort/tuplestore.c
silent-observer: Not fixed, the lock release is still here

src/backend/utils/sort/tuplestore.c
```diff
@@ -2116,12 +2182,17 @@ tuplestore_reader_waitready(TuplestoreSharingState *sstate)
...
+ * used in current state.
  */
 Tuplestorestate *
-tuplestore_open_shared_extended(SharedFileSet *fileset, const char *filename, bool skip_open)
+tuplestore_open_shared_extended(SharedFileSet *deprecated_variable, const char *filename, bool skip_open)
 {
 	Tuplestorestate *state;
 	TuplestoreSharingState *sstate;
+	SharedFileSet *fileset = NULL;
+	dsm_handle 	handle = DSM_HANDLE_INVALID;
```
silent-observer: Do we really need these new variables? The code doesn't use them much

src/backend/utils/sort/tuplestore.c
```diff
@@ -182,21 +182,20 @@ typedef struct
...
+	 * (ndone reaches nconsumers), it wakes up the producer using this same
+	 * condition variable.
 	 */
-	ConditionVariable ready_done_cv;
+	ConditionVariable ready_cv;
 
 	/* Data to identify the SharedFileSet that was used */
 	pid_t		sfs_creator_pid;
+	dsm_handle	sfs_handle;
 	uint32		sfs_number;
```
silent-observer: We actually don't need `sfs_creator_pid` and `sfs_number` anymore, I think. This was added for a rare edge case when the global file set was recreated by someone else in the background while a backend was aborting, so the aborting backend could accidentally delete the new global file set (so it has to distinguish its old fileset from newly created one). Since we are just directly referencing the fileset here, we don't need this info anymore.

src/backend/utils/sort/tuplestore.c
ShapatinAA: Can you please clarify what do you mean? I didn't quite get... Do you suggest to inline `create_shareinput_fileset()`?

silent-observer: No, I suggest removing this condition and making `create_shareinput_fileset` unconditional, since `sstate->sfs_handle` should always be NULL here (maybe add assert?)

src/backend/utils/sort/tuplestore.c
ShapatinAA: removed

src/backend/utils/sort/tuplestore.c
ShapatinAA: I'll move fileset closer to usage.

src/backend/utils/sort/tuplestore.c
ShapatinAA: Yep, haven't noticed diff contains this change

src/backend/utils/sort/tuplestore.c
silent-observer: Can't you just use `state->fileset` without additional variable?

src/backend/utils/sort/tuplestore.c
ShapatinAA: Can we remove this also from `tuplestore_open_shared_extended`?

silent-observer: No, in `tuplestore_open_shared_extended` it is actually needed, since we have `tuplestore_reader_waitready` there. It is only unnecessary in `tuplestore_make_shared_many`

ShapatinAA: To not sit on CV? But I thought that we're getting out here by abort command from QD.

BTW what's going to happen if we're getting abort command anywhere between `LWLockRelease(ShareInputScanLock);` and `lappend(local_shared_tuplestores, state)`? Wouldn't it be dead end (memory leak), since we incremented `num_total`, but haven't yet put it for `tuplestore_cleanup()`?

silent-observer: Ah, found it, that required some research. Sorry, I was misremembering things: this doesn't prevent any deadlocks anymore (that was in the older version of the code), but it is still valid as an optimization to avoid opening the tuplestore when we already know there was an error somewhere. However, for the `tuplestore_make_shared_many` I can't think of any scenario how `aborted` flag might be set at that point, so I don't think this optimization is needed.

As for your second question, cancel command from QD can arrive at any time, however it is only processed inside `CHECK_FOR_INTERRUPTS()` calls throughout the codebase, which can then throw an error with `ERRCODE_QUERY_CANCELED`, which actually cancels the execution. So a cancel command from QD can't unexpectedly happen in the middle of our code here.

ShapatinAA: done

src/backend/utils/sort/tuplestore.c
ShapatinAA: sry... done

src/backend/utils/sort/tuplestore.c
```diff
@@ -2171,15 +2208,15 @@ tuplestore_open_shared_extended(SharedFileSet *fileset, const char *filename, bo
...
-	state->fileset = fileset;
 	state->shared_filename = pstrdup(filename);
 
 	/* Only open file if requested */
 	if (!skip_open) {
 		/* Wait until writer is ready */
 		tuplestore_reader_waitready(sstate);
 
-		state->myfile = BufFileOpenShared(fileset, filename);
+		state->fileset = attach_shareinput_fileset(sstate->sfs_handle);;
```
silent-observer:
```suggestion
		state->fileset = attach_shareinput_fileset(sstate->sfs_handle);
```

ShapatinAA: fixed

silent-observer (approved):

VoidZeroNull0 (approved):

PR #450, opened 2026-05-21
CI-5713: Replace matrix with branch-driven OS config

magf:
## Replace matrix with branch-driven OS config (#450)

- Replace static matrix with workflow-level TARGET_OS and TARGET_OS_VERSION
  environment variables derived from the triggering branch:
  - 6.x  -> ubuntu 24.04 (fixes artifact mismatch: regression tests were fully switched to ubuntu24)
  - 7.x  -> ubuntu (22.04, legacy empty version suffix)
  - next -> ubuntu (22.04, legacy empty version suffix)
- Update CI README:
  - Describe these changes
  - Add 'next' branch description added in commit eaf70fc

Task: [CI-5678](https://tracker.yandex.ru/CI-5713)

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -87,65 +87,66 @@ automatically upon the completion of the `Greengage CI` workflow.
 
 - **Triggers:** `workflow_run: workflows: ["Greengage CI"],
   types: [completed]`
-- **Branch Targeting:** Runs only for the `6.x` and `7.x` branches.
-- **Version Detection:** Automatically determines the database version (6 or 7)
-  based on the triggering branch (`6.x` → version 6, `7.x` → version 7).
-  Since `Greengage SQL Dump` is not branch-specific itself, the version is
-  derived at runtime from the branch of the triggering `Greengage CI` run.
-- **Matrix Strategy:** Runs across multiple OS configurations to generate dumps
-  for all available build targets. The set of configurations mirrors what was
-  built in `Greengage CI`:
-  - `ubuntu` (default, no version suffix — compatible with Ubuntu 22.04
+- **Branch Targeting:** Runs only for the `6.x`, `7.x`, and `next` branches.
+- **Version Detection:** Automatically determines the database version (6, 7,
+  or 8) based on the triggering branch. Since `Greengage SQL Dump` is not
+  branch-specific itself, the version is derived at runtime from the branch of
+  the triggering `Greengage CI` run.
+- **Branch-driven OS Configuration:** Instead of a matrix strategy, the target
+  OS and its version are determined at runtime from the triggering branch via
+  workflow-level environment variables `TARGET_OS` and `TARGET_OS_VERSION`:
+  - `6.x` → `ubuntu 24.04`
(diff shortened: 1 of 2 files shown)


magf:
## Tests

- `6.x`  - [Passed](https://github.com/magf/greengage/actions/runs/26208821736)
- `7.x`  - [Passed](https://github.com/magf/greengage/actions/runs/26204071509)
- `next` - not required yet

ShapatinAA (dismissed):

.github/workflows/README.md
```diff
@@ -87,65 +87,66 @@ automatically upon the completion of the `Greengage CI` workflow.
...
+  - `6.x` → `ubuntu 24.04`
+  - `7.x` → `ubuntu` (no version suffix — compatible with Ubuntu 22.04
     artifact naming)
-  - `ubuntu24.04` (version 6.x only)
+  - `next` → `ubuntu` (no version suffix)
 - **Image Pull Check:** Before creating a SQL dump, the workflow attempts to
-  pull the Docker image from GHCR. This handles cases where the matrix includes
-  OS versions for which no image was built in the triggering CI run.
+  pull the Docker image from GHCR. If the pull fails, the dump creation and
+  upload steps are skipped without failing the job.
```
ShapatinAA, line 104: Why so? Why to not to fail?

magf: Because unsuccessful workflows are skipped when searching for dumps.
A workflow error indicates a dump build process error, tests failed and dumps not created, not image and build process absent.
Now this has no practical significance at all at any completion status, but when the matrix of OS/version will returned back, it will be critically important.
Example:
- we have Ubuntu and Rockylinux;
- for Ubuntu there are images for all systems, but for Rockylinux only for 6.x and 7.x (_it’s already present, just a matter of time_) and in this case, when processing the matrix, the Rockylinux image for 8.x will not be found (and Ubuntu 24.04 for 7.x too) and if the workflow is completed with an error, it will be skipped when searching for the correct dumps and will lose its meaning.

To solve this problem, a step `Pull Docker image` was specially made to check the presence of an image, which instructs the workflow to do nothing and complete successfully if image not found.

Also, `TARGET_OS_VERSION: ${{ github.event.workflow_run.head_branch == '6.x' && '24.04' || '' }}` - _this is a temporary workaround_ for the multi-threaded matrix, which is disabled for now for simplicity, but will be brought back soon.

So there is no point in changing the termination policy now, which will not affect anything.

uaqq (dismissed):

uaqq (approved):

.github/workflows/greengage-sql-dump.yml
ShapatinAA, line 91: Should we remove `.env` here also?

magf: No, this is a GitHub Actions context, and here we tell GitHub to take parameters from the `env` section.
For GitHub, this isn't an environment variable, but just a parameter of the `env` section.
It's converted into an environment variable inside the `run` section.

This parameters - `${{ <param> }}` - will be rendered before launch the workflow and will turn into regular strings.

ShapatinAA (approved):

PR #452, opened 2026-05-21
ADBDEV-7925: Improve autocompletion for CREATE TABLE

riogor:
Some gpdb specific keywords weren't supported in psql autocompletion.
And some CREATE TABLE parts weren't supported too.

- Add DISTRIBUTED autocompletion.
- Add INHERITS autocompletion (for one option).
- Add missing storage parameters in WITH option.
- Remove redundant gp_autovacuum_scope from storage parameters (now it is a GUC and using it causes an error in CREATE TABLE).
- Add classic partitioning syntax autocompletion (SUBPARTITION BY | SUBPARTITION TEMPLATE)

Because postgres lacks proper capabilities to support autocompletion for multiple options (for table and in options) without messing up existing completions, this patch doesn't address it.
Eg
```sql
-- here it works
template1=# CREATE TABLE X (c1 int) [tab]
DISTRIBUTED   INHERITS (    PARTITION BY  TABLESPACE    USING         WITH (   

-- here it doesn't
template1=# CREATE TABLE X (c1 int) WITH (appendoptimized=TRUE) [tab]


-- here it doesn't too
template1=# CREATE TABLE X (c1 int) WITH (appendoptimized=TRUE, [tab]


```

Patch also doesn't address column specification, as it would be too messy to make and maintain, and it isn't present even in postgres 18.

Tests?
I haven't found any tests about autocompletion, so i guess it's tested only by hand.


red1452 (commented):
Is it correct behavior?
```
postgres=# create table (a int) 
(             OF            PARTITION OF  
postgres=# create table (a int) of 
gp_toolkit.          information_schema.  pg_aoseg.            pg_bitmapindex.      pg_catalog.          pg_temp_01.          pg_toast.            pg_toast_temp_01.    public.
```

riogor:
No, it is not correct, but the query is incorrect too, missing a table name.
psql has mistaken column spec for a table name.

IMHO i don't think we need to fix autocomplete for wrong queries.

red1452 (dismissed):

src/bin/psql/tab-complete.c
```diff
@@ -1069,10 +1070,17 @@ static const char *const table_storage_parameters[] = {
...
+	"checksum",
+	"compresslevel",
+	"compresstype",
 	"fillfactor",
-	"gp_autovacuum_scope",
+	"gpdr_exclude",
 	"log_autovacuum_min_duration",
+	"orientation",
 	"parallel_workers",
+	"toast_tuple_target",
```
hilltracer: Is it necessary to move it? If no - I think we shouldn't touch it since it's unrelated with patch.

src/bin/psql/tab-complete.c
```diff
@@ -1069,10 +1070,17 @@ static const char *const table_storage_parameters[] = {
...
 	"autovacuum_vacuum_scale_factor",
 	"autovacuum_vacuum_threshold",
+	"blockdirectory",
+	"blocksize",
+	"checksum",
+	"compresslevel",
+	"compresstype",
 	"fillfactor",
-	"gp_autovacuum_scope",
+	"gpdr_exclude",
```
hilltracer: gpdr_exclude what is the option? I can't find it in **our** documentation
```sql
gpadmin=# create table t (a int) with ( gpdr_exclude = true);
ERROR:  unrecognized parameter "gpdr_exclude"
```

src/bin/psql/tab-complete.c
```diff
@@ -2510,12 +2524,28 @@ psql_completion(const char *text, int start, int end)
 	/* Complete PARTITION BY with RANGE ( or LIST ( or ... */
 	else if (TailMatches("PARTITION", "BY"))
 		COMPLETE_WITH("RANGE (", "LIST (", "HASH (");
+	/* Complete PARTITION BY RANGE|LIST with classic partitioning syntax */
+	else if (TailMatches("PARTITION", "BY", "RANGE|LIST", "(*)") ||
+		TailMatches("SUBPARTITION", "TEMPLATE", "(*)"))
+		COMPLETE_WITH("(", "SUBPARTITION BY");
```
hilltracer, line 2528: Looks like this condition is to broad. Can we do something with some wrong completions like
```sql
ALTER TABLE sales SET SUBPARTITION TEMPLATE () [tab]
(                SUBPARTITION BY  
```
?

src/bin/psql/tab-complete.c
```diff
@@ -1069,10 +1070,17 @@ static const char *const table_storage_parameters[] = {
 	"autovacuum_vacuum_cost_limit",
 	"autovacuum_vacuum_scale_factor",
 	"autovacuum_vacuum_threshold",
+	"blockdirectory",
```
hilltracer:
```sql
gpadmin=# create table t (a int) with ( blockdirectory = 'qwe');
ERROR:  unrecognized parameter "blockdirectory"
```

hilltracer (commented):
I added some comments but in general LGTM.

src/bin/psql/tab-complete.c
riogor: Hm, i have sorted options, but can revert that.

src/bin/psql/tab-complete.c
riogor: Hm, i've took it from greenplum 7.x docs, haven't checked that it isn't present in greengage.

src/bin/psql/tab-complete.c
riogor: Same as with gpdr_exclude

src/bin/psql/tab-complete.c
riogor: I think yes, this modification will do
```c
	else if (TailMatches("PARTITION", "BY", "RANGE|LIST", "(*)") ||
		TailMatches(MatchAnyExcept("SET"), "SUBPARTITION", "TEMPLATE", "(*)"))
		COMPLETE_WITH("(", "SUBPARTITION BY");
```
Haven't found `SUBPARTITION` anywhere else

hilltracer (approved):

red1452 (approved):

riogor:
Merge commit message:

```
Improve autocompletion for CREATE TABLE (#452)

Some gpdb-specific keywords and mechanisms weren't supported in psql 
autocompletion for CREATE TABLE command.

Add keywords autocompletion: DISTRIBUTED, INHERITS.
Add classic partitioning syntax autocompletion via SUBPARTITION BY 
and SUBPARTITION TEMPLATE.
Add missing storage parameters in WITH option.
Remove redundant gp_autovacuum_scope from storage parameters.

Don't address autocompletion support for multiple consecutive options, because 
postgres lacks proper capabilities to do it without messing up existing
completions.

Ticket: ADBDEV-7925
```

PR #455, opened 2026-05-22
CI-5581: Port CI infrastructure from 7.x to 8.x

magf:
## Port CI infrastructure from 7.x to 8.x

- [ci: add GitHub Actions](https://github.com/GreengageDB/greengage/commit/63da817dfe6f75d9b682af27491e3e969926ca91)
  - GHA Greengage CI
  - CI README
- [ci: add CODEOWNERS @GreengageDB/ci for .github](https://github.com/GreengageDB/greengage/pull/455/commits/9360d3885bee9291b2fd262b3e2c03643b599749)
- [ci: bump to latest v39](https://github.com/GreengageDB/greengage/pull/455/commits/b668c21b8cde67c5cdc0c189afe5c1254d9e0858)
- [ci: create symlik ci/ to arenadata/](https://github.com/GreengageDB/greengage/pull/455/commits/18bcb51be6e89e6b45ac63841fbb0e54d291c27c)
- [ci: add resgroup launcher script](https://github.com/GreengageDB/greengage/pull/455/commits/2c60cbe2d29d44de7a407b54c8d00b7922c9b4b9)
- [fix: handle missing git tags when generating changelog](https://github.com/GreengageDB/greengage/commit/f344cc3ff7c1271c5bba3928a7071c58e1d7471f)
  Prevent script failure when no previous tag exists in repository by
  making `git describe` gracefully fall back to empty string.
  - Use `2>/dev/null || echo ''` to avoid errors when `HEAD^` has no tag
  - Change log header to dynamically show range or full history
  - Use parameter expansion for git log range (empty `PREV_TAG` shows full history)
  This ensures `git-current-changelog.txt` is always generated even in
  repositories without any tags, displaying full git log instead.
- [fix: docker-compose --> docker compose](https://github.com/GreengageDB/greengage/commit/cd08f81fd58a7556432172545ddd0ac98c01ed59)
  Launcher 'docker-compose' deprecated; using `docker compose` plugin
- [fix: sync docker-compose with 7.x branch](https://github.com/GreengageDB/greengage/pull/455/commits/5798a7b942d339dee02c9a3383510fa2ccc4075e)
  - extract common service config into x-common-service YAML anchor
  - add sqldump and per-node log volume mounts (cdw, sdw1-3)
  - remove redundant `version: "3"` key

Task: [CI-5581](https://tracker.yandex.ru/CI-5581)


.github/workflows/greengage-abi-tests.yml
```diff
@@ -0,0 +1,162 @@
+name: Greengage ABI Tests
```
Stolb27: In my opinion, ABI tests are redundant for the next branch. We are breaking ABI every batch of Postgres commits. It's expected.

.github/workflows/greengage-abi-tests.yml
```diff
@@ -0,0 +1,162 @@
...
+      EXCEPTION_LISTS_COUNT: ${{ steps.check_exception_lists.outputs.EXCEPTION_LISTS_COUNT }}
+    steps:
+      - name: Fetch source
+        uses: actions/checkout@v3
+
+      - name: Get Greengage version variables
+        id: vars
+        run: |
+          remote_repo='https://github.com/GreengageDB/greengage.git'
+          git ls-remote --tags --refs --sort='v:refname' $remote_repo '8.*' | grep -v '-' | tail -n 1  > baseline_version_ref
```
bandetto: I think this baseline lookup comes up empty on 8.x. There are no `8.*` tags upstream in GreengageDB/greengage yet, so `git ls-remote ... '8.*'` returns nothing and `BASELINE_VERSION` ends up empty. Two knock-on effects: the `build-baseline` job then checks out GreengageDB/greengage with an empty `ref`, which silently falls back to the default branch instead of a real 8.0 baseline, and the exception-list check below runs `ls .abi-check//`, which can report a misleading count. Maybe we should gate this job on a baseline tag existing, or note that the ABI check stays dormant until the first 8.x tag lands.

.github/workflows/greengage-abi-tests.yml
```diff
@@ -0,0 +1,162 @@
...
+          ref: ${{ matrix.ref }}
+          submodules: recursive
+          fetch-depth: 1
+          path: gpdb_src
+
+      - name: Install dependencies and abi-dumper
+        run: |
+          sudo gpdb_src/README.Ubuntu.bash
+          sudo apt install -y libuv1-dev libpam0g-dev libldap-dev libipc-run-perl universal-ctags abi-dumper
+          sudo ln -fs python2 /usr/bin/python
```
bandetto: Minor: this still symlinks `python2`, and the file mixes `actions/checkout@v3` (lines 19, 64) with the `@v4` used for the artifact actions. I think both are carried over from 7.x. Could we confirm `python2` is actually needed for the 8.x build path, and maybe bump checkout to a current major (v5) for consistency?

.github/workflows/greengage-ci.yml
```diff
@@ -0,0 +1,164 @@
...
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        target_os: [ubuntu]
+    permissions:
+      contents: read  # Explicit for default behavior
+      packages: write # Required for GHCR access
+      actions: write  # Required for artifact upload
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-build.yml@CI-5720
```
bandetto: I think these reusable-workflow refs are pinned to the dev branch `@CI-5720`, while greengage-release.yml pins `@v35`. Flagging so the dev-branch ref isn't forgotten -- maybe we should move both onto stable tags before this leaves WIP.

.github/workflows/README.md
```diff
@@ -0,0 +1,245 @@
...
+**reusable workflows**) is **added, removed, or renamed**, you must contact a
+repository administrator to update the **Branch Protection Rules** accordingly.
+Without this, new, deleted, or renamed jobs will not be recognized as required
+when checking Pull Requests.
+
+## Overview
+
+The `Greengage CI` workflow triggers on:
+
+- **Push events** to versioned release branches (`6.x`, `7.x`, `next`) after
```
bandetto: This says push events trigger on `6.x`, `7.x`, `next` and tags `6.*`, `7.*`, `8.*`, but greengage-ci.yml here only triggers on push to `next` and tags `8.*`. I think this section was inherited from the 7.x docs -- maybe we should trim it to what this branch actually does.

.github/workflows/README.md
```diff
@@ -0,0 +1,245 @@
...
+  - Orca tests
+  - Resource group tests
+  - JIT tests (versions 7.x, 8.x)
+- **Upload**: Retags and pushes final Docker images to GHCR and optionally
+  DockerHub. Runs for push to the default branch (retags to `latest`) and tags
+  after build.
+- **Package**: Builds Debian packages and optionally tests deployment.
+  Currently supported for versions `6.x`, `7.x`.
+
+## Release Workflow
```
bandetto: This Release Workflow section describes a cache-based flow with manual recovery and cache-miss handling, but the actual greengage-release.yml here is a 25-line job that just uploads the `deb-packages` artifact. I think this describes the external reusable workflow rather than the file in this repo -- maybe we should note that or trim it.

.github/workflows/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+A separate workflow `Greengage release` handles the uploading of Debian package
+to GitHub releases. It is triggered when a release is published and uses a
+composite action to manage package deployment.
+
+### Key Features
+
+- **Triggers:** `release: [published]` — runs when a release is published,
```
bandetto: Small mismatch: this says the trigger is `release: [published]` including re-publishing, but greengage-release.yml uses `types: [released]`. Those differ -- `released` does not fire for prereleases, while `published` does. Maybe we should align the doc with the file (or vice versa).

.github/workflows/README.md
```diff
@@ -0,0 +1,245 @@
...
+2. **Cache Miss Scenarios:**
+   - **No previous build or previous build successful:** Provides instructions
+     to manually trigger the CI build, then restart the release workflow.
+   - **Previous build failed:** Reports the failure with a link to the failed
+     run and requires manual fixing before retrying.
+
+The release workflow is designed to be robust and provide clear feedback when
+issues occur, ensuring that releases are always consistent and reliable.
+
+## SQL Dump Workflow
```
bandetto: I do not see a SQL Dump workflow file in `.github/workflows/` on this branch, but this section (and the `behave tests gpexpand` consumer it mentions) documents one in detail. I think it is inherited from 7.x -- maybe we should drop it here until that workflow is actually ported.

.github/workflows/README.md
```diff
@@ -0,0 +1,245 @@
...
+   permissions for GHCR access and `actions: write` for artifact upload.
+2. Configure DockerHub credentials (`DOCKERHUB_TOKEN`, `DOCKERHUB_USERNAME`)
+   for DockerHub uploads:
+   - For `greengagedb/greengage`: mandatory — login failure will stop the
+     workflow.
+   - For other repositories: optional — login failure is allowed and other
+     processes (GHCR upload, etc.) are unaffected.
+3. Configure the version and target OS parameters in the branch-specific
+   workflow configuration.
+4. Create a pull request or push to the default branch (`7.x`) to trigger the
```
bandetto: I think this should say `next` rather than `7.x` -- the CI workflow on this branch triggers on push to `next`, not `7.x`.

.github/workflows/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Additional Documentation
+
+Detailed README files for each process are available in the directory [README](https://github.com/greengagedb/greengage-ci/blob/main/README/)
+of the `greengagedb/greengage-ci` repository:
+
+- Build process:
+  [README/REUSABLE-BUILD.md](https://github.com/greengagedb/greengage-ci/blob/main/README/REUSABLE-BUILD.md)
+- Package process:
+  [README/REUSABLE-PACKAGE.md](https://github.com/greengagedb/greengage-ci/blob/main/README/README/REUSABLE-PACKAGE.md)
```
bandetto: Looks like a broken link: the path has `README/` twice (`.../README/README/REUSABLE-PACKAGE.md`). I think it should be `.../README/REUSABLE-PACKAGE.md`.

.github/workflows/greengage-abi-tests.yml
magf: Removed

.github/workflows/greengage-ci.yml
magf: Yes, thanks - it's under control)
This [PR](https://github.com/GreengageDB/greengage-ci/pull/54) for branch `CI-5720` in CI is special for this task and still in progress too.

.github/workflows/greengage-release.yml
```diff
@@ -0,0 +1,25 @@
+# Release creation workflow
+name: Greengage release
```
RekGRpth: Is this workflow necessary?

RekGRpth:
> upd: add `ci_path: arenadata` to `build` and `behave`, `orca`, `resgroup` tests workflows

Is this really good for `greengage` to do this?

.github/workflows/greengage-abi-tests.yml
magf: ABI tests was removed totally

.github/workflows/greengage-abi-tests.yml
magf: ABI tests was removed totally

magf:
> > upd: add `ci_path: arenadata` to `build` and `behave`, `orca`, `resgroup` tests workflows
> 
> Is this really good for `greengage` to do this?

@Stolb27, did I understand correctly that there was a requirement to implement CI without changing the repository structure?
Then I don't understand the meaning of this question.

.github/workflows/greengage-release.yml
magf: The requirements do not point to exclude `Greengage release` CI.
@Stolb27, we need to exclude this?

.github/workflows/README.md
magf: Thanks!
Fixed.

.github/workflows/README.md
magf: Fixed

.github/workflows/README.md
magf: - `Packages` removed. From workflow file too.
- `Release Workflow` updated, truncated.

.github/workflows/README.md
magf: Updated.

.github/workflows/README.md
magf: `SQL Dump` section totally removed.

.github/workflows/README.md
magf: All about `6.x`, `7.x` was removed.
Actualized for `next` branch only.

.github/workflows/greengage-release.yml
Stolb27: We need to adapt gpAux/Makefile changes to build release artifacts. I think it's a separate task with porting other fixes from 7.x. Does it work somehow without these changes? @magf

Stolb27:
> > > upd: add `ci_path: arenadata` to `build` and `behave`, `orca`, `resgroup` tests workflows
> > 
> > 
> > Is this really good for `greengage` to do this?
> 
> @Stolb27, did I understand correctly that there was a requirement to implement CI without changing the repository structure? Then I don't understand the meaning of this question.

Yes, you should change repo as less as possible. But a symlink `ci->arendata` may be lesser evil. But it's up to you.

.github/workflows/greengage-release.yml
magf: Yes, Release creation workflow removed.

magf:
> a symlink `ci->arendata`

Done.
Also, target CI to `v36` tag (latest).

ci
```diff
@@ -0,0 +1 @@
+arenadata/
```
RekGRpth: this simlink does not work
https://github.com/GreengageDB/greengage/actions/runs/26440118735/job/77853113963?pr=455#step:11:14

magf: It's sad. I'll probably have to go back to CI parameterization.

RekGRpth: in adb simlink works https://github.com/arenadata/gpdb/blob/adb-6.x/arenadata

magf: It's not symlink. It was symlink.
This is the directory now.

RekGRpth: why not simlink like in adb?

magf: this is not a symlink `ci` --> `arenadata`
this is reverse symlink `arenadata` --> `ci`

`ci` - is the directory.

RekGRpth: maybe remove slash?

RekGRpth: like in https://github.com/arenadata/gpdb/commit/51b119a8705321d87655056fadbf1d5b92727b9e

magf: The problem has been found and fixed.
Testing now.
Cause: when creating a virtual machine, CI scripts are copied from the container as a directory. In our case, a symlink was copied instead of a directory.
Solution:
```diff
diff --git a/.github/workflows/greengage-reusable-tests-resgroup.yml b/.github/workflows/greengage-reusable-tests-resgroup.yml
index a0a78e8..ba5ddfe 100644
--- a/.github/workflows/greengage-reusable-tests-resgroup.yml
+++ b/.github/workflows/greengage-reusable-tests-resgroup.yml
@@ -104,14 +104,14 @@ jobs:
           systemctl enable --now docker
 
           cd /home/gpadmin
-          mkdir -p logs
+          mkdir -p logs ci
 
           docker load < $IMAGE_FILE
           rm $IMAGE_FILE
 
           # Extract test script from Image instaed of GIT
           docker create --name temp-container $IMAGE
-          docker cp temp-container:/home/gpadmin/gpdb_src/ci ./ci
+          docker cp temp-container:/home/gpadmin/gpdb_src/ci/. ./ci/
           docker rm temp-container
           EOF0
 ```

Comment (magf) [reply ci]: Well, in order to run the script, it must be there )

File: .github/workflows/greengage-ci.yml [hunk 13]
```diff
@@ -0,0 +1,142 @@
...
+    if: github.event_name == 'pull_request'  # Only for PR
+    strategy:
+      fail-fast: false
+      matrix:
+        target_os: [ubuntu]
+    permissions:
+      contents: read  # Explicit for default behavior
+      packages: read  # Explicit for GHCR access clarity
+      actions: write  # Required for artifact upload
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-tests-resgroup.yml@fix/semantics-docker-cp-symlink
```
bandetto: I think this resgroup ref is still pinned to a feature branch `@fix/semantics-docker-cp-symlink`, while the other reusable workflows here moved onto `@v36`. Maybe we should pin it to a stable tag before this leaves WIP, so the dev-branch ref isn't forgotten.

magf: All CI bumped to v41 latest

RekGRpth:
LGTM

bandetto (dismissed):

RekGRpth:
@magf, could you please add a skip tag to the behave tests that aren't running on GitLab?

magf:
> @magf, could you please add a skip tag to the behave tests that aren't running on GitLab?

Enabling and disabling tests is beyond my competence.
Those decisions are made by the developer.

RekGRpth (dismissed):

arenadata/scripts/run_resgroup_test.bash
```diff
@@ -0,0 +1,39 @@
...
+  set -ex
+  source \$GPHOME/greengage_path.sh
+  source gpdb_src/gpAux/gpdemo/gpdemo-env.sh
+  make -C /home/gpadmin/gpdb_src/src/test/regress
+  make PGOPTIONS='-c optimizer=$OPTIMIZER -c statement_mem=$STATEMENT_MEM' installcheck-resgroup -C gpdb_src/
+" && exitcode=0
+
+echo \$exitcode > /logs/.exitcode
+exit \$exitcode
+EOF
```
bandetto, line 39: `regression.diffs` with failed tests is not printed to stderr on failure. It's a convenienve feature that all current pipelines follow.

Stolb27 (approved):

arenadata/scripts/run_resgroup_test.bash
bandetto: This is a nit, and after a discussion we decided this imporvement would be better moved to other patch, since 6.x and 7.x missed this too.

bandetto (approved):

PR #457, opened 2026-05-25
GG-502: Fix pg_dump not assigning oids for array types

GirgerS:
Partitioning implementations in Greengage 6 and Greengage 7
differ in array type creation: 
- in Greengage 6, array type is created only for the root 
  table in a hierarchy
- in Greengage 7, they are created for every table in a
  hierarchy

This difference caused pg_restore to fail, as there where
no preassigned OIDs for these types in the dump.

This patch makes pg_dump assign unused OIDs for 
these types.

One notable exception is AO tables, because array types
are not created for them. So leave their dump logic 
unchanged.

pg_upgrade test errors.

### DIFF
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -319,7 +319,8 @@ static void binary_upgrade_set_type_oids_by_type_oid(Archive *fout,
 													 bool force_array_type);
 static void binary_upgrade_set_type_oids_by_rel(Archive *fout,
 													PQExpBuffer upgrade_buffer,
-													const TableInfo *tblinfo);
+													const TableInfo *tblinfo,
+													bool force_array_type);
 static void binary_upgrade_set_pg_class_oids(Archive *fout,
 											 PQExpBuffer upgrade_buffer,
 											 Oid pg_class_oid, bool is_index);
@@ -4937,11 +4938,12 @@ binary_upgrade_set_type_oids_by_type_oid(Archive *fout,
 static void
 binary_upgrade_set_type_oids_by_rel(Archive *fout,
 										PQExpBuffer upgrade_buffer,
-										const TableInfo *tblinfo)
+										const TableInfo *tblinfo,
+										bool force_array_type)
 {
 	TypeInfo *typinfo = findTypeByOid(tblinfo->reltype);
 	binary_upgrade_set_type_oids_by_type_oid(fout, upgrade_buffer,
-											 typinfo, false);
+											 typinfo, force_array_type);
 }
 
 static void
@@ -16891,6 +16893,7 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 	char	   *ftoptions = NULL;
 	char	   *srvname = NULL;
 	char	   *partkeydef = NULL;
+	bool	   force_array_type = false;
 
 	/* We had better have loaded per-column details about this table */
 	Assert(tbinfo->interesting);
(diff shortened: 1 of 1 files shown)


GirgerS:
All differences are due to this piece of code: https://github.com/GreengageDB/greengage/blob/6.x/src/backend/catalog/heap.c#L1557-L1578

src/bin/pg_dump/pg_dump.c
```diff
@@ -17071,8 +17075,20 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
...
+						if (!tbinfo->aotbl) {
+							/*
+							 * Array types for children of a partitioned table are created
+							 * only starting from Greengage 7, so specify arbitrary OIDs for
+							 * them.
+							 *
+							 * Also, don't do it for AO tables, array types are not created
+							 * for them.
+							 */
+							force_array_type = true;
```
bandetto, line 17087: I think `force_array_type` should be reset at the start of each loop iteration, or computed locally like `bool force_array_type = (tbinfo->aotbl == NULL);` right before the call. Right now it is declared once at the top of `dumpTableSchema()` and only ever flipped to `true`, so once a heap child sets it, every AO child that follows in the same partition list would also be dumped with a synthesised array OID. The comment one line up says "don't do it for AO tables", so I think the intent is per-iteration.

src/bin/pg_dump/pg_dump.c
```diff
@@ -17071,8 +17075,20 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 						if (tbinfo->relstorage == 'x')
 							hasExternalPartitions = true;
 
+						if (!tbinfo->aotbl) {
+							/*
+							 * Array types for children of a partitioned table are created
```
bandetto, line 17080: Maybe worth tightening "arbitrary OIDs" here, since `binary_upgrade_set_type_oids_by_type_oid()` actually walks `next_possible_free_oid` and probes `pg_type` to skip duplicates rather than picking arbitrarily. Same phrasing appears in the PR description.

src/bin/pg_dump/pg_dump.c
```diff
@@ -16905,8 +16908,9 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
 		pg_log_warning("WITH OIDS is not supported anymore (table \"%s\")",
 					   qrelname);
 
-	if (dopt->binary_upgrade)
-		binary_upgrade_set_type_oids_by_rel(fout, q, 	tbinfo);
+	if (dopt->binary_upgrade) {
```
bandetto: The rest of this file uses Allman braces, and there is also a stray hard tab between `q,` and `tbinfo` carried over from the original line. Could you match local style here and on the new block below.

bandetto (changes requested):
meow meow meow

src/bin/pg_dump/pg_dump.c
GirgerS: Thanks! I missed the fact that a partitioned hierarchy can have tables with different access methods.

src/bin/pg_dump/pg_dump.c
GirgerS: Reflexes.

src/bin/pg_dump/pg_dump.c
GirgerS: Yeah, I meant something like `we don't care which oid we pick as long as it doesn't collide with existing ones`. So I agree, `unused OIDs` or something similar fits this description better.

bandetto:
Can tests be added?

GirgerS:
> Can tests be added?

This patch is a fix of one of the errors in the pg_upgrade test `src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh`.

bandetto (approved):

red1452 (approved):

PR #463, opened 2026-05-28
Fix line endings (CR LF -> LF) for files in ggrebalance

whitehawk:
Fix line endings (CR LF -> LF) for files in ggrebalance

Errors were detected on ALT linux due to "trailing \<CR\> in interpreter" for
some files in the ggrebalance tool. This patch fixes the files of ggrebalance
where CR LF were found to LF line endings.


ShapatinAA:
But how did this happen? Ain't git converts this things according to its settings? Should we consider changing `.gitattributes` to eliminate this kind of error in the future?

bimboterminator1 (approved):

whitehawk:
> But how did this happen? Ain't git converts this things according to its settings? Should we consider changing `.gitattributes` to eliminate this kind of error in the future?

It looks that we need to change `.gitattributes` , as now it has no line ending conversion setting.

red1452 (approved):

PR #465, opened 2026-05-28
GG-410: ALTER RESOURCE GROUP inside transaction

hilltracer:
ALTER RESOURCE GROUP did not support execution inside transaction blocks
because resource group changes are hard to roll back safely: they affect both
catalog state and in-memory/shared-memory state, and uncommitted changes must
block other resource group modifications.

Introduce an opt-in postmaster GUC that lifts the restriction. With it on,
several ALTERs run in one transaction and roll back together if any of them
fails or if the user chooses to abort. The catalog write happens during the
ALTER itself, while the in-memory cgroup and shared memory apply still run at
top-level commit.

Check callbacks at PRE_COMMIT against the final pg_resgroupcapability state, so
ALTERs rolled back by subtransactions or overwritten by later ALTERs are not
applied. Preserve CPUSET callbacks whose side effects remain in the final catalog
state, for example when a later CPU_MAX_PERCENT or CPU_WEIGHT change overwrites
only part of the CPUSET change.

CREATE and DROP keep the transaction block restriction. Reject moving a query
to another resource group while resource group capabilities are being edited by
an uncommitted transaction.

This patch adapts the 6.x change 303e4a0. Most differences come from cgroup v2 
support, which added new parameters such as IO_LIMIT.

Other notable changes:

A rolled-back subtransaction won't apply its ALTER

This patch stamps each callback with its registers a callback that frees the
callbacks of an aborted subtransaction, and reassigns the callbacks of a
committed subtransaction to the parent.

Deduplication keeps the earlier duplicate and corrupts final state

The patch replaces the earlier duplicate with the later one and adds a
CPUSET-versus-CPU ordering heuristic, so a kept CPUSET callback is always
followed by the CPU callback that overrides it.

A no-op query move is silently allowed

The patch, on lock failure, resolves the target group and the session with
non-raising lookups and returns true when the move is a provable no-op.

PRE_COMMIT loop exception safety

The patch wraps the loop and marks pointers volatile across the longjmp
boundary, and moves the rejected list into transaction memory so the abort path
reclaims it with the transaction context.

Co-authored-by: Vladimir Sarmin \<v.sarmin@arenadata.io\>


hilltracer:
For testing with cgroup v2
- take fixes from #471: `git merge origin/GG-530`
- follow this message https://github.com/GreengageDB/greengage/pull/471#issuecomment-4591968773

hilltracer:
Do not add segment-level io.max checks for IO_LIMIT. The current test infrastructure does not let us reuse the existing PL/Python helper on QEs cleanly: it runs into distributed execution limitations.

This is fine for this patch. IO_LIMIT goes through the same transactional ALTER callback path as the other resource group parameters, and those parameters are already checked on segments. For IO_LIMIT we still check the catalog state and coordinator cgroup v2 io.max state.

hilltracer:
For reviewers: The main 6.x-to-7.x adaptation changes are grouped in commit c456bfdb1099.

hilltracer:
Do not add `resgroup/resgroup_io_limit_alter_in_transaction` test in v1 sheduler since io_limit works and tests only for cgroup v2

RekGRpth:
How do you plan to merge this PR? With squash? With rebase?

RekGRpth:
> (cherry picked from commit https://github.com/GreengageDB/greengage/commit/303e4a0fbbe9745be114acd29b086bfbfadf17f7)

Where is the description of changes compared to the original patch?

src/backend/commands/resgroupcmds.c
```diff
@@ -1172,6 +1176,43 @@ dropResgroupCallback(XactEvent event, void *arg)
...
+ *
+ * dumpio() returns a palloc'ed string and also allocates temporary
+ * StringInfo state, so call this only in a short-lived context.
+ */
+static bool
+resGroupIOLimitMatches(List *left, List *right)
+{
+	char   *leftStr;
+	char   *rightStr;
+	bool	result;
```
RekGRpth:
```suggestion
	char	   *leftStr;
	char	   *rightStr;
	bool		result;
```

hilltracer:
> Where is the description of changes compared to the original patch?

I'm going to squash, since we have many changes with 303e4a0.

hilltracer:
> Where is the description of changes compared to the original patch?

Added short description. I've tried to describe most of changes in code comments.
```
This patch adapts the 6.x change 303e4a0f. Most differences come from cgroup v2
support, which added new parameters such as IO_LIMIT.
```

src/backend/commands/resgroupcmds.c
hilltracer: hm.. please explain your suggestion. I don't see the reason of this change.

RekGRpth: pgindent

src/backend/utils/resgroup/resgroup.c
```diff
@@ -2919,17 +2987,40 @@ resgroupDumpWaitQueue(StringInfo str, PROC_QUEUE *queue)
...
+	appendStringInfo(str, "{\"%d\":%d},",
+					 RESGROUP_LIMIT_TYPE_MIN_COST,
+					 caps->min_cost);
+	/*
+	 * NOTE: Return a placeholder instead of calling dumpio(), which creates
+	 * palloc'ed temporary state.
+	 */
+	appendStringInfo(str, "{\"%d\":%d}",
+					 RESGROUP_LIMIT_TYPE_IO_LIMIT,
+					 0);
```
RekGRpth: What's wrong with that if the memory can be `pfree`d after adding it to `str`?

src/backend/utils/resgroup/resgroup_helper.c
```diff
@@ -479,6 +481,18 @@ pg_resgroup_move_query(PG_FUNCTION_ARGS)
...
+		 * Fail if any resource group is being edited.
+		 * ALTER holds ExclusiveLock until end of transaction.
+		 * RowShareLock allows concurrent moves but conflicts with it.
+		 */
+		if (!ConditionalLockRelationOid(ResGroupCapabilityRelationId,
+										RowShareLock))
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_OBJECT_IN_USE),
+							(errmsg("cannot move query while a resource group is being edited"))));
```
RekGRpth:
```suggestion
					(errmsg("cannot move query while a resource group is being edited"))));
```

src/backend/commands/resgroupcmds.c
```diff
@@ -1121,27 +1176,266 @@ dropResgroupCallback(XactEvent event, void *arg)
...
+		cgroupOpsRoutine->freeio(ctx->oldCaps.io_limit);
+
+	pfree(ctx);
+}
+
+/*
+ * Compare two io_limit lists through their canonical catalog representation.
+ *
+ * dumpio() returns a palloc'ed string and also allocates temporary
+ * StringInfo state, so call this only in a short-lived context.
```
RekGRpth: Why do you need this comment if you are forcibly freeing memory?

src/backend/utils/resgroup/resgroup.c
```diff
@@ -2919,17 +2987,40 @@ resgroupDumpWaitQueue(StringInfo str, PROC_QUEUE *queue)
...
+					 caps->cpuWeight);
+	appendStringInfo(str, "{\"%d\":\"%s\"},",
+					 RESGROUP_LIMIT_TYPE_CPUSET,
+					 caps->cpuset);
+	appendStringInfo(str, "{\"%d\":%d},",
+					 RESGROUP_LIMIT_TYPE_MEMORY_LIMIT,
+					 caps->memory_limit);
+	appendStringInfo(str, "{\"%d\":%d},",
+					 RESGROUP_LIMIT_TYPE_MIN_COST,
+					 caps->min_cost);
```
RekGRpth, line 3030: Why was it necessary to refactor this function?

src/backend/utils/resgroup/resgroup.c
hilltracer: The problem is that dumpio() (which references to io_limit_dump) [allocates StringInfo](https://github.com/GreengageDB/greengage/blob/3bbcacdc1ef33a20b77a7fed0596b0b6c691cad3/src/backend/utils/resgroup/cgroup_io_limit.c#L528) but doesn't free it. We can free only result but it's not the all StringInfo.

src/backend/commands/resgroupcmds.c
hilltracer: See my previous comment about dump info

src/backend/utils/resgroup/resgroup.c
hilltracer: It's not refactoring it's correction. Previous implementation was wrong since handled RESGROUP_LIMIT_TYPE_CPUSET and IO_LIMITb incorrectly, which led to problems with pointers. Because it thoughts that these were all numbers, although they were strings. This only applies to 7.x. With 6.x its fine

src/backend/utils/resgroup/resgroup.c
RekGRpth: replace `StringInfo result = makeStringInfo();` with
```c
StringInfoData result;
initStringInfo(&result);
```

src/backend/utils/resgroup/resgroup_helper.c
hilltracer: done

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/utils/resgroup/resgroup.c
hilltracer: reworked

src/backend/commands/resgroupcmds.c
hilltracer: reworked

src/backend/utils/resgroup/resgroup.c
hilltracer: I have segfault on segment when I used it in test with non-default io_limit. So, will research it.

src/backend/utils/resgroup/resgroup.c
```diff
@@ -2920,15 +2988,41 @@ resgroupDumpWaitQueue(StringInfo str, PROC_QUEUE *queue)
...
+	if (caps->io_limit != NIL)
 	{
-		appendStringInfo(str, "{\"%d\":%d}", i, caps[i]);
-		if (i < RESGROUP_LIMIT_TYPE_COUNT - 1)
-			appendStringInfo(str, ",");
+		char *io_limit_str = NULL;
+		io_limit_str = cgroupOpsRoutine->dumpio(caps->io_limit);
```
bandetto: I think this newly dereferences a dangling pointer. After `ALTER RESOURCE GROUP ... SET io_limit` commits, `ResGroupAlterOnCommit` stores `ctx->caps.io_limit` into the shared `group->caps.io_limit`, and then `freeResgroupAlterCallback` (resgroupcmds.c:1183) frees that same list. The old `resgroupDumpCaps` treated caps as an int array and never followed `io_limit`, but this version calls `dumpio(caps->io_limit)`, so a later `pg_resgroup_get_status_kv('dump')` reads freed memory. The list is also backend local, so reading it from another backend is invalid regardless of the free. Maybe we should keep `io_limit` out of the shared caps, or skip it in the dump.

src/backend/commands/resgroupcmds.c
```diff
@@ -1121,27 +1176,263 @@ dropResgroupCallback(XactEvent event, void *arg)
...
+	ListCell   *lc;
 
-	if (event == XACT_EVENT_COMMIT)
-		ResGroupAlterOnCommit(callbackCtx);
+	if (resgroup_alter_callbacks == NIL)
+		return;
 
-	if (callbackCtx->caps.io_limit != NIL)
-		cgroupOpsRoutine->freeio(callbackCtx->caps.io_limit);
+	if (event == XACT_EVENT_ABORT)
```
bandetto, line 1513: I think the prepare path is not covered here. `PrepareTransaction` fires `XACT_EVENT_PREPARE` and `FinishPreparedTransaction` fires no callbacks, so a transaction that prepares while `resgroup_alter_callbacks` is populated never clears the list. Because the list is backend static and the callback stays registered for the backend lifetime, the next ordinary COMMIT on the reused backend would apply those leftover contexts. The replaced `RegisterXactCallbackOnce` was invoked at `XACT_EVENT_PREPARE` and freed its context there, so maybe we should also handle `XACT_EVENT_PREPARE` and the parallel commit and abort events here.

src/backend/utils/resgroup/resgroup.c
hilltracer: Reverted default value here, because we can't dump. See new comment in code:
```
* IO_LIMIT is reported as the default value. ResGroupCaps is shared state,
 * but io_limit is a backend-local List * pointer. Status dump may read this
 * state from a different backend, where the pointer is invalid.
```

src/backend/commands/resgroupcmds.c
hilltracer: done

src/backend/utils/resgroup/resgroup.c
hilltracer: Reverted to only output as default value. See [this](https://github.com/GreengageDB/greengage/pull/465#discussion_r3341073331) comment

src/backend/commands/resgroupcmds.c
hilltracer: PREPARE TRANSACTION is not available as a user-facing command:

```sql
gpadmin=# BEGIN;
BEGIN
gpadmin=# ALTER RESOURCE GROUP rg_prepare_test SET CONCURRENCY 20;
ALTER RESOURCE GROUP
gpadmin=# PREPARE TRANSACTION 'rg_prepare_gid';
ERROR:  PREPARE TRANSACTION is not yet supported in Greengage Database
```

So this path is only used by the internal distributed 2PC protocol.

I also checked the segment-side flow. The callback is reached from PrepareTransaction() during internal 2PC prepare, before the final commit path. In my tests it can be called during prepare before the final commit callback is reached, so we should not clear resgroup_alter_callbacks on XACT_EVENT_PREPARE.

src/backend/utils/resgroup/resgroup.c
bandetto: I think this newly dereferences a dangling pointer. After `ALTER RESOURCE GROUP ... SET io_limit` commits, `ResGroupAlterOnCommit` stores `ctx->caps.io_limit` into the shared `group->caps.io_limit`, and then `freeResgroupAlterCallback` (resgroupcmds.c:1183) frees that same list. The old `resgroupDumpCaps` treated caps as an int array and never followed `io_limit`, but this version calls `dumpio(caps->io_limit)`, so a later `pg_resgroup_get_status_kv('dump')` reads freed memory. The list is also backend local, so reading it from another backend is invalid regardless of the free. Maybe we should keep `io_limit` out of the shared caps, or skip it in the dump.

src/backend/commands/resgroupcmds.c
bandetto, line 1513: I think the prepare path is not covered here. `PrepareTransaction` fires `XACT_EVENT_PREPARE` and `FinishPreparedTransaction` fires no callbacks, so a transaction that prepares while `resgroup_alter_callbacks` is populated never clears the list. Because the list is backend static and the callback stays registered for the backend lifetime, the next ordinary COMMIT on the reused backend would apply those leftover contexts. The replaced `RegisterXactCallbackOnce` was invoked at `XACT_EVENT_PREPARE` and freed its context there, so maybe we should also handle `XACT_EVENT_PREPARE` and the parallel commit and abort events here.

src/backend/commands/resgroupcmds.c
```diff
@@ -1121,27 +1176,263 @@ dropResgroupCallback(XactEvent event, void *arg)
...
+		 * side effects remain in the final catalog state.
+		 */
+		rel = heap_open(ResGroupCapabilityRelationId, AccessShareLock);
+
+		foreach (lc, resgroup_alter_callbacks)
+		{
+			ResourceGroupCallbackContext *ctx = lfirst(lc);
+			ResGroupCaps finalCaps;
+
+			GetResGroupCapabilities(rel, ctx->groupid, &finalCaps);
```
bandetto: I think an error thrown here turns a commit into an abort. `GetResGroupCapabilities` runs once per queued callback and can raise, for example when an `io_limit` value cannot be parsed. If it raises inside this loop, `heap_close` is skipped and the `prepared` and `rejected` lists allocated in `TopMemoryContext` leak, and the user sees a transaction that issued only ALTER statements abort at COMMIT. Maybe we should release the relation and the lists in a `PG_CATCH`, or validate earlier so this path cannot raise.

src/backend/utils/resgroup/resgroup_helper.c
```diff
@@ -479,6 +481,18 @@ pg_resgroup_move_query(PG_FUNCTION_ARGS)
 			ereport(ERROR,
 					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
 							(errmsg("cannot move myself"))));
+		/*
+		 * Fail if any resource group is being edited.
+		 * ALTER holds ExclusiveLock until end of transaction.
+		 * RowShareLock allows concurrent moves but conflicts with it.
+		 */
+		if (!ConditionalLockRelationOid(ResGroupCapabilityRelationId,
```
bandetto, line 491: I think this guard has two gaps. The matching `UnlockRelationOid` is only reached on the success path, so every `ereport` between here and the unlock holds the lock until the surrounding transaction ends rather than releasing it as the success path does. More importantly, now that `ALTER RESOURCE GROUP` can run inside a transaction, a session that already altered a group in the same transaction holds `ExclusiveLock`, so this conditional `RowShareLock` is granted against its own lock and the guard passes in exactly the case it means to block. Maybe we should detect an in-progress edit a different way.

hilltracer: 1. The lock is transaction-scoped. If an error happens, abort cleanup releases it anyway, so adding manual unlocks only for error paths unnecesorry.
2. If the same transaction does ALTER RESOURCE GROUP and then moves a query into that state. I think it's the user problem, he should know that new settings for resource group will be apply after commit. I would first test the behavior, document the current semantics, and align it with the architect.

src/backend/commands/resgroupcmds.c
```diff
@@ -1121,27 +1176,263 @@ dropResgroupCallback(XactEvent event, void *arg)
...
+{
+	char       *leftStr;
+	char       *rightStr;
+	bool        result;
+
+	if (left == NIL || right == NIL)
+		return left == NIL && right == NIL;
+
+	leftStr = cgroupOpsRoutine->dumpio(left);
+	rightStr = cgroupOpsRoutine->dumpio(right);
```
whitehawk: Is it possible to get a case when `left` and `right` consist of the same elements, but sorted in a different order? In this case `leftStr` will not be equal to `rightStr`, though the limits are the same...

src/backend/commands/resgroupcmds.c
```diff
@@ -1121,27 +1176,263 @@ dropResgroupCallback(XactEvent event, void *arg)
...
+			after = true;
+			continue;
+		}
+
+		if (!after || other->groupid != ctx->groupid)
+			continue;
+
+		/*
+		 * Example:
+		 *   SET CPUSET '0'; SET CPU_MAX_PERCENT 30;
```
whitehawk: In the documentation it is said "Consider keeping CPU core 0 unassigned. CPU core 0 is used as a fallback mechanism ". So I think that using configuration anti-patterns ("SET CPUSET '0'") in the comments should be avoided if possible.

src/backend/utils/resgroup/cgroup_io_limit.c
```diff
@@ -520,6 +520,26 @@ compare_iostat(const void *x, const void *y)
...
+	TblSpcIOLimit *b = (TblSpcIOLimit *) lfirst(*(ListCell **) y);
+
+	if (a->tablespace_oid != b->tablespace_oid)
+	{
+		if (a->tablespace_oid < b->tablespace_oid)
+			return -1;
+		return 1;
+	}
+
+	return 0;
```
RekGRpth:
```suggestion
	if (a->tablespace_oid == b->tablespace_oid)
		return 0;

	if (a->tablespace_oid < b->tablespace_oid)
		return -1;

	return 1;
```

RekGRpth:
new flaky?
```diff
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/regress/expected/gp_query_id_optimizer.out /home/gpadmin/gpdb_src/src/test/regress/results/gp_query_id.out
--- /home/gpadmin/gpdb_src/src/test/regress/expected/gp_query_id_optimizer.out	2026-06-03 21:49:10.553459361 +0000
+++ /home/gpadmin/gpdb_src/src/test/regress/results/gp_query_id.out	2026-06-03 21:49:10.599460466 +0000
@@ -789,10 +802,8 @@
     raise warning '%', 2;
     raise warning '%', 3;
 end$$; | QUERY ID: 6
-NOTICE:  END ExecutorRun | Q: select 1 / i from t | QUERY ID: 7
 NOTICE:  END ExecutorStart | Q: select 1 / i from t | QUERY ID: 7
 NOTICE:  END ProcessUtility | Q: do $$
-NOTICE:  START ExecutorRun | Q: select 1 / i from t | QUERY ID: 7
 NOTICE:  START ExecutorStart | Q: select 1 / i from t | QUERY ID: 7
 WARNING:  2
 WARNING:  3
```

src/backend/commands/resgroupcmds.c
```diff
@@ -1121,27 +1177,301 @@ dropResgroupCallback(XactEvent event, void *arg)
...
+		foreach (lc, resgroup_alter_callbacks)
+		{
+			ResourceGroupCallbackContext *ctx = lfirst(lc);
+			ResGroupAlterOnCommit(ctx);
+			freeResgroupAlterCallback(ctx);
+		}
+		list_free(resgroup_alter_callbacks);
+		resgroup_alter_callbacks = NIL;
+	}
+	else if (event == XACT_EVENT_PRE_COMMIT)
```
whitehawk, line 1535: Why don't we handle here `XACT_EVENT_PARALLEL_PRE_COMMIT`, if we've added `XACT_EVENT_PARALLEL_ABORT` and `XACT_EVENT_PARALLEL_COMMIT` above? And do we really need them? Parallel execution is disabled in GPDB.

src/backend/utils/resgroup/resgroup_helper.c
```diff
@@ -485,13 +485,25 @@ pg_resgroup_move_query(PG_FUNCTION_ARGS)
...
+			if (groupId == InvalidOid ||
+				groupId == SYSTEMRESGROUP_OID ||
+				sessionId == -1 ||
+				IsResGroupBypassedBySessionId(sessionId) ||
+				ResGroupGetGroupIdBySessionId(sessionId) != groupId)
+				ereport(ERROR,
+						(errcode(ERRCODE_OBJECT_IN_USE),
+						(errmsg("cannot move query while a resource group is being edited"))));
+
+			PG_RETURN_BOOL(true);
```
hilltracer, line 506: IMHO, we don't need this complication. We will write in the documentation that the move call is prohibited if ALTER is in progress. And this should work clearly, but here we get different behavior, which can be confusing. In addition, this is a significant difference compared to 6.x

src/backend/utils/resgroup/cgroup_io_limit.c
```diff
@@ -520,6 +520,25 @@ compare_iostat(const void *x, const void *y)
...
+
+	if (a->tablespace_oid == b->tablespace_oid)
+		return 0;
+
+	if (a->tablespace_oid < b->tablespace_oid)
+		return -1;
+
+	return 1;
+}
+
```
hilltracer, line 541: Could you please add a test for io_limit order-independent comparison?

src/backend/commands/resgroupcmds.c
```diff
@@ -1379,48 +1534,111 @@ alterResgroupCallback(XactEvent event, void *arg)
...
+		/*
+		 * GetResGroupCapabilities can raise on a malformed stored value, which
+		 * would turn this commit into an abort. Release the relation, the
+		 * TopMemoryContext list and the parsed io_limit on that path so they
+		 * do not leak, then let the abort proceed. The variables read after
+		 * the longjmp are volatile. The rejected list lives in transaction
+		 * memory and needs no cleanup here. The contexts themselves stay
+		 * owned by resgroup_alter_callbacks, which the abort callback frees.
+		 */
+		PG_TRY();
```
hilltracer, line 1559: Can GetResGroupCapabilities() really fail in normal ALTER? The value is already validated before writing to pg_resgroupcapability, so this looks like catalog corruption or an internal bug.

Do we really need this extra PG_TRY/volatile complexity justified for such a case?

Note that in 6.x we don't worry about it.

src/backend/utils/resgroup/resgroup_helper.c
bandetto: The path returns true only when the move is a provable no-op, the session is already in the target group, so nothing is read and nothing moves. Every other lock failure still raises `OBJECT_IN_USE`. I think erroring on a genuine no-op is worse than the small divergence from 6.x, so I would keep it and document the semantics.

Also I'm planning to list all diverges between this task and the 6.x task for further improvements of the latter if they're desired.

src/backend/utils/resgroup/cgroup_io_limit.c
bandetto: Done.

src/backend/commands/resgroupcmds.c
bandetto: What about OOM in the loop?

hilltracer: OOM here seems very unlikely: the loop is small and does not process user data.
If you really want to keep this complication because of TopMemoryContext, I am ok.

hilltracer:
Approved. (I can't do it in gh interface since I've created this PR, so just write this message)

whitehawk (approved):

silent-observer (approved):

PR #467, opened 2026-05-29
ADBDEV-9949: Allow background mode for abitrary command in isolation2

dkovalev1:
ADB 6.x commit 0bb081e in particular allowed to run shell commands in 
a background in isolation2 tests.

This commit adapts this feature to the 7x branch.

We need this feature for PXF tests so GG7 can run mock PXF server.

Additionally force output of shell command to the text mode to easier 
results comparison.

ADBDEV-9949

### DIFF
--- a/src/test/isolation2/sql_isolation_testcase.py
+++ b/src/test/isolation2/sql_isolation_testcase.py
@@ -696,6 +696,11 @@ class SQLIsolationExecutor(object):
         if not flag:
             if sql.startswith('!'):
                 sql = sql[1:]
+                # Support run command in background
+                bg_mode = False
+                if sql.startswith('&'):
+                    bg_mode = True
+                    sql = sql[1:]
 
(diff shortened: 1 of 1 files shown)


RekGRpth:
> soGG7

so GG7

RekGRpth:
Perhaps we should add a test for the new functionality?

dkovalev1:
> Perhaps we should add a test for the new functionality?

This is a minor change to the test infrastructure ported from other branch with a small addition. It does not even related to the sql and queries. AFAIK we do not even have specific tests for testing tests infrastructure yet. I could not find the obvious place to put the test and was hoping that we can avoid it.

Do you suggest making new test for testing test suite? What place for the test can you suggest?

RekGRpth (approved):

KnightMurloc (approved):

PR #469, opened 2026-05-29
GG-511: Fix pg_dump transferring TOASTs for partitioned tables

GirgerS:
TOAST tables are not created for partitioned tables 
starting from Greengage 7, but pg_dump still created
them during migration, resulting in a failure to
upgrade. 

Have pg_dump omit those values and pg_upgrade 
ignore them during file transfer to avoid the problem.
Also, simplify logic for TOAST tables over AOCO
relations and add an explanatory comment telling
why file transfer logic for partitioned tables diverges 
from the original code. 

pg_upgrade test errors.

### DIFF
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -4951,6 +4951,11 @@ binary_upgrade_set_pg_class_oids(Archive *fout,
 								 PQExpBuffer upgrade_buffer, Oid pg_class_oid,
 								 bool is_index)
 {
+	PGresult * res;
+	PQExpBuffer gpdb6_partition_query;
+	bool gpdb6_partitioned_table;
+	bool aoco;
+
 	if (!is_index)
 	{
 		TableInfo *tblinfo = findTableByOid(pg_class_oid);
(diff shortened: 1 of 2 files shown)


GirgerS:
In case this question would appear: it is probably possible to collect information about indices over TOAST tables directly in the query inside `get_rel_infos`, to skip additional binary search. But I fell like keeping changes more local is better, and it also seems more straightforward for me. An overall, after these changes the code is still (relatively) fast. For tables created by repeating the query from the parent ticket 1000 times, it takes around 2.5 seconds to go over them

bimboterminator1 (dismissed):
LGTM

src/bin/pg_dump/pg_dump.c
```diff
@@ -4951,6 +4951,11 @@ binary_upgrade_set_pg_class_oids(Archive *fout,
 								 PQExpBuffer upgrade_buffer, Oid pg_class_oid,
 								 bool is_index)
 {
+	PGresult * res;
```
red1452:
```suggestion
	PGresult *res;
```

red1452 (commented):
LGTM, but there is small remark of code style.

red1452 (approved):

bimboterminator1 (approved):

PR #470, opened 2026-06-01
CI-5769: Make deb artifact and cache names OS-specific 7.x

magf:
## Make deb artifact and cache names OS-specific 7.x (#470)

ci(package): adapt package matrix for OS-specific artifact naming
- convert matrix from list to include format
- add target_os_version to matrix and pass it to reusable workflow
- derive test_docker from matrix (target_os:target_os_version || 22.04)

- bump greengage-reusable-package to v40
  - append target_os+target_os_version to artifact names and cache keys
  - fix: make sigar listing optional in deb install test

Task: [CI-5769](https://tracker.yandex.ru/CI-5769)

### DIFF
--- a/.github/workflows/greengage-ci.yml
+++ b/.github/workflows/greengage-ci.yml
@@ -146,15 +146,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target_os: [ubuntu]
+        include:
+          - target_os: ubuntu
     permissions:
       contents: read  # Explicit for default behavior
       packages: write # Required for GHCR access
       actions: write  # Required for artifact upload
-    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-package.yml@v28
+    uses: greengagedb/greengage-ci/.github/workflows/greengage-reusable-package.yml@v38
     with:
       version: 7
       target_os: ${{ matrix.target_os }}
-      test_docker: ubuntu:22.04 # Docker Image  (e.g., ubuntu:22.04, ubuntu:noble) for deploy test. Skip if empty
+      target_os_version: ${{ matrix.target_os_version }}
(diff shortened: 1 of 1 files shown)


magf:
## Tests
- Developer [PR](https://github.com/magf/greengage/pull/52)
  - Package [build](https://github.com/magf/greengage/actions/runs/26745364042/job/78821735167?pr=52)
  - Package build [test](https://github.com/magf/greengage/actions/runs/26745364042/job/78823289625?pr=52)
  - Artifact [deb-packages-ubuntu22.04.zip](https://github.com/magf/greengage/actions/runs/26745364042/artifacts/7327884307)

uaqq (approved):

.github/workflows/greengage-ci.yml
```diff
@@ -146,15 +146,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target_os: [ubuntu]
+        include:
+          - target_os: ubuntu
```
dkovalev1, line 150: Isn't it the same? What the point of this change? GH docs [talks](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategymatrixinclude) about include only in the context of multi-dimensional matrix.

magf: Yes, second parameter is `target_os_version` not required for `ubuntu`.
For `24.04` only. For example from `6.x`:
```yaml
matrix:
  include:
    - target_os: ubuntu
    - target_os: ubuntu
      target_os_version: "24.04"
```

Will be used soon.

---

Also, `target_os_version: ""` could have been added for clarity, but it would have disrupted the overall picture, because in other places this parameter is omitted.

---

### upd

We still don't specify a version for Ubuntu 22.04 because it will break backward compatibility with older pipelines, but we are already on the way to solving this problem.

dkovalev1 (approved):

PR #474, opened 2026-06-03
Allow user to cancel ggrebalance if it hangs on catalog lock

whitehawk:
Allow user to cancel ggrebalance if it hangs on catalog lock

Problem description:
If there was an open transaction which blocked catalog locking (performed via
'SELECT gp_expand_lock_catalog()'), ggrebalance hanged after its schema
creation, and the user was not able to interrupt it with a signal.

Root cause:
During the long operation of a C-function (like the DB driver), if it is
executed from the main thread, the signal handler will be executed only when
the low-level function is complete.
So, we were waiting inside the driver function on 'gp_expand_lock_catalog()'
till the lock was released on the DB side, and only after that was the signal
handler invoked.

Solution:
Execute all potentially long queries (like catalog locking or checkpoints) from
a separate thread. Therefore, the main thread is able to invoke the signal
handler and cancel the pending operation.
Plus, add more logging when locking and unlocking the catalog, so the user gets
more useful hints about why the tool is blocked for a valid reason.

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/shrink.py
+++ b/gpMgmt/bin/ggrebalance_modules/shrink.py
@@ -6,6 +6,7 @@
 from transitions import Machine
 from contextlib import closing
 from typing import Any
+from queue import Queue
 
 try:
     from gppylib.commands.unix import *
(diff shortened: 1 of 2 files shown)


VoidZeroNull0 (approved):
LGTM.

GirgerS (approved):

PR #478, opened 2026-06-08
GG-540: Install deps for plpython3 in the python2 env

Stolb27:
8f9aa03 switches all regression and isolation2 tests to use plpython3.
But some tests require additional python modules. Moreover, part of them
require PyGreSQL that depends on Greengage libPQ. And currently, this
module is built only for default python interpreter as a part of
Greengage build process. 8f9aa03 tries to pass Greengage PYTHONPATH to
plpython3. As a result, tests work correctly only on python 3 builds
(e.g. Ubuntu 24). Otherwise, plpython3 environment is polluted by
python 2 modules. To run isolation2 tests in the python 2 env we should
provide tests dependencies for python 3 additionally.

The biggest problem is PyGreSQL that depends on Greengage artifacts and
customized to be wrapped in the directory of the same name. There already 
was an issue with moving shared library of the module to the proper 
directory level during migration to python 3. Also, this approach makes 
impossible to use common installation with pip. Further more, there 
is 10 years old issue with pickling PyGreSQL exceptions:

```
>>> import pickle
>>> from pygresql import pg
>>> pickle.dumps(pg.DatabaseError())
  File "<stdin>", line 1, in <module>
_pickle.PicklingError: Can't pickle <class 'pg.DatabaseError'>: import
 of module 'pg' failed
```

To solve this problem I suggest:
1. exclude moving PyGreSQL modules to the dedicated directory to build
   vanilla module as is with flat structure;
2. provide additional python module that implements shim for Greengage;
3. build PyGreSQL wheels on build Dockerfile stage if needed;
4. install them on test stage;

This approach will be useful to migrate PYTHONPATH to the virtual env
later.

Also, 
- source Greengage python environment explicitly to run python 2
cluster utilities in the plpython3 environment;
- lsof is required to source greengage_path.sh if shell is not bash/zsh;
- test relying on pickling bug reworked.

Ticket: GG-540

----------------

The PR is split on separated commits, I suggest merge them as is by rebase.


ci/Dockerfile.ubuntu
```diff
@@ -49,6 +49,15 @@ ENV TARGET_OS=ubuntu \
...
     gpdb_src/concourse/scripts/compile_gpdb.bash
 
+# Build python modules for isolation2 tests
+# PyGresSQL requires libpq, so build wheel here to install later
+RUN source /usr/local/greengage-db-devel/greengage_path.sh; \
+    unset PYTHONPATH; \
+    export PYTHON_SRC="/home/gpadmin/gpdb_src/gpMgmt/bin/pythonSrc/"; \
+    apt update && apt install -y python3-pip; \
+    python3 -m pip wheel $PYTHON_SRC/PyGreSQL-5.2.5 -w bin_gpdb/; \
+    python3 -m pip wheel $PYTHON_SRC/PyGreSQL-compat -w bin_gpdb/;
```
bandetto:
```suggestion
RUN source /usr/local/greengage-db-devel/greengage_path.sh && \
    unset PYTHONPATH && \
    export PYTHON_SRC="/home/gpadmin/gpdb_src/gpMgmt/bin/pythonSrc/" && \
    apt update && apt install -y python3-pip && \
    python3 -m pip wheel $PYTHON_SRC/PyGreSQL-5.2.5 -w bin_gpdb/ && \
    python3 -m pip wheel $PYTHON_SRC/PyGreSQL-compat -w bin_gpdb/
```

Stolb27: I'll prepend with `set -eux` as above, as was discussed

Stolb27: Fixed

silent-observer (approved):

bandetto (approved):

PR #481, opened 2026-06-09
GG-542: Check EXECUTE on coordinator functions before upgrade

GirgerS:
Starting from Greengage 7, functions marked as 
EXECUTE ON COORDINATOR or
EXECUTE ON MASTER 
should always return SETOF values.

Check for such functions before performing upgrade.

pg_upgrade test errors.

### DIFF
--- a/src/bin/pg_upgrade/greengage/check_gp.c
+++ b/src/bin/pg_upgrade/greengage/check_gp.c
@@ -1308,7 +1308,8 @@ check_execute_on_master_functions()
 			 */
 			script = fopen(output_path, "w");
 			if (!script)
-				pg_fatal("Could not create necessary file:  %s\n", output_path);
+				pg_fatal("could not open file \"%s\": %s\n",
+					output_path, strerror(errno));
 		}
 
 		fprintf(script, "Database: %s\n", active_db->db_name);
--- a/src/bin/pg_upgrade/greengage/check_gp.c
+++ b/src/bin/pg_upgrade/greengage/check_gp.c
@@ -33,6 +33,7 @@ static void check_views_with_removed_functions(void);
 static void check_views_with_removed_types(void);
 static void check_for_disallowed_pg_operator(void);
 static void check_views_with_changed_function_signatures(void);
+static void check_execute_on_master_functions(void);
 
 /*
  *	check_greengage
@@ -58,6 +59,7 @@ check_greengage(void)
 	check_views_with_removed_types();
 	check_for_disallowed_pg_operator();
 	check_views_with_changed_function_signatures();
+	check_execute_on_master_functions();
 }
 
 /*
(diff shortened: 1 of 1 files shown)


GirgerS:
An example of such function:
```sql
CREATE FUNCTION return_void() RETURNS void
    LANGUAGE plpgsql EXECUTE ON MASTER
    AS $$
BEGIN
        return;
END
$$;
```

Before this patch, pg_upgrade would give the following errors:
```
Restoring database schemas in the new cluster
  template1
*failure*

Consult the last few lines of "/home/regrig/programming/greengage-7.x/src/bin/pg_upgrade/tmp_check/datadirs/qddir/demoDataDir-1/pg_upgrade_output.d/20260609T110659.226/log/pg_upgrade_dump_1.log" for
the probable cause of the failure.
Failure, exiting
```

```
pg_restore: error: could not execute query: ERROR:  EXECUTE ON COORDINATOR is only supported for set-returning functions
```

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1247,3 +1249,98 @@ check_views_with_changed_function_signatures()
...
+	snprintf(output_path, sizeof(output_path), "%s/%s",
+			 log_opts.basedir, "execute_on_master_functions_not_returning_setof_rows.txt");
+
+	for (dbnum = 0; dbnum < old_cluster.dbarr.ndbs; dbnum++)
+	{
+		PGresult   *res;
+		int			ntups;
+		int			rowno;
+		DbInfo	   *active_db = &old_cluster.dbarr.dbs[dbnum];
+		PGconn	   *conn;
```
red1452: Too many spaces between type and name of variables.

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1247,3 +1249,98 @@ check_views_with_changed_function_signatures()
...
+	int   i_proname;
+	int   i_args;
+	int   i_nspname;
+
+	prep_status("Checking EXECUTE ON MASTER functions");
+
+	snprintf(output_path, sizeof(output_path), "%s/%s",
+			 log_opts.basedir, "execute_on_master_functions_not_returning_setof_rows.txt");
+
+	for (dbnum = 0; dbnum < old_cluster.dbarr.ndbs; dbnum++)
```
red1452: Move declarion to the "for" block.

```suggestion
	for (int dbnum = 0; dbnum < old_cluster.dbarr.ndbs; dbnum++)
```

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1247,3 +1249,98 @@ check_views_with_changed_function_signatures()
...
+
+		if (!script)
+		{
+			/*
+			 * This is the first database that has affected functions,
+			 * try to open the output file
+			 */
+			script = fopen(output_path, "w");
+			if (!script)
+				pg_fatal("Could not create necessary file:  %s\n", output_path);
```
red1452: Should you close the connection here?

red1452: Add error message to the output.

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1247,3 +1249,98 @@ check_views_with_changed_function_signatures()
...
+	if (GET_MAJOR_VERSION(old_cluster.major_version) > 904)
+		return;
+
+	char  output_path[MAXPGPATH];
+	FILE *script = NULL;
+	bool  found = false;
+	int   dbnum;
+	int   i_proname;
+	int   i_args;
+	int   i_nspname;
```
red1452: The last 3 variables are used in the "for" block only. Move declaration to there.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: I changed it for the version that also reports errno. Not sure if it is what you meant though.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: It is done the same way as in the check right above. And it so happens that is declares these variables in the outermost scope. 

The styling is pretty haphazard, and variables are declared wherever (in some places they are outside of the for loop ([here](https://github.com/GreengageDB/greengage/blob/f0067beceadc107d0e460a2bbae3329adac506b1/src/bin/pg_upgrade/check.c#L1306-L1312)), in others they are [inside](https://github.com/GreengageDB/greengage/blob/f0067beceadc107d0e460a2bbae3329adac506b1/src/bin/pg_upgrade/check.c#L1429) it. 
So I'd say that it is fine to leave this code as is.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: From what I've seen, they only connection reported by `ss` is automatically closed on `pg_upgrade` exit. So I'd guess so.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: I copied this from the check right above. It is a common [practice](https://github.com/GreengageDB/greengage/blob/f0067beceadc107d0e460a2bbae3329adac506b1/src/bin/pg_upgrade/check.c#L1366-L1367) in these files 
for some reason.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: The same thing: for some reason it is usually done this way.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Let me know if it is OK. I'll change the other PR then.

red1452: Yes, it is OK.

src/bin/pg_upgrade/greengage/check_gp.c
red1452: I saw the same functions in the file, but it does not mean, that it is a good idea to copy-paste this code. I think, you should do it more optimal (and readable).

src/bin/pg_upgrade/greengage/check_gp.c
red1452: I don't think so. `pg_fatal` ends with `exit(1)`.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Agreed.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: By the OS, not by us. And considering that similar code is still in the latest [postgres](https://github.com/postgres/postgres/blob/master/src/bin/pg_upgrade/check.c#L1129-L1132), I doubt that it caused errors.

red1452 (approved):

silent-observer (approved):

PR #482, opened 2026-06-09
Fix ggrebalance tool for Python 3.7

whitehawk:
Fix ggrebalance tool for Python 3.7

Problem description:
ggrebalance tool failed to work on systems with Python3 version less than 3.10,
while we need to support Python3 versions starting from 3.7.3.

Root cause:
ggrebalance tool used 'match/case' construction, which appeared only starting
from Python version 3.10.

Fix:
Refactor code to use 'if/elif' instead of 'match/case'.
Plus, variable declarations with a type hint and an initial value (which are
also not supported in 3.7 version) are reworked to use 'Union'.

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
+++ b/gpMgmt/bin/ggrebalance_modules/rebalance_schema.py
@@ -562,35 +562,32 @@ AS total_duration FROM cte_total""")
                                 f"SELECT stat_name, stat_value FROM {self.schema_name}.{self.rebalance_progress_view}")
             if not result.is_rollback:
                 for stat_name, stat_value in cursor:
-                    match stat_name:
-                        case self.STAT_NAME_1_1_TABLES_SHRUNK:
-                            result.tables_shrunk = stat_value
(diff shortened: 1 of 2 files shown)


KnightMurloc:
What about:
https://github.com/GreengageDB/greengage/blob/14f452b879a5a67061ebff813e94e38d1770d335/gpMgmt/bin/ggrebalance_modules/solver.py#L95-L96
and
https://github.com/GreengageDB/greengage/blob/14f452b879a5a67061ebff813e94e38d1770d335/gpMgmt/bin/ggrebalance_modules/solver.py#L323

This syntax is not supported in Python 3.7:
```
>>> best_mirror_host: int | None = None
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'

```

whitehawk:
> This syntax is not supported in Python 3.7:

Updated

dkovalev1 (approved):

KnightMurloc (approved):

PR #483, opened 2026-06-10
GG-526: Add pg_event_trigger to the list of master only tables

dkovalev1:
Event triggers are expected to work only at the master side. However gpexpand 
copies content of pg_event_trigger table to the newly created segment(s).

This commit adds pg_event_trigger to the list of master only table and adds 
behave test to verify it.

GG-526


gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -4676,3 +4676,38 @@ def impl(context):
...
+
+@given('verify that event trigger {trigger_name} not exists in segment {segment}')
+@then('verify that event trigger {trigger_name} not exists in segment {segment}')
+def impl(context, trigger_name, segment):
+    host, port = get_primary_segment_host_port_for_content(segment)
+    dburl = dbconn.DbURL(hostname=host, port=port)
+    with dbconn.connect(dburl, unsetSearchPath=False, utility=True) as conn:
+        sql = "SELECT evtname FROM pg_event_trigger WHERE evtname = '%s';" % trigger_name
+        cursor = dbconn.execSQL(conn, sql)
+        assert cursor.rowcount == 0
```
KnightMurloc: Can we just do something like this:
```sql
select count(*) from gp_dist_random('pg_event_trigger');
```

gpMgmt/bin/gpcheckcat_modules/foreign_key_check.py
```diff
@@ -67,7 +67,8 @@ def checkTableForeignKey(self, cat):
         # skip these master-only tables
         skipped_masteronly = ['gp_relation_node', 'pg_description',
                               'pg_shdescription', 'pg_stat_last_operation',
-                              'pg_stat_last_shoperation', 'pg_statistic']
+                              'pg_stat_last_shoperation', 'pg_statistic',
+                              'pg_event_trigger']
```
KnightMurloc, line 71: I don't quite understand the logic behind this list. Why do we have some tables in this list and not in `MASTER_ONLY_TABLES` and vice versa?

gpMgmt/test/behave/mgmt_utils/gpexpand.feature
```diff
@@ -820,3 +820,21 @@ Feature: expand the cluster by adding more segments
         When the user runs gpexpand with the latest gpexpand_inputfile with additional parameters "--verbose"
         Then gpexpand should print "[DEBUG]:-Skipping tar file (gpexpand_schema.tar) copy to cdw" escaped to stdout
         And verify that the cluster has 1 new segments
+
+    @gpexpand_verify_master_only
```
Stolb27, line 824: What meaning of this tag? I don't see any usages

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -4676,3 +4676,38 @@ def impl(context):
...
+
+@given('verify that event trigger {trigger_name} exists')
+def impl(context, trigger_name):    
+    with dbconn.connect(dbconn.DbURL(dbname=context.dbname), unsetSearchPath=False) as conn:
+        sql = "SELECT evtname FROM pg_event_trigger WHERE evtname = '%s';" % trigger_name
+        cursor = dbconn.execSQL(conn, sql)
+        assert cursor.rowcount == 1
+
+@given('verify that event trigger {trigger_name} not exists in segment {segment}')
+@then('verify that event trigger {trigger_name} not exists in segment {segment}')
```
Stolb27: There are multiple instructions to execute arbitrary sql. Why do we need 3 more?

gpMgmt/test/behave/mgmt_utils/gpexpand.feature
```diff
@@ -820,3 +820,21 @@ Feature: expand the cluster by adding more segments
...
+    Scenario: Verify master only tables are not copied to segment
+        Given the database is not running
+        And a working directory of the test as '/data/gpdata/gpexpand'
+        And a temporary directory under "/data/gpdata/gpexpand/expandedData" to expand into
+        And a cluster is created with no mirrors on "cdw" and "sdw1"
+        And database "gptest" exists
+        And the user creates an event trigger test_trigger
+        And verify that event trigger test_trigger exists
+        And verify that event trigger test_trigger not exists in segment 1
+        When the user runs gpexpand interview to add 1 new segment and 0 new host "news.host"
```
Stolb27: What's news.host?

Stolb27: Is it a new variant of ignore.host?

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
dkovalev1: IMHO checking for a specific event trigger which we just inserted in this test look a bit more robust then just checking the table. What if something added besides test logic? (very unlikely although)

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
Stolb27: E.g.: `below sql is executed in` / `wait until the results from boolean sql`

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
KnightMurloc: If I understood correctly, then this table should always be empty on the segments. If something appears in it, it should indicate a bug.

gpMgmt/bin/gpcheckcat_modules/foreign_key_check.py
dkovalev1: This is a skip list for foreign key checks. I considered that event trigger does not need this check as it master only. May be some other tables should be included as well, but its not related to this commit. (gp_segment_configuration included to MASTER_ONLY_TABLES twice)

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
dkovalev1: At this moment yes, unless someone adds trigger intentionally to the segment. Unlikely.

gpMgmt/test/behave/mgmt_utils/gpexpand.feature
dkovalev1: Its a unique tag similar to some other scenarios, f.e. [here](https://github.com/GreengageDB/greengage/blob/678a35e9525483db0140ddf80df601b64d94347b/gpMgmt/test/behave/mgmt_utils/gpexpand.feature#L805) and [here](https://github.com/GreengageDB/greengage/blob/678a35e9525483db0140ddf80df601b64d94347b/gpMgmt/test/behave/mgmt_utils/gpexpand.feature#L752).

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
dkovalev1: Gherkin style, tried to express what are we checking for in plain English here, not just intent of calling arbitrary sql query.

gpMgmt/test/behave/mgmt_utils/gpexpand.feature
dkovalev1: Right, ignored.host looks better here. Changed.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
KnightMurloc: And gpexpand will remove it from the new segments and we will get inconsistency. If someone needs triggers on segments, they will need to keep this in mind and this test will not let them miss it.

dkovalev1: updated

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -4676,3 +4676,36 @@ def impl(context):
 def impl(context, second):
 	cmd = Command(name='psql', cmdStr="-c 'SELECT * from generate_series(1, %s) a where pg_sleep(1) is not null;'" % second)
 	cmd.runNoWait()
+
+@given('the user creates an event trigger {trigger_name}')
+def impl(context, trigger_name):
```
bimboterminator1, line 4681: how necessary is adding new trigger specific code. Isn't it enough to use 
```
results of the sql "{sql}" db "{dbname}" are stored in the context

validate that following rows are in the stored rows
```

bimboterminator1 (dismissed):

Stolb27:
@KnightMurloc, the new test failed on CI. Did you consider porting existing patch for 7X: https://github.com/GreengageDB/greengage/commit/61daf257d1101ffb4d111f0f350ced5db1c01169?

KnightMurloc:
> @KnightMurloc, the new test failed on CI. Did you consider porting existing patch for 7X: [61daf25](https://github.com/GreengageDB/greengage/commit/61daf257d1101ffb4d111f0f350ced5db1c01169)?

I tried porting the patch from 7X, but it doesn't seem to make much sense for 2 reasons.
1. A modification of the syntax is required to create triggers in tests.
2. The test does not test the problem in the 6X case.

I think it makes sense to leave the current approach.

bimboterminator1 (approved):

Stolb27 (approved):

PR #484, opened 2026-06-10
GG-519: Improve autocomplete for CREATE FUNCTION

riogor:
Some options for CREATE FUNCTION lacked autocomplete support in psql.

Add autocomplete for CREATE FUNCTION.
Add options suggestions for CREATE FUNCTION.
Add suggestions after OR REPLACE. 

Don't address autocomplete for multiple consecutive options.

### DIFF
--- a/src/bin/psql/tab-complete.c
+++ b/src/bin/psql/tab-complete.c
@@ -2538,14 +2538,16 @@ psql_completion(const char *text, int start, int end)
 			COMPLETE_WITH("DISTRIBUTED");
 	}
 
-	else if(Matches("CREATE", "FUNCTION", MatchAny))
+	else if (Matches("CREATE", "FUNCTION", MatchAny))
 		COMPLETE_WITH("(");
-	else if(Matches("CREATE", "FUNCTION", MatchAny, "(*)"))
+	else if (Matches("CREATE", "FUNCTION", MatchAny, "(*)"))
 		COMPLETE_WITH("RETURNS");
-	else if(Matches("CREATE", "FUNCTION", MatchAny, "(*)", "RETURNS"))
+	else if (Matches("CREATE", "FUNCTION", MatchAny, "(*)", "RETURNS"))
 		COMPLETE_WITH_SCHEMA_QUERY(Query_for_list_of_datatypes,
 								   " UNION SELECT 'TABLE ('");
-	else if(Matches("CREATE", "FUNCTION", MatchAny, "(*)", "RETURNS", MatchAny) ||
+	else if (Matches("CREATE", "FUNCTION", MatchAny, "(*)", "RETURNS", "TABLE"))
+		COMPLETE_WITH("(");
+	else if (Matches("CREATE", "FUNCTION", MatchAny, "(*)", "RETURNS", MatchAnyExcept("TABLE")) ||
 			Matches("CREATE", "FUNCTION", MatchAny, "(*)", "RETURNS", "TABLE", "(*)"))
 		COMPLETE_WITH("AS", "LANGUAGE", "TRANSFORM FOR TYPE", "WINDOW",
 					  "IMMUTABLE", "STABLE", "VOLATILE", "NOT LEAKPROOF",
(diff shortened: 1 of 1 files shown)


riogor:
Patch doesn't address autocompletion for multiple consecutive options because psql can't work with strings like with parenthesis (it can skip everything in parenthesis using `(*)`). Without it, it's hard to write Matches statements.
And if strings are ignored, it will lead to lots of wrong suggestions.

VoidZeroNull0:
Why did regression tests fail? Both optimizers failed on the same test - seems like an unlikely flaky.

riogor:
That `parallel_retrieve_cursor` test is documented as flaky (ADBDEV-6184 ticket).
Guess i'm just that lucky to fail both of them.

src/bin/psql/tab-complete.c
```diff
@@ -2538,6 +2538,43 @@ psql_completion(const char *text, int start, int end)
 			COMPLETE_WITH("DISTRIBUTED");
 	}
 
+	else if(Matches("CREATE", "FUNCTION", MatchAny))
```
red1452:
```suggestion
	else if (Matches("CREATE", "FUNCTION", MatchAny))
```

red1452 (commented):
1) Add "CREATE OR REPLACE"
2) strange behavior if I enter "T", it is ok, but if I enter "t", it does not autocomplete correctly:
```
postgres=# create function asd (int)RETURNS table
TABLE (           table_am_handler  
postgres=# create function asd (int)RETURNS table 
AS                          EXECUTE ON                  LANGUAGE                    PARALLEL                    SECURITY                    STRICT                      VOLATILE                    
CALLED ON NULL INPUT        EXTERNAL SECURITY           LEAKPROOF                   RETURNS NULL ON NULL INPUT  SET                         SUPPORT                     WINDOW                      
COST                        IMMUTABLE                   NOT LEAKPROOF               ROWS                        STABLE                      TRANSFORM FOR TYPE          WITH (DESCRIBE =            
postgres=# create function asd (int)RETURNS table 
AS                          EXECUTE ON                  LANGUAGE                    PARALLEL                    SECURITY                    STRICT                      VOLATILE                    
CALLED ON NULL INPUT        EXTERNAL SECURITY           LEAKPROOF                   RETURNS NULL ON NULL INPUT  SET                         SUPPORT                     WINDOW                      
COST                        IMMUTABLE                   NOT LEAKPROOF               ROWS                        STABLE                      TRANSFORM FOR TYPE          WITH (DESCRIBE =            
postgres=# create function asd (int)RETURNS TABLE (
```

src/bin/psql/tab-complete.c
riogor: Fixed spacing

riogor:
> 1. Add "CREATE OR REPLACE"
> 
>     2. strange behavior if I enter "T", it is ok, but if I enter "t", it does not autocomplete correctly:
> 
> 
> ```
> postgres=# create function asd (int)RETURNS table
> TABLE (           table_am_handler  
> postgres=# create function asd (int)RETURNS table 
> AS                          EXECUTE ON                  LANGUAGE                    PARALLEL                    SECURITY                    STRICT                      VOLATILE                    
> CALLED ON NULL INPUT        EXTERNAL SECURITY           LEAKPROOF                   RETURNS NULL ON NULL INPUT  SET                         SUPPORT                     WINDOW                      
> COST                        IMMUTABLE                   NOT LEAKPROOF               ROWS                        STABLE                      TRANSFORM FOR TYPE          WITH (DESCRIBE =            
> postgres=# create function asd (int)RETURNS table 
> AS                          EXECUTE ON                  LANGUAGE                    PARALLEL                    SECURITY                    STRICT                      VOLATILE                    
> CALLED ON NULL INPUT        EXTERNAL SECURITY           LEAKPROOF                   RETURNS NULL ON NULL INPUT  SET                         SUPPORT                     WINDOW                      
> COST                        IMMUTABLE                   NOT LEAKPROOF               ROWS                        STABLE                      TRANSFORM FOR TYPE          WITH (DESCRIBE =            
> postgres=# create function asd (int)RETURNS TABLE (
> ```

Fixed both.

src/bin/psql/tab-complete.c
```diff
@@ -2538,6 +2539,63 @@ psql_completion(const char *text, int start, int end)
 			COMPLETE_WITH("DISTRIBUTED");
 	}
 
+	else if (Matches("CREATE", "OR", "REPLACE"))
+		COMPLETE_WITH("FUNCTION");
```
KnightMurloc: The functions are not the only thing that can be replaced. But we unconditionally auto-complete it with `FUNCTION`. Can we at least suggest them after CREATE OR REPLACE without delving into them?
Postgres supports `CREATE OR REPLACE` for `FUNCTION, AGGREGATE, LANGUAGE, PROCEDURE, RULE, TRANSFORM, and VIEW`.

riogor: Done

red1452 (approved):

KnightMurloc (approved):

PR #488, opened 2026-06-15
Fix assert if vacuum is done in parallel with the cluster shrink

whitehawk:
Fix assert if vacuum is done in parallel with the cluster shrink

Problem description:
Assert could happen if vacuum was done in parallel with the cluster shrink:
`FailedAssertion("!(stats->count == rel->rd_cdbpolicy->numsegments)"`.

Root cause:
Assert happened when vacuum processed the auxiliary tables of the ggrebalance
tool (contained in the 'ggrebalance' schema), that have replicated distribution
policy.
Assert could also happen with simpler steps:
    CREATE TABLE test_table (a int) DISTRIBUTED REPLICATED;
    ALTER TABLE test_table REBALANCE 1;
    VACUUM test_table;
The assert happened because it compared the number of stat dispatch results
with the numsegments of the table. The dispatch is done to all segments
regardless the actual numsegments of the table. Therefore, if the table was
shrunk, its numsegments was reduced, and the assert failed.

Fix:
1. Allow numsegments to be less then the number of dispatch results in the
assert, as that is an allowed situation during cluster shrink. But still emit
a warning for such a case, in order to highlight potential occasion not during
shrink.
2. During table rebalance, truncate table data for replicated tables in the
shrunk segments in order to get consistent relation statistics after vacuum for
cases:
 - the table was created on all segments and later shrunk via
'alter table rebalance' (meaning that there are some data on the shrunk
segments which can affect the statistics calculations);
 - the table was initially created on the reduced number of segments (meaning
there is no data on the shrunk segments).

### DIFF
--- a/src/backend/commands/vacuum.c
+++ b/src/backend/commands/vacuum.c
@@ -3156,7 +3156,12 @@ vac_update_relstats_from_list(VacuumStatsContext *stats_context)
 
 		if (GpPolicyIsReplicated(rel->rd_cdbpolicy))
 		{
-			Assert(stats->count == rel->rd_cdbpolicy->numsegments);
+			Assert(stats->count >= rel->rd_cdbpolicy->numsegments);
+			elogif(stats->count > rel->rd_cdbpolicy->numsegments, WARNING,
+				   "Numsegments for relation \"%s\".\"%s\" is less than stats "
+				   "dispatch results count. Valid situation only during cluster shrink.",
+				   get_namespace_name(get_rel_namespace(stats->relid)),
+				   get_rel_name(stats->relid));
 
 			stats->rel_pages = stats->rel_pages / rel->rd_cdbpolicy->numsegments;
 			stats->rel_tuples = stats->rel_tuples / rel->rd_cdbpolicy->numsegments;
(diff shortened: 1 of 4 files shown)


src/test/regress/sql/alter_rebalance.sql
ShapatinAA, line 557: Unrelated to current patch, but is this right?

whitehawk: Possibly a typo. Please create a separate task if you want it to be fixed.

src/backend/commands/vacuum.c
```diff
@@ -3156,7 +3156,12 @@ vac_update_relstats_from_list(VacuumStatsContext *stats_context)
 
 		if (GpPolicyIsReplicated(rel->rd_cdbpolicy))
 		{
-			Assert(stats->count == rel->rd_cdbpolicy->numsegments);
+			Assert(stats->count >= rel->rd_cdbpolicy->numsegments);
+			ereportif(stats->count > rel->rd_cdbpolicy->numsegments, WARNING,
```
ShapatinAA, line 3160: Do I understand correctly that this warning will be issued for every vacuum of this table after alter table rebalance?

whitehawk: It will be issued for every vacuum after `alter table rebalance` till the number of segments in `gp_segment_configuration` is shrunk (that is one of the last steps of cluster shrink process).

src/test/regress/sql/alter_rebalance.sql
ShapatinAA: Created task in parent ticket.

ShapatinAA (approved):

VoidZeroNull0 (approved):

PR #489, opened 2026-06-15
GG-559: Fix python coverage configs

ShapatinAA:
This commit fixes minor faults at coverage configs and collection process.
Changes in current commit:
1. Update behave tests coverage collection with `GPHOME` variable, so remappings are done to each test
2. Update regression tests script so it will collect and report coverage (otherwise we forced to do it in CI)
3. Add new remappings to configs for `gpperfmon` test and `gp_replica_check.py` file
4. Add to exceptions new(moved) `pg.py` and `pgdb.py` files
5. Update `$GPHOME` hardcode in `coveragerc_unit` file.

### DIFF
--- a/ci/scripts/run_behave_tests.bash
+++ b/ci/scripts/run_behave_tests.bash
@@ -78,7 +78,7 @@ run_feature() {
       cd /tmp/coverage-data
 
       if [ "$(ls "$PROJECT"-coverage-data/ | wc -l)" -gt 0 ]; then
-          coverage combine --debug=pathmap,sys,config --append \
+          coverage combine --append \
             --rcfile=/home/gpadmin/gpdb_src/gpMgmt/test/coveragerc_behave \
             "$PROJECT"-coverage-data/coverage-data*
           mv "$PROJECT"-coverage-data/coverage-data /tmp/coverage-data/coverage-data-"$PROJECT"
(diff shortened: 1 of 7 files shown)


gpMgmt/test/coveragerc_behave
```diff
@@ -36,3 +38,9 @@ source3 =
         /home/gpadmin/gpdb_src/gpMgmt/bin/gppylib
         /home/gpadmin/gpdb_src/gpMgmt/bin/gppylib
         $GPHOME/lib/python/gppylib
+
+source4 =
+        /home/gpadmin/gpdb_src/gpAux/gpperfmon/src/gpmon
+        /home/gpadmin/gpdb_src/gpAux/gpperfmon/src/gpmon
```
VoidZeroNull0, line 44: I don't understand, why do all sources have two exactly same path lines?

ci/scripts/run_behave_tests.bash
```diff
@@ -74,6 +74,7 @@ run_feature() {
     -e FEATURE="$feature" -e PROJECT="$project" \
     cdw bash -eux <<'EOF'
       set -ex
+      source /usr/local/greengage-db-devel/greengage_path.sh
```
VoidZeroNull0, line 77:
```suggestion
      source $GPHOME/greengage_path.sh
```
Can we pass it as environment variable through docker instead of hardcoding it? Should not we?

gpMgmt/test/coveragerc_behave
ShapatinAA: First line is path on which all paths below in one source will be remapped. So remapping 
```
/home/gpadmin/gpdb_src/gpAux/gpperfmon/src/gpmon -> /home/gpadmin/gpdb_src/gpAux/gpperfmon/src/gpmon
```
is indeed trivial, but correct one. I feel like specifying it.

ci/scripts/run_behave_tests.bash
ShapatinAA: Unfortunately we're in the loop here: 
- from one side we make this path hardcoded in our scripts,
- and from the other side we're declaring `GPHOME` value on CI level in docker, which later gets redefined with `sourcing greengage_path.sh`.  
I can bring this value to local variable instead if that will suit you.

VoidZeroNull0: If that is the case, I don't think that changing it is that necessary.

VoidZeroNull0 (approved):

bandetto (approved):

PR #492, opened 2026-06-17
Fix ggrebalance not failing when a parallel one is already running

whitehawk:
Fix ggrebalance not failing when a parallel one is already running

Problem description:
If one instance of ggrebalance was already running, 1st attempt to launch
2nd instance of ggrebalance failed with proper error message, while the 2nd
attempt didn't fail.

Root cause:
ggrebalance creates the 'ggrebalance.pid' file in the coordinator data
directory, and removes it before each exit. ggrebalance didn't check the actual
PID value in the file before removing. Therefore, the 1st invocation attempt
(which failed with the proper error message) removed the PID file of the
already running ggrebalance instance. And the 2nd attempt didn't find the PID
file, and, therefore, started its work.

Fix:
Check the PID value in 'ggrebalance.pid' file before file removal and remove
the file only if it was created by the same process.

### DIFF
--- a/gpMgmt/bin/ggrebalance
+++ b/gpMgmt/bin/ggrebalance
@@ -282,8 +282,13 @@ def remove_pid_file(coordinator_data_directory: str):
     """
     Removes ggrebalance pid file
     """
+    pid_filename = f'{coordinator_data_directory}/ggrebalance.pid'
     try:
-        os.unlink(coordinator_data_directory + '/ggrebalance.pid')
+        with open(pid_filename, 'r') as fp:
+            pid = int(fp.readline().strip())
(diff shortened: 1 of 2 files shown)


VoidZeroNull0:
Do I understand correctly, that only one instance of ggrebalance should be doing work? That's why we check that its PID file is removed, before starting new instance?

whitehawk:
> Do I understand correctly, that only one instance of ggrebalance should be doing work? That's why we check that its PID file is removed, before starting new instance?

Yes, only one instance is allowed.

VoidZeroNull0 (approved):

red1452 (approved):

PR #494, opened 2026-06-17
Fix wrong message on consequent calls of ggrebalance after rebalance is done

whitehawk:
Fix wrong message on consequent calls of ggrebalance after rebalance is done

Problem description:
After ggrebalance has finished cluster shrink, the 2nd call to ggrebalance
provides the proper message 'Previous run was completed successfully. Please
execute cleanup before a new run.'. But 3rd call did not do so, it launched
the last state of shrink and printed the result summary. And the next
consequent call to to rebalance provided a proper message, while the next
one after again tried to launch shrink, and so on...

Root cause:
In order to show proper message, ggrebalance checks the last finished state
taken from the ggrebalance schema. If it is 'STATE_EXECUTOR_DONE', message is
shown, and execution is stopped. The last finished state is extracted from the 
ggrebalance schema at the init stage of ggrebalance main state machine creation.
But we did the check when the main state machine had already stored it's 1st
state ('STATE_START') in the ggrebalance schema. Therefore, on the next call,
the last finished state was 'STATE_START', and ggrebalance tried to continue
interrupted operation, printed final summary, and stored again
'STATE_EXECUTOR_DONE' in the schema. Therefore, we got such loop with switching
behavior.

Fix:
Move the check of the last finished state to an earlier stage, before triggering
the first state of the main state machine. Therefore, if the state is final, we
do not store any new state in the ggrebalance schema. Also, when doing check
in the new place, we also need to consider 'clean_required' and
'rollback_required' options.

### DIFF
--- a/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
+++ b/gpMgmt/bin/ggrebalance_modules/ggrebalance_main_sm.py
@@ -32,8 +32,8 @@ class GGRebalanceMainSM:
         'STATE_PLANNING_STARTED',
         'STATE_PLANNING_DONE',
         'STATE_CHECK_PREVIOUS_RUN',
-        'STATE_END',
-        'STATE_ERROR'
+        'STATE_ERROR',
+        'STATE_END_DUMMY'
     ]
 
     states_logged = [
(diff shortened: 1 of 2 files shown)


ShapatinAA:
Can we move plain check away from run? Like "__should_skip_run()".

whitehawk:
> Can we move plain check away from run? Like "__should_skip_run()".

I think it would be a bit overcomplication at this point. So far I do not expect more checks to be added into `run()`. If more checks are needed, then I guess adding one more function will be justified.

ShapatinAA (approved):

bimboterminator1 (approved):

PR #495, opened 2026-06-17
GG-518: Improve autocomplete for some misc. commands

riogor:
Some commands in psql lacked autocompletion support.

Add options to COPY.
Add NOWAIT, COORDINATOR ONLY options to LOCK TABLE.
Add AO_AUX_ONLY option to VACUUM.
Add ROOTPARTITION, FULLSCAN options to ANALYZE.
Add RESOURCE GROUP/QUEUE list suggestion to COMMENT ON.
Add DENY options to CREATE/ALTER USER/ROLE.
Add days suggestion to CREATE/ALTER USER/ROLE ... DENY.
Add SET DEFAULT ENCODING option to ALTER TYPE.

Add unused options filtering to EXPLAIN, VACUUM, ANALYZE.

Ticket: GG-518

### DIFF
--- a/src/bin/psql/tab-complete.c
+++ b/src/bin/psql/tab-complete.c
@@ -2407,10 +2407,19 @@ psql_completion(const char *text, int start, int end)
 
 	/* Handle COPY [BINARY] <sth> FROM|TO filename */
 	else if (Matches("COPY|\\copy", MatchAny, "FROM|TO", MatchAny))
-		COMPLETE_WITH("BINARY", "DELIMITER", "NULL", "CSV",
+		COMPLETE_WITH("BINARY", "DELIMITER", "NULL", "CSV", "WITH",
 					  "ENCODING", "FREEZE", "FILL MISSING FIELDS", "NEWLINE",
 					  "ON SEGMENT", "IGNORE EXTERNAL PARTITIONS");
 	else if (Matches("COPY", "BINARY", MatchAny, "FROM|TO", MatchAny))
+		COMPLETE_WITH("CSV", "ENCODING", "FREEZE", "FILL MISSING FIELDS",
+					  "WITH", "NEWLINE", "ON SEGMENT",
+					  "IGNORE EXTERNAL PARTITIONS");
+
+	else if (Matches("COPY|\\copy", MatchAny, "FROM|TO", MatchAny, "WITH"))
+		COMPLETE_WITH("BINARY", "DELIMITER", "NULL", "CSV",
+					  "ENCODING", "FREEZE", "FILL MISSING FIELDS", "NEWLINE",
+					  "ON SEGMENT", "IGNORE EXTERNAL PARTITIONS");
+	else if (Matches("COPY", "BINARY", MatchAny, "FROM|TO", MatchAny, "WITH"))
 		COMPLETE_WITH("CSV", "ENCODING", "FREEZE", "FILL MISSING FIELDS",
 					  "NEWLINE", "ON SEGMENT", "IGNORE EXTERNAL PARTITIONS");
 
--- a/src/bin/psql/tab-complete.c
+++ b/src/bin/psql/tab-complete.c
@@ -3328,7 +3328,7 @@ psql_completion(const char *text, int start, int end)
 		 * one word, so the above test is correct.
 		 */
 		if (ends_with(prev_wd, '(') || ends_with(prev_wd, ','))
-			COMPLETE_WITH("ANALYZE", "VERBOSE", "DXL", "COSTS", "SETTINGS",
+			COMPLETE_WITH_UNUSED_OPTIONS("ANALYZE", "VERBOSE", "DXL", "COSTS", "SETTINGS",
 						  "BUFFERS", "TIMING", "SUMMARY", "FORMAT");
 		else if (TailMatches("ANALYZE|VERBOSE|DXL|COSTS|SETTINGS|BUFFERS|TIMING|SUMMARY"))
 			COMPLETE_WITH("ON", "OFF");
@@ -3976,7 +3976,7 @@ psql_completion(const char *text, int start, int end)
 		 * one word, so the above test is correct.
 		 */
 		if (ends_with(prev_wd, '(') || ends_with(prev_wd, ','))
-			COMPLETE_WITH("FULL", "FREEZE", "ANALYZE", "VERBOSE", "AO_AUX_ONLY",
+			COMPLETE_WITH_UNUSED_OPTIONS("FULL", "FREEZE", "ANALYZE", "VERBOSE", "AO_AUX_ONLY",
 						  "DISABLE_PAGE_SKIPPING", "SKIP_LOCKED",
 						  "INDEX_CLEANUP", "TRUNCATE", "PARALLEL", "SKIP_DATABASE_STATS",
 						  "ONLY_DATABASE_STATS");
--- a/src/bin/psql/tab-complete.c
+++ b/src/bin/psql/tab-complete.c
@@ -1836,20 +1836,20 @@ psql_completion(const char *text, int start, int end)
 					  "RENAME TO", "REPLICATION", "RESET", "SET", "SUPERUSER",
 					  "VALID UNTIL");
 
-	else if(TailMatches("DENY") && !TailMatches("DROP", "DENY"))
+	else if (TailMatches("DENY") && !TailMatches("DROP", "DENY"))
 		COMPLETE_WITH("DAY", "BETWEEN DAY");
-	else if(TailMatches("DROP", "DENY", "FOR"))
+	else if (TailMatches("DROP", "DENY", "FOR"))
 		COMPLETE_WITH("DAY");
-	else if(TailMatches("DAY"))
+	else if (TailMatches("DAY"))
 		COMPLETE_WITH("'Monday'", "'Tuesday'", "'Wednesday'", "'Thursday'",
 					  "'Friday'", "'Saturday'", "'Sunday'");
-	else if(TailMatches("DENY", "DAY", MatchAny) ||
+	else if (TailMatches("DENY", "DAY", MatchAny) ||
 			TailMatches("AND", "DAY", MatchAny) ||
 			TailMatches("FOR", "DAY", MatchAny))
 		COMPLETE_WITH("TIME");
-	else if(TailMatches("DENY", "BETWEEN", "DAY", MatchAny))
+	else if (TailMatches("DENY", "BETWEEN", "DAY", MatchAny))
 		COMPLETE_WITH("AND DAY", "TIME");
-	else if(TailMatches("DENY", "BETWEEN", "DAY", MatchAny, "TIME", "'*'") ||
+	else if (TailMatches("DENY", "BETWEEN", "DAY", MatchAny, "TIME", "'*'") ||
 			TailMatches("DENY", "BETWEEN", "DAY", MatchAny, "TIME", "'*", "PM'|AM'"))
 		COMPLETE_WITH("AND DAY");
 
@@ -2216,7 +2216,7 @@ psql_completion(const char *text, int start, int end)
 	/* complete ALTER TYPE <foo> ADD with actions */
 	else if (Matches("ALTER", "TYPE", MatchAny, "ADD"))
 		COMPLETE_WITH("ATTRIBUTE", "VALUE");
-	else if(Matches("ALTER", "TYPE", MatchAny, "SET"))
+	else if (Matches("ALTER", "TYPE", MatchAny, "SET"))
 		COMPLETE_WITH_SUPPRESS_APPEND("SCHEMA ", "DEFAULT ENCODING (");
 	/* ALTER TYPE <foo> RENAME	*/
 	else if (Matches("ALTER", "TYPE", MatchAny, "RENAME"))
@@ -2250,9 +2250,9 @@ psql_completion(const char *text, int start, int end)
 	else if (Matches("ALTER", "TYPE", MatchAny, "RENAME", "VALUE"))
 		COMPLETE_WITH_ENUM_VALUE(prev3_wd);
 
-	else if(Matches("ALTER", "TYPE", MatchAny, "SET", "DEFAULT", "ENCODING", "("))
+	else if (Matches("ALTER", "TYPE", MatchAny, "SET", "DEFAULT", "ENCODING", "("))
 		COMPLETE_WITH("COMPRESSTYPE =", "COMPRESSLEVEL =", "BLOCKSIZE =");
-	else if(Matches("ALTER", "TYPE", MatchAny, "SET", "DEFAULT", "ENCODING", "(", "COMPRESSTYPE", "="))
+	else if (Matches("ALTER", "TYPE", MatchAny, "SET", "DEFAULT", "ENCODING", "(", "COMPRESSTYPE", "="))
 		COMPLETE_WITH("ZLIB", "ZSTD", "RLE_TYPE", "NONE");
 
 /*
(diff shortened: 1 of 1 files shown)


src/bin/psql/tab-complete.c
```diff
@@ -1613,7 +1613,7 @@ psql_completion(const char *text, int start, int end)
 			COMPLETE_WITH_FUNCTION_ARG(prev2_wd);
 	}
 	/* ALTER PROTOCOL */
-	else if(Matches("ALTER", "PROTOCOL", MatchAny))
+	else if (Matches("ALTER", "PROTOCOL", MatchAny))
```
KnightMurloc: Remove unnecessary changes.

src/bin/psql/tab-complete.c
```diff
@@ -2362,16 +2408,28 @@ psql_completion(const char *text, int start, int end)
 	}
 
 	/* Handle COPY [BINARY] <sth> FROM|TO filename */
-	else if (Matches("COPY|\\copy", MatchAny, "FROM|TO", MatchAny) ||
-			 Matches("COPY", "BINARY", MatchAny, "FROM|TO", MatchAny))
+	else if (Matches("COPY|\\copy", MatchAny, "FROM|TO", MatchAny))
```
KnightMurloc: What about `WITH` keyword?

src/bin/psql/tab-complete.c
```diff
@@ -2362,16 +2408,28 @@ psql_completion(const char *text, int start, int end)
...
 		COMPLETE_WITH("BINARY", "DELIMITER", "NULL", "CSV",
-					  "ENCODING");
-
-	/* Handle COPY [BINARY] <sth> FROM|TO filename CSV */
-	else if (Matches("COPY|\\copy", MatchAny, "FROM|TO", MatchAny, "CSV") ||
-			 Matches("COPY", "BINARY", MatchAny, "FROM|TO", MatchAny, "CSV"))
-		COMPLETE_WITH("HEADER", "QUOTE", "ESCAPE", "FORCE QUOTE",
+					  "ENCODING", "FREEZE", "FILL MISSING FILEDS", "NEWLINE",
+					  "ON SEGMENT", "IGNORE EXTERNAL PARTITIONS");
```
KnightMurloc: What about `LOG ERRORS`?

KnightMurloc:
```
postgres=# copy t2 to '/tmp/2.txt' FILL MISSING FILEDS ;
ERROR:  syntax error at or near "FILEDS"
LINE 1: copy t2 to '/tmp/2.txt' FILL MISSING FILEDS ;
```

src/bin/psql/tab-complete.c
```diff
@@ -2362,16 +2408,28 @@ psql_completion(const char *text, int start, int end)
...
+	else if (Matches("COPY", "BINARY", MatchAny, "FROM|TO", MatchAny))
+		COMPLETE_WITH("CSV", "ENCODING", "FREEZE", "FILL MISSING FILEDS",
+					  "NEWLINE", "ON SEGMENT", "IGNORE EXTERNAL PARTITIONS");
```
KnightMurloc: The `CSV` is not compatible with `BINARY`

src/bin/psql/tab-complete.c
```diff
@@ -2362,16 +2408,28 @@ psql_completion(const char *text, int start, int end)
...
+		COMPLETE_WITH("HEADER", "QUOTE", "ESCAPE", "FORCE NULL",
 					  "FORCE NOT NULL");
+	else if (Matches("COPY|\\copy", MatchAny, "FROM", MatchAny, "CSV", "FORCE") ||
+			 Matches("COPY", "BINARY", MatchAny, "FROM", MatchAny, "CSV", "FORCE"))
+		COMPLETE_WITH("NULL", "NOT NULL");
+
+	/* Handle COPY [BINARY] <sth> TO filename CSV */
+	else if (Matches("COPY|\\copy", MatchAny, "TO", MatchAny, "CSV") ||
+			 Matches("COPY", "BINARY", MatchAny, "TO", MatchAny, "CSV"))
+		COMPLETE_WITH("HEADER", "QUOTE", "ESCAPE", "FORCE QUOTE");
```
KnightMurloc, line 2456: Do we really need to handle the CSV case separately? BTW, CSV is the default, so it's not necessary to specify it to use the options below.

src/bin/psql/tab-complete.c
```diff
@@ -2245,14 +2284,14 @@ psql_completion(const char *text, int start, int end)
...
-		else if (TailMatches("VERBOSE|SKIP_LOCKED"))
+			COMPLETE_WITH_UNUSED_OPTIONS("VERBOSE", "SKIP_LOCKED", "FULLSCAN", "ROOTPARTITION");
+		else if (TailMatches("VERBOSE|SKIP_LOCKED|FULLSCAN|ROOTPARTITION"))
 			COMPLETE_WITH("ON", "OFF");
 	}
 	else if (HeadMatches("ANALYZE") && TailMatches("("))
 		/* "ANALYZE (" should be caught above, so assume we want columns */
 		COMPLETE_WITH_ATTR(prev2_wd, "");
-	else if (HeadMatches("ANALYZE"))
+	else if (HeadMatches("ANALYZE") && !TailMatches("ALL"))
```
KnightMurloc: What is the purpose of this condition?

src/bin/psql/tab-complete.c
```diff
@@ -3256,11 +3314,11 @@ psql_completion(const char *text, int start, int end)
 
 /*
  * EXPLAIN [ ( option [, ...] ) ] statement
- * EXPLAIN [ ANALYZE ] [ VERBOSE ] statement
+ * EXPLAIN [ ANALYZE ] [ VERBOSE ] [ DXL ] statement
  */
 	else if (Matches("EXPLAIN"))
 		COMPLETE_WITH("SELECT", "INSERT", "DELETE", "UPDATE", "DECLARE",
-					  "ANALYZE", "VERBOSE");
+					  "ANALYZE", "VERBOSE", "DXL", " (");
```
KnightMurloc: DXL looks like an undocumented parameter intended primarily for developers. Are we sure we want to display it to users?

src/bin/psql/tab-complete.c
```diff
@@ -3579,6 +3643,23 @@ psql_completion(const char *text, int start, int end)
...
+	else if (Matches("LOCK", MatchAny, "IN", "ACCESS", "SHARE", "MODE", "NOWAIT") ||
+			 Matches("LOCK", "TABLE", MatchAny, "IN", "ACCESS", "SHARE", "MODE", "NOWAIT"))
+		COMPLETE_WITH("COORDINATOR ONLY");
+
+	else if (Matches("LOCK", MatchAny, "IN", "SHARE|EXCLUSIVE", "MODE") ||
+			 Matches("LOCK", MatchAny, "IN", "ACCESS|ROW", "SHARE|EXCLUSIVE", "MODE") ||
+			 Matches("LOCK", MatchAny, "IN", "SHARE", "ROW|UPDATE", "EXCLUSIVE", "MODE") ||
+			 Matches("LOCK", "TABLE", MatchAny, "IN", "SHARE|EXCLUSIVE", "MODE") ||
+			 Matches("LOCK", "TABLE", MatchAny, "IN", "ACCESS|ROW", "SHARE|EXCLUSIVE", "MODE") ||
+			 Matches("LOCK", "TABLE", MatchAny, "IN", "SHARE", "ROW|UPDATE", "EXCLUSIVE", "MODE"))
```
KnightMurloc: Can we just check `LOCK [TABLE]` at the beginning and `MODE` at the end?

src/bin/psql/tab-complete.c
```diff
@@ -3579,6 +3643,23 @@ psql_completion(const char *text, int start, int end)
 			 Matches("LOCK", "TABLE", MatchAny, "IN", "SHARE"))
 		COMPLETE_WITH("MODE", "ROW EXCLUSIVE MODE",
 					  "UPDATE EXCLUSIVE MODE");
+
```
KnightMurloc, line 3737: After `lock TABLE t1 in`, I got a suggestion for `nowait`, which is invalid.

src/bin/psql/tab-complete.c
riogor: Done

src/bin/psql/tab-complete.c
riogor: Text is default format, as stated in [docs](https://techdocs.broadcom.com/us/en/vmware-tanzu/data-solutions/tanzu-greenplum/7/greenplum-database/ref_guide-sql_commands-COPY.html) (and in [code](https://github.com/GreengageDB/greengage/blob/a73316ce7d5b3e5c9ff635adbadfaacd29f5951c/src/backend/commands/copy.c#L1422-L1427) too).

I think we have to handle it separately, as these 6 options are available only in csv mode.

src/bin/psql/tab-complete.c
riogor: Good point.
Guess it's better to stick to docs.

src/bin/psql/tab-complete.c
riogor: Hm, its the remains of old code, tried to adapt them. Looks useless, as ANALYZE doesn't accept multiple tables.

src/bin/psql/tab-complete.c
riogor: Fixed

src/bin/psql/tab-complete.c
riogor: Removed

src/bin/psql/tab-complete.c
riogor: Fixed

src/bin/psql/tab-complete.c
riogor: Simplified

src/bin/psql/tab-complete.c
KnightMurloc: At least `HEADER` and `ESCAPE` can be used with text mode.

KnightMurloc: Also `FORCE QUOTE`, `FORCE_NOT_NULL` and `FORCE_NULL` only work with `WITH` keyword.

src/bin/psql/tab-complete.c
riogor: Added

src/bin/psql/tab-complete.c
riogor: Fixed

src/bin/psql/tab-complete.c
riogor: Added HEADER/ESCAPE

src/bin/psql/tab-complete.c
riogor: Added

riogor:
On note about `WITH` keyword.

I made patch, according to old `COPY` syntax, where options aren't in brackets and `WITH` is optional (there are lots of examples with all options in our tests). And it was already in tab-complete, so i've expanded it.

New syntax is what in docs currently, where all options are in brackets and comma-separated. 

Supporting both old and new syntax seems weird, so i stick with expanding what already was here.

src/bin/psql/tab-complete.c
```diff
@@ -2363,16 +2406,43 @@ psql_completion(const char *text, int start, int end)
...
+	/* Handle COPY [BINARY] <sth> TO filename CSV */
+	else if (Matches("COPY|\\copy", MatchAny, "TO", MatchAny, "CSV"))
+		COMPLETE_WITH("HEADER", "QUOTE", "ESCAPE", "FORCE QUOTE");
+
+	else if(HeadMatches("COPY|\\copy", MatchAny) && TailMatches("LOG", "ERRORS"))
+		COMPLETE_WITH("", "SEGMENT REJECT LIMIT");
+	else if(HeadMatches("COPY|\\copy", MatchAny) &&
+			TailMatches("LOG", "ERRORS", "SEGMENT", "REJECT", "LIMIT", MatchAny))
+		COMPLETE_WITH("ROWS", "PERCENT");
+
```
KnightMurloc: I don't really like it. It looks messy (mostly because copy parameters are messy themselves in ggdb...) What if we leave auto-completion for WITH only? It looks much simpler and more stable. Something like that:
```c
	else if(Matches("COPY|\\copy", MatchAny, "FROM|TO", MatchAny))
		COMPLETE_WITH("WITH (");
	else if (HeadMatches("COPY|\\copy", MatchAny, "FROM|TO", MatchAny , "WITH", "(*"))
	{
		if (ends_with(prev_wd, '(') || ends_with(prev_wd, ','))
			COMPLETE_WITH_UNUSED_OPTIONS("FORMAT",
										 "FREEZE",
										 "DELIMITER",
										 "NULL",
										 "HEADER",
										 "QUOTE",
										 "ESCAPE",
										 "FORCE_QUOTE",
										 "FORCE_NOT_NULL",
										 "FORCE_NULL",
										 "ENCODING",
										 "FILL MISSING FIELDS",
										 "LOG ERRORS",
										 "IGNORE EXTERNAL PARTITIONS");
		else if (TailMatches("FREEZE|HEADER"))
			COMPLETE_WITH("ON", "OFF");
	}
```

riogor: Main problem with this way is that there 3 options (`FORCE_QUOTE`, `FORCE_NULL`, `FORCE_NOT_NULL`), that allow column list as its argument, which includes commas as separator. 
I don't know a way to easily distinguish between column list and `COPY` options comma.
Without it, i'll give wrong suggestions.

riogor: The best simple way (although still very cumbersome) is to track amount of open brackets at current position. That'll distinguish between top `WITH` and column list.
But there can be unclosed brackets in column names (within quotes), `DELIMETER`, `QUOTE`, `ESCAPE`, which will ruin algo.
And it looks hard to properly escape quoted strings. (psql doesn't do that)

KnightMurloc: What about something like this:
```c
		if (!(HeadMatches("COPY|\\copy", MatchAny, "FROM|TO", MatchAny , "WITH") &&
			  TailMatches("FORCE_NULL|FORCE_NOT_NULL|FORCE_QUOTE", "(*") &&
			  !TailMatches("FORCE_NULL|FORCE_NOT_NULL|FORCE_QUOTE", "(*)")) &&
				(ends_with(prev_wd, '(') || ends_with(prev_wd, ',')))
			COMPLETE_WITH_UNUSED_OPTIONS("FORMAT",
										 "FREEZE",
										 "DELIMITER",
										 "NULL",
										 "HEADER",
										 "QUOTE",
										 "ESCAPE",
										 "FORCE_QUOTE",
										 "FORCE_NOT_NULL",
										 "FORCE_NULL",
										 "ENCODING");
		else if (TailMatches("FREEZE|HEADER"))
			COMPLETE_WITH("ON", "OFF");
```

riogor: It won't work. `TailMatches` looks only at last n words (so e.g., `FORCE_NULL` should be second last word here). And when i write at least one colname, it will shift by one word, now making `FORCE_NULL` third last and missing it completely.
Matches and wildcard here are not like in regex, and they won't try checking other positions.

E.g., this toy match:
```c
else if(TailMatches("FORCE_NULL|FORCE_NOT_NULL|FORCE_QUOTE", "(*"))
		COMPLETE_WITH("c1", "c2", "c3", "c4");
```
will act as follows:
```
template1=# FORCE_NULL(
C1  C2  C3  C4 

template1=# FORCE_NULL(C
C1  C2  C3  C4 

template1=# FORCE_NULL(C1 


template1=# FORCE_NULL(C1 , 

```

KnightMurloc:
What about: `SEGMENT REJECT LIMIT <n> PERCENT` (without `LOG ERRORS`)?

riogor:
> What about: `SEGMENT REJECT LIMIT <n> PERCENT` (without `LOG ERRORS`)?

Somehow forgot that, added.
Split `FROM|TO`, because `SEGMENT REJECT LIMIT` works only in `COPY x FROM ...`

KnightMurloc (approved):
Don't forget to update the description.

red1452 (approved):

PR #496, opened 2026-06-18
GG-550: Fix RANK(1) clause appearing in the dump

GirgerS:
Before this patch, pg_dump would dump RANK(1) 
clause for subpartition templates, if the 
tables at the required partitioning level 
are range partitions themselves .

This syntax has been removed from Greengage 7, 
making it not possible to restore such dump.

The root cause of this issue was 
pg_get_partition_template_def function from 
Greengage 6.

This patch makes pg_dump avoid calling this 
function directly, by building subpartition templates 
from the catalog entries.

pg_upgrade test errors.


GirgerS:
This patch also fixes GG-557, as there is no need to dump `WITH (tablename=...)` clause anymore.

To test the changes, create the following table (it is the table that cause pg_upgrade to fail) in Greengage 6 cluster, and dump it using pg_dump from Greengage 7.
```
CREATE TABLE mpp5992 (a int, b date, c char,
                     d char(4), e varchar(20), f timestamp)
WITH (orientation=column,appendonly=true)
partition by range (b)
subpartition by list (a)
subpartition template (
subpartition l1 values (1,2,3,4,5),
subpartition l2 values (6,7,8,9,10) )
subpartition by list (e)
subpartition template (
subpartition ll1 values ('Engineering'),
subpartition ll2 values ('QA') )
subpartition by list (c)
subpartition template (
subpartition lll1 values ('M'),
subpartition lll2 values ('F') )
(
  start (date '2007-01-01')
  end (date '2010-01-01') every (interval '1 year')
);
``` 

Then, you can run the same command in Greengage 7, and compare it to the results of the command in the dump. They should result in the same table. `gp_partition_template` catalog table should be the same for both commands, and, for example, 
```
CREATE TABLE public.mpp5992 (
    a integer,
    b date,ALTER TABLE mpp5992 ADD PARTITION
            START (date '2017-02-01') INCLUSIVE
            END (date '2017-03-01') EXCLUSIVE;
```
command should create the same tables.

bandetto:
> Then, you can run the same command in Greengage 7

Can you please post a `diff` between DDLs dumped by both versions for clarity? An example of how output has changed.

src/bin/pg_dump/pg_dump.c
```diff
@@ -7418,11 +7418,56 @@ getPartitionDefs(Archive *fout, TableInfo tblinfo[], int numTables)
 	appendPQExpBufferChar(tbloids, '}');
 	resetPQExpBuffer(query);
 
+	/*
+	 * Collect partitions via build-in function,
+	 * and partition templates using handmade query,
+	 * because pg_get_partition_template_def may use
+	 * syntax not compatible with Greengage 7
```
bandetto, line 7425: Can this bug be addressed by actually fixing this functions output rather than adding a workaround?

GirgerS: In short, It is a regular function from Greengage 6 (as it is executed in the source cluster), and there is no way to make it work for both version at the same time (because only alternative is `FOR (<value>)` syntax, and it is not possible to reliably find a value with a range). Directly altering tables doesn't work for partitions in Greengage 6. So I doubt we allowed to change it.

GirgerS:
> > Then, you can run the same command in Greengage 7
> 
> Can you please post a `diff` between DDLs dumped by both versions for clarity? An example of how output has changed.

For the table I mentioned above, old dump:
```sql
 ALTER TABLE public.mpp5992 ALTER PARTITION FOR (RANK(1)) ALTER PARTITION l1 
SET SUBPARTITION TEMPLATE 
          (
          SUBPARTITION lll1 VALUES('M') WITH (tablename='mpp5992'), 
          SUBPARTITION lll2 VALUES('F') WITH (tablename='mpp5992')
          );

ALTER TABLE public.mpp5992 ALTER PARTITION FOR (RANK(1)) 
SET SUBPARTITION TEMPLATE 
          (
          SUBPARTITION ll1 VALUES('Engineering') WITH (tablename='mpp5992') , 
          SUBPARTITION ll2 VALUES('QA') WITH (tablename='mpp5992')
          );

ALTER TABLE public.mpp5992 
SET SUBPARTITION TEMPLATE 
          (
          SUBPARTITION l1 VALUES(1, 2, 3, 4, 5) WITH (tablename='mpp5992') , 
          SUBPARTITION l2 VALUES(6, 7, 8, 9, 10) WITH (tablename='mpp5992')
          )
;
```
New one:
```sql
 ALTER TABLE public.mpp5992_1_prt_1_2_prt_l1 SET SUBPARTITION TEMPLATE
(
SUBPARTITION lll1 VALUES('M'),
SUBPARTITION lll2 VALUES('F')
);
ALTER TABLE public.mpp5992_1_prt_1 SET SUBPARTITION TEMPLATE
(
SUBPARTITION ll1 VALUES('Engineering'),
SUBPARTITION ll2 VALUES('QA')
);
ALTER TABLE public.mpp5992 SET SUBPARTITION TEMPLATE
(
SUBPARTITION l1 VALUES(1, 2, 3, 4, 5),
SUBPARTITION l2 VALUES(6, 7, 8, 9, 10)
);
```

Also, I didn't keep the formatting, as I like this version better. But we can keep it if necessary.

src/bin/pg_dump/pg_dump.c
bandetto: So the goal is to make `pg_dump` version-agnostic. How about comparing `current_setting('server_version_num')`?

GirgerS: It is not, per PostgreSQL docs:
> Because pg_dump is used to transfer data to newer versions of PostgreSQL, the output of pg_dump can be expected to load into PostgreSQL server versions newer than pg_dump's version. pg_dump can also dump from PostgreSQL servers older than its own version. (Currently, servers back to version 9.2 are supported.) However, pg_dump cannot dump from PostgreSQL servers newer than its own major version; it will refuse to even try, rather than risk making an invalid dump. Also, it is not guaranteed that pg_dump's output can be loaded into a server of an older major version — not even if the dump was taken from a server of that version. Loading a dump file into an older server may require manual editing of the dump file to remove syntax not understood by the older server. Use of the --quote-all-identifiers option is recommended in cross-version cases, as it can prevent problems arising from varying reserved-word lists in different PostgreSQL versions.

GirgerS: So the goal is to make this code work only for 7.x target cluster. And this code path is always [executed](https://github.com/GreengageDB/greengage/blob/7.x/src/bin/pg_dump/pg_dump.c#L7373-L7387) inside source cluster with version 5.x (or below, which we don't support), and 6.x, where it will always work.

bandetto (dismissed):

src/bin/pg_dump/pg_dump.c
```diff
@@ -7418,11 +7418,56 @@ getPartitionDefs(Archive *fout, TableInfo tblinfo[], int numTables)
...
+						"		SELECT src.tbloid src,"
+						"			template.partitionlevel level,"
+						"			template.partitionboundary bounds,"
+						"			template.partitionposition position,"
+						"			COALESCE(part.name, src_table.relname) name,"
+						"			COALESCE(part.nspname, nsp.nspname) nspname"
+						"		FROM unnest('%s'::pg_catalog.oid[]) AS src(tbloid)"
+						"			JOIN pg_class src_table ON src_table.oid = src.tbloid"
+						"			JOIN pg_namespace nsp ON src_table.relnamespace = nsp.oid"
+						"			JOIN pg_partition_templates template ON template.tablename = src_table.relname"
```
hilltracer, line 7448: Here we join only by name without schema name. So it can be problem with identical names in different schemas.

<details>
<summary>spoiler</summary>

```sql
CREATE SCHEMA dump_bug_s1;
CREATE SCHEMA dump_bug_s2;

CREATE TABLE dump_bug_s1.same_name
(
    a int,
    b int
)
PARTITION BY RANGE (b)
SUBPARTITION BY LIST (a)
SUBPARTITION TEMPLATE
(
    SUBPARTITION s1_a VALUES (1)
)
(
    START (1) END (3) EVERY (1)
);

CREATE TABLE dump_bug_s2.same_name
(
    a int,
    b int
)
PARTITION BY RANGE (b)
SUBPARTITION BY LIST (a)
SUBPARTITION TEMPLATE
(
    SUBPARTITION s2_a VALUES (2)
)
(
    START (1) END (3) EVERY (1)
);
```

old 6.x dump:
```sql
CREATE TABLE dump_bug_s1.same_name (
    a integer,
    b integer
)
 DISTRIBUTED BY (a) PARTITION BY RANGE(b)
          SUBPARTITION BY LIST(a) 
          (
          START (1) END (2) EVERY (1) WITH (tablename='same_name_1_prt_1', appendonly='false')
                  (
                  SUBPARTITION s1_a VALUES(1) WITH (tablename='same_name_1_prt_1_2_prt_s1_a', appendonly='false')
                  ), 
          START (2) END (3) EVERY (1) WITH (tablename='same_name_1_prt_2', appendonly='false')
                  (
                  SUBPARTITION s1_a VALUES(1) WITH (tablename='same_name_1_prt_2_2_prt_s1_a', appendonly='false')
                  )
          );
 ALTER TABLE dump_bug_s1.same_name 
SET SUBPARTITION TEMPLATE 
          (
          SUBPARTITION s1_a VALUES(1) WITH (tablename='same_name')
          )
;

```

7.x after patch:
```sql
CREATE TABLE dump_bug_s1.same_name (
    a integer,
    b integer
) DISTRIBUTED BY (a) PARTITION BY RANGE(b)
          SUBPARTITION BY LIST(a) 
          (
          START (1) END (2) EVERY (1) WITH (tablename='same_name_1_prt_1', appendonly='false')
                  (
                  SUBPARTITION s1_a VALUES(1) WITH (tablename='same_name_1_prt_1_2_prt_s1_a', appendonly='false')
                  ), 
          START (2) END (3) EVERY (1) WITH (tablename='same_name_1_prt_2', appendonly='false')
                  (
                  SUBPARTITION s1_a VALUES(1) WITH (tablename='same_name_1_prt_2_2_prt_s1_a', appendonly='false')
                  )
          );
 ALTER TABLE dump_bug_s1.same_name SET SUBPARTITION TEMPLATE
(
SUBPARTITION s1_a VALUES(1),
SUBPARTITION s2_a VALUES(2)
);
;

```

also see this in the end of query:
```
);
;
```
can we fix it?

</details>

src/bin/pg_dump/pg_dump.c
```diff
@@ -7418,11 +7418,56 @@ getPartitionDefs(Archive *fout, TableInfo tblinfo[], int numTables)
...
+						"			LEFT JOIN LATERAL ("
+						"				SELECT"
+						"					MIN(part.partitiontablename) name,"
```
hilltracer, line 7452: Looks like MIN(partitiontablename) is not equivalent to the old RANK(1) addressing.

<details>
<summary>spoiler</summary>

```sql
CREATE TABLE public.rank_name_bug2
(
    a int,
    b int,
    c int
)
PARTITION BY RANGE (b)

SUBPARTITION BY LIST (a)
SUBPARTITION TEMPLATE
(
    SUBPARTITION l1 VALUES (1),
    SUBPARTITION l2 VALUES (2)
)

SUBPARTITION BY LIST (c)
SUBPARTITION TEMPLATE
(
    SUBPARTITION c1 VALUES (1),
    SUBPARTITION c2 VALUES (2)
)

(
    PARTITION z_first_range START (1) END (10),
    PARTITION a_second_range START (10) END (20)
);
```

6.x result

```sh
pg_dump -s -t public.rank_name_bug2 postgres
```

```sql
CREATE TABLE public.rank_name_bug2 (
    a integer,
    b integer,
    c integer
)
 DISTRIBUTED BY (a) PARTITION BY RANGE(b)
          SUBPARTITION BY LIST(a)
                  SUBPARTITION BY LIST(c) 
          (
          PARTITION z_first_range START (1) END (10) WITH (tablename='rank_name_bug2_1_prt_z_first_range', appendonly='false')
                  (
                  SUBPARTITION l1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l1', appendonly='false')
                          (
                          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l1_3_prt_c1', appendonly='false'), 
                          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l1_3_prt_c2', appendonly='false')
                          ), 
                  SUBPARTITION l2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l2', appendonly='false')
                          (
                          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l2_3_prt_c1', appendonly='false'), 
                          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l2_3_prt_c2', appendonly='false')
                          )
                  ), 
          PARTITION a_second_range START (10) END (20) WITH (tablename='rank_name_bug2_1_prt_a_second_range', appendonly='false')
                  (
                  SUBPARTITION l1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l1', appendonly='false')
                          (
                          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l1_3_prt_c1', appendonly='false'), 
                          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l1_3_prt_c2', appendonly='false')
                          ), 
                  SUBPARTITION l2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l2', appendonly='false')
                          (
                          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l2_3_prt_c1', appendonly='false'), 
                          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l2_3_prt_c2', appendonly='false')
                          )
                  )
          );
 ALTER TABLE public.rank_name_bug2 ALTER PARTITION z_first_range 
SET SUBPARTITION TEMPLATE 
          (
          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2'), 
          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2')
          );

ALTER TABLE public.rank_name_bug2 
SET SUBPARTITION TEMPLATE 
          (
          SUBPARTITION l1 VALUES(1) WITH (tablename='rank_name_bug2') , 
          SUBPARTITION l2 VALUES(2) WITH (tablename='rank_name_bug2')
          )
;
```



7.x after patch

```sh
$GREENGAGE7_INSTALLATION/bin/pg_dump -s -t public.rank_name_bug2 postgres
```

```sql
CREATE TABLE public.rank_name_bug2 (
    a integer,
    b integer,
    c integer
) DISTRIBUTED BY (a) PARTITION BY RANGE(b)
          SUBPARTITION BY LIST(a)
                  SUBPARTITION BY LIST(c) 
          (
          PARTITION z_first_range START (1) END (10) WITH (tablename='rank_name_bug2_1_prt_z_first_range', appendonly='false')
                  (
                  SUBPARTITION l1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l1', appendonly='false')
                          (
                          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l1_3_prt_c1', appendonly='false'), 
                          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l1_3_prt_c2', appendonly='false')
                          ), 
                  SUBPARTITION l2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l2', appendonly='false')
                          (
                          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l2_3_prt_c1', appendonly='false'), 
                          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_z_first_range_2_prt_l2_3_prt_c2', appendonly='false')
                          )
                  ), 
          PARTITION a_second_range START (10) END (20) WITH (tablename='rank_name_bug2_1_prt_a_second_range', appendonly='false')
                  (
                  SUBPARTITION l1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l1', appendonly='false')
                          (
                          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l1_3_prt_c1', appendonly='false'), 
                          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l1_3_prt_c2', appendonly='false')
                          ), 
                  SUBPARTITION l2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l2', appendonly='false')
                          (
                          SUBPARTITION c1 VALUES(1) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l2_3_prt_c1', appendonly='false'), 
                          SUBPARTITION c2 VALUES(2) WITH (tablename='rank_name_bug2_1_prt_a_second_range_2_prt_l2_3_prt_c2', appendonly='false')
                          )
                  )
          );
 ALTER TABLE public.rank_name_bug2_1_prt_a_second_range SET SUBPARTITION TEMPLATE
(
SUBPARTITION c1 VALUES(1),
SUBPARTITION c2 VALUES(2)
);

ALTER TABLE public.rank_name_bug2 SET SUBPARTITION TEMPLATE
(
SUBPARTITION l1 VALUES(1),
SUBPARTITION l2 VALUES(2)
);
;
```

So in last output we set template for rank_name_bug2_1_prt_a_second_range but it should be rank_name_bug2_1_prt_z_first_range.

</details>

src/bin/pg_dump/pg_dump.c
```diff
@@ -7418,11 +7418,56 @@ getPartitionDefs(Archive *fout, TableInfo tblinfo[], int numTables)
...
+	 * because pg_get_partition_template_def may use
+	 * syntax not compatible with Greengage 7
+	 * (for example, RANK(1) clause)
+	 */
 	appendPQExpBuffer(query,
+						"WITH templates_per_level AS ("
+						"	SELECT src,"
+						"		entry.level,"
+						"		FORMAT ("
+						"			'ALTER TABLE %%s.%%s SET SUBPARTITION TEMPLATE\n(\n%%s\n);\n',"
```
hilltracer: Looks like this produces extra `;`. See in my examples:
```sql
...
    );
    ;
```

hilltracer:
Since this patch has no tests, please share manual before/after dump output for edge cases smth like: DEFAULT partitions, dropped/re-added partitions, empty templates, quoted/mixed-case names, non-public schemas, and 3+ subpartition levels...

src/bin/pg_dump/pg_dump.c
GirgerS: Thanks for noticing, I forgot to push this change.

src/bin/pg_dump/pg_dump.c
GirgerS: Thanks for noticing, it would've been a disaster otherwise. [pg_partitions](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/catalog/system_views.sql#L765) and [pg_partition_templates](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/catalog/system_views.sql#L887) store the schema of the root table, so a combination of shema name + table name should suffice.

src/bin/pg_dump/pg_dump.c
GirgerS: Yes, and it is fine, because subpartition templates are [attached](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/commands/tablecmds_gp.c#L1337-L1408) to partitioning levels, not tables.

As long as these tables are on the same level, we can pick either of them.

GirgerS: So, when running the query from the latter dump you've provided everything works fine:

```sql
-- assuming that public.rank_name_bug2 is created
ALTER TABLE public.rank_name_bug2 ADD PARTITION t_third_range START(20) END(30);
\dt
 Schema |                         Name                          |       Type        | Owner  |  Storage
--------+-------------------------------------------------------+-------------------+--------+-----------
 public | rank_name_bug2                                        | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_a_second_range                   | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_a_second_range_2_prt_l1          | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_a_second_range_2_prt_l1_3_prt_c1 | table             | regrig | heap
 public | rank_name_bug2_1_prt_a_second_range_2_prt_l1_3_prt_c2 | table             | regrig | heap
 public | rank_name_bug2_1_prt_a_second_range_2_prt_l2          | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_a_second_range_2_prt_l2_3_prt_c1 | table             | regrig | heap
 public | rank_name_bug2_1_prt_a_second_range_2_prt_l2_3_prt_c2 | table             | regrig | heap
 public | rank_name_bug2_1_prt_t_third_range                    | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_t_third_range_2_prt_l1           | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_t_third_range_2_prt_l1_3_prt_c1  | table             | regrig | heap
 public | rank_name_bug2_1_prt_t_third_range_2_prt_l1_3_prt_c2  | table             | regrig | heap
 public | rank_name_bug2_1_prt_t_third_range_2_prt_l2           | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_t_third_range_2_prt_l2_3_prt_c1  | table             | regrig | heap
 public | rank_name_bug2_1_prt_t_third_range_2_prt_l2_3_prt_c2  | table             | regrig | heap
 public | rank_name_bug2_1_prt_z_first_range                    | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_z_first_range_2_prt_l1           | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_z_first_range_2_prt_l1_3_prt_c1  | table             | regrig | heap
 public | rank_name_bug2_1_prt_z_first_range_2_prt_l1_3_prt_c2  | table             | regrig | heap
 public | rank_name_bug2_1_prt_z_first_range_2_prt_l2           | partitioned table | regrig | heap
 public | rank_name_bug2_1_prt_z_first_range_2_prt_l2_3_prt_c1  | table             | regrig | heap
 public | rank_name_bug2_1_prt_z_first_range_2_prt_l2_3_prt_c2  | table             | regrig | heap
```

GirgerS:
> Since this patch has no tests, please share manual before/after dump output for edge cases smth like: DEFAULT partitions, dropped/re-added partitions, empty templates, quoted/mixed-case names, non-public schemas, and 3+ subpartition levels...

I've tried to create every concerning object in the attached files, except for empty templates and private schemas. I have little experience with this part of database, so can you give an example of such schema? My understanding is that all schemas are private by default (except for superusers). 

And is there a way to create empty template? Attempts to create one give me syntax error:
```sql
CREATE TABLE public.rank_name_bug2
(
    a int,
    b int,
    c int
)
PARTITION BY RANGE (b)

SUBPARTITION BY LIST (a)
SUBPARTITION TEMPLATE
()

SUBPARTITION BY LIST (c)
SUBPARTITION TEMPLATE
()

(
    PARTITION z_first_range START (1) END (10),
    PARTITION a_second_range START (10) END (20)
);
ERROR:  syntax error at or near ")"
LINE 11: ()
```

Regardless, these clauses shouldn't change at all. `pg_partition_tempaltes` uses the same [function](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/catalog/system_views.sql#L938) under the hood.

Also, to give me more context, why do you consider:
- DEFAULT partitions an edge case? They live on the same level with other partitions, so we can select them just like others.
- 3+ subpartition levels an edge case? The logic is different only for the first level, others work the same.

Regarding dropped/re-added partitions, the only edge-case that I can think of is a level with no partitions, but there is no way to make it (create_tables.sql has such command, Greenage will complain and refuse to execute it)

[create_tables.sql](https://github.com/user-attachments/files/29289079/create_tables.sql)
[dump_old.sql](https://github.com/user-attachments/files/29289077/dump_old.sql)
[dump_new.sql](https://github.com/user-attachments/files/29289075/dump_new.sql)

hilltracer:
> private schemas

I meant non-public/user-created schemas, especially same table names in different schemas. So I'm ok.

> Attempts to create one give me syntax error

I was wrong, forget about empty templates.

> [create_tables.sql](https://github.com/user-attachments/files/29289079/create_tables.sql)
> [dump_old.sql](https://github.com/user-attachments/files/29289077/dump_old.sql)
> [dump_new.sql](https://github.com/user-attachments/files/29289075/dump_new.sql)

Thanks for tests. LGTM and enough for me.

hilltracer (approved):

bandetto (approved):

PR #498, opened 2026-06-19
GG-545: Fix handling path with trailing slashes

ShapatinAA:
Previously `gpstart` utility couldn't start standby on directory specified with trailing slash, like that:
`/path/to/standby/`. 

It was enough to add normalization on path for further usage, to fix the issue.

### DIFF
--- a/gpMgmt/test/behave/mgmt_utils/gpstart.feature
+++ b/gpMgmt/test/behave/mgmt_utils/gpstart.feature
@@ -179,7 +179,8 @@ Feature: gpstart behave tests
           And the standby is not initialized
 
          When the user runs gpinitstandby with options "-S /tmp/standby_data/"
-         Then verify the standby master entries in catalog
+         Then gpinitstandby should return a return code of 0
+          And verify the standby master entries in catalog
         
          When the master goes down
           And the user runs "gpstart -a"
--- a/gpMgmt/test/behave/mgmt_utils/gpstart.feature
+++ b/gpMgmt/test/behave/mgmt_utils/gpstart.feature
@@ -172,3 +172,19 @@ Feature: gpstart behave tests
 
           When the user runs psql with "-c 'drop user foouser;'" against database "postgres"
           Then psql should return a return code of 0
+
+    Scenario: gpstart on dir with trailing slash
+        Given the database is running
+          And the catalog has a standby master entry
+          And the standby is not initialized
+
+         When the user runs gpinitstandby with options "-S /tmp/standby_data/"
+         Then verify the standby master entries in catalog
+        
+         When the master goes down
+          And the user runs "gpstart -a"
+         Then gpstart should return a return code of 0
+          And verify the standby master entries in catalog
+        
+         When execute sql "select datadir from gp_segment_configuration where content = -1 and role = 'm'" in db "postgres" and store result in the context
+         Then validate that "/tmp/standby_data/" is in the stored rows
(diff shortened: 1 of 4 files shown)


GirgerS:
One small question.

Can you confirm that [this](https://github.com/GreengageDB/greengage/pull/498/changes#diff-de83eb981749fe06f9bea4dc5cec1abe50e51a47dac67fbfafad4dcfe4906999R175) test fails before the patch? Local attempt to run gpactivatestandby feature fail for me locally for some reason.

GirgerS:
Sorry, missclick.

ShapatinAA:
> One small question.
> 
> Can you confirm that [this](https://github.com/GreengageDB/greengage/pull/498/changes#diff-de83eb981749fe06f9bea4dc5cec1abe50e51a47dac67fbfafad4dcfe4906999R175) test fails before the patch? Local attempt to run gpactivatestandby feature fail for me locally for some reason.

After private discussion I think this question can be closed. 

Thou, if something still unclear please ask me, I would be glad to clarify!

gpMgmt/test/behave/mgmt_utils/gpstart.feature
```diff
@@ -172,3 +172,19 @@ Feature: gpstart behave tests
...
           When the user runs psql with "-c 'drop user foouser;'" against database "postgres"
           Then psql should return a return code of 0
+
+    Scenario: gpstart on dir with trailing slash
+        Given the database is running
+          And the catalog has a standby master entry
+          And the standby is not initialized
+
+         When the user runs gpinitstandby with options "-S /tmp/standby_data/"
+         Then verify the standby master entries in catalog
```
riogor:
```suggestion
         When the user runs gpinitstandby with options "-S /tmp/standby_data/"
         Then gpinitstandby should return a return code of 0
         And verify the standby master entries in catalog
```

Shouldn't we also check for return code here? (like in gpactivatestandby test)

riogor:
Other than that, LGTM.

Unrelated to patch, shouldn't we also fix trailing slashes in `gp_segment_configuration` table? Just for consistency.

GirgerS:
> > One small question.
> > Can you confirm that [this](https://github.com/GreengageDB/greengage/pull/498/changes#diff-de83eb981749fe06f9bea4dc5cec1abe50e51a47dac67fbfafad4dcfe4906999R175) test fails before the patch? Local attempt to run gpactivatestandby feature fail for me locally for some reason.
> 
> After private discussion I think this question can be closed.
> 
> Thou, if something still unclear please ask me, I would be glad to clarify!

Yes, I have no questions left.

gpMgmt/test/behave/mgmt_utils/gpstart.feature
ShapatinAA: Yes, thank you for suggestion! I should have add this check also.

ShapatinAA:
> `gp_segment_configuration`

I think we can discuss this one, because answer on this question can be ambiguous.
 
We either should have only normalized paths in our system, or we just should be able to work with non-normalized paths... 
And to stick to the first seems ideal to me and consistent, but I can not estimate amount of work here to be done (should we only fix writes to `gp_segment_configuration` or make normalization at _init/config_read_ stage or what will be affected etc.).

Right now I want to just stick to the rule "If it works - do not touch it".

riogor:
> Right now I want to just stick to the rule "If it works - do not touch it".

Yeah, this patch doesn't need fixing paths in table. After all, everything works. And it doesn't seem very important.

Guess you can make a task for it in backlog, to not forget about it completely.

riogor (approved):

GirgerS (approved):

PR #499, opened 2026-06-19
GG-567: Fix handling path with trailing slashes

ShapatinAA:
Previously `gpstart` utility couldn't start standby on directory specified with trailing slash, like that:
`/path/to/standby/`.

It was enough to add normalization on path for further usage, to fix the issue.

### DIFF
--- a/gpMgmt/bin/gppylib/commands/gp.py
+++ b/gpMgmt/bin/gppylib/commands/gp.py
@@ -67,6 +67,7 @@ RECOVERY_REWIND_APPNAME = '__gprecoverseg_pg_rewind__'
 PGDATABASE_FOR_COMMON_USE= 'postgres'
 
 def get_postmaster_pid_locally(datadir):
+    datadir = os.path.normpath(datadir)
     cmdStr = "ps -ef | grep 'postgres -D %s' | grep -v grep" % (datadir)
     name = "get postmaster"
     cmd = Command(name, cmdStr)
@@ -78,6 +79,7 @@ def get_postmaster_pid_locally(datadir):
         return -1
 
 def getPostmasterPID(hostname, datadir):
+    datadir = os.path.normpath(datadir)
     cmdStr="echo 'START_CMD_OUTPUT';ps -ef | grep 'postgres -D %s' | grep -v grep" % (datadir)
     name="get postmaster pid"
     cmd=Command(name,cmdStr,ctxt=REMOTE,remoteHost=hostname)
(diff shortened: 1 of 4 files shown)


gpMgmt/test/behave/mgmt_utils/gpinitstandby.feature
```diff
@@ -140,6 +140,15 @@ Feature: Tests for gpinitstandby feature
         Then gpinitstandby should return a return code of 0
         And verify that pg_hba.conf file has "standby" entries in each segment (primary and mirror) data directories
 
+    Scenario: gpinitstandby on dir with trailing slash
+        Given the database is running
+          And the standby is not initialized
+
+         When the user runs gpinitstandby with options "-S /tmp/standby_data/"
+         Then verify the standby coordinator entries in catalog
```
GirgerS: One small question: shouldn't we check for error code here as in [6X](https://github.com/GreengageDB/greengage/pull/498/changes) version?

ShapatinAA: Yep, Thank you!

gpMgmt/bin/gppylib/commands/gp.py
```diff
@@ -67,24 +67,26 @@
...
         sout = cmd.get_results().stdout.lstrip(' ')
-        return int(sout.split()[1])
+        comm, pid = sout.split()
+        return int(pid if comm == "postgres" else -1)
     except:
         return -1
 
 def getPostmasterPID(hostname, datadir):
-    cmdStr="echo 'START_CMD_OUTPUT';ps -ef | grep 'postgres -D %s' | grep -v grep" % (datadir)
+    cmdStr="echo 'START_CMD_OUTPUT';ps -p \"$(head -1 %s/postmaster.pid)\" -o comm= -o pid=" % (datadir)
```
GirgerS: I guess there is one more question. Why dollar sign doesn't need to be escaped? I thought that is is done [here](https://github.com/GreengageDB/greengage/blob/17dd13b61a72bb7fbebb92c01e57cb10251c4556/gpMgmt/bin/gppylib/commands/base.py#L495-L515), but it doesn't seem to be the case.

ShapatinAA: Got rid of this changes.

GirgerS (approved):
LGTM

red1452 (approved):

PR #501, opened 2026-06-19
Add experimental Arrow Flight FDW

iamlapa:
## Summary

This PR adds a new experimental `gpcontrib/arrowflight` extension that provides an `arrowflight_fdw` foreign data wrapper for Apache Arrow Flight data exchange.

The extension supports segment-parallel reads and writes between Greengage Database segments and external Arrow Flight services.

## Motivation

Existing external data paths such as `gpfdist` are reliable, but they move typed data through text-oriented formats. That adds CPU cost for type output/input conversion and can increase transferred bytes for many typed datasets.
Arrow Flight provides a typed, binary, columnar streaming protocol over gRPC.

## What Changed

- Added `gpcontrib/arrowflight` extension:
  - `arrowflight_fdw_handler`
  - `arrowflight_fdw_validator`
  - `arrowflight_build_info`
  - `CREATE FOREIGN DATA WRAPPER arrowflight_fdw`
- Added FDW read path:
  - `GetFlightInfo -> FlightEndpoint/Ticket -> DoGet`
  - segment-index endpoint selection
  - projection pushdown metadata for the `GetFlightInfo` path
  - retry/timeout handling before streaming starts
- Added FDW write path:
  - `INSERT` support through Arrow Flight `DoPut`
  - staged `FinalizeOperation` / `AbortOperation`
  - operation metadata
  - batch row/byte flushing controls
- Added server-level connection/security options:
  - host/port defaults
  - TLS, mTLS client cert/key, CA file
  - Bearer token file
- Added strict URL/path/ticket validation and safer diagnostics without logging secrets.
- Added regression tests for option validation, planning, write fail-fast behavior, and extension setup/teardown.
- Added development/test tooling:
  - deterministic `arrowflightd` smoke service
  - FDW read/write smoke scripts
  - benchmark scripts
  - DuckDB compose integration helper
- Registered the extension under `gpcontrib` only when `with_arrow_flight=yes`.

## Review Notes

- No Greengage core parser/planner changes are required by this version.
- Source distribution hints for connector-created temporary tables use column options (`insert_dist_by_key`, `insert_dist_by_key_weight`).
- `tools/arrowflightd.cpp` and `tools/arrowflight_smoke_server.cpp` are development/test tooling, not production data services.

## Validation

Arrow-enabled build and regression checks require a prepared Greengage build
tree and Apache Arrow C++ development packages (`arrow.pc` and `arrow-flight.pc`) available through `pkg-config`. They were validated in a Docker-based Greengage build environment:

```bash
make -C gpcontrib/arrowflight USE_ARROW_FLIGHT=1
make -C gpcontrib/arrowflight install USE_ARROW_FLIGHT=1
make -C gpcontrib/arrowflight installcheck USE_ARROW_FLIGHT=1
```
## Current Limitations and Assumptions

- The extension is experimental.
- `tools/arrowflightd.cpp` is a development/test service, not a production
  Arrow Flight service.
- `tls=false` is suitable only for trusted dev/test networks. Production should
  use `tls=true` with `tls_ca_file`, `tls_client_cert_file`,
  `tls_client_key_file`, and `auth_token_file`.
- TLS certificates must cover the configured Flight host in their SAN. Token
  files can be rotated in place; certificate, key, and CA changes require a
  process restart.
- IPv6 Arrow Flight URLs are not supported.
- Writes use staged finalization. Durable commit/rollback behavior is
  implemented by the receiving service.
- Retry is limited to the initial read/open path or write stream opening. After
  rows have been consumed or sent, the FDW does not retry automatically.
- Remote projection pushdown is implemented for the FDW `GetFlightInfo` path
  with `projection_pushdown=try|require`. Direct `DoGet(ticket)` URLs and
  zero-column scans still use the full remote schema.
- `max_batch_bytes` controls writer flushing, but it is not a full memory limit
  for all Arrow/gRPC buffering.
- `timestamp` accepts Arrow timestamps without timezone. `timestamptz` accepts
  empty/UTC timezones on read and emits UTC on write. Non-UTC timezone semantics
  are rejected.
- Native Arrow temporal mappings do not encode Greengage temporal infinities.
  `date`, `timestamp`, and `timestamptz` infinities fail deterministically
  instead of using private sentinel values.
- Arrow `date64` values must be day-aligned. Arrow timestamp values must fit the
  finite Greengage timestamp range after unit conversion and epoch adjustment.
- `json`, `jsonb`, and arrays currently use UTF-8 text exchange, not native
  nested Arrow vectors.


gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+        |
+        +-- write path
+              INSERT on QE segments -> DoPut
+              RecordBatch stream + schema metadata
+              FinalizeOperation / AbortOperation actions
+```
+
+`arrowflight_fdw` is created with `OPTIONS (mpp_execute 'all segments')`, so
+servers inherit segment execution unless they override `mpp_execute`. With this
+placement, each segment reads or writes its own part of the Flight stream.
```
AndrewSavitskiy: What will happen in terms of data consistency if some segment(s) will throw an error and whole transaction will abort?

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+Coordinator-only execution is supported by the FDW framework, but it removes
+segment parallelism.
+
+Execution placement affects the Flight contract:
+
+- `all segments` is the extension default. The FDW runs on QE segments; each
+  segment has its own `GpIdentity.segindex`, consumes the matching
+  `FlightEndpoint` for reads, and opens its own `DoPut` stream for writes.
```
AndrewSavitskiy: There might be the cases when number of readers/writers (executors) won't be matched number of segments, does this handled?

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+  placement.
+
+### Read Sequence
+
+The normal read path uses `mpp_execute 'all segments'` and one endpoint per
+Greengage segment.
+
+```mermaid
+sequenceDiagram
+    autonumber
```
AndrewSavitskiy: What's autonumber?

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+segment parallelism.
+
+Execution placement affects the Flight contract:
+
+- `all segments` is the extension default. The FDW runs on QE segments; each
+  segment has its own `GpIdentity.segindex`, consumes the matching
+  `FlightEndpoint` for reads, and opens its own `DoPut` stream for writes.
+- `coordinator` can be set explicitly for single-endpoint read tests. The
+  foreign scan runs only on the coordinator, so there is no segment id for
+  `flight_endpoint_policy=segment_index`. `INSERT` is not supported in this
```
AndrewSavitskiy: As I understand connector won't be fully functional in master/coordinator only mode (in terms of `mpp_execute` setting)?

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+    participant SQL as User SQL
+    participant QD as Coordinator
+    participant QE as QE segment i
+    participant FDW as arrowflight_fdw
+    participant SVC as Arrow Flight service
+
+    SQL->>QD: SELECT ... FROM foreign_table
+    QD->>QE: Dispatch foreign scan
+    QE->>FDW: Open scan with GpIdentity.segindex = i
+    FDW->>SVC: GetFlightInfo(path/segments/N[/columns/...])
```
AndrewSavitskiy: Does this structure (path/segments/N[/columns/...]) imply some kind of same structure of data placement (or maybe metadata description) on the source (Arrow Flight Service) side?

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+    SVC-->>FDW: N FlightEndpoint entries
+    FDW->>FDW: Pick endpoint i using segment_index
+    FDW->>SVC: DoGet(ticket for segment i)
+    SVC-->>FDW: Arrow RecordBatches
+    FDW-->>QE: Greengage tuples
+    QE-->>QD: Segment result slice
+    QD-->>SQL: Query result
+```
+
+With a direct `url` read, the FDW treats the URL as a ready `DoGet` ticket by
```
AndrewSavitskiy: All segments will read single URL in this mode? How does segment will know what data partition should it read?

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+default and skips `GetFlightInfo`.
+
+### Write Sequence
+
+`INSERT` also runs on QE segments. Each segment opens its own `DoPut` stream and
+the receiving service uses the operation metadata to stage, finalize, or abort
+the load.
```
AndrewSavitskiy: I've noted consistency related question above - finalize or abort dedicated segment loaded data or the whole transaction?

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+    SVC-->>FDW: Final ack metadata
+    alt Segment finished
+        FDW->>SVC: FinalizeOperation
+    else Segment error or cancel
+        FDW->>SVC: AbortOperation
+    end
+```
+
+Finalize and abort actions are sent by the segment writer. The external service
+owns the durable commit/rollback semantics for the shared `operation_id`.
```
AndrewSavitskiy: As I understood the context of data loading operation is defined by `operation_id`. Who/when commits or aborts this distributed (since there are multiple segments) transaction?

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+Do not include the segment count in table DDL. For example, on a cluster with
+two primary segments the table still uses a stable path:
+
+```sql
+OPTIONS
+    (path 'dataset/bench')
+```
+
+When a query runs, the FDW reads the current segment count with
+`getgpsegmentCount()` and sends this `GetFlightInfo` request to the service:
```
AndrewSavitskiy: It needs to be checked how is going to work with `ForeignServer.num_segments` parameter.
Since I've seen mentioned above `insert_dist_by_key` redistribution capability it's worth check too.

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+static.source=pxf_like_ingest,static.job_id=job_001,static.tenant=analytics
+```
+
+After the stream completes, the client expects a final metadata ack:
+
+```text
+af.ack.final=true
+```
+
+The client then sends a `FinalizeOperation` action. On error or interruption it
```
AndrewSavitskiy: Unfortunately transaction semantic isn't described here, so related above questions are sill relevant.

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+- Writes use staged finalization. Durable commit/rollback behavior is
+  implemented by the receiving service.
+- Retry is limited to the initial read/open path or write stream opening. After
+  rows have been consumed or sent, the FDW does not retry automatically.
+- Remote projection pushdown is implemented for the FDW `GetFlightInfo` path
+  with `projection_pushdown=try|require`. Direct `DoGet(ticket)` URLs and
+  zero-column scans still use the full remote schema.
+- `max_batch_bytes` controls writer flushing, but it is not a full memory limit
+  for all Arrow/gRPC buffering.
+- `timestamp` accepts Arrow timestamps without timezone. `timestamptz` accepts
```
AndrewSavitskiy: timestamp/timestamptz related limitations and assumptions need to be researched and validated in production dev.

gpcontrib/arrowflight/README.md
```diff
@@ -0,0 +1,936 @@
...
+- `tls=false` is suitable only for trusted dev/test networks. Production should
+  use `tls=true` with `tls_ca_file`, `tls_client_cert_file`,
+  `tls_client_key_file`, and `auth_token_file`.
+- TLS certificates must cover the configured Flight host in their SAN. Token
+  files can be rotated in place; certificate, key, and CA changes require a
+  process restart.
+- IPv6 Arrow Flight URLs are not supported.
+- FDW writes support `INSERT` only; `UPDATE`, `DELETE`, and `COPY` are not
+  implemented.
+- Writes use staged finalization. Durable commit/rollback behavior is
```
AndrewSavitskiy: Staged and overall transactional semantics needs to be researched and validated in production dev.

gpcontrib/arrowflight/src/arrowflight_fdw.cpp
```diff
@@ -0,0 +1,939 @@
...
+											  server->options,
+											  true);
+}
+
+static ArrowFlightFdwPlanState *
+af_fdw_get_plan_state(Oid foreigntableid)
+{
+	ForeignTable *table = GetForeignTable(foreigntableid);
+	ArrowFlightFdwPlanState *state =
+		(ArrowFlightFdwPlanState *) palloc0(sizeof(*state));
```
AndrewSavitskiy: I would write as `sizeof(ArrowFlightFdwPlanState)`.

gpcontrib/arrowflight/src/arrowflight_fdw.cpp
```diff
@@ -0,0 +1,939 @@
...
+	char	   *rows_value;
```
AndrewSavitskiy: Common code convention implies that definitions like these should be placed at the beginning of the function. I note that only here.

gpcontrib/arrowflight/src/arrowflight_fdw.cpp
```diff
@@ -0,0 +1,939 @@
...
+									 slot,
+									 exec_state->project_all,
+									 exec_state->projected_attrs,
+									 &exec_state->security_options);
+}
+
+static void
+af_fdw_ReScanForeignScan(ForeignScanState *node)
+{
+	ArrowFlightFdwExecState *exec_state =
```
AndrewSavitskiy: Since `path.rescannable` been set to `false` in `af_fdw_GetForeignPaths` before, I guess it's cleaner to throw an ERROR here.

gpcontrib/arrowflight/src/arrowflight_fdw.cpp
```diff
@@ -0,0 +1,939 @@
...
+		(ArrowFlightFdwExecState *) node->fdw_state;
+
+	if (exec_state == NULL)
+		return;
+
+	if (exec_state->flight_state != NULL)
+	{
+		af_flight_stream_close(exec_state->flight_state);
```
AndrewSavitskiy: As I see cleanup relies entirely on `EndForeignScan`/`EndForeignModify` and the local `PG_CATCH` blocks. But on an error or user cancel that originates outside the FDW's (for example SIGINT in a motion), these nodes are not run — abort goes through resource-owner release and memory-context reset. I think that may lead to: leaked connections, C++ heap leak, and reader cancels never runs?

As I don't yet understand the transactional semantics of this connector, my hypothesis is that server-side staging operation is never aborted/finalized.

AndrewSavitskiy (commented):
I think that the main question is what are the transactional requirements to the connector. Second one, is what are the scenarios and counterparts of the connector, since it's not generic and implements it's own Arrow Flight based protocol (distribution of endpoints to QE, commit decision on intermediate server side).

gpcontrib/arrowflight/src/flightsql_reader.cpp
```diff
@@ -0,0 +1,2512 @@
...
+	if (schema == nullptr)
+		throw std::runtime_error("Arrow Flight stream returned no schema");
+
+	if (state == nullptr)
+		throw std::runtime_error("Arrow Flight stream state is not initialized");
+
+	state->attr_batch_indexes.assign(tupdesc->natts, -1);
+	state->expected_batch_columns = schema->num_fields();
+
+	if (schema->num_fields() == tupdesc->natts)
```
AndrewSavitskiy, line 743: There is a problem related to this comarison:

```
postgres=# ALTER FOREIGN TABLE af_dbg_read DROP COLUMN label;
ALTER FOREIGN TABLE
postgres=# select * from af_dbg_read where amount >= 0.6 and ts <= now() and active = true;
ERROR:  Flight SQL read failed: Arrow Flight stream has 6 columns, external table expects 7  (seg0 slice1 127.0.0.1:10111 pid=1577122)
```

I guess that is because this line compares `schema->num_fields()` against `tupdesc->natts`, which counts dropped attributes.

Selecting an explicit subset still works, because that takes the name-matching branch: generated remote SQL is correct, it requests exactly the six live columns. Looks like only the validation arithmetic is wrong.

gpcontrib/arrowflight/src/flightsql_reader.cpp
```diff
@@ -0,0 +1,2512 @@
...
+	{
+		if (state->projection_pushdown_required)
```
AndrewSavitskiy, line 745: I guess that a well-behaving SQL server will honor the projection, but what if it doesn't, so it makes the FDW fragile against any server that treats the query text loosely. I think it's better to match by name and discarding the extras.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+
+	if (plan == NULL || plan->finished)
+		return;
+
+	error = complete ?
+		af_flightsql_mpp_complete_plan(
+			plan->url, plan->serialized_plan, plan->timeout_ms,
+			plan->max_plan_bytes, &plan->security_options) :
+		af_flightsql_mpp_abort_plan(
+			plan->url, plan->serialized_plan, plan->timeout_ms,
```
AndrewSavitskiy, line 1504: Inherited configured timeout_ms (default is -1) and both connect paths set a deadline only if (timeout_ms > 0).

So an unreachable server will forever block transaction abort (hanging in abort I guess is bad and exactly where the user cannot intervene).

This one is called in callbacks, that handle `XACT_EVENT_ABORT` and `XACT_EVENT_PARALLEL_ABORT`.

It also fires on `ROLLBACK TO SAVEPOINT` and on error recovery for any failed statement.

gpcontrib/arrowflight/src/flightsql_mpp.cpp
```diff
@@ -0,0 +1,940 @@
...
+			"Flight SQL MPP plan route count does not match Greengage segment count");
+	if (require_cluster_transaction &&
+		response.transaction_scope() !=
+		mpp::TRANSACTION_SCOPE_CLUSTER)
+		throw std::runtime_error(
+			"Flight SQL MPP planned ingest requires cluster-scoped transactions");
+
+	std::vector<bool> seen((size_t) segment_count, false);
+
+	for (const mpp::IngestRoute& route : response.routes())
```
AndrewSavitskiy, line 403: This is what AI had found regarding the `endpoint_location_allowlist` in write path:

`endpoint_location_allowlist` is consulted only in the reader. `ValidatePlan` checks route count, uniqueness, lease window and that no route changes the origin's TLS mode — but never the host. A compromised or malicious control endpoint can therefore name any `grpc+tls://host:port` in its create response, and every segment QE will connect there and hand over the bearer token, the mTLS client identity and the full `INSERT` payload.

This is precisely the redirection the README says the allowlist prevents. That guarantee holds for reads and not for write_routing_mode=planned.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+			for (plan = flightsql_remote_plans;
+				 plan != NULL;
+				 plan = plan->next)
+			{
+				if (!plan->finished)
+					ereport(ERROR,
+							(errcode(ERRCODE_FDW_ERROR),
+							 errmsg("Flight SQL MPP ingest plan was not completed")));
+			}
+			for (transaction = flightsql_remote_transactions;
```
AndrewSavitskiy, line 1595: I guess that since `XACT_EVENT_PRE_COMMIT` commits each registered remote transaction in a loop with `error_on_failure=true` (third argument). If a later commit fails, the earlier ones are already committed and cannot be undone; the local transaction then aborts, leaving the two sides inconsistent. 

I understand that nobody promised anything like ACID here, but in production version the README's atomicity should considers the case.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+	}
+
+	if (error_on_failure)
+		ereport(ERROR,
+				(errcode(ERRCODE_FDW_ERROR),
+				 errmsg("could not %s Flight SQL transaction",
+						commit ? "commit" : "rollback"),
+				 errdetail_internal("%s", error)));
+
+	ereport(WARNING,
```
AndrewSavitskiy, line 1439: That's what AI had noticed here:

`flightsql_finish_remote_transaction` always logs "could not rollback Flight SQL transaction" in its non-fatal branch regardless of the commit flag; flightsql_finish_remote_plan likewise always says "could not abort". Latent today — only abort paths pass error_on_failure=false — but it will mislead the first time a non-fatal commit path is added.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("Flight SQL server does not support planned MPP ingest"),
+				 errdetail("Required actions: %s, %s, and %s.",
+						   AF_FLIGHT_SQL_MPP_CREATE_ACTION,
+						   AF_FLIGHT_SQL_MPP_COMPLETE_ACTION,
+						   AF_FLIGHT_SQL_MPP_ABORT_ACTION)));
+
+	if (strcmp(mode, AF_FLIGHT_SQL_WRITE_TRANSACTION_REQUIRED) == 0)
+	{
+		char	   *transaction_id =
```
AndrewSavitskiy, line 788: This one is called for every `INSERT`. 

`flightsql_register_remote_transaction` adds it to a linked list (`transaction->next = flightsql_remote_transactions`) there is no lookup/no reuse. 
Several INSERTs in one local transaction against one server opened same number of remote transactions.

I thing these consequences follow (some of them are obvious, but anyway should be noticed):

1. Statements are not atomic related to each other. Each commits independently at `PRE_COMMIT`.

2. There is no way to guarantee an ordering. Commit order is reverse-registration, since the list is prepended, so the last statement commits first. 

3. Two open transactions from one session writing the same table can self-deadlock on a locking engine. Since there are branched with missing deadline, that blocks forever. I think it's the most critical potential issue.

4. Possible resource pressure scale with statement count, not transaction count.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+			plan_link = &plan->next;
+			continue;
+		}
+
+		if (event == SUBXACT_EVENT_COMMIT_SUB)
+		{
+			plan->subtransaction_id = parent_subid;
+			plan_link = &plan->next;
+		}
+		else if (event == SUBXACT_EVENT_ABORT_SUB)
```
AndrewSavitskiy, line 1659: Related to timeout and blocking potential issue:

Since PL/pgSQL exceptions basically is a subtransactions, and `SUBXACT_EVENT_ABORT_SUB` issues the remote rollback with no deadline (we're inside the handle here). 

So an exception handler for recover from a failing remote server will hang forever on exactly the failure it exists to catch.

+1 to properly handling unbounded network I/O in abort paths.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+	if (capabilities->sql_transactions !=
+		AF_FLIGHT_SQL_CAPABILITY_SUPPORTED)
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("write_transaction_mode=required needs SQL transaction support"),
+				 errdetail("Server: %s %s.", server_name, server_version)));
+}
+
+static void
+flightsql_register_remote_transaction(
```
AndrewSavitskiy, line 1374: Found by AI.

There are two lists:

```
static FlightSqlRemoteTransaction *flightsql_remote_transactions = NULL;                                                                          
static FlightSqlRemotePlan        *flightsql_remote_plans        = NULL;  
```

Both lists only set `finished = true`; entries are freed when `TopTransactionContext` goes away. Inside one long transaction running many statements, every subsequent flightsql_complete_query_plans and every callback walks all prior entries — quadratic in statement count.

Who pays for it:

```
flightsql_complete_query_plans()  — walks the entire list on EVERY query on the dispatcher (called from the ExecutorFinish hook, :590)                                                                     
XACT_EVENT_PRE_COMMIT     — walks all plans            
XACT_EVENT_PRE_COMMIT     — walks all transactions                                                                                  
```

`flightsql_complete_query_plans` is the expensive one: it runs from the `ExecutorFinish` hook for every query on the coordinator, and re-scans every already-finished entry from every earlier statement. With N statements in a transaction that's O(N²) list traversal, and combined with S1 — where each `INSERT` adds a fresh transaction node — the list grows by one node per write statement.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+							 makeString(pstrdup("")),
+							 projected_attrs,
+							 makeInteger(project_all ? 1 : 0));
+
+	return make_foreignscan(tlist, local_exprs, baserel->relid, NIL,
+							fdw_private, NIL, remote_exprs, outer_plan);
+}
+
+static void
+flightsql_ExecutorStart(QueryDesc *query_desc, int eflags)
```
AndrewSavitskiy, line 562: Personally, I found the design based on using `ExecutorStart`/`ExecutorEnd` hooks rather smelly.

Why is it smelly from my point for view:
1. Every statement in the backend walks its plan tree once the FDW has been touched even once.                                   
2. The chain is captured at first use, so if another extension replaces `ExecutorStart_hook` without chaining, every arrowflight sql query fails with flightsql_fdw scan was not prepared on the dispatcher.

As I understood the key it that FlightInfo is computed on the QD but consumed on the QEs, and the dispatched plan is the channel the FDW uses to get it there.

Maybe there is another way to push it to QE? As we do in other proprietary connectors?

AI additional comments:

Once the FDW has been touched once in a session, `flightsql_ExecutorStart` calls `flightsql_plan_contains_scan` for every non-EXPLAIN query on the dispatcher. The walker recurses the entire plan tree including subplans, and performs three catalog lookups per `ForeignScan`/`ModifyTable` found. A session that reads a foreign table once and then runs unrelated OLTP queries pays this on every statement, forever.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+	}
+
+	if (next_ExecutorStart_hook != NULL)
+		next_ExecutorStart_hook(query_desc, eflags);
+	else
+		standard_ExecutorStart(query_desc, eflags);
+}
+
+static void
+flightsql_ExecutorFinish(QueryDesc *query_desc)
```
AndrewSavitskiy, line 582: This hook is skipped on statement error and on abandoned portals. 

The` PRE_COMMIT` guard then converts a missed completion into "Flight SQL MPP ingest plan was not completed" at commit time rather than completing or aborting it.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+}
+
+static void
+flightsql_ExecutorStart(QueryDesc *query_desc, int eflags)
+{
+	if (Gp_role == GP_ROLE_DISPATCH &&
+		(eflags & EXEC_FLAG_EXPLAIN_ONLY) == 0 &&
+		flightsql_plan_contains_scan(query_desc->plannedstmt))
+	{
+		PlannedStmt *stmt =
```
AndrewSavitskiy, line 568: Found by AI:

`copyObjectImpl` runs on every execution, including one-shot plans where there is no cached plan to protect. The copy itself is correct and necessary for cached plans (see below) — only its unconditional application is wasteful.

gpcontrib/arrowflight/src/flightsql_mpp.cpp
```diff
@@ -0,0 +1,940 @@
...
+	{
+		if (ch < 0x20 || ch == 0x7f)
+			throw std::runtime_error(
+				"Flight SQL auth_token_file contains control characters");
+	}
+	return token;
+}
+
+MppClient
+Connect(const char *url, int timeout_ms,
```
AndrewSavitskiy, line 125: Looks like that `Connect` builds a fresh `FlightClient` on every call.

A single required + planned `INSERT` therefore costs six separate coordinator connections:

capability discovery > ListActions > BeginTransaction > MPP create > MPP complete > Commit

Each one re-reads the certificates related entities and repeat the TLS handshake. 
Also in case of many severs and statements it should be multiplid by servers and by statements count. 
With TLS this is may be a significant per-statement overhead.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
```diff
@@ -0,0 +1,2190 @@
...
+	const char *query;
+	char	   *serialized;
+	ArrowFlightSqlCapabilities capabilities;
+
+	flightsql_validate_mpp_contract(foreigntableid);
+	if (list_length(node->fdw_private) != 4)
+		elog(ERROR, "flightsql_fdw: invalid private scan state");
+
+	query = strVal(list_nth(node->fdw_private, FLIGHTSQL_PRIVATE_QUERY));
+	af_flightsql_get_capabilities(url, &security, &capabilities);
```
AndrewSavitskiy, line 732: Found by AI:

`flightsql_PrepareForeignScanForDispatch` calls `af_flightsql_get_capabilities` and never reads the result; reads do not check capabilities. Softened by a per-backend cache keyed on URL plus security options (flightsql_reader.cpp:182), so the real cost is one `GetSqlInfo` per backend per distinct server, not per query.

gpcontrib/arrowflight/src/flightsql_mpp.cpp
```diff
@@ -0,0 +1,940 @@
...
+	if (authority.front() == '[')
+		throw std::runtime_error(
+			"IPv6 Flight SQL MPP routes are not supported");
+
+	std::string url =
+		std::string(AF_SCHEME) + authority +
+		(tls ? "?tls=true" : "?tls=false");
+	ArrowFlightConnection parsed;
+
+	af_parse_flight_connection(url.c_str(), &parsed);
```
AndrewSavitskiy, line 367: Found by AI:

`af_parse_flight_connection` ereports while `std::strings` are in scope. The longjmp skips their destructors, leaking malloc'd memory that no PostgreSQL memory context tracks. Error paths only, and bounded per failed statement.

AndrewSavitskiy (commented):
I think that Arrow Flight **SQL** connector is the right direction (rather then previous rather low-level Arrow Flight only), but I think that for it's little bit over-engineered in the part of transactional support in `required` mode. As I understand there are no real world servers, that satisfy the required transactional capabilities.
Implementation of the planned contract exists as synthetic server in tools folder.

For my point of view that main obstacle, is that transactions must survive across connections. So the main requirement - begin on one connection, ingest on N, commit on a fourth. Now no SqlInfo flag expresses it. Real world servers can advertise all four flags honestly and still fail, because almost every engine binds transactions to sessions.

I think that for now the `auto_commit` is enough for real-world now.

That my overall opinion. Other comments about current implementation.

gpcontrib/arrowflight/src/flightsql_reader.cpp
iamlapa: Fixed. Schema validation now matches requested live attributes by name instead of comparing the field count with `tupdesc->natts`. Added tests for dropping the first, middle and last columns, including `SELECT *`, with both optimizers.

gpcontrib/arrowflight/src/flightsql_reader.cpp
iamlapa: changed to name-based matching for both full and projected reads. Unrequested fields are ignored; missing, duplicate or incompatible requested fields still produce an error.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: fixed. Remote abort/rollback RPCs now have finite deadlines within a shared five-second cleanup budget, even with `timeout_ms=-1`. A shorter configured timeout still applies.

gpcontrib/arrowflight/src/flightsql_mpp.cpp
iamlapa: fixed. Planned write routes now use the same origin-or-allowlist check as read endpoints, together with the TLS-mode check.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: removed the multiple-remote-transaction case for required writes - one local transaction now reuses one remote transaction, and a second foreign server is rejected before its remote execution.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: fixed both messages

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: Fixed. Subsequent required INSERTs through the same foreign server reuse the existing transaction ID and commit once.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: fixed. Subtransaction cleanup now uses the same bounded cleanup budget.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: fixed the accumulation of completed entries. Completed plans are unlinked and their memory contexts deleted after each statement; subtransaction abort also removes its plans. Repeated inserts reuse one remote transaction entry, so later statements no longer scan all previously completed writes.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: Partially addressed. FDW detection now uses a marker in `fdw_private`, eliminating the catalog lookups, and stops at the first match. The hooks and plan-tree walk remain. They currently provide the pre-dispatch point for attaching runtime FlightInfo and ingest state; replacing them needs a separate dispatch design. The hook-chaining concern remains open.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: I checked the lifecycle paths. Statement errors use transaction/subtransaction abort callbacks to abort unfinished plans rather than completing them. Normal cleanup of a non-failed portal calls ExecutorFinish. Added partial read-cursor close and exception-recovery tests. The `PRE_COMMIT` guard remains intentionally: an unfinished ingest must not be silently committed.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: Agreed that copying a one-shot plan is avoidable overhead. Fixed.

gpcontrib/arrowflight/src/flightsql_mpp.cpp
iamlapa: Client reuse is still outstanding. Transaction reuse reduces BeginTransaction/Commit calls across multiple inserts, and capability discovery is cached, but control RPCs still create new clients.

gpcontrib/arrowflight/src/flightsql_fdw.cpp
iamlapa: removed the unused capability lookup from read dispatch.

gpcontrib/arrowflight/src/flightsql_mpp.cpp
iamlapa: fixed

PR #507, opened 2026-06-24
GG-580: Ensure proper order of statements in fts_segment_reset test

KnightMurloc:
Ensure proper order of statements in the test fts_segment_reset.

The aforementioned isolation test contains statements that are started in
parallel at the same time. Under some circumstances, e.g., when using a Docker
container based on Ubuntu 22.04, these statements are executed in an unexpected
order, causing the test to fail.

This commit adds a fault injector to enforce the order of statements by
introducing a dependency between getting segment master into reset and the start
of next command.

(cherry picked from commit 2ada266e665f310f0ca91ed8220dba213648a52d)

---

Merge with rebase


Stolb27 (approved):
as clean cherry pick

silent-observer (approved):

PR #509, opened 2026-06-25
CI-5908: Directory `sqldump` absent for 8.x

magf:
## Directory `sqldump` absent for 8.x

- Create directory immediately before create dump with `777` permissions
- Set `777` permissions on `sqldump` directory if exists
- No need to pack SQL dump in CI

Task:[ CI-5908](https://tracker.yandex.ru/CI-5908)

### DIFF
--- a/concourse/scripts/dumpdb.bash
+++ b/concourse/scripts/dumpdb.bash
@@ -25,5 +25,6 @@ psql \
     echo ""
 done
 
+mkdir -pm 777 sqldump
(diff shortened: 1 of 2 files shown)


Stolb27 (approved):

silent-observer (approved):

Stolb27:
Merged as is, because CI doesn't check affected code branches. Wait for the dump

PR #512, opened 2026-06-26
GG-574: Fix scan for AO table without columns

hilltracer:
Commit d3dcb9b taught AO row scans to determine the number of physical
attributes from metadata for fast ALTER TABLE ADD COLUMN. That code assumed
that each scanned tuple had at least one physical attribute and incorrectly
started from 1. For a zero-column AO row table, the scan tuple descriptor had
natts = 0, but the read block kept curLargestAttnum at 1 and passed
expected_natts = 1 to create_memtuple_binding(). Assertion builds failed
because create_memtuple_binding() required expected_natts to be between 0 and
tupdesc->natts. Release builds could create an invalid binding and crash later
while deforming the memtuple.

Initialize curLargestAttnum to 0 when starting a new AO read block, and allow
AppendOnlyExecutorReadBlock_BindingInit() to create a binding with zero
expected attributes.

Ticket: GG-574


dkovalev1 (approved):

GirgerS (approved):
LGTM

PR #514, opened 2026-06-28
GG-551: Report tables and columns removed from Greengage 7 (6.x)

GirgerS:
This patch makes several adjustments to pg_upgrade_support
functions to report more objects that would make upgrade fail.

These changes are as follows:
- backport pg_upgrade_support functions that check for removed
tables and columns.
- report removed objects in materialized views, in addition to regular
  views.
- also, adapt static/dynamic OID logic for schemas other than
   pg_catalog, as it was done in 5.x.

Don't add tests because they belong to the ggupgrade repo
and will be added there after this patch.


contrib/pg_upgrade_support/pg_upgrade_support.c
```diff
@@ -66,13 +78,278 @@ PG_FUNCTION_INFO_V1(view_has_removed_operators);
...
+static Oid __gp_aocsseg_oids[]         = {2205};
+static Oid __gp_aocsseg_history_oids[] = {2205};
+static Oid __gp_aoseg_oids[]           = {2205};
+static Oid __gp_aoseg_history_oids[]   = {2205};
+
+static RemovedFunctionDynamic functions_with_changed_signatures_dynamic[] =
+{
+	{"gp_toolkit",  "__gp_aocsseg",          __gp_aocsseg_oids,          1},
+	{"gp_toolkit",  "__gp_aocsseg_history",  __gp_aocsseg_history_oids,  1},
+	{"gp_toolkit",  " __gp_aoseg",           __gp_aoseg_oids,            1},
```
hilltracer:
```suggestion
	{"gp_toolkit",  "__gp_aoseg",           __gp_aoseg_oids,            1},
```
Right?
Did you test that your patch works fine with all removed objects? Can you please add commands sequence to manual test your patch. And show result of testing.

contrib/pg_upgrade_support/pg_upgrade_support.c
```diff
@@ -18,6 +18,16 @@
 #include "catalog/oid_dispatch.h"
 #include "catalog/pg_authid.h"
 #include "catalog/pg_class.h"
+#include "catalog/pg_constraint.h"
+#include "catalog/pg_authid.h"
```
hilltracer:
```suggestion
```
it's a duplicate.

contrib/pg_upgrade_support/pg_upgrade_support.c
```diff
@@ -823,3 +1190,195 @@ check_node_changed_function_signatures_walker(Node *node, void *context)
...
+	}
+
+	if (IsA(node, Var))
+	{
+		Var * var = (Var *) node;
+		List *rtable = (List *) list_nth(context->rtableStack, var->varlevelsup);
+		RangeTblEntry *rte = (RangeTblEntry *) list_nth(rtable, var->varno - 1);
+		return is_removed_column(rte->relid, var->varattno);
+
+	}
```
hilltracer, line 1516: Please look at the Var handling in find_expr_references_walker() so this implementation is different. Your walker directly does list_nth(rtable, varno - 1) and checks rte->relid/varattno, but does not validate varlevelsup/varno, does not process InvalidAttrNumber and does not expose RTE_JOIN via joinaliasvars. Is it expected? Maybe use similar implementation?

https://github.com/GreengageDB/greengage/blob/e3119106f568ccfc4c01a099229f9ee9c0ab9c87/src/backend/catalog/dependency.c#L1539-L1591

hilltracer:
So now we inform user only about name of views and mviews that have removed ralations/attributes. But it would be more informative to add information about which specific relations/attributes were removed in these views, this could help rewrite these views. Does this make sense? How difficult would it be to return this?

GirgerS:
Some test for: column walker, removed tables, columns, and objects that were moved to 'dynamic' category. Assuming that there are no questions to table walker. 

Scripts for testing removed tables and columns are generated the same way it is described in the ticket. Reporting code was slightly changed to duplicate output to stdout (to make it easier to see changes):

```diff
diff --git a/src/bin/pg_upgrade/greengage/check_gp.c b/src/bin/pg_upgrade/greengage/check_gp.c
index a654dc1f5eb..c7508e1cc4d 100644
--- a/src/bin/pg_upgrade/greengage/check_gp.c
+++ b/src/bin/pg_upgrade/greengage/check_gp.c
@@ -1525,6 +1525,8 @@ check_views_with_removed_columns()
 		}
 
 		fprintf(script, "Database: %s\n", active_db->db_name);
+		printf("Database: %s\n", active_db->db_name);
+
 
 		i_viewname = PQfnumber(res, "badviewname");
 		i_removed_columns = PQfnumber(res, "removed_columns");
@@ -1534,6 +1536,9 @@ check_views_with_removed_columns()
 			fprintf(script, "  %s\n%s\n",
 					PQgetvalue(res, rowno, i_viewname),
 					PQgetvalue(res, rowno, i_removed_columns));
+			printf("  %s\n%s\n",
+					PQgetvalue(res, rowno, i_viewname),
+					PQgetvalue(res, rowno, i_removed_columns));
 		}
 
 		PQclear(res);
@@ -1630,6 +1635,7 @@ check_views_with_removed_relations()
 		}
 
 		fprintf(script, "Database: %s\n", active_db->db_name);
+		printf("Database: %s\n", active_db->db_name);
 
 		i_viewname = PQfnumber(res, "badviewname");
 		i_removed_tables = PQfnumber(res, "removed_tables");
@@ -1639,6 +1645,10 @@ check_views_with_removed_relations()
 			fprintf(script, "  %s\n%s\n",
 					PQgetvalue(res, rowno, i_viewname),
 					PQgetvalue(res, rowno, i_removed_tables));
+			printf("  %s\n%s\n",
+					PQgetvalue(res, rowno, i_viewname),
+					PQgetvalue(res, rowno, i_removed_tables));
+
 		}
 
 		PQclear(res);

```

[test_misc_objects.sql](https://github.com/user-attachments/files/29691066/test_misc_objects.sql)
[test_misc_objects_out.txt](https://github.com/user-attachments/files/29691143/test_misc_objects_out.txt)
[test_walker.sql](https://github.com/user-attachments/files/29691158/test_walker.sql)
[test_walker_out.txt](https://github.com/user-attachments/files/29691119/test_walker_out.txt)
[test_columns.sql](https://github.com/user-attachments/files/29691229/test_columns.sql)
[test_columns_out.txt](https://github.com/user-attachments/files/29691231/test_columns_out.txt)
[test_tables.sql](https://github.com/user-attachments/files/29691232/test_tables.sql)
[test_tables_out.txt](https://github.com/user-attachments/files/29691235/test_tables_out.txt)

contrib/pg_upgrade_support/pg_upgrade_support.c
GirgerS: You are right, stupid mistake by me. I have fixed it and provided the tests.

contrib/pg_upgrade_support/pg_upgrade_support.c
GirgerS: > Your walker directly does list_nth(rtable, varno - 1) and checks rte->relid/varattno, but does not validate varlevelsup/varno

It should be fine (as long as catalog and query are correct), but I've added additional checks just in case.

> does not process InvalidAttrNumber

That's an interesting topic. It can appear for queries like:

```sql
SELECT pg_class FROM pg_class;
```

I decided to report such cases (even though they can be upgraded successfully, see comment in the code on why, I wouldn't explain it better than there). BUT! We can report false positives, which is not great to say the least. So we might thing about postponing these changes, or searching for a more clever fix.


```
and does not expose RTE_JOIN via joinaliasvars
```

Yes, that was a mistake, my walker wouldn't handle joins well. I've added this code. And some explanatory comments.

GirgerS:
> So now we inform user only about name of views and mviews that have removed ralations/attributes. But it would be more informative to add information about which specific relations/attributes were removed in these views, this could help rewrite these views. Does this make sense? How difficult would it be to return this?

I've added bare minimum of reporting, so we tell which removed columns/tables are used. It would be too inconvenient for the user otherwise. We can also report where in the query these objects are located, but I decided not to.

contrib/pg_upgrade_support/pg_upgrade_support.c
GirgerS: Thanks.

hilltracer (dismissed):

contrib/pg_upgrade_support/pg_upgrade_support.c
```diff
@@ -66,13 +77,298 @@ PG_FUNCTION_INFO_V1(view_has_removed_operators);
...
+	Assert(matview->rd_rel->relkind == RELKIND_MATVIEW);
+
+	if (matview->rd_rel->relhasrules == false ||
+		matview->rd_rules->numLocks < 1)
+		elog(ERROR,
+			 "materialized view \"%s\" is missing rewrite information",
+			 RelationGetRelationName(matview));
+
+	if (matview->rd_rules->numLocks > 1)
+	elog(ERROR,
```
bimboterminator1: Intend elog, please

contrib/pg_upgrade_support/pg_upgrade_support.c
```diff
@@ -823,3 +1209,419 @@ check_node_changed_function_signatures_walker(Node *node, void *context)
...
+check_and_report_removed_columns(RemovedColumnsWalkerContext *context, Oid reloid, int attnum)
+{
+	int i;
+	Oid schema_oid;
+	int removed_column_attnum;
+
+	for (i = 0; i < num_removed_columns_static; i++)
+	{
+		removed_column_attnum = removed_columns_static[i].attnum;
+		if (reloid == removed_columns_static[i].reloid &&
```
bimboterminator1, line 1393: Maybe aviod reporting for negatice attnums and add `removed_column_attnum > 0` ?
ame consideration for the dynamic loop just below.

bimboterminator1 (commented):
LGTM, just a few insignificant remarks

contrib/pg_upgrade_support/pg_upgrade_support.c
GirgerS: Can you give me more context, please? I'm missing the reason why we wouldn't want to report them. From what I've seen, they behave like regular columns, and cause upgrade to fail too.

contrib/pg_upgrade_support/pg_upgrade_support.c
GirgerS: Thanks!

bimboterminator1 (approved):

hilltracer (approved):

PR #515, opened 2026-06-28
GG-551: Report tables and columns removed from Greengage 7 (7.x)

GirgerS:
Check for views with removed relations and columns before 
performing upgrade. They will cause pg_restore to fail in the 
middle of the upgrade, and it might get a lot of time to get 
to this point. So report them to the user in advance.

The same problems can appear in materialized views,
therefore it is necessary to checks them in addition 
to regular ones.

Don't add tests because they belong to the ggupgrade repo
and will be added there after this patch.

Ticket: GG-551

### DIFF
--- a/src/bin/pg_upgrade/greengage/check_gp.c
+++ b/src/bin/pg_upgrade/greengage/check_gp.c
@@ -37,6 +37,8 @@ static void check_views_with_changed_function_signatures(void);
 static void check_execute_on_master_functions(void);
 static void check_for_missing_support_function_for_partitions(void);
 static void check_for_incompatible_guc_settings(void);
+static void check_views_with_removed_columns(void);
+static void check_views_with_removed_relations(void);
 
 /*
  *	check_greengage
@@ -60,6 +62,8 @@ check_greengage(void)
 	check_views_with_removed_operators();
 	check_views_with_removed_functions();
 	check_views_with_removed_types();
+	check_views_with_removed_columns();
+	check_views_with_removed_relations();
 	check_for_disallowed_pg_operator();
 	check_views_with_changed_function_signatures();
 	check_execute_on_master_functions();
@@ -882,7 +886,7 @@ check_views_with_removed_operators()
 		res = executeQueryOrDie(conn,
 								"SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname) AS badviewname "
 								"FROM pg_class c JOIN pg_namespace n on c.relnamespace=n.oid "
-								"WHERE c.relkind = 'v' AND "
+								"WHERE c.relkind IN ('v', 'm') AND "
 								"view_has_removed_operators(c.oid) = TRUE;");
 
 		PQclear(executeQueryOrDie(conn, "DROP FUNCTION view_has_removed_operators(OID);"));
@@ -969,7 +973,7 @@ check_views_with_removed_functions()
 		res = executeQueryOrDie(conn,
 								"SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname) AS badviewname "
 								"FROM pg_class c JOIN pg_namespace n on c.relnamespace=n.oid "
-								"WHERE c.relkind = 'v' "
+								"WHERE c.relkind IN ('v', 'm') "
 								"AND c.oid >= 16384 "
 								"AND view_has_removed_functions(c.oid) = TRUE;");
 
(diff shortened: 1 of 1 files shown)


src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1447,3 +1451,213 @@ check_for_missing_support_function_for_partitions()
...
+								"          public.get_removed_columns(c.oid) AS removed_columns "
+								"	FROM pg_class c JOIN pg_namespace n on c.relnamespace=n.oid "
+								"	WHERE c.oid >= %d AND "
+								"	c.relkind IN ('v', 'm') "
+								") views "
+								"WHERE removed_columns != ''",
+								FirstNormalObjectId);
+
+		PQclear(executeQueryOrDie(conn, "DROP FUNCTION public.get_removed_columns(OID);"));
+		PQclear(executeQueryOrDie(conn, "SET search_path to 'pg_catalog';"));
```
bimboterminator1: Do we set searc path to public somewhere? I guess it can be dropped here and below

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1447,3 +1451,213 @@ check_for_missing_support_function_for_partitions()
...
+		pg_log(PG_REPORT, "fatal\n");
+		gp_fatal_log(
+		   "| Your installation contains views using removed columns.\n"
+		   "| These columns are no longer present on the target version.\n"
+		   "| These views must be updated to use columns supported in the\n"
+		   "| target version or removed before upgrade can continue. A list\n"
+		   "| of the problem views is in the file:\n\t%s\n\n", output_path);
+	}
+
+	check_ok();
```
bimboterminator1: How this will work in `--continue-check-on-fatal` ? Maybe add else?

GirgerS: To be fail, I completely missed this flag, big thanks for noticing. I've also fixed previous fix added by us (as this issue seems to be too small for a separate ticket).

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Right, it is an artifact from copypasted code that I forgot to remove. Should be fixed now.

bimboterminator1 (dismissed):

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1343,8 +1347,8 @@ check_execute_on_master_functions()
 			"| return SETOF rows.\n"
 			"| A list of the problem functions is in the file:\n\t%s\n\n", output_path);
 	}
-
-	check_ok();
+	else
+		check_ok();
```
hilltracer: Do we need also fix this?
https://github.com/GreengageDB/greengage/blob/284ad82d08b8a40fc98ad6442052aa24ca1e8832/src/bin/pg_upgrade/greengage/check_gp.c#L575-L586

hilltracer: Please manually test your fix and show result, how it worked before and after your patch (with flag --continue-check-on-fatal)

hilltracer (commented):
Please update in patch description info about tests e.g. like it was done in 6.x PR

> Don't add tests because they belong to the ggupgrade repo
and will be added there after this patch.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Yes. But I reverted changes to functions not added in this patch, as per internal discussion.

GirgerS: For new functions it seems to work. Before:
```
Checking for views with removed relations                   fatal
| Your installation contains views using removed relations.
| These relations are no longer present on the target version.
| These views must be updated to use relations supported in the
| target version or removed before upgrade can continue. A list
| of the problem views is in the file:
	/home/regrig/greengage/7.x-data/data/gpseg-1/pg_upgrade_output.d/20260709T135423.445/view_referencing_removed_relations.txt

ok
```

After:

```
Checking for views with removed relations                   fatal
| Your installation contains views using removed relations.
| These relations are no longer present on the target version.
| These views must be updated to use relations supported in the
| target version or removed before upgrade can continue. A list
| of the problem views is in the file:
	/home/regrig/greengage/7.x-data/data/gpseg-1/pg_upgrade_output.d/20260709T142132.834/view_referencing_removed_relations.txt
```

hilltracer (dismissed):

bimboterminator1 (dismissed):

bimboterminator1 (approved):

hilltracer (approved):

PR #517, opened 2026-06-29
GG-571: Implement 'rows out' in EXPLAIN ANALYZE

bandetto:
Implement 'rows out' in EXPLAIN ANALYZE

Add the gp_enable_explain_rows_out GUC, off by default. When it is on,
EXPLAIN ANALYZE reports the spread of rows produced across segments for
each plan node. The figures make uneven data distribution visible per
node.

Changes:

1. Add the gp_enable_explain_rows_out boolean GUC in the experimental
   feature group, exposed as USERSET and hidden from SHOW ALL.
2. Emit a "Rows out" line in cdbexplain_showExecStats when the GUC is
   on, reporting the average rows across workers with the maximum and
   minimum rows and the segments that produced them, in both text and
   structured output formats.
3. Compute the figures as per-loop rows (ntuples/nloops) to match the
   row count already shown on the node.
4. Skip non-participating (T_Invalid) segment slots, taking the
   average, the worker count, and the minimum only over the segments
   that ran the node.
5. Cover the feature in the gp_explain regression test on a uniform and
   a skewed table, and assert the line is emitted.

EXPLAIN ANALYZE now reports per-node row-count skew across segments.

Original commit is taken from open-gpdb/gpdb@1d4123048fbf5482c636eb193876d517edaaae81.

Co-authored-by: Vladimir Sarmin \<v.sarmin@arenadata.io\>

---

This commit will be rebased onto the target branch to preserve authorship.

### DIFF
--- a/src/test/regress/expected/gp_explain.out
+++ b/src/test/regress/expected/gp_explain.out
@@ -29,13 +29,26 @@ begin
   end loop;
 end;
 $$ language plpgsql;
+create or replace function get_explain_analyze_xml_output(explain_query text)
+returns xml as
+$$
+declare
+  x xml;
+begin
+  execute 'EXPLAIN (ANALYZE, VERBOSE, FORMAT XML) ' || explain_query
+  into x;
+  return x;
+end;
+$$ language plpgsql;
 --
 -- Test explain_memory_verbosity option
 --
+set gp_use_legacy_hashops=off;
 CREATE TABLE explaintest (id int4);
 NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'id' as the Greengage Database data distribution key for this table.
 HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
 INSERT INTO explaintest SELECT generate_series(1, 10);
+reset gp_use_legacy_hashops;
 ANALYZE explaintest;
 EXPLAIN ANALYZE SELECT * FROM explaintest;
                                                                QUERY PLAN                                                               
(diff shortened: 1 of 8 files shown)


whitehawk:
It looks that the commit hash of the original commit in the description is wrong.

src/test/regress/sql/gp_explain.sql
```diff
@@ -203,9 +203,27 @@ explain (slicetable, costs off) SELECT * FROM explaintest;
...
+-- Test explain rows out in JSON format, exercising the structured branch.
+explain (analyze, format json) SELECT * FROM explaintest;
+set gp_enable_explain_rows_out=DEFAULT;
+
+-- Rows out on a skewed distribution, so max and min land on different segments.
+CREATE TABLE explain_rows_skew (id int) DISTRIBUTED BY (id);
+INSERT INTO explain_rows_skew SELECT 1 FROM generate_series(1, 100);
+INSERT INTO explain_rows_skew SELECT g FROM generate_series(2, 9) g;
+ANALYZE explain_rows_skew;
+set gp_enable_explain_rows_out=on;
```
whitehawk: If we add new test cases comparing to the original commit, I think we can just remove `set gp_enable_explain_rows_out=DEFAULT;` above and remove `set gp_enable_explain_rows_out=on;` here, as it is already set at the beginning of the test.

src/test/regress/sql/gp_explain.sql
```diff
@@ -203,9 +203,27 @@ explain (slicetable, costs off) SELECT * FROM explaintest;
...
+explain analyze SELECT * FROM explain_rows_skew;
+set gp_enable_explain_rows_out=DEFAULT;
```
whitehawk:
```suggestion
reset gp_enable_explain_rows_out;
```
?

src/test/regress/expected/gp_explain.out
```diff
@@ -447,6 +447,118 @@ explain (slicetable, costs off, format json) SELECT * FROM explaintest;
...
 (1 row)
 
+-- Test explain rows out.
+set gp_enable_explain_rows_out=on;
+explain analyze SELECT * FROM explaintest;
+                                                          QUERY PLAN                                                           
+-------------------------------------------------------------------------------------------------------------------------------
+ Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..3.10 rows=10 width=4) (actual time=0.266..0.384 rows=5 loops=2)
+   ->  Seq Scan on explaintest  (cost=0.00..3.10 rows=4 width=4) (actual time=0.011..0.013 rows=2 loops=2)
+         Rows out: 3.33 rows avg x 3 workers, 5 rows max (seg1), 2 rows min (seg2).
```
whitehawk: From the documentation, `rows=2` means:
> The maximum number of rows returned by the segment that produced the most rows for the operation. If multiple segments produce an equal number of rows, EXPLAIN ANALYZE shows the segment with the longest <time> to end.

But `Rows out` shows `5 rows max (seg1)`.

If I understand the documentation correctly, these 2 numbers should match.

whitehawk: Hm, I've added `SELECT *, gp_segment_id FROM explaintest order by gp_segment_id;` into the test and it says that minimum amount of rows is 1:
```
SELECT *, gp_segment_id FROM explaintest order by gp_segment_id;
 id | gp_segment_id 
----+---------------
  2 |             0
  3 |             0
  4 |             0
  7 |             0
  8 |             0
  1 |             1
  5 |             2
  6 |             2
  9 |             2
 10 |             2
(10 rows)
```
while `Rows out:` shows min is 2.

bandetto: Fixed.

bandetto: Fixed.

bimboterminator1 (dismissed):

whitehawk:
LGTM, but please add the PR ID to the commit summary.

whitehawk (dismissed):

whitehawk (approved):

bimboterminator1 (approved):

PR #518, opened 2026-06-30
GG-531: Fixup pg_dump array type name creation

VoidZeroNull0:
What happened?
Dumping 6.x with forceful array type creation option leads to type names
collision and otherwise impossible type names in 7.x. When trying to restore
from dump, they lead to errors.

Why it happens?
Array type name creation logic inside pg_dump.c is much simpler and not
consistent with the one in pg_type.c.  

How do we fix this?
Let's resolve name collisions like we do in makeArrayTypeName() from pg_type.c,
but without using backend specific code. Also, we need to store names assigned
by the dump.

Performance considerations?
Running a query each time to check collision seems like a major slowdown. HTAB
could be a nice way to speed up things, but it cannot be used in frontend code
because of missing memory contexts code. Comparing current type name against
array of strings with all type names is kind of slow. So, making an assumption
that collisions are pretty rare, running the query 1 or 2 times does not seem
that bad.

Tests?
No auto tests as this is the one of many to fix them. Testing could be done by
manually dumping and looking and resulting type names.

Ticket: GG-531

### DIFF
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -12709,7 +12709,8 @@ truncate_array_type_name(char *typeName, int encoding, int limit)
 	int 	byte_len = 0;
 	int		char_len = 0;
 
-	while(byte_len < limit) {
+	while(byte_len < limit)
+	{
 		char_len = PQmblen(typeName + byte_len, encoding);
 		/* Non-valid encoding or broken string */
 		Assert(char_len > 0);
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -96,13 +96,13 @@ typedef struct
 	int			objsubid;		/* subobject (table column #) */
 } SecLabelItem;
 
-typedef struct TypeNamesCache {
+typedef struct TypeNamesList {
 	Oid 	schema_oid;
 	char  **names;
 	int 	count;
 	int 	capacity;
-	struct 	TypeNamesCache *next;
-} TypeNamesCache;
+	struct 	TypeNamesList *next;
+} TypeNamesList;
 
 typedef enum OidOptions
 {
@@ -12715,15 +12715,15 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
  * 
  * Analogous to makeArrayTypeName() from pg_type.c, but does not use
  * backend functions.
- * Also manages its own cache of assigned names, which lives till
+ * Also manages its own list of assigned names, which lives till
  * the end of pg_dump.
  */
 static char *
 make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 {
-	static TypeNamesCache *type_names_cache = NULL;
-	TypeNamesCache *cache = NULL;
-	TypeNamesCache *prev = NULL;
+	static TypeNamesList *type_names_list = NULL;
+	TypeNamesList *list = NULL;
+	TypeNamesList *prev = NULL;
 	char 		   *arr;
 	int 			namelen = strlen(typeName);
 	int 			i;
@@ -12737,14 +12737,14 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 	if (!arr)
 		fatal("out of memory");
 
-	/* Search for schema's cache */
-	cache = type_names_cache;
-	while (cache)
+	/* Search for schema's list */
+	list = type_names_list;
+	while (list)
 	{
-		if (cache->schema_oid == typeNamespace)
+		if (list->schema_oid == typeNamespace)
 			break;
-		prev = cache;
-		cache = cache->next;
+		prev = list;
+		list = list->next;
 	}
 
 	/* Try to find a unique name */
@@ -12772,9 +12772,9 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 
 		/* Check if dump already assigned such name for this schema */
 		is_assigned = false;
-		for (j = 0; cache && j < cache->count; j++)
+		for (j = 0; list && j < list->count; j++)
 		{
-			if (strcmp(cache->names[j], arr) == 0)
+			if (strcmp(list->names[j], arr) == 0)
 			{
 				is_assigned = true;
 				break;
@@ -12789,37 +12789,37 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 	if (i >= NAMEDATALEN - 1)
 		fatal("could not form array type name for type \"%s\"", typeName);
 
-	/* If no cache found for this schema, create one */
-	if (!cache)
+	/* If no list found for this schema, create one */
+	if (!list)
 	{
-		cache = (TypeNamesCache *) pg_malloc0(sizeof(TypeNamesCache));
-		if (!cache)
+		list = (TypeNamesList *) pg_malloc0(sizeof(TypeNamesList));
+		if (!list)
 			fatal("out of memory");
-		cache->schema_oid = typeNamespace;
-		cache->capacity = 16;
-		cache->count = 0;
-		cache->names = (char **) pg_malloc(cache->capacity * sizeof(char *));
-		if (!cache->names)
+		list->schema_oid = typeNamespace;
+		list->capacity = 16;
+		list->count = 0;
+		list->names = (char **) pg_malloc(list->capacity * sizeof(char *));
+		if (!list->names)
 			fatal("out of memory");
-		cache->next = NULL;
+		list->next = NULL;
 
 		/* Add to linked list */
 		if (prev)
-			prev->next = cache;
+			prev->next = list;
 		else
-			type_names_cache = cache;
+			type_names_list = list;
 	}
 
-	if (cache->count == cache->capacity)
+	if (list->count == list->capacity)
 	{
-		cache->capacity = cache->capacity * 2;
-		cache->names = (char **) pg_realloc(cache->names,
-											cache->capacity * sizeof(char *));
-		if (!cache->names)
+		list->capacity = list->capacity * 2;
+		list->names = (char **) pg_realloc(list->names,
+											list->capacity * sizeof(char *));
+		if (!list->names)
 			fatal("out of memory");
 	}
-	cache->names[cache->count] = arr;
-	cache->count++;
+	list->names[list->count] = arr;
+	list->count++;
 
 	return arr;
 }
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -301,7 +301,7 @@ static char *format_function_arguments(const FuncInfo *finfo, const char *funcar
 									   bool is_agg);
 static char *format_function_signature(Archive *fout,
 									   const FuncInfo *finfo, bool honor_quotes);
-static void truncate_array_type_name(char *typeName, int encoding);
+static void truncate_array_type_name(char *typeName, int encoding, int limit);
 static char *make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout);
 static char *convertRegProcReference(const char *proc);
 static char *getFormattedOperatorName(const char *oproid);
@@ -12711,16 +12711,16 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
 }
 
 static void
-truncate_array_type_name(char *typeName, int encoding)
+truncate_array_type_name(char *typeName, int encoding, int limit)
 {
 	int 	byte_len = 0;
 	int		char_len = 0;
 
-	while(byte_len < NAMEDATALEN - 1) {
+	while(byte_len < limit) {
 		char_len = PQmblen(typeName + byte_len, encoding);
 		/* Non-valid encoding or broken string */
 		Assert(char_len > 0);
-		if (byte_len + char_len > NAMEDATALEN - 1)
+		if (byte_len + char_len > limit)
 			break;
 		byte_len += char_len;
 	}
@@ -12744,16 +12744,13 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 	TypeNamesList  *prev = NULL;
 	char		   *arr;
 	int				namelen = strlen(typeName);
-	int				i;
-	int				j;
+	int				i, j;
 	PGresult	   *res;
 	PQExpBuffer		query;
 	bool			is_dup;
 	bool			is_assigned;
 
 	arr = (char *) pg_malloc(NAMEDATALEN);
-	if (!arr)
-		fatal("out of memory");
 
 	/* Search for schema's list */
 	list = type_names_list;
@@ -12781,7 +12778,7 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 			 * inside pg_upgrade (with --binary-upgrade), which uses default
 			 * database encoding.
 			 */
-			truncate_array_type_name(arr, fout->encoding);
+			truncate_array_type_name(arr, fout->encoding, NAMEDATALEN - 1);
 		}
 
 		/* Check existence in pg_type */
@@ -12821,14 +12818,10 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 	if (!list)
 	{
 		list = (TypeNamesList *) pg_malloc0(sizeof(TypeNamesList));
-		if (!list)
-			fatal("out of memory");
 		list->schema_oid = typeNamespace;
 		list->capacity = 16;
 		list->count = 0;
 		list->names = (char **) pg_malloc(list->capacity * sizeof(char *));
-		if (!list->names)
-			fatal("out of memory");
 		list->next = NULL;
 
 		/* Add to linked list */
@@ -12843,8 +12836,6 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 		list->capacity = list->capacity * 2;
 		list->names = (char **) pg_realloc(list->names,
 											list->capacity * sizeof(char *));
-		if (!list->names)
-			fatal("out of memory");
 	}
 	list->names[list->count] = arr;
 	list->count++;
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -12769,7 +12769,8 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 		arr[i - 1] = '_';
 		if (i + namelen < NAMEDATALEN)
 			strcpy(arr + i, typeName);
-		else {
+		else 
+		{
 			memcpy(arr + i, typeName, NAMEDATALEN - i);
 			/*
 			 * Be aware, that if archive's encoding is passed with --encoding
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -96,14 +96,6 @@ typedef struct
 	int			objsubid;		/* subobject (table column #) */
 } SecLabelItem;
 
-typedef struct TypeNamesList {
-	Oid 	schema_oid;
-	char  **names;
-	int 	count;
-	int 	capacity;
-	struct 	TypeNamesList *next;
-} TypeNamesList;
-
 typedef enum OidOptions
 {
 	zeroAsOpaque = 1,
@@ -4942,6 +4934,7 @@ binary_upgrade_set_type_oids_by_type_oid(Archive *fout,
 						  "'%u'::pg_catalog.oid, $_GPDB_$%s$_GPDB_$::text);\n\n",
 						  pg_type_array_oid, pg_type_array_ns_oid,
 						  pg_type_array_name);
+		pg_free(pg_type_array_name);
 	}
 
 	destroyPQExpBuffer(upgrade_query);
@@ -12734,32 +12727,34 @@ truncate_array_type_name(char *typeName, int encoding, int limit)
  * Analogous to makeArrayTypeName() from pg_type.c, but does not use
  * backend functions.
  * Also manages its own list of assigned names, which lives till
- * the end of pg_dump.
+ * the end of current schema dump.
  */
 static char *
 make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 {
-	static TypeNamesList *type_names_list = NULL;
-	TypeNamesList  *list = NULL;
-	TypeNamesList  *prev = NULL;
-	char		   *arr;
-	int				namelen = strlen(typeName);
-	int				i, j;
-	PGresult	   *res;
-	PQExpBuffer		query;
-	bool			is_dup;
-	bool			is_assigned;
+	static SimpleStringList	   *type_names = NULL;
+	static Oid 					names_schema_oid = InvalidOid;
+	SimpleStringListCell	   *next;
+	SimpleStringListCell	   *cell;
+	char					   *arr;
+	int							namelen = strlen(typeName);
+	int							i;
+	PGresult				   *res;
+	PQExpBuffer					query;
+	bool						is_dup, is_assigned;
 
 	arr = (char *) pg_malloc(NAMEDATALEN);
 
-	/* Search for schema's list */
-	list = type_names_list;
-	while (list)
+	/* We have encountered new schema, have to clear the old list */
+	if (type_names && names_schema_oid != typeNamespace)
 	{
-		if (list->schema_oid == typeNamespace)
-			break;
-		prev = list;
-		list = list->next;
+		names_schema_oid = typeNamespace;
+		for(cell = type_names->head; cell; cell = next)
+		{
+			next = cell->next;
+			pg_free(cell);
+		}
+		type_names = NULL;
 	}
 
 	/* Try to find a unique name */
@@ -12797,13 +12792,14 @@ make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
 
 		/* Check if dump already assigned such name for this schema */
 		is_assigned = false;
-		for (j = 0; list && j < list->count; j++)
+		if (type_names)
 		{
-			if (strcmp(list->names[j], arr) == 0)
-			{
-				is_assigned = true;
-				break;
-			}
+			for (cell = type_names->head; cell; cell = cell->next)
+				if (strcmp(cell->val, arr) == 0)
+				{
+					is_assigned = true;
+					break;
+				}
 		}
 		if (is_assigned)
 			continue;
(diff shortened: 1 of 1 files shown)


src/bin/pg_dump/pg_dump.c
```diff
@@ -12698,6 +12701,57 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
...
+    for (i = 1; i < NAMEDATALEN - 1; i++)
+    {
+        arr[i - 1] = '_';
+        if (i + namelen < NAMEDATALEN)
+            strcpy(arr + i, typeName);
+        else
+        {
+            memcpy(arr + i, typeName, NAMEDATALEN - i - 1);
+            arr[NAMEDATALEN - 1] = '\0';
+        }
```
red1452: I think, that you may change the all "if" block by `strlcpy`.

src/bin/pg_dump/pg_dump.c
```diff
@@ -4917,7 +4918,9 @@ binary_upgrade_set_type_oids_by_type_oid(Archive *fout,
 
 		pg_type_array_oid = next_possible_free_oid;
 		pg_type_array_ns_oid = tyinfo->dobj.namespace->dobj.catId.oid;
-		pg_type_array_name = psprintf("_%s", tyinfo->dobj.name);
+		pg_type_array_name = make_array_type_name(tyinfo->dobj.name,
```
red1452, line 4931: Should we free `pg_type_array_name` after using?

VoidZeroNull0: I guess we should, but in 5fbe862 we still need that memory to detect collisions till the end of dumping:
https://github.com/GreengageDB/greengage/blob/379b2a608ef27eb1cd9df57c4a65a3f9a222d0f2/src/bin/pg_dump/pg_dump.c#L12821-L12824

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: Good idea, thanks!

GirgerS: I believe that not using [truncate_identifier](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/catalog/pg_type.c#L914) can cause some errors too. It handles multibyte encodings, because it is not enough to just null-terminate them. For utf-8:

```sql
postgres=# CREATE DATABASE unicode_test WITH ENCODING 'UTF8';
CREATE DATABASE
postgres=# \q
regrig@fourier:~/greengage/6.x$ psql unicode_test;
psql (9.4.26)
Type "help" for help.

unicode_test=# CREATE table test (a int)
PARTITION BY RANGE (a) (
        PARTITION xxxxxxxxxxxxxxxxx_string_with_russian_multibite_char_Ъ
        START (10) END (11)
);
NOTICE:  CREATE TABLE will create partition "t_1_prt_xxxxxxxxxxxxxxxxx_string_with_russian_multibite_char_Ъ" for table "test"
CREATE TABLE
unicode_test=# DROP TABLE test;
DROP TABLE
unicode_test=# CREATE table test (a int)
PARTITION BY RANGE (a) (
        PARTITION xxxxxxxxxxxxxxx_string_with_сyrillic_multibite_char_Д
        START (10) END (11)
);
NOTICE:  CREATE TABLE will create partition "t_1_prt_xxxxxxxxxxxxxxx_string_with_сyrillic_multibite_char_Д" for table "test"
CREATE TABLE
unicode_test=# \q
regrig@fourier:~/greengage/6.x$ ../7.x/bin/pg_dump postgres --binary-upgrade > dump.sql
regrig@fourier:~/greengage/6.x$ ../7.x/bin/pg_dump unicode_test --binary-upgrade > dump.sql
pg_dump: error: query failed: ERROR:  invalid byte sequence for encoding "UTF8": 0xd0 0x27
pg_dump: error: query was: SELECT EXISTS(SELECT 1 FROM pg_catalog.pg_type WHERE typname = '_t_1_prt_xxxxxxxxxxxxxxx_string_with_сyrillic_multibite_char_� AND typnamespace = '2200'::pg_catalog.oid);
```


Meanwhile in 7.x cluster:
```sql
unicode_test=# CREATE table test (a int)
PARTITION BY RANGE (a) (
        PARTITION xxxxxxxxxxxxxxx_string_with_сyrillic_multibite_char_Д
        START (10) END (11)
);
CREATE TABLE
unicode_test=# SELECT typname FROM pg_type WHERE typname LIKE '%string_with_сyrillic_multibite_char%';
-[ RECORD 1 ]----------------------------------------------------------
typname | t_1_prt_xxxxxxxxxxxxxxx_string_with_сyrillic_multibite_char_Д
-[ RECORD 2 ]----------------------------------------------------------
typname | _t_1_prt_xxxxxxxxxxxxxxx_string_with_сyrillic_multibite_char_
```
Also, the last time I checked this, functions for handling multibyte string weren't linking with `pg_upgrade`...

GirgerS:
There might be another (most likely unrelated) issue with selecting names from partitioned tables:

```
CREATE table test (a int)
PARTITION BY RANGE (a) (
        PARTITION xxxxxxxxxxxxxxx_name_with_exactly_63_characters_xxxxxxxxxxxxxxx
        START (10) END (11)
);


FATAL:  Unexpected internal error (indexcmds.c:2036)
DETAIL:  FailedAssertion("!(availchars > 0)", File: "indexcmds.c", Line: 2036)
HINT:  Process 294039 will wait for gp_debug_linger=120 seconds before termination.
Note that its locks and other resources will not be released until then.
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Succeeded.
```

If it indeed the case, we might want to create a ticket for it.

GirgerS:
And also, we can potentially hit an edge case when we can't create array types for existing tables in 7.x,
even though in 6.x everything worked fine.

What do we think about it?

```
-- 6.x
postgres=# CREATE TABLE _____________________________________________________________ (a int);
postgres=# CREATE table edge_case (a int)
PARTITION BY RANGE (a) (
        PARTITION d START (10) END (11) WITH (tablename='______________________________________________________________')
postgres=# \q
regrig@fourier:~/greengage/6.x$ ../7.x/bin/pg_dump postgres --binary-upgrade > dump.sql
pg_dump: error: could not form array type name for type "______________________________________________________________"
);
```

```
-- 7.x
postgres=# CREATE TABLE _____________________________________________________________ (a int);
CREATE table edge_case (a int)
PARTITION BY RANGE (a) (
        PARTITION d START (10) END (11) WITH (tablename='______________________________________________________________')
);
CREATE TABLE
ERROR:  could not form array type name for type "______________________________________________________________"
);
```

VoidZeroNull0:
> even though in 6.x everything worked fine

Guess we cannot do much for 6.x in that case inside `pg_dump`. But, we can prohibit such names inside backend code of 6.x.

VoidZeroNull0:
Probably, it is even better idea to add a check to `gpcheckcat`, `ggcheckmigrate` or other simmilar utilities to notify user about the problem. Then, they can manually change the name of the table if they need to upgrade.

GirgerS:
> Probably, it is even better idea to add a check to gpcheckcat, ggcheckmigrate or other simmilar utilities to notify user about the problem. Then, they can manually change the name of the table if they need to upgrade.

One solutions that comes to mind is to create `binary_upgrade_set_next_array_type_name` support functon or something similar. Of course, if these is no other logic that assumes that we create names the way we do. 

But it's hard for me to tell if this case can appear during regular database usage, not when trying to intentionally trigger it. It might be that a check would more that enough.

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: > Also, the last time I checked this, functions for handling multibyte string weren't linking with pg_upgrade...

I found out, that `PQmblen()` seems to fix the issue and correctly links to `pg_dump` at least.

VoidZeroNull0:
Could not reproduce on neither of versions. But got something similar on 6.x:
```SQL
psql (9.4.26)
Type "help" for help.

postgres=# CREATE table test (a int)
PARTITION BY RANGE (a) (
        PARTITION xxxxxxxxxxxxxxx_name_with_exactly_63_characters_xxxxxxxxxxxxxxx
        START (10) END (11)
);
NOTICE:  CREATE TABLE will create partition "__prt_xxxxxxxxxxxxxxx_name_with_exactly_63_characters_xxxxxxxxxxx" for table "test"
ERROR:  relation "public.__prt_xxxxxxxxxxxxxxx_name_with_exactly_63_characters_xxxxxxxxxxx" does not exist
```
Judging by the code assert (`DETAIL:  FailedAssertion("!(availchars > 0)", File: "indexcmds.c", Line: 2036)`), it seems to be related to incorrect naming of the object. Could it be, that the partition name was truncated wrongly?
https://github.com/GreengageDB/greengage/blob/1d4adb9ecd91c68257071f5e3adb1a9272f387e5/src/backend/commands/indexcmds.c#L2669
Anyway, seems to do nothing with `pg_dump`, but worth research in its own ticket if the bug is real.

GirgerS:
LGTM, with two minor things. First off all, we don't take into account that archive encoding can be overwritten:

```
../7.x/bin/pg_dump postgres --binary-upgrade --encoding=SQL_ASCII > out.sql
pg_dump: error: query failed: ERROR:  invalid byte sequence for encoding "UTF8": 0xd0 0x27
pg_dump: error: query was: SELECT EXISTS(SELECT 1 FROM pg_catalog.pg_type WHERE typname = '_t_1_prt_xxxxxxxxxxxxxxxxx_string_with_russian_multibite_char_� AND typnamespace = '2200'::pg_catalog.oid);
```

But, it [seems](https://github.com/GreengageDB/greengage/blob/7.x/src/bin/pg_upgrade/dump.c#L58-L65) that pg_upgrade always uses the default one. And `--binary-upgrade` flag is documented as `for internal usage`, so it should work. But maybe we should add a comment explaining this.

And also I got a strange error for the dump in the attachment when testing reallocations. Can you confirm if it the case? It is not related to array type name creation though.

[schema.sql](https://github.com/user-attachments/files/29792735/schema.sql)

VoidZeroNull0:
> And also I got a strange error for the dump

I have tried dumping it, and migrating using `pg_upgrade_run_6X_to_7X_migration.sh` but got nothing, except for diffs:
```diff
 -- For binary upgrade, set heap's relfrozenxid and relminmxid
 UPDATE pg_catalog.pg_class
-SET relfrozenxid = '733', relminmxid = '1'
+SET relfrozenxid = '1513', relminmxid = '1'
 WHERE oid = 's3.t'::pg_catalog.regclass;
 RESET allow_system_table_mods;
 ```
 Could you please, attach the error itself?

**Comment by GirgerS** [comment]
> I have tried dumping it, and migrating using pg_upgrade_run_6X_to_7X_migration.sh but got nothing, except for diffs:

Interesting, I'll need to think why it can be the case... Because I reliably get this error (with Greengage 6 from 6.x branch and Greengage 7 from either 7.x or this branch), when running `pg_upgrade_run_6X_to_7X_migration.sh`. Anyway, it is a seperate issue, I think we are good.

> Could you please, attach the error itself?

[error.txt](https://github.com/user-attachments/files/29804074/error.txt)

P.S.: does `configure` config in Greengage 7 have: `--enable-cassert --enable-debug`? Without these options, upgrade works for me too.

File: src/bin/pg_dump/pg_dump.c [hunk 3]
```diff
@@ -12698,6 +12710,147 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
...
+
+	if (list->count == list->capacity)
+	{
+		list->capacity = list->capacity * 2;
+		list->names = (char **) pg_realloc(list->names,
+											list->capacity * sizeof(char *));
+		if (!list->names)
+			fatal("out of memory");
+	}
+	list->names[list->count] = arr;
```
red1452, line 12843: Where is place of freeing this memory?

src/bin/pg_dump/pg_dump.c
```diff
@@ -12698,6 +12710,147 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
...
+			type_names_list = list;
+	}
+
+	if (list->count == list->capacity)
+	{
+		list->capacity = list->capacity * 2;
+		list->names = (char **) pg_realloc(list->names,
+											list->capacity * sizeof(char *));
+		if (!list->names)
+			fatal("out of memory");
```
red1452: You may remove this check, because `pg_realloc` makes it inside and execs `exit(EXIT_FAILURE);` in case of error.

src/bin/pg_dump/pg_dump.c
```diff
@@ -12698,6 +12710,147 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
...
+	{
+		list = (TypeNamesList *) pg_malloc0(sizeof(TypeNamesList));
+		if (!list)
+			fatal("out of memory");
+		list->schema_oid = typeNamespace;
+		list->capacity = 16;
+		list->count = 0;
+		list->names = (char **) pg_malloc(list->capacity * sizeof(char *));
+		if (!list->names)
+			fatal("out of memory");
```
red1452: The same as `pg_realloc`.

src/bin/pg_dump/pg_dump.c
```diff
@@ -12698,6 +12710,147 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
...
+static char *
+make_array_type_name(const char *typeName, Oid typeNamespace, Archive *fout)
+{
+	static TypeNamesList *type_names_list = NULL;
+	TypeNamesList  *list = NULL;
+	TypeNamesList  *prev = NULL;
+	char		   *arr;
+	int				namelen = strlen(typeName);
+	int				i;
+	int				j;
```
red1452: What about `int i,j;`?

src/bin/pg_dump/pg_dump.c
```diff
@@ -12698,6 +12710,147 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
 	return fn.data;
 }
 
+static void
+truncate_array_type_name(char *typeName, int encoding)
+{
+	int 	byte_len = 0;
+	int		char_len = 0;
+
+	while(byte_len < NAMEDATALEN - 1) {
```
red1452: Use correct code style: open bracket at the next line.

src/bin/pg_dump/pg_dump.c
```diff
@@ -12698,6 +12710,147 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
...
+		char_len = PQmblen(typeName + byte_len, encoding);
```
red1452, line 12722: How does `PQmblen` correctly calculate length for string with maximum length? Such string here is without "NULL" symbol in the allocated memory.

src/bin/pg_dump/pg_dump.c
```diff
@@ -12698,6 +12710,147 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
...
+		if (list->schema_oid == typeNamespace)
+			break;
+		prev = list;
+		list = list->next;
+	}
+
+	/* Try to find a unique name */
+	for (i = 1; i < NAMEDATALEN - 1; i++)
+	{
+		arr[i - 1] = '_';
```
red1452, line 12770: Is it correct to add one-byte symbol to strings, which are two (or more) bytes encoding?

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: Thanks! Fixed.

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: Fixed.

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: Done!

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: Done!

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: `PQmblen()` calculates only the length of a character (it can be multi-byte) on address `(typeName + byte_len)`. By moving through byte array and adding length of another char, we are determining the position where null-terminator should be to not accidentally cut in half a multi-byte string.

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: It is not correct for two (or more) bytes encoding, but:
1. To my knowledge, [all](https://github.com/GreengageDB/greengage/blob/1d4adb9ecd91c68257071f5e3adb1a9272f387e5/src/backend/utils/mb/wchar.c#L1771-L1820) present in Postgres encodings are ASCII-compatible, so underscore in all of them takes exactly 1 byte.
2. We are trying to replicate how `makeArrayTypeName()` from `pg_type.c` behaves and it adds just plain underscores:
https://github.com/GreengageDB/greengage/blob/1d4adb9ecd91c68257071f5e3adb1a9272f387e5/src/backend/catalog/pg_type.c#L937-L946

VoidZeroNull0:
> P.S.: does configure config in Greengage 7 have: --enable-cassert --enable-debug? Without these options, upgrade works for me too.

Yup, now it does not work.

> Interesting, I'll need to think why it can be the case...

I have tried to change dump you have attached a little bit and managed to reduce it for reproduction, but have not found why it happens:
```SQL
CREATE SCHEMA s0;
CREATE SCHEMA s1;
CREATE SCHEMA s2;
CREATE SCHEMA s3;

CREATE TABLE s0.t (a int);
CREATE TABLE s1.t (a int);
CREATE TABLE s2.t (a int);
CREATE TABLE s3.t (a int);
CREATE TABLE s0._t (a int);
CREATE TABLE s1._t (a int);
CREATE TABLE s2._t (a int);
CREATE TABLE s3._t (a int);
```

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: Sorry, it is indeed needed. I have misunderstood how pg_dump dumps schemas - we don't need to keep all the names for all the schemas at a time. We now free it [here](https://github.com/GreengageDB/greengage/blob/7d810cd6d039046958d376c48d1b6d5fb8609e8f/src/bin/pg_dump/pg_dump.c#L4937).

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: Now we free it all here:
https://github.com/GreengageDB/greengage/blob/7d810cd6d039046958d376c48d1b6d5fb8609e8f/src/bin/pg_dump/pg_dump.c#L12748-L12758

VoidZeroNull0:
In [7d810cd](https://github.com/GreengageDB/greengage/pull/518/commits/7d810cd6d039046958d376c48d1b6d5fb8609e8f) was a rewrite of how assigned names are remembered. Turns out, there is no need to store all names for all schemas, because schemas are processed sequentially. So, we can free when we meet new schema.

red1452 (commented):
I get example from the ticket, run it:
```
evgeniy@evgeniy-pc:~/gpdb/greengage7$ createdb test
evgeniy@evgeniy-pc:~/gpdb/greengage7$ psql test
psql (12.22)
Type "help" for help.

test=# create table _sales_1_prt_sales_2_prt_cr1_3_prt_ca (a int);
CREATE TABLE
test=# CREATE TABLE sales (id int, year int, qtr int, c_rank int, code char(1), region text)
DISTRIBUTED BY (id)
PARTITION BY LIST (code)
  SUBPARTITION BY RANGE (c_rank)
    SUBPARTITION by LIST (region)

( PARTITION sales VALUES ('S')
   ( SUBPARTITION cr1 START (1) END (2)
      ( SUBPARTITION ca VALUES ('CA') ),
      SUBPARTITION cr2 START (3) END (4)
        ( SUBPARTITION ca VALUES ('CA') ) ),

 PARTITION returns VALUES ('R')
   ( SUBPARTITION cr1 START (1) END (2)
      ( SUBPARTITION ca VALUES ('CA') ),
     SUBPARTITION cr2 START (3) END (4)
        ( SUBPARTITION ca VALUES ('CA') ) )
);
CREATE TABLE

test=# select typname from pg_type where typname like '%sales%' ORDER BY 1;
                 typname                 
-----------------------------------------
 ___sales_1_prt_sales_2_prt_cr1_3_prt_ca
 __sales_1_prt_sales_2_prt_cr1_3_prt_ca
 _sales
 _sales_1_prt_returns
 _sales_1_prt_returns_2_prt_cr1
 _sales_1_prt_returns_2_prt_cr1_3_prt_ca
 _sales_1_prt_returns_2_prt_cr2
 _sales_1_prt_returns_2_prt_cr2_3_prt_ca
 _sales_1_prt_sales
 _sales_1_prt_sales_2_prt_cr1
 _sales_1_prt_sales_2_prt_cr1_3_prt_ca
 _sales_1_prt_sales_2_prt_cr2
 _sales_1_prt_sales_2_prt_cr2_3_prt_ca
 sales
 sales_1_prt_returns
 sales_1_prt_returns_2_prt_cr1
 sales_1_prt_returns_2_prt_cr1_3_prt_ca
 sales_1_prt_returns_2_prt_cr2
 sales_1_prt_returns_2_prt_cr2_3_prt_ca
 sales_1_prt_sales
 sales_1_prt_sales_2_prt_cr1
 sales_1_prt_sales_2_prt_cr1_3_prt_ca
 sales_1_prt_sales_2_prt_cr2
 sales_1_prt_sales_2_prt_cr2_3_prt_ca
(24 rows)

test=# \q
```

There is type with name ` ___sales_1_prt_sales_2_prt_cr1_3_prt_ca`.
Than I do:
```
evgeniy@evgeniy-pc:~/gpdb/greengage7$ pg_dump test | grep "___sales_1_prt_sales_2_prt_cr1_3_prt_ca"
evgeniy@evgeniy-pc:~/gpdb/greengage7$
```
There is not command for creating any object with such name. Is it correct?

VoidZeroNull0:
> evgeniy@evgeniy-pc:~/gpdb/greengage7$ pg_dump test | grep "___sales_1_prt_sales_2_prt_cr1_3_prt_ca"

To see what is added by patch, you have to run something like `pg_dump --binary-upgrade -d test > dump.sql` against the running 6.x cluster.  `pg_dump` has to be from patch's branch. The option `--binary-upgrade` is needed as else `pg_dump` won't dump (or, in our case) create array types.

> There is not command for creating any object with such name. Is it correct?

If you mean for the type itself, here are the lines inside the dump to create it when restoring (before the patch there weren't for this type):
```SQL
-- For binary upgrade, must preserve pg_type array oid
SELECT pg_catalog.binary_upgrade_set_next_array_pg_type_oid('16401'::pg_catalog.oid, '2200'::pg_catalog.oid, $_GPDB_$___sales_1_prt_sales_2_prt_cr1_3_prt_ca$_GPDB_$::text);
```
We expect that the same dumped schema both on 6.x and 7.x will have `pg_catalog.binary_upgrade_set_next_array_pg_type_oid()` with the same exact type names.

src/bin/pg_dump/pg_dump.c
red1452: still actual...

src/bin/pg_dump/pg_dump.c
```diff
@@ -12698,6 +12703,132 @@ format_function_signature(Archive *fout, const FuncInfo *finfo, bool honor_quote
...
+			pg_free(cell);
+		}
+		type_names = NULL;
+	}
+
+	/* Try to find a unique name */
+	for (i = 1; i < NAMEDATALEN - 1; i++)
+	{
+		arr[i - 1] = '_';
+		if (i + namelen < NAMEDATALEN)
```
red1452, line 12771: `NAMEDATALEN - 1` ? for `\0`?

VoidZeroNull0: Yup. We also want to keep at least one symbol in the string besides underscores (or how else they will be indistiguishable).

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: Should be fixed now.

GirgerS:
> In [7d810cd](https://github.com/GreengageDB/greengage/pull/518/commits/7d810cd6d039046958d376c48d1b6d5fb8609e8f) was a rewrite of how assigned names are remembered. Turns out, there is no need to store all names for all schemas, because schemas are processed sequentially. So, we can free when we meet new schema.

Are they? The only similar logic I see is [here](https://github.com/GreengageDB/greengage/blob/e3bee96bfa9eda54295c868f9244744a5ea5c549/src/bin/pg_dump/pg_dump.c#L1068-L1078). First, they are indeed sorted by schema name. But then there is topological sort, so dependent objects are restored in the correct order. So for the following database:
```sql
create schema t1;
create schema t2;
CREATE VIEW t1.v1 AS (SELECT 1);
CREATE VIEW t2.v2 AS (SELECT * FROM t1.v1);
CREATE VIEW t1.v3 AS (SELECT * FROM t2.v2);
```

Schemas will alternate in the [dump](https://github.com/user-attachments/files/29886471/dump.sql). Am I missing something?

src/bin/pg_dump/pg_dump.c
VoidZeroNull0: We now again do not free it as it is needed to the end of pg_dump's job.

src/bin/pg_dump/pg_dump.c
```diff
@@ -96,6 +96,14 @@ typedef struct
 	int			objsubid;		/* subobject (table column #) */
 } SecLabelItem;
 
+typedef struct TypeNamesList {
```
red1452: one more place with bracket

VoidZeroNull0: FIxed.

src/bin/pg_dump/pg_dump.c
red1452: it was suggestion to change condition: `if (i + namelen < NAMEDATALEN -1)`.

VoidZeroNull0: > We also want to keep at least one symbol in the string besides underscores

I thought it was a question about stopping condition inside `for`, sorry.

> it was suggestion to change condition: `if (i + namelen < NAMEDATALEN -1)`.

I think, that it is correct and matches the behavior of the `makeArrayTypeName()`, no changes needed. Here, we leave a place for `\0` and don't truncate too early, else it would be `<= NAMEDATALEN` .

red1452: ok.

GirgerS:
I sorry for not noticing this earlier, but it seems that we collide with already existing array names. For example, simple dumps like:

```
CREATE TABLE t0 (a int)
PARTITION BY range(a) (
    PARTITION p START (10) END (20) WITH (tablename='p')
);

CREATE TABLE t1 (a int)
PARTITION BY range(a) (
    PARTITION p START (10) END (20) WITH (tablename='_p')
);


```

Still fail. In the source cluster, we have two array types: `p` and `_p`, so we assume that the array type name for the first table would be `__p`. But! When recreating this table in the target cluster, there are no `t1` and `_p` tables yes, to it picks `_p` as an array name... Back to the square one...

VoidZeroNull0:
As solution to the problem had to be rewritten, this PR will be closed. New one is [here](https://github.com/GreengageDB/greengage/pull/542).

PR #520, opened 2026-06-30
Fix initialization of unlogged tables at recovery

red1452:
Fix initialization of unlogged tables at recovery

Function heap_create_with_catalog creates file for new relation. It always
creates "main" fork file and creates "init" fork files for unlogged relation.

If segment was down, we can recovery it by gprecoverseg, when, simultaneously,
unlogged table is being created by parallel process. Basebackup copies all files
at time, when "main" fork was created by the heap_create_with_catalog function,
but "init" fork wasn't, "main" fork will not be filtered and will be added to
basebackup (without "init" fork). Next, "Init" fork will be created little later
by parallel process, and new record will be added to the WAL. Basebackup with
WAL will be sent to segment to start recovery process. At recovery, segment will
apply WAL to basebackup, create "init" fork, try to copy from "init" to "main"
fork and get error at the ResetUnloggedRelations function, because main fork was
created earlier.

Patch adds removing "main" fork file if it exists when "init" fork is copied to
main at recovery at the ResetUnloggedRelations function.
Test checks recovery of segment, if it contains just "main" fork. "Init" fork
should be created correctly at recovery.


src/backend/storage/file/reinit.c
```diff
@@ -335,6 +336,17 @@ ResetUnloggedRelationsInDbspaceDir(const char *dbspacedirname, int op)
...
+			/*
+			 * Sometimes, before recovery there is not init fork, but there is
+			 * main fork. So at cleanup it was not removed. Next, at
+			 * recovery init fork will be created and if we will not remove main
+			 * fork here, function copy_file will return error. Let's check main
+			 * fork here and remove if it is presented.
+			 */
+			struct stat statbuf;
+			if (lstat(dstpath, &statbuf) == 0)
+				unlink(dstpath);
```
bimboterminator1: Should we add `elog(DEBUG2)` for unsuccesful unlink? Or even 'ERROR' ? Or even remove the lstat frim if clause and use : `if (unlink(dstpath) < 0 && errno != ENOENT)`?

src/test/recovery/t/014_unlogged_reinit.pl
```diff
@@ -79,3 +79,36 @@
...
+$node->safe_psql('postgres',
+	'CREATE UNLOGGED TABLE ts1_unlogged2 (id int) TABLESPACE ts1');
+
+$baseUnloggedPath = $node->safe_psql('postgres',
+	q{select pg_relation_filepath('base_unlogged2')});
+$ts1UnloggedPath = $node->safe_psql('postgres',
+	q{select pg_relation_filepath('ts1_unlogged2')});
+
+# Crash the postmaster.
+$node->stop('immediate');
```
bimboterminator1, line 102: A checkpoint could happen between the table creation and the crash. In this case the init fork creation WAL will come before the redo point and recovery won't recreate it. So, that can lead to test flakiness

src/backend/storage/file/reinit.c
red1452: Case when we should remove file happens extremely rarely. It means that using `lstat` is faster, than using unlink for all files wihout checking by lstat.

src/test/recovery/t/014_unlogged_reinit.pl
red1452: fixed.

bimboterminator1 (dismissed):

bimboterminator1:
Why have the tests failed? because of `elog(ERROR...`?

src/backend/storage/file/reinit.c
```diff
@@ -335,6 +336,25 @@ ResetUnloggedRelationsInDbspaceDir(const char *dbspacedirname, int op)
 					 dbspacedirname, oidbuf, de->d_name + oidchars + 1 +
 					 strlen(forkNames[INIT_FORKNUM]));
 
+			/*
+			 * Sometimes, before recovery there is not init fork, but there is
```
whitehawk:
```suggestion
			 * Sometimes, before recovery there is no init fork, but there is
```

src/backend/storage/file/reinit.c
```diff
@@ -335,6 +336,25 @@ ResetUnloggedRelationsInDbspaceDir(const char *dbspacedirname, int op)
...
+			 * main fork. So at cleanup it was not removed. Next, at
+			 * recovery init fork will be created and if we will not remove main
+			 * fork here, function copy_file will return error. Let's check main
+			 * fork here and remove if it is presented.
```
whitehawk:
```suggestion
			 * fork here and remove it if presented.
```

whitehawk:
I think the description is a bit confusing.

1. The summary simply talks about 'initialization of unlogged tables', which I think could be interpreted as something related to the unlogged tables creation process.  While we actually fix the recovery scenario. So I would suggest to adjust summary to something like 'Fix initialization of unlogged tables at recovery'.

2. The original scenario involved 'gprecoverseg', but it is not mentioned in the description. The description mentions basebackup without mentioning gprecoverseg, but the tests do not involve basebackup. So, a person, who is not in context of this issue, may not understand the entire chain 'gprecoverseg' -> 'basebackup' -> 'StartupXLOG()'. I suggest to add all the details into the PR description.

whitehawk:
BTW, is it possible to face this issue with pure Postgres + basebackup? If yes, maybe we can propose this patch to Postgres community?

src/test/recovery/t/014_unlogged_reinit.pl
```diff
@@ -79,3 +79,42 @@
 	'vm fork in tablespace removed at startup');
 ok( !-f "$pgdata/${ts1UnloggedPath}_fsm",
 	'fsm fork in tablespace removed at startup');
+
+
+# Create new unlogged tables to check recovery when DB contains main fork only.
+
+# Make unreal get checkpoint after table creation before crash.
```
whitehawk:
```suggestion
# Make it impossible to get a checkpoint after table creation before a crash.
```

red1452:
> BTW, is it possible to face this issue with pure Postgres + basebackup? If yes, maybe we can propose this patch to Postgres community?

At Postgres (and GG7) the `heap_create_with_catalog` functions was rewrited, all forks are created at the begging and time between when main fork is creating and init fork is creating is less, than the same time at GG6.

whitehawk:
In the description: 'we can to recovery it by gprecoverseg' -> 'we can recover it by gprecoverseg'

src/backend/storage/file/reinit.c
```diff
@@ -335,6 +336,25 @@ ResetUnloggedRelationsInDbspaceDir(const char *dbspacedirname, int op)
...
+			 * fork here and remove it if presented.
+			 * Such situations may happen when pg_basebackup checks main fork
+			 * file at time when init fork is not created yet. pg_basebackup
+			 * does not filter such file and adds it to the backup. Next, init
+			 * fork will be created by reading xlog at recovery.
+			 */
+			struct stat statbuf;
+			if (lstat(dstpath, &statbuf) == 0)
+			{
+				if (unlink(dstpath) != 0)
```
whitehawk, line 353: Maybe call unlink(dstpath) unconditionally and ignore ENOENT or error on anything else? Therefore, we can have one invocation of syscalls instead of two.

red1452: @bimboterminator1 asked me the same question. I think, that case when we should remove file happens extremely rarely. It means that using lstat is faster, than using unlink for all files wihout checking by lstat.

whitehawk (approved):

bimboterminator1 (approved):

PR #523, opened 2026-07-02
ADBDEV-9961: Research why Orca don't use index scan after exsits transformation

riogor:
There exists a simple query with exists in it, for which ORCA doesn't use index scan.

Add LeftSemiJoin to Join2IndexApply transform to fix that.


VoidZeroNull0:
Should not we close this PR?

riogor:
Yep, totally forgot about it

PR #524, opened 2026-07-03
GG-570: Implement 'rows out' in EXPLAIN ANALYZE [6.x]

bandetto:
Implement 'rows out' in EXPLAIN ANALYZE

Backport the gp_enable_explain_rows_out GUC, off by default. When it is
on, EXPLAIN ANALYZE reports the spread of rows produced across segments
for each plan node. The figures make uneven data distribution visible
per node.

Changes:

1. Add the gp_enable_explain_rows_out boolean GUC in the experimental
   feature group, exposed as USERSET and hidden from SHOW ALL.
2. Emit a "Rows out" line in cdbexplain_showExecStats when the GUC is
   on, reporting the average rows across workers with the maximum and
   minimum rows and the segments that produced them, in both text and
   structured output formats.
3. Compute the figures as per-loop rows (ntuples/nloops) to match the
   row count already shown on the node.
4. Skip non-participating (T_Invalid) segment slots, taking the
   average, the worker count, and the minimum only over the segments
   that ran the node.
5. Cover the feature in the gp_explain regression test on a uniform and
   a skewed table, and assert the line is emitted.

EXPLAIN ANALYZE now reports per-node row-count skew across segments.

Backported from GreengageDB/greengage@cc5a070, originally taken from
open-gpdb/gpdb@1d41230.

Co-authored-by: Vladimir Rachkin \<33569237+robozmey@users.noreply.github.com\>

---

This commit will be rebased onto the target branch to preserve authorship.

### DIFF
--- a/src/test/regress/expected/gp_explain.out
+++ b/src/test/regress/expected/gp_explain.out
@@ -29,13 +29,27 @@ begin
   end loop;
 end;
 $$ language plpgsql;
+-- Return EXPLAIN ANALYZE result as xml to manipulate it further.
+create or replace function get_explain_analyze_xml_output(explain_query text)
+returns xml as
+$$
+declare
+  x xml;
+begin
+  execute 'EXPLAIN (ANALYZE, VERBOSE, FORMAT XML) ' || explain_query
+  into x;
+  return x;
+end;
+$$ language plpgsql;
 --
 -- Test explain_memory_verbosity option
 -- 
+set gp_use_legacy_hashops=off;
 CREATE TABLE explaintest (id int4);
 NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'id' as the Greengage Database data distribution key for this table.
 HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
 INSERT INTO explaintest SELECT generate_series(1, 10);
+reset gp_use_legacy_hashops;
 EXPLAIN ANALYZE SELECT * FROM explaintest;
                                                                QUERY PLAN                                                               
 ----------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


src/test/regress/expected/gp_explain.out
```diff
@@ -337,6 +337,112 @@ explain analyze SELECT * FROM explaintest;
...
 
 set gp_enable_explain_allstat=DEFAULT;
+-- Test explain rows out.
+set gp_enable_explain_rows_out=on;
+explain analyze SELECT * FROM explaintest;
+                                                          QUERY PLAN
+-------------------------------------------------------------------------------------------------------------------------------
+ Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..3.10 rows=10 width=4) (actual time=0.266..0.384 rows=5 loops=2)
+   ->  Seq Scan on explaintest  (cost=0.00..3.10 rows=4 width=4) (actual time=0.011..0.013 rows=2 loops=2)
+         Rows out: 3.33 rows avg x 3 workers, 5 rows max (seg1), 2 rows min (seg2).
```
whitehawk: In the 7x commit:
```
Rows out: 3.33 rows avg x 3 workers, 5 rows max (seg0), 1 rows min (seg1).
```
why does the `min` differ? And the segids are also different.
The distribution looks matching the values from 7x:
```
SELECT *, gp_segment_id FROM explaintest order by gp_segment_id;
 id | gp_segment_id 
----+---------------
  3 |             0
  4 |             0
  7 |             0
  8 |             0
  2 |             0
  1 |             1
  6 |             2
  9 |             2
 10 |             2
  5 |             2
(10 rows)
```

bandetto: Fixed the output

whitehawk: It appears that these explain output is ignored during test. For ex, I may set random numbers:
```diff
diff --git a/src/test/regress/expected/gp_explain.out b/src/test/regress/expected/gp_explain.out
index 5fd5bcc7d8d..4a2c9d6fc02 100644
--- a/src/test/regress/expected/gp_explain.out
+++ b/src/test/regress/expected/gp_explain.out
@@ -343,8 +343,8 @@ explain analyze SELECT * FROM explaintest;
                                                           QUERY PLAN
 -------------------------------------------------------------------------------------------------------------------------------
  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..3.10 rows=10 width=4) (actual time=0.266..0.384 rows=5 loops=2)
-   ->  Seq Scan on explaintest  (cost=0.00..3.10 rows=4 width=4) (actual time=0.011..0.013 rows=2 loops=2)
-         Rows out: 3.33 rows avg x 3 workers, 5 rows max (seg0), 1 rows min (seg1).
+   ->  Seq Scan on explaintest  (cost=0.00..3.10 rows=4 width=4) (actual time=0.011..0.013 rows=987 loops=2)
+         Rows out: 3.33 rows avg x 3 workers, 500 rows max (seg34), 1000 rows min (seg123).
    (slice0)    Executor memory: 322K bytes.
    (slice1)    Executor memory: 50K bytes avg x 3 workers, 50K bytes max (seg0).
  Memory used:  128000kB
@@ -388,14 +388,14 @@ explain (analyze, format json) SELECT * FROM explaintest;
            "Plan Width": 4,                   +
            "Actual Startup Time": 0.011,      +
            "Actual Total Time": 0.013,        +
-           "Actual Rows": 2,                  +
+           "Actual Rows": 223,                +
            "Actual Loops": 2,                 +
            "Workers": 3,                      +
-           "Average Rows": 3.33,              +
-           "Max Rows": 5,                     +
-           "Max Rows Segment": 0,             +
-           "Min Rows": 1,                     +
-           "Min Rows Segment": 1              +
+           "Average Rows": 8.12,              +
+           "Max Rows": 10,                    +
+           "Max Rows Segment": 88,            +
+           "Min Rows": 1333,                  +
+           "Min Rows Segment": 56             +
          }                                    +
        ]                                      +
      },                                       +
(END)

```
and it still will pass. So what is the point of this test? It will not catch any breakage of calculation logic for the metrics. And I'm afraid that the same question is valid for the 7x patch (#517).

bandetto: Used XML to retrieve the row count from EXPLAIN.

whitehawk: Why not to use something like:
```
-- start_matchsubs
-- m/ Gather Motion [12]:1  \(slice1; segments: [12]\)/
-- s/ Gather Motion [12]:1  \(slice1; segments: [12]\)/ Gather Motion XXX/
-- m/Execution time: \d+\.\d+ ms/
-- s/Execution time: \d+\.\d+ ms/Execution time: ##.### ms/
-- m/Planning time: \d+\.\d+ ms/
-- s/Planning time: \d+\.\d+ ms/Planning time: ##.### ms/
-- m/\(slice\d+\)    Executor memory: (\d+)\w bytes avg x \d+ workers, \d+\w bytes max \(seg\d+\)\./
-- s/Executor memory: .../Executor memory: ####K bytes avg x #### workers, ####K bytes max (seg#)./
-- end_matchsubs

explain (analyze, costs off, timing off) SELECT * FROM explaintest;
```
? Like it is done in [explain_analyze.sql](https://github.com/GreengageDB/greengage/blob/187ea238dfa092724fe7bd236477f7a6062216e2/src/test/regress/sql/explain_analyze.sql#L2). It seems more simple than XML parsing.

bandetto: I don't see how it's simpler, and matters that much. I'm using the same approach as the rest of the tests in gp_explain.sql do.

src/test/regress/sql/gp_explain.sql
```diff
@@ -34,12 +34,30 @@ begin
...
+$$ language plpgsql;
+
 
 --
 -- Test explain_memory_verbosity option
 -- 
+set gp_use_legacy_hashops=off;
 CREATE TABLE explaintest (id int4);
+set allow_system_table_mods=true;
+update gp_distribution_policy set numsegments = 3 where localoid = 'explaintest'::regclass;
```
whitehawk: Why do we need to update system tables in the test?

bandetto: This was an artifact left from me testing things. Removed.

src/test/regress/sql/gp_explain.sql
```diff
@@ -34,12 +34,27 @@ begin
...
+  into x;
+  return x;
+end;
+$$ language plpgsql;
+
 
 --
 -- Test explain_memory_verbosity option
 -- 
+set gp_use_legacy_hashops=off;
```
whitehawk, line 54: Do we need this delta in existing test case?

bandetto: It fixes the hash placement used by the expected maximum and minimum segment values.

whitehawk (approved):

bimboterminator1 (approved):

PR #525, opened 2026-07-03
Avoid shallow-copying duplicated subplan roots

Alena0704:
Avoid shallow-copying duplicated subplan roots

The duplicated subplan's subroot is only read (never mutated through its
glob->subroots slot), and makeNode+memcpy produced only a shallow copy
anyway.  Append the existing subroot to keep glob->subroots aligned with
glob->subplans, dropping the allocation.  Apply the same change to both
fixup_subplan_walker() (cdbmutate.c) and nested_subplan_mutator()
(prepunion.c).

Co-Authored-By: Andrey Sokolov <sokolov.andrey.yurevich@gmail.com>

### DIFF
--- a/src/backend/cdb/cdbmutate.c
+++ b/src/backend/cdb/cdbmutate.c
@@ -3184,20 +3184,21 @@ fixup_subplan_walker(Node *node, SubPlanWalkerContext *context)
 			 * there is more than one subplan node which refer to the same
 			 * plan_id. In this case create a duplicate subplan, append it to
 			 * the glob->subplans and update the plan_id of the subplan to
-			 * refer to the new copy of the subplan node. Note since subroot
-			 * is not used there is no need of new subroot, and initplans are
+			 * refer to the new copy of the subplan node. glob->subroots must
+			 * stay aligned with glob->subplans (same length, indexed by
+			 * plan_id); the duplicate's subroot is only ever read (e.g. by
+			 * apply_shareinput_xslice), never mutated through this slot, so we
+			 * reuse the original subroot instead of copying it. Initplans are
 			 * not needed to be duplicated.
 			 */
 			PlannerInfo *root = (PlannerInfo *) context->base.node;
 			Plan	    *dupsubplan = (Plan *) copyObject(planner_subplan_get_plan(root, subplan));
+			PlannerInfo *subroot = planner_subplan_get_root(root, subplan);
 			int			 newplan_id = list_length(root->glob->subplans) + 1;
-			PlannerInfo *dupsubroot = makeNode(PlannerInfo);
-
-			memcpy(dupsubroot, planner_subplan_get_root(root, subplan), sizeof(PlannerInfo));
-			root->glob->subroots = lappend(root->glob->subroots, dupsubroot);
 
 			subplan->plan_id = newplan_id;
 			root->glob->subplans = lappend(root->glob->subplans, dupsubplan);
+			root->glob->subroots = lappend(root->glob->subroots, subroot);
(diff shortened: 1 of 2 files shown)


hilltracer:
@Alena0704, thanks for interesting to our project!

The shallow copy in fixup_subplan_walker was added in [this](https://github.com/arenadata/gpdb/pull/342) PR, more specifically by [this](https://github.com/arenadata/gpdb/pull/342/changes/0500f91b893e432523f89e1562130d61752cc3fe) fix commit. In addition, we have a similar shallow copy of a subroot in [`nested_subplan_mutator`](https://github.com/GreengageDB/greengage/blob/75eb38e283ebb0ccd8c9b795345211508fbc548f/src/backend/optimizer/prep/prepunion.c#L1850).

A shallow copy only performs top-level copying. It creates the false impression of fully independent duplicates, but it still provides some isolation, even if only for the fields stored directly in the top-level object. Before removing it, we need to make sure that we do not modify any of that isolated data. For example, we need to verify that [here](https://github.com/GreengageDB/greengage/blob/a5eb6a32c5406274109bdd66ea9ef5e86eb4f14c/src/backend/cdb/cdbmutate.c#L1924), after we [pass](https://github.com/GreengageDB/greengage/blob/a5eb6a32c5406274109bdd66ea9ef5e86eb4f14c/src/backend/cdb/cdbmutate.c#L2013) it into the plan_tree_walker context, no top-level PlannerInfo fields are modified later.

This is not obvious to me. I also do not like that we would have two different approaches in fixup_subplan_walker and nested_subplan_mutator, where the safety check (if we want to remove shallow copy here) appears to be much more complex.

Therefore, I need strong justification before accepting this patch. @Alena0704, could you please explain what the actual benefits of changing fixup_subplan_walker would be?

Alena0704:
Thanks! The makeNode and memcpy here only makes a shallow copy - it copies the top-level fields, but parse, glob, rtable, rowMarks are pointers that the copy still shares with the original. So it doesn't really isolate much.

After the subroot is added to glob->subroots, the only thing that reads it is the loop in standard_planner that passes each subroot to set_plan_references:
https://github.com/GreengageDB/greengage/blob/64dd42f3363077a008b4e4a2ec2580b2867e4eb8/src/backend/optimizer/plan/planner.c#L383-L389

And [set_plan_references](https://github.com/GreengageDB/greengage/blob/64dd42f3363077a008b4e4a2ec2580b2867e4eb8/src/backend/optimizer/plan/setrefs.c#L342) never changes a top-level field of root - everything it writes goes into root->glob, which is shared either way, and the rest it only reads. So reusing the subroot behaves exactly like copying it. The ORCA path already works this way: its fake subroots all share one glob.

So this isn't about performance (it's one PlannerInfo per duplicated subplan) - the point is that the copy looks independent but isn't, which is easy to trip over later.

The same is true for [nested_subplan_mutator](https://github.com/GreengageDB/greengage/blob/64dd42f3363077a008b4e4a2ec2580b2867e4eb8/src/backend/optimizer/prep/prepunion.c#L1834-L1868), so updated both for consistency.

hilltracer:
Fixes LGTM. @Alena0704, please fix patch description and append info about nested_subplan_mutator. We will use it as commit description after squash. We are using postgres like style, width of 80 characters. 

The title should be succinct and capacious so that we can use it when preparing release description. Smth like: Avoid shallow-copying duplicated subplan roots.

hilltracer (approved):

andreyaksenov:
> Smth like: Avoid shallow-copying duplicated subplan roots.

LGTM

RekGRpth (approved):

PR #532, opened 2026-07-13
Fix wrong table data distribution after gpexpand

whitehawk:
Fix wrong table data distribution after gpexpand

Problem description:
In case 'synchronous_commit' was "off", after a table had been expanded and
a primary segment had gone down, the data distribution could be wrong (tuples
with the same distribution hash could be duplicated over several segments).

Root cause:
If 'synchronous_commit' is off, the PREPARE of the transaction doesn't wait
for the mirror to receive all WAL records. The issue happened if the
corresponding mirror didn't receive all WAL records by the moment the primary
had gone down. The mirror was promoted, but it had the table's data prior to the
expand operation.

Fix:
Emit an error at ALTER TABLE EXPAND TABLE and ALTER TABLE SET DISTRIBUTED BY if
'synchronous_commit' is off.


src/backend/commands/tablecmds.c
```diff
@@ -15051,6 +15052,21 @@ ATExecExpandTable(List **wqueue, Relation rel, AlterTableCmd *cmd)
...
+	 * lose/duplicate the rows this command just moved, if not all WAL was sent
+	 * to the mirror.
+	 * Force it "on" for this transaction only, so the
+	 * PREPARE below always either gets a real replication ack or a
+	 * catalog-visible mirror-down signal, regardless of the ambient GUC.
+	 */
+	set_config_option("synchronous_commit", "on",
+					   PGC_USERSET, PGC_S_SESSION,
+					   GUC_ACTION_LOCAL, true, 0);
+
```
bandetto, line 15068: Am I right that `synchronous_commit` is being set only for the time of the query?

Why not simplify and just throw an error that says 'Cannot proceed, set `synchronous_commit=on` to avoid potential loss of data'?

whitehawk: Ok, sounds reasonable, updated.

ShapatinAA:
In case of turning `synchronous_commit` to off - do we need to guaranty something? 

Here I get

```sql
create extension if not exists gp_inject_fault;
drop table if exists test;
create table test(a int) distributed by (a);
insert into test select generate_series(1, 100);

set synchronous_commit = off;

select gp_inject_fault('wal_sender_loop', 'suspend', dbid) from gp_segment_configuration where content=0 and role='p';

insert into test select generate_series(101, 150);

select '\! pg_ctl -D ' || datadir || ' stop -m immediate' as cmd_stop_primary_0 from gp_segment_configuration where content = 0 and role = 'p'
\gset 
:cmd_stop_primary_0

select gp_request_fts_probe_scan();
``` 

```sql
postgres=# select count(*) from test;
 count 
-------
   130
(1 row)
```
but should we try to fix it?

whitehawk:
> In case of turning `synchronous_commit` to off - do we need to guaranty something?
> 
> Here I get
> 
> ```sql
> create extension if not exists gp_inject_fault;
> drop table if exists test;
> create table test(a int) distributed by (a);
> insert into test select generate_series(1, 100);
> 
> set synchronous_commit = off;
> 
> select gp_inject_fault('wal_sender_loop', 'suspend', dbid) from gp_segment_configuration where content=0 and role='p';
> 
> insert into test select generate_series(101, 150);
> 
> select '\! pg_ctl -D ' || datadir || ' stop -m immediate' as cmd_stop_primary_0 from gp_segment_configuration where content = 0 and role = 'p'
> \gset 
> :cmd_stop_primary_0
> 
> select gp_request_fts_probe_scan();
> ```
> 
> ```sql
> postgres=# select count(*) from test;
>  count 
> -------
>    130
> (1 row)
> ```
> 
> but should we try to fix it?

`synchronous_commit` is inheritted from Postgres, it is not Greengage specific. And it is stated in the Postgres documentation that standby may not be durable if the `synchronous_commit` is off.  "_turning synchronous_commit off can be a useful alternative when performance is more important than exact certainty about the durability of a transaction_." So, I would say that we should not protect the usual inserts as the user is supposed to understand the risks when turning `synchronous_commit` off (and if we protect all writing operations, then there will be no sense in this GUC at all).

But, regarding gpexpand and underlying internal Greengage-specific commands - they are not mentioned by the documentation of  `synchronous_commit` at all, and the user may not be aware that it is affected by it as well. So I think it is better to protect them explicetly.

ShapatinAA (approved):

bandetto:
What if we set `synchronous_commit=on` inside a transaction, do the ALTER, then before issuing a final COMMIT, we issue `synchronous_commit=off`?

src/backend/commands/tablecmds.c
```diff
@@ -15394,6 +15409,17 @@ ATExecSetDistributedBy(Relation rel, Node *node, AlterTableCmd *cmd)
 			(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
 			errmsg("permission denied: \"%s\" is a system catalog", RelationGetRelationName(rel))));
 
+	/*
+	 * Like ATExecExpandTable, this command can rewrite and redistribute the
+	 * table's data via an internal CTAS. See the comment there.
+	 * And, forbid operation if synchronous_commit is not fully enabled here
+	 * as well.
+	 */
+	if (synchronous_commit != SYNCHRONOUS_COMMIT_ON)
```
bandetto, line 15418: Can you test if the current placement also rejects a same-key SET without REORGANIZE and EXPAND on a readable external table, even though those paths perform no redistribution?

whitehawk:
> What if we set `synchronous_commit=on` inside a transaction, do the ALTER, then before issuing a final COMMIT, we issue `synchronous_commit=off`?

These command are supposed to be issue by the `gpexpand` tool, not manually. So the protection is more about system-wide setting of the GUC.

src/backend/commands/tablecmds.c
whitehawk: I suppose that both will reject, as the `relkind` check is done after this delta

bandetto (approved):

PR #533, opened 2026-07-13
GG-627: metadata exchange communication channel (#440)

dkovalev1:
This patch is to add possibility of sending some arbitrary binary data 
(metadata) from backends to the frontend. In out particular case it could
be segment and coordinator.

The data is sent using direct libpq messages with M as a message type. Format 
of the message is pretty simple - just message length and message body.

To send messages there is a new API function pq_metadatasend which can be called 
from an extension.

The messages are processed on the frontend (coordinator in our case) side,
and collected into a linked list ggMetadataList in the cdbconn.c.

The code for message processing is inspired by the notice processing mechanism 
which works in a similar way. To lower the number of allocations,
flexible array is used.

To test the new API there is a test extension and a code for retrieving messages
implemented as well. This code uses methods PQMetadataWalk, PQgetNextMetadata, 
PQgetMetadata and PQCleanMetadata. These methods can be called from an extension 
as well.

New metadata API is available only at the backend (no FRONTEND macro defined).

ADBDEV-9854

(cherry picked from commit 907059a)

Changes from the original commit: resolved few minor conflicts.

GG-627

### DIFF
--- a/src/backend/access/transam/xact.c
+++ b/src/backend/access/transam/xact.c
@@ -93,6 +93,7 @@
 #include "utils/workfile_mgr.h"
 #include "utils/vmem_tracker.h"
 #include "cdb/cdbdisp.h"
+#include "cdb/cdbconn.h"
 #include "postmaster/autovacuum.h"
 
 /*
@@ -2978,6 +2979,8 @@ CommitTransaction(void)
 	if(Gp_role == GP_ROLE_DISPATCH || Gp_role == GP_ROLE_UTILITY)
 		MoveDbSessionLockRelease();
 
+	AtCommit_MetadataQueues();
+
 	AtCommit_TablespaceStorage();
 
 	AtCommit_Notify();
@@ -3587,6 +3590,8 @@ AbortTransaction(void)
 
 		DoPendingDbDeletes(false);
 		DatabaseStorageResetSessionLock();
+
+		AtAbort_MetadataQueues();
 
 		AtAbort_TablespaceStorage();
 		gp_guc_need_restore = true;
--- a/src/backend/cdb/dispatcher/cdbconn.c
+++ b/src/backend/cdb/dispatcher/cdbconn.c
@@ -1049,6 +1049,10 @@ PQCreateMetadataQueue(ggMetadataQueueId queue_id)
 {
 	Assert(TopTransactionContext != NULL);
 
+	if (Gp_role != GP_ROLE_DISPATCH)
+		ereport(ERROR, (errcode(ERRCODE_GP_COMMAND_ERROR),
+						errmsg("Metadata queue can only be created on QD")));
+
 	MemoryContext oldcontext = MemoryContextSwitchTo(TopTransactionContext);
 
 	ggMetadataQueue *queue = palloc0(sizeof(ggMetadataQueue));
--- a/src/backend/cdb/dispatcher/cdbconn.c
+++ b/src/backend/cdb/dispatcher/cdbconn.c
@@ -1047,6 +1047,8 @@ AtAbort_MetadataQueues(void)
 void
 PQCreateMetadataQueue(ggMetadataQueueId queue_id)
 {
+	Assert(TopTransactionContext != NULL);
+
 	MemoryContext oldcontext = MemoryContextSwitchTo(TopTransactionContext);
 
 	ggMetadataQueue *queue = palloc0(sizeof(ggMetadataQueue));
@@ -1068,6 +1070,6 @@ PQDeleteMetadataQueue(ggMetadataQueueId queue_id)
 
 	PQCleanMetadataInternal(queue);
 
-	ggMetadataQueues = list_delete(ggMetadataQueues, queue);
+	ggMetadataQueues = list_delete_ptr(ggMetadataQueues, queue);
 	pfree(queue);
 }
--- a/src/backend/cdb/dispatcher/cdbconn.c
+++ b/src/backend/cdb/dispatcher/cdbconn.c
@@ -35,6 +35,7 @@ static uint32 cdbconn_get_motion_listener_port(PGconn *conn);
 static void cdbconn_disconnect(SegmentDatabaseDescriptor *segdbDesc);
 
 static void MPPnoticeReceiver(void *arg, const PGresult *res);
+static void MPPmetadataReceiver(void *arg, ggMetadataChunk *metadata_chunk, ggMetadataQueueId queue_id);
 
 static const char *
 transStatusToString(PGTransactionStatusType status)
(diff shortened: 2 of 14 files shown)


red1452:
Please, make `git cherry-pick -x` and update link to the PR at commit message.

red1452 (dismissed):

src/backend/access/transam/xact.c
```diff
@@ -2978,6 +2979,8 @@ CommitTransaction(void)
 	if(Gp_role == GP_ROLE_DISPATCH || Gp_role == GP_ROLE_UTILITY)
 		MoveDbSessionLockRelease();
 
+	AtCommit_MetadataQueues();
```
bimboterminator1, line 2982: May the queues be created on segments as well?  Maybe prohibit that ?

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -871,3 +890,184 @@ forwardQENotices(void)
...
+	SegmentDatabaseDescriptor *segdbDesc = arg;
+
+	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
+
+	/* 
+	 * queue is expected to be created by the extension
+	 */
+	if (queue == NULL)
+	{
+		elog(WARNING, "Could not recieve chunk. No metadata queue with id %u", queue_id);
```
bimboterminator1, line 931: Should the warning be used? In case of QE failures without gang dropping, QD will have 'M' messages in connection and the next command will show those warnings

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -871,3 +890,184 @@ forwardQENotices(void)
...
+void
+AtAbort_MetadataQueues(void)
+{
+	PQDeleteMetadataQueues();
+}
+
+void
+PQCreateMetadataQueue(ggMetadataQueueId queue_id)
+{
+	MemoryContext oldcontext = MemoryContextSwitchTo(TopTransactionContext);
```
bimboterminator1, line 1056: If we are not running active transaction the palloc0 will segfault. We should make API safe for such usages.

src/backend/cdb/dispatcher/cdbconn.c
```diff
@@ -871,3 +890,184 @@ forwardQENotices(void)
...
+PQDeleteMetadataQueue(ggMetadataQueueId queue_id)
+{
+	ggMetadataQueue *queue = PQMetadataFindQueue(queue_id);
+
+	if (!queue)
+		elog(ERROR, "No metadata queue with id %u", queue_id);
+
+	PQCleanMetadataInternal(queue);
+
+	ggMetadataQueues = list_delete(ggMetadataQueues, queue);
```
bimboterminator1: it's quite unsafe using list_delete, since it uses equal() , which compares nodetags, but queue does not have one

src/backend/libpq/pqformat.c
```diff
@@ -668,3 +669,44 @@ pq_sendint(StringInfo buf, uint32 i, int b)
...
+	 */
+	if (len > (size_t)(INT_MAX - sizeof(int32) - sizeof(int)))
+		ereport(ERROR,
+				(errcode(ERRCODE_PROGRAM_LIMIT_EXCEEDED),
+				 errmsg("custom metadata message too large")));
+
+	/*
+	 * Initialize the message buffer. 'M' is the message type tag.
+	 */
+	pq_beginmessage(&buf, 'M');
```
bimboterminator1, line 699: There is a machinery which may be broken by specific message type https://github.com/GreengageDB/greengage/blob/770674d690532017eaa5b53963c1cbd6c445bb19/src/interfaces/libpq/fe-protocol3.c#L1772-L1813 . I guess, currently, it's irrelevant?

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2380,3 +2390,54 @@ build_startup_packet(const PGconn *conn, char *packet,
...
+	}
+
+	int32 queue_id;
+
+	if (pqGetInt(&queue_id, sizeof(ggMetadataQueueId), conn))
+	{
+		return 1;
+	}
+
+	int payload_len = length - sizeof(ggMetadataQueueId);
```
bimboterminator1, line 2419: Is it guaranteed that length >= sizeof(ggMetadataQueueId)

src/interfaces/libpq/fe-protocol3.c
```diff
@@ -2380,3 +2390,54 @@ build_startup_packet(const PGconn *conn, char *packet,
...
+
+	/*
+	 * Since the metadata might be pretty long, we create an own buffer
+	 * rather than using conn->workBuffer.  workBuffer is intended
+	 * for stuff that is expected to be short.
+	 */
+
+	ggMetadataChunk *chunk = pqPalloc(sizeof(ggMetadataChunk) + payload_len);
+	if (chunk == NULL)
```
bimboterminator1, line 2428: Currently, pqPalloc() throws ERROR instead of NULL.

src/backend/access/transam/xact.c
dkovalev1: In my understanding, metadata queues are orthogonal to the GG distributed architecture and queues can be created on any host. In theory, it is still possible to make an application or an extension which can benefit from metadata on the segments although this is not implemented yet. I don't think we need to prohibit it intentionally.

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: Do you suggest to do ERROR instead?

src/backend/libpq/pqformat.c
dkovalev1: I think this is unrelated code processing COPY.

src/interfaces/libpq/fe-protocol3.c
dkovalev1: Technically yes, but it is safe to check it twice and it is double checked in some other places over the code.

src/interfaces/libpq/fe-protocol3.c
dkovalev1: yes, by design. Added assert.

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: Good catch, list_delete_ptr() seems to suit better.

src/backend/cdb/dispatcher/cdbconn.c
dkovalev1: Queue lives in the transaction, so active transaction is expected. Added Assert since it is a programmer's error.

src/backend/cdb/dispatcher/cdbconn.c
bimboterminator1: do DEBUG2 ?

src/backend/access/transam/xact.c
bimboterminator1: Then we should add the queries cleanup into the PrepareTransaction. It means if a query has been created on segments, PREPARE transaction will cause toptransactioncontext drop, and we get a dangling pointer

dkovalev1: Agreed, let's stick to the current use case and allow queue creation on QD only.

bimboterminator1 (approved):

red1452 (approved):

PR #536, opened 2026-07-16
GG-613: Allow running the pg_upgrade test on a regression dump

GirgerS:
Prepare pg_upgrade test scripts so they can be run on the CI with regression dump.

The changes are as follows:
- change how the compared dumps are collected use: pg_dump instead of
  pg_dumpall to be able to exclude objects from them.
- move pg_upgrade_run_6X_to_7X_migration.sh script to the ci folder, closer to
  other CI scripts.
- move most of the test instructions to the readme.ci file, and the rest to
  the README.gpdb file. Also, describe how to test regression dump inside 
  the docker container.
- add the CLEANUP_SCRIPT and DUMP_OPTIONS options to the
  ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash file
  the -f, -d and -v flags to test_gpdb.sh.
- add the src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql script, 
  to cleanup regression dump before upgrading it.

It is worth noting that with the current pg_dump state, partitioned, 
append-optimized, foreign tables and several other objects 
would cause pre- and post-upgrade dumps to differ, making the test fail.
As a workaround, partitioned table are compared separately, and pg_dumpall
during testing is invoked with --data-only option.


GirgerS:
Before performing the test, it is necessary to merge in-progress pull requests, that fix remaining pg_upgrade issues:
```bash
# Assuming that current working directory is the root of the Greengage 7 source tree
git checkout GG-613
git rebase origin/GG-531
git rebase origin/GG-626
docker build -f ci/Dockerfile.pg_upgrade -t gpdb7_pgupgrade:latest --build-arg GGDB7_IMAGE=gpdb7_u22:latest .


mkdir dump_dir
mkdir logs

# Download any regression dump from 6.x, and copy it to ./dump_dir/dump.sql
# For example, from https://github.com/GreengageDB/greengage/actions/runs/26749706317

docker run --rm \
	-v $(pwd)/dump_dir:/dump_dir \
	-v $(pwd)/logs:/logs \
	gpdb7_pgupgrade bash -c \
	"export SQL_SCHEMA=/dump_dir/dump.sql; \
	 export CLEANUP_SCRIPT=/home/gpadmin/gpdb_src/src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql; \
	 export DUMP_OPTIONS='--data-only --extra-float-digits=-3'; \
	 gpdb_src/concourse/scripts/setup_gpadmin_user.bash; \
	 su gpadmin /home/gpadmin/gpdb_src/ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump1.sql /logs/dump1.sql; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump2.sql /logs/dump2.sql; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/regression.diffs /logs/regression.diffs;"
```

ci/readme.md
```diff
@@ -176,3 +176,94 @@ All tests are run in one way or another on the demo cluster, wherever possible.
...
+export GREENGAGE7_SRC=/home/gpadmin/gpdb7_src
+export GREENGAGE6_INSTALLATION=/usr/local/greengage-db-6X
+export GREENGAGE7_INSTALLATION=/usr/local/greengage-db-7X
+export SQL_SCHEMA=/home/gpadmin/dump.sql
+export CLEANUP_SCRIPT=/home/gpadmin/cleanup.sql
+export DUMP_OPTIONS="--data-only"
+bash   ${GREENGAGE7_SRC}/src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh
+```
+
+## Running pg_upgrade test from the image
```
hilltracer:
```suggestion
## Running pg_upgrade test from the docker image
```

src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql
```diff
@@ -0,0 +1,400 @@
...
+ALTER DATABASE dsp2 RESET gp_default_storage_options;
+
+
+-- The way how tsquery is displayed was chagned between versions, so is may differ between dumps.
+-- See "tsqueryout" and "infix" functions for more details.
+\c regression
+DROP TABLE public.test_tsquery;
+
+
+-- Drop ALL partitioned tables, as the they are dumped differently based on the version of the source cluster.
```
hilltracer: Why we drop all partition tables? How will we test p.tables upgrade?

ci/readme.md
```diff
@@ -176,3 +176,94 @@ All tests are run in one way or another on the demo cluster, wherever possible.
...
+To run pg_upgrade tests with the regression dump, it is necessary to specify `cleanup_regression_dump_from_6X.sql` cleanup script. Also, only --data-only dumps can be compared without any difference. The --extra-float-digits option is specified to synchronize the formatting of floats in the dumps.
+```bash
+mkdir logs
+docker run --rm \
+	-v $(pwd)/dump_dir:/dump_dir \
+	-v $(pwd)/logs:/logs \
+	gpdb7_pgupgrade bash -c \
+	"export SQL_SCHEMA=/dump_dir/dump.sql; \
+	 export CLEANUP_SCRIPT=/home/gpadmin/gpdb_src/src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql; \
+	 export DUMP_OPTIONS='--data-only --extra-float-digits=-3'; \
```
hilltracer, line 265: why do we need `--extra-float-digits`? Is there possibility of missing some important error because of this? Is it possible to make this WA rounding more precise?

src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql
```diff
@@ -0,0 +1,400 @@
...
+DROP FUNCTION qp_misc_rio.func_plpythonu2(x integer);
+DROP FUNCTION qp_misc_rio.t18_pytest();
+DROP FUNCTION qp_query_execution.qx_count_operator(query text, planner_operator text, optimizer_operator text);
+DROP FUNCTION qp_with_functional_inlining.cte_func3();
+DROP FUNCTION qp_with_functional_noinlining.cte_func3();
+DROP FUNCTION sort_schema.has_sortmethod(explain_analyze_query text);
+
+
+-- Views with removed functions
+\c regression;
```
hilltracer:
```suggestion
\c regression
```

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -331,6 +337,12 @@ main() {
 			B )
 				realpath OLD_BINDIR "${OPTARG}"
 				;;
+			f )
+				realpath cleanup_script "${OPTARG}"
+				;;
+			d )
+				DUMP_OPTS+="${OPTARG}"
```
hilltracer:
```suggestion
				DUMP_OPTS+=" ${OPTARG}"
```
?

src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql
```diff
@@ -0,0 +1,400 @@
+-- Tables partitioned by multiple columns
+\c regression
+DROP TABLE bfv_partition.t26002_t1;
```
hilltracer, line 3: The drop commands don't include `IF EXISTS` clause, and it is [ran](https://github.com/GreengageDB/greengage/blob/e285224bdc85856e88c77e4623149afaf4acb463/src/bin/pg_upgrade/test_gpdb.sh#L408) with ON_ERROR_STOP=1. So any missed object will stop execution. Is it ok?

maybe add 'IF EXISTS'? Because regression dump  can be changed.

ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash
hilltracer, line 67: As far as I understand here we will have ON_ERROR_STOP=0. So any failed SQL commands won't stop execution. How will we know that we test something in regression dump?

ci/readme.md
```diff
@@ -176,3 +176,94 @@ All tests are run in one way or another on the demo cluster, wherever possible.
...
+
+For example:
+```bash
+export GREENGAGE6_SRC=/home/gpadmin/ggdb6_src
+export GREENGAGE7_SRC=/home/gpadmin/gpdb7_src
+export GREENGAGE6_INSTALLATION=/usr/local/greengage-db-6X
+export GREENGAGE7_INSTALLATION=/usr/local/greengage-db-7X
+export SQL_SCHEMA=/home/gpadmin/dump.sql
+export CLEANUP_SCRIPT=/home/gpadmin/cleanup.sql
+export DUMP_OPTIONS="--data-only"
```
hilltracer: Using `--data-only` prevents testing any DDL operations. Is it ok? How will we test it?

ci/readme.md
```diff
@@ -176,3 +176,94 @@ All tests are run in one way or another on the demo cluster, wherever possible.
...
+docker run --rm \
+	-v $(pwd)/dump_dir:/dump_dir \
+	-v $(pwd)/logs:/logs \
+	gpdb7_pgupgrade bash -c \
+	"export SQL_SCHEMA=/dump_dir/dump.sql; \
+	 gpdb_src/concourse/scripts/setup_gpadmin_user.bash; \
+	 su gpadmin /home/gpadmin/gpdb_src/ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump1.sql /logs/dump1.sql; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump2.sql /logs/dump2.sql; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/regression.diffs /logs/regression.diffs;"
```
hilltracer: test_gpdb run without `-r`. Do I understand right that with success execution all artifacts will be removed? Is it ok?

ci/readme.md
```diff
@@ -176,3 +176,94 @@ All tests are run in one way or another on the demo cluster, wherever possible.
...
+For example:
+```bash
+export GREENGAGE6_SRC=/home/gpadmin/ggdb6_src
+export GREENGAGE7_SRC=/home/gpadmin/gpdb7_src
+export GREENGAGE6_INSTALLATION=/usr/local/greengage-db-6X
+export GREENGAGE7_INSTALLATION=/usr/local/greengage-db-7X
+export SQL_SCHEMA=/home/gpadmin/dump.sql
+export CLEANUP_SCRIPT=/home/gpadmin/cleanup.sql
+export DUMP_OPTIONS="--data-only"
+bash   ${GREENGAGE7_SRC}/src/bin/pg_upgrade/pg_upgrade_run_6X_to_7X_migration.sh
```
hilltracer:
```suggestion
bash   ${GREENGAGE7_SRC}/ci/scripts/pg_upgrade_run_6X_to_7X_migration.sh
```

src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql
hilltracer: It seems that we will suffer to maintain the cleanup list after changing the regression tests and generating a new dump. Is it possible to automate something here? Do we really need to delete all partitioned tables this way?

GirgerS: > Why we drop all partition tables? 

They are dumped by different code paths depending on the version (i.e. by Greengage-specific code from 6X, and PostgreSQL code from 7.x),
and rewriting this seems to give no benefits for pg_upgrade, besides being able to run this test.

>  How will we test p.tables upgrade?

And it is a good question. I asked something similar internally, but it got no feedback. So let's think about it here.

One low-effort solution that comes to mind is to to two separate runs of the test:
	- one without partitioned tables, with dump comparison
	- one with partitioned tables, but without dump comparison (and then eyeballing it when necessary)

ci/readme.md
GirgerS: No, the only difference in how floats are displayed:
	https://github.com/GreengageDB/greengage/blob/7.x/src/backend/utils/adt/float.c#L306-L321
	https://github.com/GreengageDB/greengage/blob/6.x/src/backend/utils/adt/float.c#L313-L342


The workaround is taken from here:
	https://github.com/GreengageDB/greengage/blob/7.x/src/bin/pg_upgrade/TESTING#L62-L66

src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql
GirgerS: I'd say that this should not cause that much troubles (on its own), because we don't remove tests that often.

ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash
GirgerS: There are still unresolved `pg_dump` issues (for example, GG-555), that cause some (minor) errors when restoring the dump, and make the test fail if `ON_ERROR_STOP=1` is specified.

ci/readme.md
GirgerS: The reason to use `--data-only` option is to work around the fact that DDL for AO and external tables is different between versions/dumps

> How will we test it?

Good question, again. But I'd say that the current idea is to test at least something (and the fact that we can upgrade regression schema much better than no tests at all). And we can always iterate on it.

src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql
GirgerS: > It seems that we will suffer to maintain the cleanup list after changing the regression tests and generating a new dump.

I'll take a step back and say that there is a worse problem with this test. The dump that we test it collected on 6.x branch, and the test is performed on 7.x branch.
Meaning that it is possible to create an incompatible object on 6.x branch (because all the tests are passing), and silently kill this test on 7.x. And if it is mandatory... then we stalled all PRs! The easiest workaround is to run pg_upgrade tests only when pg_upgrade/pg_dump/pg_restore chagnes, or on demand.

And I don't see an easy way to collect removed/failing objects. We can run SQL queries to collect them, but this is basically duplicating pg_dump logic.

ci/readme.md
GirgerS: Fixed

ci/readme.md
GirgerS: Fixed

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: Fixed

src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql
GirgerS: Fixed

ci/readme.md
GirgerS: Fixed

ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash
hilltracer: Do I understand right that we will do ON_ERROR_STOP=1 eventually? Is it possible to do it? Is there task for it?
If no, I think we need to decide how we will check that pg_upgrade test is really testing restoring of objects.

GirgerS: > Do I understand right that we will do ON_ERROR_STOP=1 eventually

Well, it should be the goal. 

> Is there task for it?

These are two tasks related to the errors themselves (GG-554, GG-555), but I'm not sure if we are planning to do them.

> If no, I think we need to decide how we will check that pg_upgrade test is really testing restoring of objects

And also, these errors are not really related to `pg_upgrade`, as they are more about `pg_dump` from 6.x being able to do it's job correctly. And it doesn't participate in upgrade.

hilltracer: Ok, but still `how we will check that pg_upgrade test is really testing restoring of objects`?

Now regression dump execution may fail to load many objects and we won't know about it.

ci/readme.md
```diff
@@ -176,3 +176,94 @@ All tests are run in one way or another on the demo cluster, wherever possible.
 For example, cross_subnet tests or tests with tag `concourse_cluster` currently not worked because of too complex cluster preconditions.
 
 Tests in a docker-compose cluster use the same ssh keys for `gpadmin` user and pre-add the cluster hosts to `.ssh/know_hosts` and `/etc/hosts`.
+
+## Running pg_upgrade test locally
+
+To run `pg_upgrade_run_6X_to_7X_migration.sh` locally, it is necessary to specify the following environment variables:
```
bimboterminator1: `.bash` here and in other places

ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash
bimboterminator1, line 2: Maybe add `#!/usr/bin/env bash` for beauty

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -392,8 +404,8 @@ main() {
 	fi
 	
 	# Run any pre-upgrade tasks to prep the cluster
-	if [ -f "test_gpdb_pre.sql" ]; then
-		if ! ${OLD_BINDIR}/psql -f test_gpdb_pre.sql -v ON_ERROR_STOP=1 postgres; then
+	if [ -n "$cleanup_script" ]; then
+		if ! ${OLD_BINDIR}/psql -f $cleanup_script -v ON_ERROR_STOP=1 postgres; then
```
bimboterminator1: I suggest quote the `"$cleanup_script"` - a path with spaces breaks the unquoted form

bimboterminator1 (commented):
Minor comments. The solution itself LGTM

ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash
GirgerS: It is done this way in the main script, so why not.

ci/readme.md
GirgerS: I should be more careful next time. 

`grep -riI pg_upgrade_run_6X_to_7X_migration.sh` shows none for me.

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: Should be done.

`grep -riI cleanup_script` shows that they are quoted for me.

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -215,11 +223,14 @@ usage()
...
 	echo " -b <dir>     new cluster executable directory"
 	echo " -B <dir>     old cluster executable directory (defaults to new binaries)"
+	echo " -f <file>    Script to run in the old cluster before performing upgrade (defaults to none)"
+	echo " -d <string>  Specify additional pg_dump options"
 	echo " -s           Run smoketest only"
 	echo " -C           Skip gpcheckcat test"
 	echo " -m           Upgrade mirrors"
 	echo " -r           Retain temporary installation after test, even on success"
 	echo " -p           pg_upgrade performance checking only"
+	echo " -v           test pg_upgrade from Greenagage 6 to Greengage 7"
```
hilltracer:
```suggestion
	echo " -v           test pg_upgrade from Greengage 6 to Greengage 7"
```

src/bin/pg_upgrade/test_gpdb.sh
hilltracer: README.test.md was removed

ci/readme.md
```diff
@@ -176,3 +176,97 @@ All tests are run in one way or another on the demo cluster, wherever possible.
...
+	 export CLEANUP_SCRIPT=/home/gpadmin/gpdb_src/src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql; \
+	 export DUMP_OPTIONS='--data-only --extra-float-digits=-3'; \
+	 gpdb_src/concourse/scripts/setup_gpadmin_user.bash; \
+	 su gpadmin /home/gpadmin/gpdb_src/ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump1.sql /logs/dump1.sql; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump2.sql /logs/dump2.sql; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump_partitions1.sql /logs/dump_partitions1.sql; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump_partitions2.sql /logs/dump_partitions2.sql; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/regression.diffs /logs/regression.diffs; \
+	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/partitions_regression.diffs /logs/partitions_regression.diffs;"
```
hilltracer: looks like this doesn't work with success result. since regression.diffs will be removed and partitions_regression.diffs will create only with [error](https://github.com/GreengageDB/greengage/blob/b679f1b981fd8a06ec29c8368f42de24a415b11a/src/bin/pg_upgrade/test_gpdb.sh#L376-L375).

src/bin/pg_upgrade/test_gpdb.sh
hilltracer, line 452: should we remove partitions_regression.diffs here also? And the same for Makefile [cleanup](https://github.com/GreengageDB/greengage/blob/e285224bdc85856e88c77e4623149afaf4acb463/src/bin/pg_upgrade/Makefile#L41)

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -410,10 +506,76 @@ main() {
...
+								WHERE a.attnum > 0::pg_catalog.int2
+									AND EXISTS (
+										SELECT 1
+										FROM pg_partition p
+										WHERE p.parrelid = c.oid
+									)
+							GROUP BY (n.nspname, c.relname)
+						) subq
+						ORDER BY (out)
+					) TO STDOUT;
```
hilltracer: We can have problem with dropped columns. Repro on 7.x:

<details>
<summary>spoiler</summary>

```sql
CREATE TABLE part_table (id int, a int, b int, c int)
DISTRIBUTED BY (id)
PARTITION BY RANGE (a)
    SUBPARTITION BY RANGE (b)
        SUBPARTITION TEMPLATE (START (1) END (3) EVERY (1))
    (START (1) END (3) EVERY (1));

INSERT INTO part_table
    SELECT i+1, mod(i/4,2)+1, mod(i/2,2)+1
    FROM generate_series(0,100) AS i;
```

```sql
ALTER TABLE part_table DROP COLUMN c;
```

```sql
COPY (
	SELECT FORMAT('COPY %s (%s) TO STDOUT;', name, attrs) COLLATE "default" out
	FROM (
		SELECT FORMAT('%s.%s', quote_ident(n.nspname),
		quote_ident(c.relname)) name,
		string_agg(quote_ident(a.attname), ',' ORDER BY a.attnum) attrs
		FROM pg_class c
			JOIN pg_namespace n ON c.relnamespace = n.oid
			JOIN pg_attribute a ON a.attrelid = c.oid
		WHERE c.relkind = 'p'
			AND a.attnum > 0::pg_catalog.int2
			AND EXISTS (
				SELECT 1
				FROM pg_partition_tree(c.oid)
				WHERE parentrelid IS NULL
			)
		GROUP BY (n.nspname, c.relname)
	) subq
	ORDER BY (out)
) TO STDOUT;
```
result:
```sql
COPY public.part_table (id,a,b,"........pg.dropped.4........") TO STDOUT;
```
execution:
```sql
gpadmin=# COPY public.part_table (id,a,b,"........pg.dropped.4........") TO STDOUT;
ERROR:  column "........pg.dropped.4........" of relation "part_table" does not exist
```

</details>

src/bin/pg_upgrade/test_gpdb.sh
hilltracer, line 354: main() and diff_and_exit() contain a lot of duplicated dump logic. Consider extracting the common database iteration and partition dump handling into a helper function, while keeping the version-specific SQL queries separate.

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -265,8 +276,76 @@ diff_and_exit() {
...
+					) subq
+					ORDER BY (out)
+				) TO STDOUT;
+			EOF
+			)
+			queries_string=$(PGOPTIONS="${pgopts}" psql "${database}" -c "${queries_query}")
+			readarray -t queries <<< ${queries_string}
+			for query in "${queries[@]}"; do
+				echo "${query}" >> "$temp_root/dump_partitions2.sql"
+				PGOPTIONS="${pgopts}" psql "${database}" -c "${query}" | sort >> "$temp_root/dump_partitions2.sql"
```
hilltracer: looks like `| sort >> "` can mask dump-generation failures and let the test pass by comparing incomplete dump files. Since the script don't use `pipefail`.

hilltracer: Here we don't use float processing like `--extra-float-digits=-3` for pg_dump. Is it problem?

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: Should be [fixed](https://github.com/GreengageDB/greengage/pull/536/changes/b56785453907a79e6122402e85b0e9697a8fba17)

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: Should be [fixed](https://github.com/GreengageDB/greengage/pull/536/changes/e52231c2bb1b590d98e87a03854db6b0c6201182)

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: Should be [fixed](https://github.com/GreengageDB/greengage/pull/536/changes/d3fdaa21c98ac4c732b7a165f5d5fce93ae18504#r3686859529)

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: It funny how I explicitly mentioned `dumpTableData_copy`, but forgot to implement this check. Should be [fixed](https://github.com/GreengageDB/greengage/pull/536/changes/fb6a9c137149a732383fda8dea1d1663d1e10df3).

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: Should be fixed as a part of [this](https://github.com/GreengageDB/greengage/pull/536/changes/1657e554c87b66bb647aeccc84ea0e8fffdfe8aa) commit

GirgerS: I'd guess it may become a problem in the future. Should be fixed after [this](https://github.com/GreengageDB/greengage/pull/536/changes/d0cd7eb121e8be8683b0a6b2984bbef0b9f0b011) commit.

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: It seems to be better with a helper function (though most of the code is still queries). Please take a look at the [commit](https://github.com/GreengageDB/greengage/pull/536/changes/1657e554c87b66bb647aeccc84ea0e8fffdfe8aa).

ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash
GirgerS: As it was decided on the internal discussion, the best solution is to fix underlying pg_dump issues and enable `ON_ERROR_STOP` after it is done. [Here](https://tracker.yandex.ru/GG-659) is the task for it. 

For now, leave this unchanged.

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: Makefile should also be [fixed](https://github.com/GreengageDB/greengage/pull/536/changes/3898e804287280ee2cee0c1a3cb9203a3dc10b78)

ci/readme.md
GirgerS: Should be [fixed](https://github.com/GreengageDB/greengage/pull/536/commits/8e99abb1ba9929b948a010c09cbfcec39afac27a)

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -175,6 +175,61 @@ check_vacuum_worked()
...
+dump_database_schema()
+{
+	dump_path="$1"
+	partitions_dump_path="$2"
+	partitions_query="$3"
+	queries_query="$4"
+	pgopts="$5"
+
+	# To dump schemas, use pg_dump from NEW_BINDIR. Note, that we intentionally don't use pg_dumpall,
+	# and dump each database by hand, because we need to be able to exclude specific
```
hilltracer, line 189: `because we need to be able to exclude specific`
looks like sentence isn't continued.

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -175,6 +175,61 @@ check_vacuum_worked()
...
+				# And now, create a separate dump for each partitioned table, by manually getting
+				# their contents. Do this the way as in the pg_dump (see dumpTableData_copy).
+				# Use the --extra-float-digits option to make sure that floats
+				# are dumped identically.
+				# Also, sort the rows, as their order may differ between the versions.
+				queries_string=$(PGOPTIONS="${pgopts}" psql "${database}" -c "${queries_query}")
+				readarray -t queries <<< ${queries_string}
+				for query in "${queries[@]}"; do
+					result=$(PGOPTIONS="-c extra_float_digits=-3 ${pgopts}" psql "${database}" -c "${query}")
+					result=$(echo "${result}" | sort)
```
hilltracer: This now captures partition table data in a shell variable, unlike the previous streaming used
```sh
				PGOPTIONS="${pgopts}" psql "${database}" -c "${query}" | sort >> "$temp_root/dump_partitions2.sql"
```

For regression dumps all data in patition tables can be large; can we keep this path file/stream-based while still checking the psql status?

src/bin/pg_upgrade/test_gpdb.sh
hilltracer: Looks like the issue is still actual. 

Fine example:

<details>
<summary>spoiler</summary>

init data:
```sql
DROP TABLE IF EXISTS part_table;
CREATE TABLE part_table (id int, a int, b int, c int)
DISTRIBUTED BY (id)
PARTITION BY RANGE (a)
    SUBPARTITION BY RANGE (b)
        SUBPARTITION TEMPLATE (START (1) END (3) EVERY (1))
    (START (1) END (3) EVERY (1));

INSERT INTO part_table
    SELECT i+1, mod(i/4,2)+1, mod(i/2,2)+1
    FROM generate_series(0,10) AS i;
```

```sh
#!/usr/bin/env bash

set -euo pipefail

perf_test=0
cross_version_upgrade=1
pgopts=''

DUMP_OPTS='--data-only --extra-float-digits=-3'

rm -f ./dump2.sql ./dump_partitions2.sql

temp_root="."
dump_path="$temp_root/dump2.sql"
partitions_dump_path="$temp_root/dump_partitions2.sql"
NEW_BINDIR='/usr/local/greengage-db-devel/bin'

partitions_query=$(cat <<- EOF
	COPY (
		WITH partitions AS (
			SELECT DISTINCT (pg_partition_tree(oid)).relid
			FROM pg_class
		) SELECT FORMAT('%s.%s', quote_ident(n.nspname), quote_ident(c.relname)) out
		FROM partitions p
			JOIN pg_class c ON p.relid = c.oid
			JOIN pg_namespace n ON c.relnamespace = n.oid
		ORDER BY(out)
	) TO STDOUT;
EOF
)

queries_query=$(cat <<- EOF
	COPY (
		SELECT FORMAT('COPY %s (%s) TO STDOUT;', name, attrs) COLLATE "default" out
		FROM (
			SELECT FORMAT('%s.%s', quote_ident(n.nspname),
			quote_ident(c.relname)) name,
			string_agg(quote_ident(a.attname), ',' ORDER BY a.attnum) attrs
			FROM pg_class c
				JOIN pg_namespace n ON c.relnamespace = n.oid
				JOIN pg_attribute a ON a.attrelid = c.oid
			WHERE c.relkind = 'p'
				AND a.attnum > 0::pg_catalog.int2
				AND a.attisdropped = false
				AND EXISTS (
					SELECT 1
					FROM pg_partition_tree(c.oid)
					WHERE parentrelid IS NULL
				)
			GROUP BY (n.nspname, c.relname)
		) subq
		ORDER BY (out)
	) TO STDOUT;
EOF
)

dump_database_schema()
{
	dump_path="$1"
	partitions_dump_path="$2"
	partitions_query="$3"
	queries_query="$4"
	pgopts="$5"

	# To dump schemas, use pg_dump from NEW_BINDIR. Note, that we intentionally don't use pg_dumpall,
	# and dump each database by hand, because we need to be able to exclude specific
	if (( !$perf_test )) ; then
		databases_string=$(PGOPTIONS="${pgopts}" psql template1 -c "COPY (SELECT datname FROM pg_database WHERE datname != 'template0' ORDER BY (datname)) TO STDOUT;");
		readarray -t databases <<< ${databases_string}
		for database in "${databases[@]}"; do
			# We are getting '\' symbol automatically escaped to '\\', convert is back
			database=$(echo "${database}" | sed 's/\\\\/\\/g')

			# When upgrading Greengage 6 to Greengage 7, partitioned tables would cause
			# pre- and post-upgrade dumps to differ, because the way they are dumped depends
			# on the version of the source cluster. So, make pg_dump ignore them and compare
			# them seperately.
			args_to_ignore_partitions=()
			if (( $cross_version_upgrade )); then
				partitions_string=$(PGOPTIONS="${pgopts}" psql "${database}" -c "${partitions_query}")
				readarray -t partitions <<< ${partitions_string}
				for partition in "${partitions[@]}"; do
					if [ -z "${partition}" ]; then
						continue
					fi
					args_to_ignore_partitions+=(-T)
					args_to_ignore_partitions+=("${partition}")
				done

				# And now, create a separate dump for each partitioned table, by manually getting
				# their contents. Do this the way as in the pg_dump (see dumpTableData_copy).
				# Use the --extra-float-digits option to make sure that floats
				# are dumped identically.
				# Also, sort the rows, as their order may differ between the versions.
				queries_string=$(PGOPTIONS="${pgopts}" psql "${database}" -c "${queries_query}")
				readarray -t queries <<< ${queries_string}
				for query in "${queries[@]}"; do
					result=$(PGOPTIONS="-c extra_float_digits=-3 ${pgopts}" psql "${database}" -c "${query}")
					result=$(echo "${result}" | sort)

					echo "${query}"  >> "${partitions_dump_path}"
					echo "${result}" >> "${partitions_dump_path}"
				done
			fi

			PGOPTIONS="${pgopts}" ${NEW_BINDIR}/pg_dump "${database}" "${args_to_ignore_partitions[@]}" ${DUMP_OPTS} >> "${dump_path}"
			echo ${dump_path}
		done
		echo done
	fi
}

dump_database_schema "${dump_path}" "${partitions_dump_path}" "${partitions_query}" "${queries_query}" "${pgopts}"
```

result:
```
COPY public.part_table (id,a,b,c) TO STDOUT;
10	1	1	\N
11	1	2	\N
1	1	1	\N
2	1	1	\N
3	1	2	\N
4	1	2	\N
5	2	1	\N
6	2	1	\N
7	2	2	\N
8	2	2	\N
9	1	1	\N




```

</details>

example with `qwe` iinside queries_query

<details>
<summary>spoiler</summary>

```sh
gpadmin@gpdb7:~/src/tasks/GG-613/tmpdir$ ./script.sh 
ERROR:  syntax error at or near "qwe"
LINE 1: qwe
        ^
./dump2.sql
ERROR:  syntax error at or near "qwe"
LINE 1: qwe
        ^
./dump2.sql
ERROR:  syntax error at or near "qwe"
LINE 1: qwe
        ^
./dump2.sql
done

```

but the empty dump is still created

```






```

</details>

So we will compare two empty dump and will say all is ok.

GirgerS: Ah, I see, these is no `set -e` in the script, so we have to check exit codes manually

Should work [now](https://github.com/GreengageDB/greengage/pull/536/changes/2111934a84adadb58a06834a3562ed114c853b11)

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: [Right](https://github.com/GreengageDB/greengage/pull/536/changes/773ec7d7871f7883fef45ef0cc2766a846be2f8a)

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: The easiest thing to do is to set pipefail inside the function then.

ci/readme.md
```diff
@@ -176,3 +176,102 @@ All tests are run in one way or another on the demo cluster, wherever possible.
...
+
+To run `pg_upgrade_run_6X_to_7X_migration.bash` locally, it is necessary to specify the following environment variables:
+- GREENGAGE6_SRC - directory of the Greengage6 source code.
+- GREENGAGE7_SRC - directory of the Greengage7 source code.
+- GREENGAGE6_INSTALLATION - directory where the Greengage6 installation is located (i.e., after make install).
+- GREENGAGE7_INSTALLATION - directory where the Greengage7 installation is located (i.e., after make install).
+- SQL_SCHEMA - optional path to an sql file that will be loaded before performing the upgrade. If it is not specified,
+  the upgrade will be performed on an empty cluster.
+- CLEANUP_SCRIPT - optional path to an sql file that will be executed after loading SQL_SCHEMA, before performing pg_upgrade. It can be used to remove deprecated or otherwise failing objects from SQL_SCHEMA. If it is not specified, no cleanup will occur.
+- DUMP_OPTIONS - optional parameters that will be passed to pg_dumpall to collect pre- and post-upgrade dumps. These dumps are compared at the end of the test, to determine whether it was successful. If this option is not specified, pg_dumpall will be executed with no parameters.
```
hilltracer:
```suggestion
- DUMP_OPTIONS - optional parameters that will be passed to pg_dump to collect pre- and post-upgrade dumps. These dumps are compared at the end of the test, to determine whether it was successful. If this option is not specified, pg_dump will be executed with no parameters.
```

ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash
```diff
@@ -33,6 +34,20 @@ else
...
+    echo "CLEANUP_SCRIPT environment variable is not set. No cleanup will occur."
+else
+    echo "CLEANUP_SCRIPT is set. This test will load \"${CLEANUP_SCRIPT}\" before performing upgrade."
+fi
+
+DUMP_OPTIONS=${DUMP_OPTIONS:-}
+if [ -z "${DUMP_OPTIONS}" ]; then
+    echo "DUMP_OPTIONS environment variable is not set. No additional options will be passed to pg_dump."
+else
+    echo "DUMP_OPTIONS is set. This test will pass ${DUMP_OPTIONS} to pg_dumpall."
```
hilltracer:
```suggestion
    echo "DUMP_OPTIONS is set. This test will pass ${DUMP_OPTIONS} to pg_dump."
```

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -167,6 +175,87 @@ check_vacuum_worked()
...
 
+dump_database_schema()
+{
+	set -o pipefail
+
+	dump_path="$1"
+	partitions_dump_path="$2"
+	partitions_query="$3"
+	queries_query="$4"
+	pgopts="$5"
```
hilltracer: The function changed many variables: dump_path, partitions_dump_path, partitions_query, queries_query, pgopts, databases_string, partitions_string, queries_string, database, partition, query. And arrays: databases, partitions, queries, args_to_ignore_partitions.

I think better to set it as `local` for safety. As in other functions in file. Consider using `local -a` for arrays.

Comething like:
```sh
local dump_path="$1"
local partitions_dump_path="$2"
local partitions_query="$3"
local queries_query="$4"
local pgopts="$5"

local databases_string
local partitions_string
local queries_string

local database
local partition
local query

local -a databases=()
local -a partitions=()
local -a queries=()
local -a args_to_ignore_partitions=()
```

GirgerS: > I think better to set it as local for safety. As in other functions in file. Consider using local -a for arrays.

I think it makes sense too.

Should be [done](https://github.com/GreengageDB/greengage/pull/536/changes/23e5f7d6f94e43b1d4b4436a76ba61899cb8c78f), hopefully I didn't miss anything

hilltracer (commented):
Test passed with last 6.x regression dump.
https://github.com/GreengageDB/greengage/actions/runs/29922570317/artifacts/8535443934

<details>
<summary>test execution</summary>

```sh
tar xvf ubuntu24.04_postgres_sqldump.tar

mkdir dump_dir
mv ./sqldump/dump.sql ./dump_dir/dump.sql

mkdir logs
cd ~/greengage
git checkout GG-613
git merge GG-531-new

docker build -f ci/Dockerfile.pg_upgrade -t gpdb7_pgupgrade:latest --build-arg GGDB7_IMAGE=gpdb7_u22:latest .
cd ..

docker run \
	-v $(pwd)/dump_dir:/dump_dir \
	-v $(pwd)/logs:/logs \
	gpdb7_pgupgrade bash -c \
	"export SQL_SCHEMA=/dump_dir/dump.sql; \
	 export CLEANUP_SCRIPT=/home/gpadmin/gpdb_src/src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql; \
	 export DUMP_OPTIONS='--data-only --extra-float-digits=-3'; \
	 gpdb_src/concourse/scripts/setup_gpadmin_user.bash; \
	 su gpadmin /home/gpadmin/gpdb_src/ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash; \
	 test_exit_code=\$?; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump1.sql /logs/dump1.sql; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump2.sql /logs/dump2.sql; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump_partitions1.sql /logs/dump_partitions1.sql; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/tmp_check/dump_partitions2.sql /logs/dump_partitions2.sql; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/regression.diffs /logs/regression.diffs; \
	 cp /home/gpadmin/gpdb_src/src/bin/pg_upgrade/partitions_regression.diffs /logs/partitions_regression.diffs; \
	 (exit \$test_exit_code);"
```

</details>

hilltracer (dismissed):

ci/readme.md
bimboterminator1: line 257 currently says schema comparison is impossible. Say it's disabled for now, and why.

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -167,6 +175,99 @@ check_vacuum_worked()
...
+					exit 1
+				fi
+
+				local -a partitions
+				readarray -t partitions <<< ${partitions_string}
+
+				local partition
+				for partition in "${partitions[@]}"; do
+					if [ -z "${partition}" ]; then
+						continue
```
bimboterminator1, line 226: line 228 removes the table from the dump (-T), lines 232-253 put back only the rows. So the definition is never compared. All of this can change unnoticed: AO -> heap, compression dropped, DISTRIBUTED BY -> RANDOMLY, column encodings lost, partition bounds changed.

if someone later removes --data-only, partitions still won't be covered, but will look like they are.

ci/readme.md
GirgerS: > line 257 currently says schema comparison is impossible. Say it's disabled for now, and why.

I've added a part of this PR description to the comment

src/bin/pg_upgrade/test_gpdb.sh
GirgerS: I've mentioned this in the option description

bimboterminator1 (dismissed):

hilltracer (dismissed):

ci/readme.md
```diff
@@ -176,3 +176,101 @@ All tests are run in one way or another on the demo cluster, wherever possible.
...
+- GREENGAGE6_SRC - directory of the Greengage6 source code.
+- GREENGAGE7_SRC - directory of the Greengage7 source code.
+- GREENGAGE6_INSTALLATION - directory where the Greengage6 installation is located (i.e., after make install).
+- GREENGAGE7_INSTALLATION - directory where the Greengage7 installation is located (i.e., after make install).
+- SQL_SCHEMA - optional path to an sql file that will be loaded before performing the upgrade. If it is not specified,
+  the upgrade will be performed on an empty cluster.
+- CLEANUP_SCRIPT - optional path to an sql file that will be executed after loading SQL_SCHEMA, before performing pg_upgrade. It can be used to remove deprecated or otherwise failing objects from SQL_SCHEMA. If it is not specified, no cleanup will occur.
+- DUMP_OPTIONS - optional parameters that will be passed to pg_dump to collect pre- and post-upgrade dumps. These dumps are compared at the end of the test, to determine whether it was successful. If this option is not specified, pg_dump will be executed with no parameters. Please note that partitioned tables are not affected by this option, because they are excluded from the dump and their data is compared seperately.
```
magf:
```suggestion
- DUMP_OPTIONS - optional parameters that will be passed to pg_dump to collect pre- and post-upgrade dumps. These dumps are compared at the end of the test, to determine whether it was successful. If this option is not specified, pg_dump will be executed with no parameters. Please note that partitioned tables are not affected by this option, because they are excluded from the dump and their data is compared separately.
```

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -167,6 +175,99 @@ check_vacuum_worked()
...
+
+		local database
+		for database in "${databases[@]}"; do
+			# We are getting '\' symbol automatically escaped to '\\', convert is back
+			database=$(echo "${database}" | sed 's/\\\\/\\/g')
+
+			# When upgrading Greengage 6 to Greengage 7, partitioned tables would cause
+			# pre- and post-upgrade dumps to differ, because the way they are dumped depends
+			# on the version of the source cluster. So, make pg_dump ignore them and compare
+			# them seperately.
```
magf:
```suggestion
			# them separately.
```

src/bin/pg_upgrade/test_gpdb.sh
```diff
@@ -66,6 +65,15 @@ retain_tempdir=0
 # GPDB_UPGRADE_FIXME: what about the pg_upgrade precheck for upgrade?
 perf_test=0
 
+# This script is run to test pg_upgrade from Greengage 6, to Greengage 7,
+# so perform additional steps to enusure that it completes successfully
```
magf:
```suggestion
# so perform additional steps to ensure that it completes successfully
```

src/bin/pg_upgrade/cleanup_regression_dump_from_6X.sql
```diff
@@ -0,0 +1,144 @@
...
+DROP SCHEMA equal_operator_not_in_search_path_schema CASCADE;
+
+
+-- Incompatible GUC settings
+\c regression
+ALTER DATABASE dsp1 RESET gp_default_storage_options;
+ALTER DATABASE dsp2 RESET gp_default_storage_options;
+
+
+-- The way how tsquery is displayed was chagned between versions, so is may differ between dumps.
```
magf:
```suggestion
-- The way how tsquery is displayed was changed between versions, so is may differ between dumps.
```

magf (changes requested):
Overall, LGTM.

I only suggested a few typo fixes and have one minor comment: using `-v` for the cross-version mode doesn't feel very intuitive, since `-v` is commonly associated with "verbose".

I'd probably use `-x` for the cross-version option instead.

GirgerS:
> I only suggested a few typo fixes and have one minor comment: using -v for the cross-version mode doesn't feel very intuitive, since -v is commonly associated with "verbose".

Makes sense, I've change it

magf (approved):

bimboterminator1 (approved):

hilltracer (approved):

GirgerS:
@andreyaksenov, please review the PR title

andreyaksenov:
> @andreyaksenov, please review the PR title

my suggestion:

```
Allow running the pg_upgrade test on a regression dump
```

GirgerS:
> > @andreyaksenov, please review the PR title
> 
> my suggestion:
> 
> ```
> Allow running the pg_upgrade test on a regression dump
> ```

Thanks!

PR #541, opened 2026-07-24
ADBDEV-7442: Fix error during WAL compression 

riogor:
Sometimes when WAL compression happend, it could produce an error indicating
that destination buffer is too small.
Error occurs because compression produces image, larger than uncompressed one
and larger than destination buffer.
Better way to handle this is to fall back to using uncompressed image.

Return false when it fails to compress WAL, falling back to uncompressed image.
Replace elog(ERROR) with elog(LOG).

No tests because of unpredictable WAL writing and good zstd compression, 
compressing pretty much everything.

Chose cloudberry's patch because it contains more elog messages and is more error-tolerant.

Ticket: ADBDEV-7442

(cherry-picked from commits
apache/cloudberry@55d2cacddbcfa1363a948e82fea3ded8fe247a86,
apache/cloudberry@ec75156498c01d2ef2e228d1782b9b8b6d2c0851)

Patch should be merged with rebase.

### DIFF
--- a/src/backend/access/transam/xloginsert.c
+++ b/src/backend/access/transam/xloginsert.c
@@ -641,6 +641,8 @@ XLogRecordAssemble(RmgrId rmid, uint8 info,
 											cbimg.hole_length,
 											regbuf->compressed_page,
 											&compressed_len);
+				if (!is_compressed)
+					elog(LOG, "WAL compression failed, using uncompressed image");
 			}
 
 			/*
@@ -852,7 +854,10 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 	{
 		cxt = ZSTD_createCCtx();
 		if (!cxt)
-			elog(ERROR, "out of memory");
+		{
+			elog(LOG, "out of memory");
+			return false;
+		}
 	}
 
 	len = ZSTD_compressCCtx(cxt,
@@ -861,8 +866,11 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 							COMPRESS_LEVEL);
 
 	if (ZSTD_isError(len))
-		elog(ERROR, "compression failed: %s uncompressed len %d",
+	{
+		elog(LOG, "compression failed: %s uncompressed len %d",
 			 ZSTD_getErrorName(len), orig_len);
+		len = -1;		/* failure */
+	}
 
 	/*
 	 * We recheck the actual size even if ZSTD reports success and
--- a/src/backend/access/transam/xloginsert.c
+++ b/src/backend/access/transam/xloginsert.c
@@ -641,8 +641,6 @@ XLogRecordAssemble(RmgrId rmid, uint8 info,
 											cbimg.hole_length,
 											regbuf->compressed_page,
 											&compressed_len);
-				if (!is_compressed)
-					elog(LOG, "WAL compression failed, using uncompressed image");
 			}
 
 			/*
--- a/src/backend/access/transam/xloginsert.c
+++ b/src/backend/access/transam/xloginsert.c
@@ -36,6 +36,7 @@
 #ifdef USE_ZSTD
 /* Zstandard library is provided */
 #include <zstd.h>
+#include <zstd_errors.h>
 /* zstandard compression level to use. */
 #define COMPRESS_LEVEL 3
 #endif
@@ -824,6 +825,7 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 {
 #ifdef USE_ZSTD
 	static ZSTD_CCtx  *cxt = NULL;      /* ZSTD compression context */
+	static bool	compression_failure_logged = false;
 	int32		orig_len = BLCKSZ - hole_length;
 	int32		len;
 	int32		extra_bytes = 0;
@@ -853,9 +855,14 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 		cxt = ZSTD_createCCtx();
 		if (!cxt)
 		{
-			elog(LOG, "out of memory");
+			if (!compression_failure_logged)
+			{
+				elog(LOG, "out of memory while allocating ZSTD compression context");
+				compression_failure_logged = true;
+			}
 			return false;
 		}
+		compression_failure_logged = false;
 	}
 
 	len = ZSTD_compressCCtx(cxt,
@@ -865,10 +872,17 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 
 	if (ZSTD_isError(len))
 	{
-		elog(LOG, "compression failed: %s uncompressed len %d",
-			 ZSTD_getErrorName(len), orig_len);
-		len = -1;		/* failure */
+		if (ZSTD_getErrorCode(len) != ZSTD_error_dstSize_tooSmall && !compression_failure_logged)
+		{
+			elog(LOG, "compression failed: %s uncompressed len %d",
+				ZSTD_getErrorName(len), orig_len);
+
+			compression_failure_logged = true;
+		}
+		return false;
 	}
+
+	compression_failure_logged = false;
 
 	/*
 	 * We recheck the actual size even if ZSTD reports success and
--- a/src/backend/access/transam/xloginsert.c
+++ b/src/backend/access/transam/xloginsert.c
@@ -30,6 +30,7 @@
 #include "replication/origin.h"
 #include "storage/bufmgr.h"
 #include "storage/proc.h"
+#include "utils/faultinjector.h"
 #include "utils/memutils.h"
 #include "pg_trace.h"
 
@@ -853,6 +854,15 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 	if (!cxt)
 	{
 		cxt = ZSTD_createCCtx();
+
+#ifdef FAULT_INJECTOR
+		/*
+		 * Forget the context to imitate context creation failure, producing 
+		 * an out of memory log record.
+		 */
+		if (SIMPLE_FAULT_INJECTOR("xlog_compress_backup_block") == FaultInjectorTypeSkip)
+			cxt = NULL;
+#endif
 		if (!cxt)
 		{
 			if (!compression_failure_logged)
@@ -865,8 +875,20 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 		compression_failure_logged = false;
 	}
 
+	size_t dest_capacity = BLCKSZ;
+
+#ifdef FAULT_INJECTOR
+	/*
+	 * Shrink the output buffer so ZSTD_compressCCtx() genuinely returns
+	 * ZSTD_error_dstSize_tooSmall below, regardless of how compressible
+	 * the input actually is.
+	 */
+	if (SIMPLE_FAULT_INJECTOR("xlog_compress_backup_block_dstsize_too_small") == FaultInjectorTypeSkip)
+		dest_capacity = 1;
+#endif
+
 	len = ZSTD_compressCCtx(cxt,
-							dest, BLCKSZ,
+							dest, dest_capacity,
 							source, orig_len,
 							COMPRESS_LEVEL);
 
--- a/src/test/isolation2/expected/xlog_compress_backup_block_dstsize_too_small.out
+++ b/src/test/isolation2/expected/xlog_compress_backup_block_dstsize_too_small.out
@@ -1,0 +1,42 @@
+-- ZSTD_compressCCtx() reporting ZSTD_error_dstSize_tooSmall is a routine
+-- outcome (the page just didn't compress small enough to fit), not a real
+-- failure -- it is expected to happen regularly on ordinary data, not only
+-- on pathological input. Verify XLogCompressBackupBlock() still handles it
+-- gracefully: no crash, and the page falls back to being stored
+-- uncompressed.
+SET wal_compression = on;
+SET
+
+-- Force ZSTD_compressCCtx() to genuinely report dstSize_tooSmall on every
+-- call, on every segment, by shrinking its output buffer.
+SELECT gp_inject_fault_infinite('xlog_compress_backup_block_dstsize_too_small', 'skip', dbid) FROM gp_segment_configuration WHERE content > -1 AND role = 'p';
+ gp_inject_fault_infinite 
+--------------------------
+ Success:                 
+ Success:                 
+ Success:                 
+(3 rows)
+
+CREATE TABLE xlog_compress_backup_block_dstsize_too_small(a int, b text) DISTRIBUTED BY (a);
+CREATE TABLE
+
+INSERT INTO xlog_compress_backup_block_dstsize_too_small SELECT i, repeat('x', 200) FROM generate_series(1, 3000) i;
+INSERT 0 3000
+
+SELECT gp_inject_fault('xlog_compress_backup_block_dstsize_too_small', 'reset', dbid) FROM gp_segment_configuration WHERE content > -1 AND role = 'p';
+ gp_inject_fault 
+-----------------
+ Success:        
+ Success:        
+ Success:        
+(3 rows)
+
+-- Segments must still be up, and every row must have made it in.
+SELECT count(*) FROM xlog_compress_backup_block_dstsize_too_small;
+ count 
+-------
+ 3000  
+(1 row)
+
+DROP TABLE xlog_compress_backup_block_dstsize_too_small;
+DROP TABLE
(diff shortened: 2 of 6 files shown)


RekGRpth:
use `git cherry-pick -x` to auto put commit hash into commit message

RekGRpth:
point in pr desr that merge should be via rebase

riogor:
Done

RekGRpth:
> (cherry picked from commit 55d2cacddbcfa1363a948e82fea3ded8fe247a86)

(cherry picked from commit apache/cloudberry@55d2cac)

> (cherry picked from commit ec75156498c01d2ef2e228d1782b9b8b6d2c0851)

(cherry picked from commit apache/cloudberry@ec75156)

RekGRpth (dismissed):

src/backend/access/transam/xloginsert.c
```diff
@@ -861,8 +864,11 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 							COMPRESS_LEVEL);
 
 	if (ZSTD_isError(len))
-		elog(ERROR, "compression failed: %s uncompressed len %d",
```
bandetto, line 864: I think this LOG call can create a log storm. Destination buffer is too small is an expected compression miss for an incompressible full-page image because the destination capacity is only `BLCKSZ`. With `wal_compression` enabled, one log entry can be emitted for every affected image.

The caller already falls back to the uncompressed image when false is returned. PostgreSQL handles this error silently (https://github.com/postgres/postgres/blob/REL_18_STABLE/src/backend/access/transam/xloginsert.c#L988-L993). Could we set `len = -1` without `calling elog(LOG)`?

riogor: Yeah, good idea.

riogor: Done

riogor:
Also added ticket to commit messages.

bandetto (dismissed):

RekGRpth:
1) The ZSTD_isError error becomes completely silent. Previously, elog(ERROR) would display the specific cause (ZSTD_getErrorName(len), orig_len). Now, this error branch produces no log at all—neither locally nor at the call site. The actual compression failure (not "no ZSTD," but an error within zstd) will become invisible to the operator: WAL will simply start writing full-page images uncompressed, without a single line in the log explaining why.

2) Asymmetry between two error branches for the same function. The failure of ZSTD_createCCtx() is logged (elog(LOG, "out of memory")), while the failure of ZSTD_compressCCtx() (via ZSTD_isError) is not. Both branches mean the same thing to the caller (failed to compress the block), but the diagnostics for them are different for no apparent reason.

2) The elog(LOG, "out of memory") log is unbounded. cxt is a static variable and remains NULL if context creation fails. This means that if memory pressure persists, each subsequent full-page image will attempt to create the context again and write to the log again—potentially causing a flood at a time when the server is already experiencing memory pressure.

3) Error information is lost. Even if the ZSTD_isError crash log needs to be restored later, ZSTD_getErrorName(len) cannot be restored—the error code is overwritten by assigning len = -1, rather than being stored separately.

bandetto:
@RekGRpth tl;dr?

A singular `elog()` can be left near the `len = -1` and made contain `elog(DEBUG1, ...)`. But I'd rather remove all `elog()` calls since I cannot think of a reason these errors would be helpful.

riogor:
`elog(DEBUG1)` (or `DEBUG2`) seems as a tradeoff - it doesn't remove logs completely and usually this shouldn't create log storm (unless user will enable it).

riogor:
@RekGRpth @bandetto what do you think?

RekGRpth:
I propose to make the first two commits pure cherry-picks, and the third commit the rest of our improvements.

RekGRpth:
> `elog(DEBUG1)` (or `DEBUG2`) seems as a tradeoff - it doesn't remove logs completely and usually this shouldn't create log storm (unless user will enable it).

No, it's better to create a LOG level and also provide flood protection in both places, for example, logging no more than once every N seconds, or
```c
static bool
XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
						char *dest, uint16 *dlen)
{
#ifdef USE_ZSTD
	static ZSTD_CCtx  *cxt = NULL;
	static bool	compression_failure_logged = false;
	...

	if (!cxt)
	{
		cxt = ZSTD_createCCtx();
		if (!cxt)
		{
			if (!compression_failure_logged)
			{
				elog(LOG, "out of memory while allocating ZSTD compression context");
				compression_failure_logged = true;
			}
			return false;
		}
	}

	len = ZSTD_compressCCtx(cxt, dest, BLCKSZ, source, orig_len, COMPRESS_LEVEL);

	if (ZSTD_isError(len))
	{
		if (!compression_failure_logged)
		{
			elog(LOG, "WAL page compression failed: %s uncompressed len %d",
				 ZSTD_getErrorName(len), orig_len);
			compression_failure_logged = true;
		}
		return false;
	}

	compression_failure_logged = false;	/* recovered, re-arm the log */
	...
```

riogor:
I think this won't solve issue completely
- Log storm still can occur from `Destination buffer is too small` - it can happen periodically (e.g. every second/third page), bypassing your check.
- And in either way (timer or flag) some messages will be lost, e.g. when zstd error changes, or when context error resolves, but zstd fails to compress.

Maybe a better way is to skip logging error `Destination buffer is too small`, but everything else is still logged as `elog(LOG)`? If that error is the only cause for log storm.

And for context creation errors your fix can be used.

RekGRpth:
> I think this won't solve issue completely
> 
>     * Log storm still can occur from `Destination buffer is too small` - it can happen periodically (e.g. every second/third page), bypassing your check.
> 
>     * And in either way (timer or flag) some messages will be lost, e.g. when zstd error changes, or when context error resolves, but zstd fails to compress.
> 
> 
> Maybe a better way is to skip logging error `Destination buffer is too small`, but everything else is still logged as `elog(LOG)`? If that error is the only cause for log storm.
> 
> And for context creation errors your fix can be used.

Yes, you can try that, but in a separate commit after two pure cherry-picks.

src/backend/access/transam/xloginsert.c
```diff
@@ -853,20 +855,29 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 		cxt = ZSTD_createCCtx();
 		if (!cxt)
 		{
-			elog(LOG, "out of memory");
+			if(!context_failure_logged)
```
RekGRpth:
```suggestion
			if (!context_failure_logged)
```

src/backend/access/transam/xloginsert.c
```diff
@@ -853,20 +855,29 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
...
 	if (ZSTD_isError(len))
 	{
-		elog(LOG, "compression failed: %s uncompressed len %d",
-			 ZSTD_getErrorName(len), orig_len);
+		if(ZSTD_getErrorCode(len) != ZSTD_error_dstSize_tooSmall)
+		{
+			elog(LOG, "compression failed: %s uncompressed len %d",
+				ZSTD_getErrorName(len), orig_len);
+		}
 		len = -1;		/* failure */
```
RekGRpth:
```suggestion
		if (ZSTD_getErrorCode(len) != ZSTD_error_dstSize_tooSmall &&
			!context_failure_logged)
		{
			elog(LOG, "compression failed: %s uncompressed len %d",
				ZSTD_getErrorName(len), orig_len);
			context_failure_logged = true;
		}
		return false;
```

RekGRpth:
How about the test that Claude suggested?
```diff
diff --git a/src/backend/access/transam/xloginsert.c b/src/backend/access/transam/xloginsert.c
index a1f9f44b173..f7bf0894113 100644
--- a/src/backend/access/transam/xloginsert.c
+++ b/src/backend/access/transam/xloginsert.c
@@ -30,6 +30,7 @@
 #include "replication/origin.h"
 #include "storage/bufmgr.h"
 #include "storage/proc.h"
+#include "utils/faultinjector.h"
 #include "utils/memutils.h"
 #include "pg_trace.h"
 
@@ -866,6 +867,18 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 
    context_failure_logged = false;
 
+#ifdef FAULT_INJECTOR
+   if (SIMPLE_FAULT_INJECTOR("xlog_compress_backup_block") == FaultInjectorTypeSkip)
+   {
+       if (!context_failure_logged)
+       {
+           elog(LOG, "out of memory while allocating ZSTD compression context");
+           context_failure_logged = true;
+       }
+       return false;
+   }
+#endif
+
    len = ZSTD_compressCCtx(cxt,
                            dest, BLCKSZ,
                            source, orig_len,
```
src/test/isolation2/sql/xlog_compress_backup_block_fail.sql
```sql
-- XLogCompressBackupBlock() used to elog(ERROR) when ZSTD compression of a
-- full-page image failed. That ERROR happens inside a critical section
-- during WAL record assembly, so it gets promoted to PANIC and crashes the
-- segment. Verify that a compression failure is now handled gracefully:
-- the segment stays up and falls back to storing the page uncompressed.
CREATE TABLE xlog_compress_backup_block_fail(a int, b text) DISTRIBUTED BY (a);

SET wal_compression = on;

-- Force every call to XLogCompressBackupBlock() to fail, on every segment.
SELECT gp_inject_fault_infinite('xlog_compress_backup_block', 'skip', dbid)
    FROM gp_segment_configuration WHERE content > -1 AND role = 'p';

-- Freshly extended pages always need a full-page image on first write, so
-- this reliably drives many calls into XLogCompressBackupBlock() on each
-- segment. Before the fix, this would PANIC the segments.
INSERT INTO xlog_compress_backup_block_fail
    SELECT i, repeat('x', 200) FROM generate_series(1, 3000) i;

SELECT gp_inject_fault('xlog_compress_backup_block', 'reset', dbid)
    FROM gp_segment_configuration WHERE content > -1 AND role = 'p';

-- Segments must still be up, and every row must have made it in.
SELECT count(*) FROM xlog_compress_backup_block_fail;

DROP TABLE xlog_compress_backup_block_fail;
```

RekGRpth:
and another
```diff
diff --git a/src/backend/access/transam/xloginsert.c b/src/backend/access/transam/xloginsert.c
index a1f9f44b173..01d82e3968d 100644
--- a/src/backend/access/transam/xloginsert.c
+++ b/src/backend/access/transam/xloginsert.c
@@ -30,6 +30,7 @@
 #include "replication/origin.h"
 #include "storage/bufmgr.h"
 #include "storage/proc.h"
+#include "utils/faultinjector.h"
 #include "utils/memutils.h"
 #include "pg_trace.h"
 
@@ -866,10 +867,24 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 
    context_failure_logged = false;
 
-   len = ZSTD_compressCCtx(cxt,
-                           dest, BLCKSZ,
-                           source, orig_len,
-                           COMPRESS_LEVEL);
+   {
+       size_t      dest_capacity = BLCKSZ;
+
+#ifdef FAULT_INJECTOR
+       /*
+        * Shrink the output buffer so ZSTD_compressCCtx() genuinely returns
+        * ZSTD_error_dstSize_tooSmall below, regardless of how compressible
+        * the input actually is.
+        */
+       if (SIMPLE_FAULT_INJECTOR("xlog_compress_backup_block_dstsize_too_small") == FaultInjectorTypeSkip)
+           dest_capacity = 1;
+#endif
+
+       len = ZSTD_compressCCtx(cxt,
+                               dest, dest_capacity,
+                               source, orig_len,
+                               COMPRESS_LEVEL);
+   }
 
    if (ZSTD_isError(len))
    {
```
src/test/isolation2/sql/xlog_compress_backup_block_dstsize_too_small.sql
```sql
-- ADBDEV-7442
-- ZSTD_compressCCtx() reporting ZSTD_error_dstSize_tooSmall is a routine
-- outcome (the page just didn't compress small enough to fit), not a real
-- failure -- it is expected to happen regularly on ordinary data, not only
-- on pathological input. Verify XLogCompressBackupBlock() still handles it
-- gracefully: no crash, and the page falls back to being stored
-- uncompressed.
CREATE TABLE xlog_compress_backup_block_dstsize_too_small(a int, b text) DISTRIBUTED BY (a);

SET wal_compression = on;

-- Force ZSTD_compressCCtx() to genuinely report dstSize_tooSmall on every
-- call, on every segment, by shrinking its output buffer.
SELECT gp_inject_fault_infinite('xlog_compress_backup_block_dstsize_too_small', 'skip', dbid)
    FROM gp_segment_configuration WHERE content > -1 AND role = 'p';

-- Freshly extended pages always need a full-page image on first write, so
-- this reliably drives many calls into XLogCompressBackupBlock() on each
-- segment.
INSERT INTO xlog_compress_backup_block_dstsize_too_small
    SELECT i, repeat('x', 200) FROM generate_series(1, 3000) i;

SELECT gp_inject_fault('xlog_compress_backup_block_dstsize_too_small', 'reset', dbid)
    FROM gp_segment_configuration WHERE content > -1 AND role = 'p';

-- Segments must still be up, and every row must have made it in.
SELECT count(*) FROM xlog_compress_backup_block_dstsize_too_small;

DROP TABLE xlog_compress_backup_block_dstsize_too_small;
```

src/backend/access/transam/xloginsert.c
riogor: Fixed

src/backend/access/transam/xloginsert.c
riogor: Why we need `context_failure_logged` flag here? I thought of using it only for `ZSTD_createCCtx()` errors.

riogor:
Tests look interesting, i can add them.

src/backend/access/transam/xloginsert.c
RekGRpth: to avoid log spam

RekGRpth: you may rename `context_failure_logged` to `failure_logged` or add another `error_failure_logged`

src/backend/access/transam/xloginsert.c
```diff
@@ -848,21 +851,62 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
...
-			elog(ERROR, "out of memory");
+		{
+			if (!context_failure_logged)
+			{
+				elog(LOG, "out of memory while allocating ZSTD compression context");
+				context_failure_logged = true;
+			}
+			return false;
+		}
 	}
```
RekGRpth: Is there any way to combine and optimize this? There's a lot of code duplication?

src/backend/access/transam/xloginsert.c
```diff
@@ -848,21 +851,62 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
...
 
-	len = ZSTD_compressCCtx(cxt,
-							dest, BLCKSZ,
-							source, orig_len,
-							COMPRESS_LEVEL);
+	context_failure_logged = false;
+
+	{
```
RekGRpth: this is redundant

src/backend/access/transam/xloginsert.c
riogor: Fair enough, fixed

src/backend/access/transam/xloginsert.c
riogor: Combined two ifs into one

src/backend/access/transam/xloginsert.c
RekGRpth: still actual

src/backend/access/transam/xloginsert.c
```diff
@@ -850,19 +853,56 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
...
 	if (!cxt)
 	{
+#ifdef FAULT_INJECTOR
+		if (SIMPLE_FAULT_INJECTOR("xlog_compress_backup_block") == FaultInjectorTypeSkip)
+			cxt = 0;
+		else
+			cxt = ZSTD_createCCtx();
+#else
 		cxt = ZSTD_createCCtx();
+#endif
```
RekGRpth, line 865:
```suggestion
		cxt = ZSTD_createCCtx();
#ifdef FAULT_INJECTOR
		/* some comment as below */
		if (SIMPLE_FAULT_INJECTOR("xlog_compress_backup_block") == FaultInjectorTypeSkip)
			cxt = NULL;
#endif
```

riogor: Done

riogor:
Also found that in tests `INSERT` doesn't trigger WAL compression, but `CREATE TABLE` does.
Placed table creation below fault injector.

src/backend/access/transam/xloginsert.c
```diff
@@ -851,18 +854,57 @@ XLogCompressBackupBlock(char *page, uint16 hole_offset, uint16 hole_length,
 	if (!cxt)
 	{
 		cxt = ZSTD_createCCtx();
+
+#ifdef FAULT_INJECTOR
+		/*
+		 * Forget the context to imitiate context creation failure, producing
```
RekGRpth:
```suggestion
		 * Forget the context to imitate context creation failure, producing 
```

riogor: Fixed

RekGRpth (approved):

bandetto (approved):

PR #542, opened 2026-07-24
GG-531: Fix array type name collisions that caused restore errors when upgrading from Greengage 6 to 7

VoidZeroNull0:
What happened?
Dumping 6.x with forceful array type creation option leads to type names
collision and otherwise impossible type names in 7.x. When trying to restore
from dump, they lead to errors.

Why it happens?
Array type name creation logic inside pg_dump.c is much simpler and not
consistent with the one in pg_type.c.

How do we fix this?
Let's  generate missing type names not during dump, but during restore. This
allows to get exact type names as on the source cluster. For this purpose, use
makeArrayTypeName() from pg_type.c, also keep assigned type names to prevent
collisions. Keep assigned names inside HTAB.
Also, change the query that is used to get partitioned tables inside 
dumpTableSchema function. With the current approach, when creating a 
partitioned table, it is necessary for binary_upgrade_set_next_pg_type_oid and 
binary_upgrade_set_next_array_pg_type_oid function calls to appear in the same 
order as the partitions in the subsequent CREATE TABLE ... PARTITION BY 
command.
The order of partitions is determined by the get_parts function,
so imitate its logic with RECURSIVE query.

Freeing up shenanigans?
HTAB is allocated inside TopMemoryContext, and not freed intentionally. As
restore creates a backend process with its libpq connection, all its memory
(including HTAB) will be cleared when work is done. The reason for this is that
the moment clear up is needed is not obvious - we can't tell if we don't need
generated names anymore.

Tests?
No auto tests as this is the one of many to fix them. Testing could be done by
manually dumping and looking and resulting type names.

Ticket: GG-531

### DIFF
--- a/src/backend/utils/adt/pg_upgrade_support.c
+++ b/src/backend/utils/adt/pg_upgrade_support.c
@@ -26,7 +26,6 @@
 #include "miscadmin.h"
 #include "utils/array.h"
 #include "utils/builtins.h"
-
 #include "utils/memutils.h"
 #include "utils/syscache.h"
 
--- a/src/backend/utils/adt/pg_upgrade_support.c
+++ b/src/backend/utils/adt/pg_upgrade_support.c
@@ -66,11 +66,10 @@ InitCreatedArrayNamesHash()
 static void
 RememberCreatedName(char *typname, Oid schema)
 {
-	CreatedName key;
+	/* Zero out padding bytes for HASH_BLOBS */
+	CreatedName key = {0};
 	if (created_names == NULL)
 		InitCreatedArrayNamesHash();
-	/* Zero out padding bytes for HASH_BLOBS */
-	MemSet(&key, 0, sizeof(CreatedName));
 	key.schema_oid = schema;
 	strlcpy(key.name, typname, NAMEDATALEN);
 	hash_search(created_names, &key, HASH_ENTER, NULL);
@@ -99,7 +98,7 @@ makeArrayTypeNameUpgrade(const char *typeName, Oid typeNamespace)
 		pfree(arr);
 		arr = new_arr;
 
-		MemSet(&key, 0, sizeof(CreatedName));
+		key = (CreatedName) {0};
 		key.schema_oid = typeNamespace;
 		strlcpy(key.name, arr, NAMEDATALEN);
 		if (!hash_search(created_names, &key, HASH_FIND, NULL))
@@ -127,7 +126,7 @@ makeArrayTypeNameUpgrade(const char *typeName, Oid typeNamespace)
 Datum
 binary_upgrade_set_next_pg_type_oid(PG_FUNCTION_ARGS)
 {
-	CreatedName   key;
+	CreatedName   key = {0};
 	MemoryContext oldctx;
 	Oid			typoid = PG_GETARG_OID(0);
 	Oid			typnamespaceoid = PG_GETARG_OID(1);
@@ -144,7 +143,6 @@ binary_upgrade_set_next_pg_type_oid(PG_FUNCTION_ARGS)
 	in_catalog = OidIsValid(old_type_oid);
 
 	oldctx = MemoryContextSwitchTo(TopMemoryContext);
-	MemSet(&key, 0, sizeof(CreatedName));
 	key.schema_oid = typnamespaceoid;
 	strlcpy(key.name, typname, NAMEDATALEN);
 	if (in_catalog || (created_names && hash_search(created_names, &key, HASH_FIND, NULL)))
--- a/src/backend/utils/adt/pg_upgrade_support.c
+++ b/src/backend/utils/adt/pg_upgrade_support.c
@@ -86,35 +86,40 @@ static char *
 makeArrayTypeNameUpgrade(const char *typeName, Oid typeNamespace)
 {
 	CreatedName key;
-	char *arr, *modifiedTypeName;
-	int underscores = 0;
+	char *arr, *new_arr;
+	int iteration = 0;
 
-	modifiedTypeName = palloc(NAMEDATALEN);
-	arr = makeArrayTypeName(typeName, typeNamespace);
-	MemSet(&key, 0, sizeof(CreatedName));
-	key.schema_oid = typeNamespace;
-	strlcpy(key.name, arr, NAMEDATALEN);
-	while (created_names && hash_search(created_names, &key, HASH_FIND, NULL))
+	/* Make sure that the initial name is allocated by us */
+	arr = palloc(NAMEDATALEN);
+	strlcpy(arr, typeName, NAMEDATALEN);
+
+	while (true)
 	{
-		underscores++;
-		/* 
-		 * We tried to create new name NAMEDATALEN times,
-		 * effectivly meaning that it is impossible to create one.
-		 */
-		if (underscores >= NAMEDATALEN)
-			ereport(ERROR,
-				   (errcode(ERRCODE_DUPLICATE_OBJECT),
-					errmsg("could not form array type name for type \"%s\"",
-							typeName)));
-		MemSet(modifiedTypeName, '_', underscores);
-		strlcpy(modifiedTypeName + underscores, typeName, NAMEDATALEN - underscores);
+		new_arr = makeArrayTypeName(arr, typeNamespace);
 		pfree(arr);
-		arr = makeArrayTypeName(modifiedTypeName, typeNamespace);
+		arr = new_arr;
+
 		MemSet(&key, 0, sizeof(CreatedName));
 		key.schema_oid = typeNamespace;
 		strlcpy(key.name, arr, NAMEDATALEN);
+		if (!hash_search(created_names, &key, HASH_FIND, NULL))
+			break;
+
+		if (iteration >= NAMEDATALEN)
+		{
+			/*
+			 * If we end up with a name that consists entirely of underscores
+			 * ((NAMEDATALEN - 1) underscores, and one character for a null terminator),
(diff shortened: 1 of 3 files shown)


GirgerS:
In case you are wondering why changing [this](https://github.com/GreengageDB/greengage/pull/542/changes#diff-36e3806266f5e832b422f878e1d7fef4ebeea2143b1733ea3fa2f90a99bad5c5R17131) query is necessary:

when getting partitions DDL after [this](https://github.com/GreengageDB/greengage/blob/5503ae42cf807c7fa1361a7339ddb8a5c7f31238/src/bin/pg_dump/pg_dump.c#L7468) call, we end up querying partitions [here](https://github.com/GreengageDB/greengage/blob/88351ee392923b8198624c7aa0c3d0cc4d41e2f6/src/backend/cdb/cdbpartition.c#L2523-L2637), and
`systable_beginscan` seems to return rows sorted on the columns of the indices used (PartitionParrelidParlevelParistemplateIndexId, PartitionRuleParoidParparentruleParruleordIndexId).

My best guess on why this can happend is that index scans on `bree` indices are sorted [by default](https://www.postgresql.org/docs/current/indexes-ordering.html). This further confirms by the fact that `systable_beginscan_ordered` has essentially the same code, minus that it gives an error when an index cannot be used. ([systable_beginscan](https://github.com/GreengageDB/greengage/blob/88351ee392923b8198624c7aa0c3d0cc4d41e2f6/src/backend/access/index/genam.c#L312-L395), [systable_beginscan_ordered](https://github.com/GreengageDB/greengage/blob/88351ee392923b8198624c7aa0c3d0cc4d41e2f6/src/backend/access/index/genam.c#L506-L581))

Of course, the given query is not a direct copy of the scans above, but I decided not to change the proposed query without further discussion, as it seems to work either way.

src/backend/utils/adt/pg_upgrade_support.c
```diff
@@ -38,14 +40,136 @@ do {															\
...
+	arr = palloc(NAMEDATALEN);
+	strlcpy(arr, typeName, NAMEDATALEN);
+
+	while (true)
+	{
+		new_arr = makeArrayTypeName(arr, typeNamespace);
+		pfree(arr);
+		arr = new_arr;
+
+		MemSet(&key, 0, sizeof(CreatedName));
```
red1452: I think, it is better to do `MemSet(&key, 0, sizeof(key)); `. In the future type of this variable may be changed, and who will do it may forget to change "memset".

src/backend/utils/adt/pg_upgrade_support.c
```diff
@@ -38,14 +40,136 @@ do {															\
...
+}
+
+static void
+RememberCreatedName(char *typname, Oid schema)
+{
+	CreatedName key;
+	if (created_names == NULL)
+		InitCreatedArrayNamesHash();
+	/* Zero out padding bytes for HASH_BLOBS */
+	MemSet(&key, 0, sizeof(CreatedName));
```
red1452: Here, you may change two lines:
```
CreatedName key;
MemSet(&key, 0, sizeof(CreatedName));
```
by:
```
CreatedName key = {0};
```

src/backend/utils/adt/pg_upgrade_support.c
```diff
@@ -38,14 +40,136 @@ do {															\
...
+
+	old_type_oid = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid,
+								   CStringGetDatum(typname),
+								   ObjectIdGetDatum(typnamespaceoid));
+	in_catalog = OidIsValid(old_type_oid);
+
+	oldctx = MemoryContextSwitchTo(TopMemoryContext);
+	MemSet(&key, 0, sizeof(CreatedName));
+	key.schema_oid = typnamespaceoid;
+	strlcpy(key.name, typname, NAMEDATALEN);
```
red1452, line 147: what about use `sizeof(key.name)` instead of `NAMEDATALEN`?

red1452 (commented):
Add note to the PR's description about modifing query at pg_dump.

src/backend/utils/adt/pg_upgrade_support.c
GirgerS: I can't think of a reason for not using `NAMEDATALEN` as a key length, so I doubt it would end up changing here.

src/backend/utils/adt/pg_upgrade_support.c
GirgerS: Good idea, I like this approach more

src/backend/utils/adt/pg_upgrade_support.c
GirgerS: I'd say that `key = (CreatedName) {0};` is more consistent with other change.

red1452 (dismissed):

src/backend/utils/adt/pg_upgrade_support.c
```diff
@@ -38,14 +40,134 @@ do {															\
...
+	while (true)
+	{
+		new_arr = makeArrayTypeName(arr, typeNamespace);
+		pfree(arr);
+		arr = new_arr;
+
+		key = (CreatedName) {0};
+		key.schema_oid = typeNamespace;
+		strlcpy(key.name, arr, NAMEDATALEN);
+		if (!hash_search(created_names, &key, HASH_FIND, NULL))
```
bimboterminator1: in `in_catalog` case the `makeArrayTypeNameUpgrade` may be called with NULL created names. Prabably it's unreachable in practice, but since functions are sql callable it'd be better to guard such cases (likely with laze initialization).

src/backend/utils/adt/pg_upgrade_support.c
```diff
@@ -38,14 +40,134 @@ do {															\
...
+		{
+			/*
+			 * If we end up with a name that consists entirely of underscores
+			 * ((NAMEDATALEN - 1) underscores, and one character for a null terminator),
+			 * and it is not present in the catalog, makeArrayTypeName will return
+			 * this name over and over again, leaving us stack in an infinite loop.
+			 *
+			 * In such case, abort when we know that we've seen every possible name.
+			 */
+			break;
```
bimboterminator1, line 117: and what name do we return in this case?

src/backend/utils/adt/pg_upgrade_support.c
```diff
@@ -38,14 +40,134 @@ do {															\
...
+		 */
+		moved_array_typname = makeArrayTypeNameUpgrade(typname, typnamespaceoid);
+		RememberCreatedName(moved_array_typname, typnamespaceoid);
+	}
+
+	/* Remember that we've already taken this name */
+	RememberCreatedName(typname, typnamespaceoid);
+	MemoryContextSwitchTo(oldctx);
+
 	AddPreassignedOidFromBinaryUpgrade(typoid, TypeRelationId, typname,
```
bimboterminator1, line 171: What if some segments derive different names?

src/backend/utils/adt/pg_upgrade_support.c
GirgerS: > Prabably it's unreachable in practice

In this case, we need a type to collide with, and it would likely be already stored in the hashmap. But let's be safe, it never hurts.

(and also, these functions all callable only when the server is started in binary upgrade mode, which is far from regular usage 
```
postgres=# SELECT binary_upgrade_set_next_pg_type_oid(1, 2, '3');
ERROR:  function can only be called when server is in binary upgrade mode
```
)

src/backend/utils/adt/pg_upgrade_support.c
GirgerS: I didn't mention it, but it doesn't matter.

This is the cornercase when partitioned table from 6.x to 7.x, because
there are not enough names after we add array types missing from 6.x to the catalog.

Consider upgrading the following schema, it should error out (but not end up being stuck in an infinite loop):
```
CREATE table test (a int)
PARTITION BY RANGE (a) (
        PARTITION test1 START (8)  END (9)  WITH (tablename='______________________________________________________________x'),
        PARTITION test2 START (10) END (11) WITH (tablename='_____________________________________________________________xx')
);
```
We [decided](https://github.com/GreengageDB/greengage/pull/518#issuecomment-4894065286) to fix it separately, because this corner case is not present in the regression dump.

src/backend/utils/adt/pg_upgrade_support.c
GirgerS: [Schema is restored only on the coordinator](https://github.com/GreengageDB/greengage/blob/322b80f908aecd14d60d2bff297e796595a37182/src/bin/pg_upgrade/pg_upgrade.c#L237-L242), and then it is copied to the segments. So we dodged this bullet.

GirgerS:
Also, what an interesting test [failure](https://github.com/GreengageDB/greengage/actions/runs/31005366604/job/92303852647?pr=542).

bimboterminator1 (approved):

red1452 (approved):

andreyaksenov:
Release note suggestion:

```
Fixed array type name collisions that caused restore errors when upgrading from Greengage 6 to 7
```

GirgerS:
> Release note suggestion:
> 
> ```
> Fixed array type name collisions that caused restore errors when upgrading from Greengage 6 to 7
> ```

I've changed the title.

Also, if I remember remember correctly, we are reviewing only the title of the PR, right?

andreyaksenov:
> Also, if I remember remember correctly, we are reviewing only the title of the PR, right?

yep - only the info that goes to RN

PR #451, opened 2026-05-21
In-Memory Catalog for Temporary Tables

e-ivkov:
See motivation behind this change: https://wiki.yandex.ru/homepage/r--d/greengage/adr-fast-temp-tables/

This PR introduces in-memory catalog for temporary table metadata.

Temporary table metadata (pg_class rows, pg_attribute rows, index entries, etc.) is stored entirely in backend-private memory instead of on-disk pg_catalog tables.  Virtual tuples are mixed with regular on-disk tuples during system catalog scans so the rest of the system sees a unified view. 

Because temporary tables are visible only within a single session, there is no need for shared memory, locks, or MVCC bookkeeping (xmin/xmax). Transaction support uses a simple snapshot stack: each transaction or subtransaction pushes a copy of the current state, and commit/abort pops or discards it.

This solves the pg_catalog bloating problem caused by applications that create and drop many temporary tables.  Without this, every temp table leaves dead tuples in shared catalog tables, triggering expensive autovacuum runs that affect the entire database.

The feature is gated by the GUC gp_enable_temp_memory_catalog.  When enabled, DDL functions wrap temp-table operations in BEGIN_TEMP_TABLE_SCOPE / END_TEMP_TABLE_SCOPE (see tempcat.h), and the catalog DML layer (CatalogTupleInsert, etc.) redirects writes to this module.  System catalog scans (systable_beginscan, etc.) merge virtual tuples from tempcat alongside regular on-disk tuples.

PR #543, opened 2026-07-27
Avoid 2PC when only locks are present in WAL

e-ivkov:
Initially introduced in: https://github.com/GreengageDB/greengage/pull/451
And tempcat will hugely benefit from this optimization

## Rationale

`CREATE TEMP TABLE + INSERT` transaction with tempcat on emits exactly four records: three XLOG_STANDBY_LOCK (table, toast table, toast index) and the commit record. Zero with block references.

Those standby-lock records are the ones that were flipping didLogXid and forcing 2PC, and they can't be suppressed cheaply:

- LockAcquireExtended() decides to log from the locktag alone (lock.c:955) - just (dbid, reloid). It has no way to know the relation is temporary.
- On CREATE, there is no relation yet at all: heap_create_with_catalog() locks a freshly allocated OID at heap.c:1578, before any catalog tuple or relcache entry exists - and under tempcat a shared-catalog row never appears.
- Everywhere else, persistence lives behind the lock; looking it up first inverts PG's lock-then-lookup ordering in the hottest lock-manager path.

Plus the commit record is unavoidable regardless (temp tuples need a real XID), so it is not possible to write no WAL.

But "wrote any WAL" was being used as a proxy for "did durable work needing a distributed prepare (2PC)", and tempcat decoupled the two. Hence fixing the predicate (max_registered_block_id > 0) rather than trying to write zero WAL.

### DIFF
--- a/src/backend/tcop/dest.c
+++ b/src/backend/tcop/dest.c
@@ -266,8 +266,18 @@ ReadyForQuery(CommandDest dest)
 					pq_sendint64(&buf, VmemTracker_GetMaxReservedVmemBytes());
 					pq_endmessage(&buf);
 
+					/*
+					 * Tell the dispatcher whether this QE did durable work that
+					 * requires a two-phase commit.  We report whether the
+					 * transaction wrote WAL for a permanent relation rather than
+					 * whether it wrote any WAL at all: a transaction that only
+					 * touched temporary objects (e.g. temp tables via the in-memory
+					 * catalog) still emits lock/invalidation WAL, but that does not
+					 * need a two-phase commit, so the dispatcher can use a cheaper
+					 * one-phase commit.
+					 */
 					pq_beginmessage(&buf, 'x'); /* wrote_xlog */
-					pq_sendbyte(&buf, TransactionDidWriteXLog());
+					pq_sendbyte(&buf, TransactionWroteWalForPermanentRel());
 					pq_endmessage(&buf);
 				}
 
--- a/src/backend/access/transam/xact.c
+++ b/src/backend/access/transam/xact.c
@@ -148,6 +148,20 @@ int32 gp_subtrans_warn_limit = 16777216; /* 16 million */
 bool		seqXlogWrite;
 
 /*
+ * Set when the current transaction emits a WAL record that modifies a
+ * permanent relation, i.e. a record carrying registered block references.
+ * Temporary and unlogged relations never WAL-log their data pages
+ * (RelationNeedsWAL is false for them), so such records always pertain to a
+ * permanent relation and mean the transaction did durable work that requires a
+ * distributed two-phase commit.  A transaction that only emits
+ * lock/invalidation/commit records -- e.g. one that touches only temporary
+ * tables through the in-memory catalog -- leaves this false and may commit with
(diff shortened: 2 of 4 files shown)


Stolb27:
> Doesn't make sense without: https://github.com/GreengageDB/greengage/pull/451

In my opinion, this statement is incorrect. As we discussed earlier, the problem may be reproduced without tempcat at all with:

```
BEGIN;
LOCK TABLE ON AEL;
SELECT
COMMIT;
```

Such readonly transactions requires 2PC too now. Am I correct?

@e-ivkov

e-ivkov:
> > Doesn't make sense without: #451
> 
> In my opinion, this statement is incorrect. As we discussed earlier, the problem may be reproduced without tempcat at all with:
> 
> ```
> BEGIN;
> LOCK TABLE ON AEL;
> SELECT
> COMMIT;
> ```
> 
> Such readonly transactions requires 2PC too now. Am I correct?
> 
> @e-ivkov

Yes, you are correct, in this case the optimization will work and use 1 phase commit

e-ivkov:
Updated PR description

RekGRpth:
> Updated PR description

Try to avoid the words `tempcat` and `in-memory catalog` in the PR description and comments, as the patch is not directly related to them.

src/backend/tcop/dest.c
```diff
@@ -266,8 +266,18 @@ ReadyForQuery(CommandDest dest)
...
+					 * transaction wrote WAL for a permanent relation rather than
+					 * whether it wrote any WAL at all: a transaction that only
+					 * touched temporary objects (e.g. temp tables via the in-memory
+					 * catalog) still emits lock/invalidation WAL, but that does not
+					 * need a two-phase commit, so the dispatcher can use a cheaper
+					 * one-phase commit.
+					 */
 					pq_beginmessage(&buf, 'x'); /* wrote_xlog */
-					pq_sendbyte(&buf, TransactionDidWriteXLog());
+					pq_sendbyte(&buf, TransactionWroteWalForPermanentRel());
```
RekGRpth, line 280: The `TransactionDidWriteXLog` function is being deprecated in core, but third-party extensions can still rely on it. How good is this?

RekGRpth:
Could you add tests:

1) A transaction with only temporary tables is truly 1PC.
2) A mixed transaction (temporary + regular table, AO table) is still 2PC.
3) ANALYZE temp-table

src/backend/tcop/dest.c
Stolb27: @RekGRpth, what's the problem here? TransactionDidWriteXLog still reports if xact writes to WAL anything (including lockings info). TransactionWroteWalForPermanentRel should report only (and only) if a permanent relation is modified.

Stolb27:
> A transaction with only temporary tables is truly 1PC.

@RekGRpth, Even temp tables are persisted to the catalog tables, so they still requires 2PC with or without this patch.

bimboterminator1:
@e-ivkov Thanks for your contribution! This optimization is quite valuable. We enhanced a little bit xloginsert condition for covering cases when registered block is modified but with no  2pc need. **Preserved your authorship in #572**

bimboterminator1:
Closed in favor of #572

PR #471, opened 2026-06-01
GG-530: Fix resource group auxiliary tools for cgroup v2 tests

hilltracer:
Calling check_cgroup_io_max() could cause SIGSEGV because ctypes used the
default int return type for get_tablespace_path(), but the C function returns
char *.

Set get_tablespace_path.restype to ctypes.c_char_p.

PR #544, opened 2026-07-28
GG-615: Fix handling custom cgroup v2 path for resource groups

GirgerS:
When changing a guc via gp_config, it may perform its own 
validations, in addition to the ones in the cluster. 

Such check for gp_resource_manager guc
wasn't properly updated to reflect changes made in
ffee673bd734cd4811e060fcf83595642fcf8e7b
making it verify hardcoded /sys/fs/cgroup/gpdb path,
instead of the actual one, specified via 
gp_resource_group_cgroup_parent guc.

Update the check to use gp_resource_group_cgroup_parent.

### DIFF
--- a/gpMgmt/bin/gpcheckresgroupv2impl
+++ b/gpMgmt/bin/gpcheckresgroupv2impl
@@ -86,7 +86,7 @@ class CgroupValidationVersionTwo(CgroupValidation):
 
 
 if __name__ == '__main__':
-    # We should always be given exactly one arg
+    assert len(sys.argv) == 2, "Internal error: this script should be given exactly one argument."
     cgroup_parent = sys.argv[1]
 
     try:
--- a/gpMgmt/bin/gpcheckresgroupv2impl
+++ b/gpMgmt/bin/gpcheckresgroupv2impl
@@ -2,6 +2,7 @@
 # Copyright (c) 2017, VMware, Inc. or its affiliates.
 
 import os
+import sys
 from functools import reduce
 
 
@@ -30,7 +31,7 @@ class CgroupValidationVersionTwo(CgroupValidation):
         self.mount_point = self.detect_cgroup_mount_point()
         self.tab = {"r": os.R_OK, "w": os.W_OK, "x": os.X_OK, "f": os.F_OK}
 
-    def validate_all(self):
+    def validate_all(self, cgroup_parent):
         """
         Check the permissions of the toplevel gpdb cgroup dirs.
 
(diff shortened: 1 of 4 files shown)


GirgerS:
The way how to test these changes is described [here](https://github.com/GreengageDB/greengage/pull/471#issuecomment-4591968773), with one small difference: don't create `/sys/fs/cgroup/gpdb.service` directory, `gpconfig -c gp_resource_manager -v group-v2` should correctly verify only `/sys/fs/cgroup/gpdb` because we [change](https://github.com/GreengageDB/greengage/pull/544/changes#diff-21440cc7642a7f9bfa8248b969be3abbeac836f3327c4dcc0f5eebc4e1afd938R23) the guc earlier.

Also, consider following `Steps to reproduce` in the parent ticket, the error should go away.

src/test/isolation2/sql/resgroup/resgroup_auxiliary_tools_v2.sql
```diff
@@ -20,8 +20,9 @@ CREATE LANGUAGE plpython3u;
 
 -- enable resource group and restart cluster.
 -- start_ignore
+! gpconfig -c gp_resource_group_cgroup_parent -v "gpdb";
+! gpstop -rai;
```
bimboterminator1: is it necessary to restart the cluster? gpstop -u is not enough for the test?

gpMgmt/bin/gpcheckresgroupv2impl
```diff
@@ -85,7 +86,10 @@ class CgroupValidationVersionTwo(CgroupValidation):
 
 
 if __name__ == '__main__':
+    # We should always be given exactly one arg
```
red1452: If it is, what about adding check here?

src/test/isolation2/sql/resgroup/resgroup_auxiliary_tools_v2.sql
```diff
@@ -20,8 +20,9 @@ CREATE LANGUAGE plpython3u;
...
 ! gpconfig -c gp_resource_manager -v group-v2;
-! gpconfig -c gp_resource_group_cgroup_parent -v "gpdb"
 ! gpconfig -c max_connections -v 250 -m 25;
 ! gpconfig -c runaway_detector_activation_percent -v 100;
 ! gpstop -rai;
```
red1452, line 28: Does it not enought to do restart just here?

gpMgmt/bin/gpconfig
```diff
@@ -167,9 +167,18 @@ class Guc:
                 if msg is not None:
                     return msg
             elif newval == "'group-v2'":
-                msg = GpResGroup().validate_v2()
-                if msg is not None:
-                    return msg
+                try:
+                    dburl = dbconn.DbURL()
+                    conn = dbconn.connect(dburl, True)
```
bimboterminator1: Guc.validate() is reached only from `validate_change_options(options, conn, guc)`, which already holds an open connection and just used it to fetch this GUC out of pg_settings. Passing that `conn` down (or reading gp_resource_group_cgroup_parent alongside the existing lookup) is better:
- no redundant connect/close
- the DatabaseError handler below becomes unnecessary

src/test/isolation2/sql/resgroup/resgroup_auxiliary_tools_v2.sql
bimboterminator1: Seems to be so, add the one line comment about that

GirgerS: > is it necessary to restart the cluster

As you already verified, it is.

But it took me some time to understand why. The `gp_resource_group_cgroup_parent` is [marked](https://github.com/GreengageDB/greengage/commit/ffee673bd734cd4811e060fcf83595642fcf8e7b#diff-3d03368dae4bd5ea009d6cdc0ce2ddc4620e2b776429532d32fde1c3ba63c515R4580) as `PGC_POSTMASTER`, not `PGC_SIGHUP`, so it wouldn't be reloaded on `gpstop -u` or `pg_ctl reload`.

src/test/isolation2/sql/resgroup/resgroup_auxiliary_tools_v2.sql
GirgerS: Not really. After the patch, `gpconfig -c gp_resource_manager -v group-v2;` verifies cgroup directory based on the value of `gp_resource_group_cgroup_parent` (which defaults to `gpdb.service`, while we have `gpdb`), so we need restart to make sure that this guc has the proper value.

gpMgmt/bin/gpconfig
GirgerS: Agreed, it seems better this way.

gpMgmt/bin/gpcheckresgroupv2impl
GirgerS: The default exception that Python gives seems good enough to me:
```
20260730:17:41:14:255548 gpconfig:fourier:regrig-[CRITICAL]:-new GUC value failed validation: [localhost:Traceback (most recent call last):
  File "/home/regrig/greengage/7.x/bin/gpcheckresgroupv2impl", line 90, in <module>
    cgroup_parent = sys.argv[1]
                    ~~~~~~~~^^^
IndexError: list index out of range],[fourier:Traceback (most recent call last):
  File "/home/regrig/greengage/7.x/bin/gpcheckresgroupv2impl", line 90, in <module>
    cgroup_parent = sys.argv[1]
                    ~~~~~~~~^^^
IndexError: list index out of range]
new GUC value failed validation: [localhost:Traceback (most recent call last):
  File "/home/regrig/greengage/7.x/bin/gpcheckresgroupv2impl", line 90, in <module>
    cgroup_parent = sys.argv[1]
                    ~~~~~~~~^^^
IndexError: list index out of range],[fourier:Traceback (most recent call last):
  File "/home/regrig/greengage/7.x/bin/gpcheckresgroupv2impl", line 90, in <module>
    cgroup_parent = sys.argv[1]
                    ~~~~~~~~^^^
IndexError: list index out of range]
```

But is still added an assert (because this script shouldn't be run by the user), I guess that the new error is better: 

```
20260730:17:46:47:265985 gpconfig:fourier:regrig-[CRITICAL]:-new GUC value failed validation: [localhost:Traceback (most recent call last):
  File "/home/regrig/greengage/7.x/bin/gpcheckresgroupv2impl", line 89, in <module>
    assert len(sys.argv) == 2, "Internal error: this script should be given exactly one argument."
           ^^^^^^^^^^^^^^^^^^
AssertionError: Internal error: this script should be given exactly one argument.],[fourier:Traceback (most recent call last):
  File "/home/regrig/greengage/7.x/bin/gpcheckresgroupv2impl", line 89, in <module>
    assert len(sys.argv) == 2, "Internal error: this script should be given exactly one argument."
           ^^^^^^^^^^^^^^^^^^
AssertionError: Internal error: this script should be given exactly one argument.]
```

bimboterminator1 (approved):

red1452 (approved):

PR #546, opened 2026-07-29
Keep no-match rows when pulling up a correlated aggregate subquery

Alena0704:
Keep no-match rows when pulling up a correlated aggregate subquery
With the Postgres planner (optimizer=off or an ORCA fallback), a correlated scalar subquery with an aggregate is pulled up into an INNER join with a grouped subquery (convert_EXPR_to_join), which drops outer rows that have no match. The original subquery keeps them: it computes the aggregate over empty input, so e.g. COUNT yields 0 there:

	select ... from t1
		where t1.a > (select count(*) from t2 where t2.a = t1.d);

A row with no match in t2 must be compared as "t1.a > 0" and can pass, but the INNER join dropped it.

To fix this, pull the subquery up into a LEFT join, so no-match rows survive as null-extended rows, and rewrite the comparison to return the same value the subquery would:

    outer OP CASE WHEN match_flag THEN expr ELSE empty_input_default END

match_flag is a constant TRUE column added to the subquery. For a matched row the CASE returns the real expression; for a null-extended row the flag is NULL and the CASE returns the empty-input default (0 for COUNT, NULL for other aggregates).

The comparison runs above the LEFT join as a filter, not as the join condition: as a join qual it would null-extend matched rows that fail it, and the default would let them back in.

The LEFT join is not always needed. If a no-match row cannot pass the comparison anyway -- e.g. "1 = (select count(*) ...)" turns into "1 = 0" for it -- dropping it is fine and the INNER join is kept as before. This is detected by substituting the empty-input default into the comparison and constant-folding it. Ordinary sum/avg/min/max comparisons fall into this group: their empty-input value is NULL, and a comparison with NULL does not pass, so those plans do not change.

If the comparison cannot be placed above the join (the sublink is in an outer join's ON clause) or the subquery's targetlist is correlated, the pull-up bails out and the sublink runs as a SubPlan, as before.

Adapted from open-gpdb/gpdb#397.
Co-Authored-By: excaliiibur <excaliiibur@foxmail.com>

### DIFF
--- a/src/backend/cdb/cdbsubselect.c
+++ b/src/backend/cdb/cdbsubselect.c
@@ -17,6 +17,7 @@
 #include "access/htup_details.h"
 #include "access/skey.h"
 #include "catalog/pg_operator.h"
+#include "catalog/pg_proc.h"
 #include "catalog/pg_type.h"
 #include "nodes/makefuncs.h"
 #include "optimizer/clauses.h"
@@ -42,6 +43,11 @@ static JoinExpr *make_join_expr(Node *larg, int r_rtindex, int join_type);
 static Node *make_lasj_quals(PlannerInfo *root, SubLink *sublink, int subquery_indx);
 
 static Node *add_null_match_clause(Node *clause);
+static Expr *build_match_flag_case_expr(Var *flagVar, Var *aggVar, Expr *defaultExpr);
+static Expr *build_empty_input_default_expr(Node *expr, bool *sawNonNullDefault);
+static Node *replace_agg_with_empty_default_mutator(Node *node, void *context);
+static bool no_match_row_survives(PlannerInfo *root, OpExpr *opexp,
+								  Expr *defaultExpr);
 
 typedef struct NonNullableVarsContext
 {
@@ -576,6 +582,26 @@ safe_to_convert_EXPR(SubLink *sublink, ConvertSubqueryToJoinContext *ctx1)
 	if (list_length(subselect->targetList) != 1)
 		return false;
 
+	/*
+	 * A window function cannot survive the pull-up.  Without it the subquery
+	 * has a plain aggregate and so produces exactly one row per outer row, and
+	 * the window runs over that single row.  The pulled-up subquery is grouped
+	 * by the correlation columns, so the same window would run over every
+	 * group at once and compute a different value.  (The rewrite of the
+	 * comparison below would also copy the WindowFunc into a qual above the
+	 * join, where there is no WindowAgg node to evaluate it.)
+	 */
+	if (subselect->hasWindowFuncs)
+		return false;
+
+	/**
+	 * Correlation in the targetlist cannot be handled: the pulled-up
+	 * expression (and the empty-input default derived from it) would carry
+	 * upper-level Vars out of the subquery.
+	 */
+	if (contain_vars_of_level_or_above((Node *) subselect->targetList, 1))
+		return false;
+
 
 	/**
 	 * Walk the quals of the subquery to do a more fine grained check as to whether this subquery
@@ -639,6 +665,72 @@ convert_EXPR_to_join(PlannerInfo *root, OpExpr *opexp)
 
 		subselect->jointree->quals = ctx1.innerQual;
 
+		/*
+		 * An INNER join drops outer rows that have no matching inner
+		 * rows.  Without the pull-up they are kept: the subquery
+		 * computes its expression over empty input (COUNT = 0, other
+		 * aggregates NULL) and the comparison may still pass.
+		 *
+		 * So plug the empty-input value into the comparison and run
+		 * eval_const_expressions() on it.  FALSE or NULL means no-match
+		 * rows cannot pass and the INNER join is correct; otherwise use
+		 * a LEFT join to keep them.
+		 */
+		Expr	   *defaultExpr;
+		TargetEntry *flagTLE = NULL;
+		bool		use_left_join;
+		bool		sawNonNullDefault;
+
+		defaultExpr = build_empty_input_default_expr((Node *) origSubqueryTLE->expr,
+													&sawNonNullDefault);
+
+		/*
+		 * defaultExpr ends up in a qual of the outer query, where the planner
+		 * folds constant expressions at plan time.  Substituting 0 for a COUNT
+		 * can turn a subexpression the original query only ever evaluated per
+		 * row into a constant one: "1/count(*)" becomes "1/0" and raises
+		 * "division by zero" while planning, even for a query that returns no
+		 * rows at all.  A NULL default cannot do that -- it only propagates
+		 * through strict functions, which the planner folds without calling
+		 * them, and anything it does evaluate was already constant in the
+		 * original expression.
+		 *
+		 * So take the LEFT-join path only when the invented value is a plain
+		 * constant; otherwise leave the sublink to be planned as a SubPlan,
+		 * which keeps the original semantics.
+		 */
+		if (sawNonNullDefault && !IsA(defaultExpr, Const))
+			return NULL;
+
+		use_left_join = no_match_row_survives(root, opexp, defaultExpr);
+
+		if (use_left_join)
+		{
+			/*
+			 * After the LEFT join the expression column is NULL both for a
+			 * no-match row and for a matched group whose expression is
+			 * genuinely NULL.  To tell them apart, add a constant-TRUE
+			 * match-flag column to the subquery: it can be NULL only when
+			 * the LEFT join found no match and filled the subquery's
+			 * columns with NULLs.
+			 *
+			 * The flag goes BEFORE the expression column: with this
+			 * order the planner can drop the SubqueryScan node from the
+			 * plan.
+			 */
+			TargetEntry *aggTLE = (TargetEntry *) llast(subselect->targetList);
+
+			flagTLE = makeTargetEntry((Expr *) makeBoolConst(true, false),
+									  aggTLE->resno,
+									  pstrdup("csq_count_flag"),
+									  false);
+			aggTLE->resno++;
+			subselect->targetList = list_truncate(subselect->targetList,
+												  list_length(subselect->targetList) - 1);
+			subselect->targetList = lappend(subselect->targetList, flagTLE);
+			subselect->targetList = lappend(subselect->targetList, aggTLE);
+		}
+
 		/**
 		 * Construct a new range table entry for the new pulled up subquery.
 		 */
@@ -660,7 +752,8 @@ convert_EXPR_to_join(PlannerInfo *root, OpExpr *opexp)
 
 		join_expr->quals = joinQual;
 
-		TargetEntry *subselectAggTLE = (TargetEntry *) list_nth(subselect->targetList, list_length(subselect->targetList) - 1);
+		/* The pulled-up expression column is last in either layout. */
+		TargetEntry *subselectAggTLE = (TargetEntry *) llast(subselect->targetList);
 
 		/**
 		 *	modify the op expr to involve the column that has the computed aggregate that needs to compared.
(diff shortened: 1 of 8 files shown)


Alena0704:
An issue without this commit is:
```
postgres=# drop table t_out;
DROP TABLE
postgres=# drop table t_in;
DROP TABLE
postgres=# create table t_out as select 1 as a distributed by (a);                                          
SELECT 1
postgres=# create table t_in  as select 2 as a distributed by (a);                                          
SELECT 1
postgres=# set optimizer=0;                                                                                 
SET
postgres=# select * from t_out where a > (select count(*) from t_in where t_in.a = t_out.a);
 a 
---
(0 rows)

postgres=# set optimizer=1;
SET
postgres=# select * from t_outwhere a > (select count(*) from t_in where t_in.a = t_out.a);
 a 
---
 1
(1 row)
```
As you see the Postgres planner returns incorrect result.

red1452:
Thanks for contributing!
Do you research: why Postgresql generates plan with subplan, while Greengage uses Hash join?
```
postgres=# select * from t_out where a > (select count(*) from t_in where t_in.a = t_out.a);
 a 
---
 1
(1 row)

postgres=# explain (costs off, verbose, analyze) select * from t_out where a > (select count(*) from t_in where t_in.a = t_out.a);
                                   QUERY PLAN                                    
---------------------------------------------------------------------------------
 Seq Scan on public.t_out (actual time=0.042..0.045 rows=1 loops=1)
   Output: t_out.a
   Filter: (t_out.a > (SubPlan 1))
   SubPlan 1
     ->  Aggregate (actual time=0.017..0.018 rows=1 loops=1)
           Output: count(*)
           ->  Seq Scan on public.t_in (actual time=0.011..0.011 rows=0 loops=1)
                 Output: t_in.a
                 Filter: (t_in.a = t_out.a)
                 Rows Removed by Filter: 1
 Planning Time: 0.229 ms
 Execution Time: 0.134 ms
(12 rows)

```
There is error at the test:
```diff
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/regress/expected/subselect_gp.out /home/gpadmin/gpdb_src/src/test/regress/results/subselect_gp.out
--- /home/gpadmin/gpdb_src/src/test/regress/expected/subselect_gp.out	2026-07-30 06:50:37.480701798 +0000
+++ /home/gpadmin/gpdb_src/src/test/regress/results/subselect_gp.out	2026-07-30 06:50:37.583698198 +0000
@@ -953,22 +957,22 @@
     on t1.a > (select count(*) from t_in t2 where t2.a = t1.a);
 QUERY PLAN
 ___________
- Aggregate
+ Finalize Aggregate
    ->  Gather Motion 3:1  
-         ->  Aggregate
+         ->  Partial Aggregate
                ->  Nested Loop Left Join
-                     Join Filter: (t1.a > (SubPlan 1))
+                     Join Filter: (t1.a > ((SubPlan 1)))
                      ->  Seq Scan on t_out t1
+                           SubPlan 1
+                             ->  Aggregate
+                                   ->  Result
+                                         Filter: (t2.a = t1.a)
+                                         ->  Materialize
+                                               ->  Broadcast Motion 3:3  
+                                                     ->  Seq Scan on t_in t2
                      ->  Materialize
                            ->  Broadcast Motion 3:3  
                                  ->  Seq Scan on t_out t3
-                     SubPlan 1  
-                       ->  Aggregate
-                             ->  Result
-                                   Filter: (t2.a = t1.a)
-                                   ->  Materialize
-                                         ->  Broadcast Motion 3:3  
-                                               ->  Seq Scan on t_in t2
 GP_IGNORE:(17 rows)
 
 select count(*) from t_out t1 left join t_out t3
```
```diff
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/regress/expected/subselect_gp_optimizer.out /home/gpadmin/gpdb_src/src/test/regress/results/subselect_gp.out
--- /home/gpadmin/gpdb_src/src/test/regress/expected/subselect_gp_optimizer.out	2026-07-30 06:50:57.060763829 +0000
+++ /home/gpadmin/gpdb_src/src/test/regress/results/subselect_gp.out	2026-07-30 06:50:57.202761658 +0000
@@ -924,22 +928,22 @@
     on t1.a > (select count(*) from t_in t2 where t2.a = t1.a);
 QUERY PLAN
 ___________
- Aggregate
+ Finalize Aggregate
    ->  Gather Motion 3:1  
-         ->  Aggregate
+         ->  Partial Aggregate
                ->  Nested Loop Left Join
-                     Join Filter: (t1.a > (SubPlan 1))
+                     Join Filter: (t1.a > ((SubPlan 1)))
                      ->  Seq Scan on t_out t1
+                           SubPlan 1
+                             ->  Aggregate
+                                   ->  Result
+                                         Filter: (t2.a = t1.a)
+                                         ->  Materialize
+                                               ->  Broadcast Motion 3:3  
+                                                     ->  Seq Scan on t_in t2
                      ->  Materialize
                            ->  Broadcast Motion 3:3  
                                  ->  Seq Scan on t_out t3
-                     SubPlan 1  
-                       ->  Aggregate
-                             ->  Result
-                                   Filter: (t2.a = t1.a)
-                                   ->  Materialize
-                                         ->  Broadcast Motion 3:3  
-                                               ->  Seq Scan on t_in t2
 GP_IGNORE:(17 rows)
 
 select count(*) from t_out t1 left join t_out t3
```

Alena0704:
> Thanks for contributing! Do you research: why Postgresql generates plan with subplan, while Greengage uses Hash join?
> 
> ```
> postgres=# select * from t_out where a > (select count(*) from t_in where t_in.a = t_out.a);
>  a 
> ---
>  1
> (1 row)
> 
> postgres=# explain (costs off, verbose, analyze) select * from t_out where a > (select count(*) from t_in where t_in.a = t_out.a);
>                                    QUERY PLAN                                    
> ---------------------------------------------------------------------------------
>  Seq Scan on public.t_out (actual time=0.042..0.045 rows=1 loops=1)
>    Output: t_out.a
>    Filter: (t_out.a > (SubPlan 1))
>    SubPlan 1
>      ->  Aggregate (actual time=0.017..0.018 rows=1 loops=1)
>            Output: count(*)
>            ->  Seq Scan on public.t_in (actual time=0.011..0.011 rows=0 loops=1)
>                  Output: t_in.a
>                  Filter: (t_in.a = t_out.a)
>                  Rows Removed by Filter: 1
>  Planning Time: 0.229 ms
>  Execution Time: 0.134 ms
> (12 rows)
> ```
> 
> There is error at the test:
> 
> ```diff
> diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/regress/expected/subselect_gp.out /home/gpadmin/gpdb_src/src/test/regress/results/subselect_gp.out
> --- /home/gpadmin/gpdb_src/src/test/regress/expected/subselect_gp.out	2026-07-30 06:50:37.480701798 +0000
> +++ /home/gpadmin/gpdb_src/src/test/regress/results/subselect_gp.out	2026-07-30 06:50:37.583698198 +0000
> @@ -953,22 +957,22 @@
>      on t1.a > (select count(*) from t_in t2 where t2.a = t1.a);
>  QUERY PLAN
>  ___________
> - Aggregate
> + Finalize Aggregate
>     ->  Gather Motion 3:1  
> -         ->  Aggregate
> +         ->  Partial Aggregate
>                 ->  Nested Loop Left Join
> -                     Join Filter: (t1.a > (SubPlan 1))
> +                     Join Filter: (t1.a > ((SubPlan 1)))
>                       ->  Seq Scan on t_out t1
> +                           SubPlan 1
> +                             ->  Aggregate
> +                                   ->  Result
> +                                         Filter: (t2.a = t1.a)
> +                                         ->  Materialize
> +                                               ->  Broadcast Motion 3:3  
> +                                                     ->  Seq Scan on t_in t2
>                       ->  Materialize
>                             ->  Broadcast Motion 3:3  
>                                   ->  Seq Scan on t_out t3
> -                     SubPlan 1  
> -                       ->  Aggregate
> -                             ->  Result
> -                                   Filter: (t2.a = t1.a)
> -                                   ->  Materialize
> -                                         ->  Broadcast Motion 3:3  
> -                                               ->  Seq Scan on t_in t2
>  GP_IGNORE:(17 rows)
>  
>  select count(*) from t_out t1 left join t_out t3
> ```
> 
>  ```diff
> diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/regress/expected/subselect_gp_optimizer.out /home/gpadmin/gpdb_src/src/test/regress/results/subselect_gp.out
> --- /home/gpadmin/gpdb_src/src/test/regress/expected/subselect_gp_optimizer.out	2026-07-30 06:50:57.060763829 +0000
> +++ /home/gpadmin/gpdb_src/src/test/regress/results/subselect_gp.out	2026-07-30 06:50:57.202761658 +0000
> @@ -924,22 +928,22 @@
>      on t1.a > (select count(*) from t_in t2 where t2.a = t1.a);
>  QUERY PLAN
>  ___________
> - Aggregate
> + Finalize Aggregate
>     ->  Gather Motion 3:1  
> -         ->  Aggregate
> +         ->  Partial Aggregate
>                 ->  Nested Loop Left Join
> -                     Join Filter: (t1.a > (SubPlan 1))
> +                     Join Filter: (t1.a > ((SubPlan 1)))
>                       ->  Seq Scan on t_out t1
> +                           SubPlan 1
> +                             ->  Aggregate
> +                                   ->  Result
> +                                         Filter: (t2.a = t1.a)
> +                                         ->  Materialize
> +                                               ->  Broadcast Motion 3:3  
> +                                                     ->  Seq Scan on t_in t2
>                       ->  Materialize
>                             ->  Broadcast Motion 3:3  
>                                   ->  Seq Scan on t_out t3
> -                     SubPlan 1  
> -                       ->  Aggregate
> -                             ->  Result
> -                                   Filter: (t2.a = t1.a)
> -                                   ->  Materialize
> -                                         ->  Broadcast Motion 3:3  
> -                                               ->  Seq Scan on t_in t2
>  GP_IGNORE:(17 rows)
>  
>  select count(*) from t_out t1 left join t_out t3
> ```

Thanks for the review!

The original fix was written for open-gpdb, where the planner already pulls this subquery into a join, so it was done as a LEFT join + CASE. When porting to Greengage I only verified that the wrong results were gone, not the ORCA plan - so this question didn't come up on my side. I've found Greengage's commit 2ef1914b4df (#1658), which disables ORCA's decorrelation of such COUNT subqueries and keeps them as a SubPlan for the same reason; it took me a while to understand.

Rethinking it, I now think keeping the correlated SubPlan is the safer and more consistent choice here, so I'm reworking the planner path to keep the SubPlan too. Thanks!

silent-observer:
I actually think your current solution is better than the one I made in 2ef1914b4df1418befff7443ea5c2b22be74a726. In case of ORCA, the decorrelation logic was broken in other ways too, so as a temporary solution we decided to disable it for this case. However, decorrelation is useful for performance, and while your solution is not complete, it does allow decorrelation in more cases than completely disabling it. (For a more complete solution search "Magic Decorrelation", for example.)

I wasn't able to find queries where the current patch gives incorrect results, so I think it is logically correct. Are there other reasons to change it, other than consistency with ORCA? I think it would be better to port this solution to ORCA in the future instead.

Alena0704:
> I actually think your current solution is better than the one I made in [2ef1914](https://github.com/GreengageDB/greengage/commit/2ef1914b4df1418befff7443ea5c2b22be74a726). In case of ORCA, the decorrelation logic was broken in other ways too, so as a temporary solution we decided to disable it for this case. However, decorrelation is useful for performance, and while your solution is not complete, it does allow decorrelation in more cases than completely disabling it. (For a more complete solution search "Magic Decorrelation", for example.)

Thank you. I'm looking at this but I haven't yet figured out how to add transformation in ORCA optimizer to fix the case from 2ef1914 commit and anyway it deserves the separate PR I think, so inconsistency stays.

> I wasn't able to find queries where the current patch gives incorrect results, so I think it is logically correct. Are there other reasons to change it, other than consistency with ORCA? I think it would be better to port this solution to ORCA in the future instead.

I don't find neither and agree to port it to ORCA in the future.

silent-observer:
Porting this patch to ORCA is out of scope for this (and also lower priority, since there is no observable incorrect behavior to fix there, unlike the current issue). Instead, please just update plans in `subselect_gp.out` and `subselect_gp_optimizer.out` expected output files, so that the new test actually passes.

Alena0704:
I updated the commit because I ported bugfixes from https://github.com/apache/cloudberry/pull/1933#discussion_r3932828105

PR #549, opened 2026-07-29
ADBDEV-10515: Add debug info collection for interconnect code

silent-observer:
Add debug info collection for interconnect code

To assist with further investigation of a segfault, more info should be
collected in interconnect. Specifically, add elog(PANIC) to collect a core
dump if a connection/buffer mismatch is ever detected. Also, add buffer list
event tracing using a fixed-size circular buffer. Previous events such as
adding/removing ICBuffers to/from interconnect lists will be recorded in the
buffer, and so the trace of past events will be present in the core dump if a
segfault happens. This would allow to verify if the sequence of ICBuffer
operations leading up to the segfault was valid.

No test is added because this doesn't fix any specific bug.

### DIFF
--- a/src/backend/cdb/motion/ic_udpifc.c
+++ b/src/backend/cdb/motion/ic_udpifc.c
@@ -5698,6 +5698,8 @@ SendChunkUDPIFC(ChunkTransportState *transportStates,
 		}
 		checkExceptions(transportStates, pEntry, conn, retry++, timeout);
 		doCheckExpiration = false;
+
+		now = getCurrentTime();
 	}
 
 	conn->pBuff = (uint8 *) conn->curBuff->pkt;
--- a/src/backend/cdb/motion/ic_udpifc.c
+++ b/src/backend/cdb/motion/ic_udpifc.c
@@ -664,13 +664,13 @@ typedef struct ICTrace
 static ICTrace* ic_trace;
 static uint32 ic_trace_idx = 0;
 
-#define IC_TRACE(o,b,now) do { \
+#define IC_TRACE(o,b,now,r) do { \
 	ICTrace *t = &ic_trace[ic_trace_idx++ % IC_TRACE_MAX_SIZE]; \
 	t->timestamp = (now); \
 	t->operation = (o); \
 	t->buf = (b); \
 	t->numOutStanding = unack_queue_ring.numOutStanding; \
-	t->route = t->buf->conn? t->buf->conn->route : -1; \
+	t->route = (r); \
 	t->seq = t->buf->pkt->seq; \
 } while(0)
 
@@ -739,7 +739,7 @@ static inline ICBuffer *icBufferListDelete(ICBufferList *list, ICBuffer *buf);
 static inline ICBuffer *icBufferListPop(ICBufferList *list);
 static void icBufferListFree(ICBufferList *list);
 static inline ICBuffer *icBufferListAppend(ICBufferList *list, ICBuffer *buf);
-static void icBufferListReturn(ICBufferList *list, bool isUnackQueue, uint64 now);
+static void icBufferListReturn(ICBufferList *list, bool isUnackQueue, uint64 now, uint16 route);
 
 static ChunkTransportState *SetupUDPIFCInterconnect_Internal(SliceTable *sliceTable);
 static inline TupleChunkListItem RecvTupleChunkFromAnyUDPIFC_Internal(ChunkTransportState *transportStates,
@@ -2629,7 +2629,7 @@ icBufferListAppend(ICBufferList *list, ICBuffer *buf)
  *
  */
 static void
-icBufferListReturn(ICBufferList *list, bool isUnackQueue, uint64 now)
+icBufferListReturn(ICBufferList *list, bool isUnackQueue, uint64 now, uint16 route)
 {
 #ifdef USE_ASSERT_CHECKING
 	icBufferListCheck("icBufferListReturn", list);
@@ -2639,7 +2639,7 @@ icBufferListReturn(ICBufferList *list, bool isUnackQueue, uint64 now)
 	while ((buf = icBufferListPop(list)) != NULL)
 	{
 		if (isUnackQueue)
-			IC_TRACE(IC_OP_UNACK_QUEUE_DELETE, buf, now);
+			IC_TRACE(IC_OP_UNACK_QUEUE_DELETE, buf, now, route);
 
 		if (isUnackQueue && Gp_interconnect_fc_method == INTERCONNECT_FC_METHOD_LOSS)
 		{
@@ -2648,11 +2648,11 @@ icBufferListReturn(ICBufferList *list, bool isUnackQueue, uint64 now)
 			unack_queue_ring.numOutStanding--;
 			if (icBufferListLength(list) >= 1)
 				unack_queue_ring.numSharedOutStanding--;
-			IC_TRACE(IC_OP_RING_DELETE, buf, now);
+			IC_TRACE(IC_OP_RING_DELETE, buf, now, route);
 		}
 
 		icBufferListAppend(&snd_buffer_pool.freeList, buf);
-		IC_TRACE(IC_OP_FREELIST_APPEND, buf, now);
+		IC_TRACE(IC_OP_FREELIST_APPEND, buf, now, route);
 	}
 }
 
@@ -2769,7 +2769,7 @@ getSndBuffer(MotionConn *conn, uint64 now)
 	if (icBufferListLength(&snd_buffer_pool.freeList) > 0)
 	{
 		ret = icBufferListPop(&snd_buffer_pool.freeList);
-		IC_TRACE(IC_OP_FREELIST_DELETE, ret, now);
+		IC_TRACE(IC_OP_FREELIST_DELETE, ret, now, conn->route);
 		return ret;
 	}
 	else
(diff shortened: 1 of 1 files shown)


src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -1579,6 +1616,10 @@ InitMotionUDPIFC(int *listenerSocketFd, uint16 *listenerPort)
 	snd_control_info.minCwnd = 0;
 	snd_control_info.ackBuffer = palloc0(MIN_PACKET_SIZE);
 
+	/* Allocate debug tracing circular buffer */
+	ic_trace = palloc0(sizeof(ICTrace) * IC_TRACE_MAX_SIZE);
+	ic_trace_idx = 0;
```
RekGRpth, line 1621: Why not do this under the `USE_ASSERT_CHECKING` macro?

silent-observer: This is intentional, because we want this as part of core dumps from customers. So this will be present even on release builds (hence the need for this to not affect performance)

src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -1652,6 +1693,9 @@ CleanupMotionUDPIFC(void)
 	pfree(snd_control_info.ackBuffer);
 	snd_control_info.ackBuffer = NULL;
 
+	/* free the debug trace buffer*/
+	pfree(ic_trace);
```
RekGRpth, line 1697:
```suggestion
	pfree(ic_trace);
	ic_trace = NULL;
```

src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -637,6 +637,43 @@ static ICStatistics ic_statistics;
...
+	uint32 seq;
+} ICTrace;
+
+#define IC_TRACE_MAX_SIZE 1024
+static ICTrace* ic_trace;
+static uint32 ic_trace_idx = 0;
+
+#define IC_TRACE(o,b) do { \
+	ICTrace *t = &ic_trace[ic_trace_idx++ % IC_TRACE_MAX_SIZE]; \
+	t->timestamp = getCurrentTime(); \
```
RekGRpth:
```suggestion
#define IC_TRACE(o,b,now) do { \
	ICTrace *t = &ic_trace[ic_trace_idx++ % IC_TRACE_MAX_SIZE]; \
	t->timestamp = now; \
```

RekGRpth:
How about hiding the new functionality under the `gp_log_interconnect` or new `gp_trace_interconnect` GUC?

silent-observer:
The new GUC will have to be enabled by default, since we want the future core dumps to have tracing. If so, do we even need a new GUC (which would have to be documented)? In what situations would customers need to disable tracing?

RekGRpth:
> The new GUC will have to be enabled by default, since we want the future core dumps to have tracing. If so, do we even need a new GUC (which would have to be documented)? In what situations would customers need to disable tracing?

No, it's better to disable the new GUC by default. The error has only occurred with one customer so far, so we recommend enabling it. Why slow everyone down?

src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -2584,24 +2629,31 @@ icBufferListAppend(ICBufferList *list, ICBuffer *buf)
  *
  */
 static void
-icBufferListReturn(ICBufferList *list, bool inExpirationQueue)
+icBufferListReturn(ICBufferList *list, bool isUnackQueue)
```
RekGRpth:
```suggestion
icBufferListReturn(ICBufferList *list, bool isUnackQueue, uint64 now)
```

src/backend/cdb/motion/ic_udpifc.c
RekGRpth:
```suggestion
getSndBuffer(MotionConn *conn, uint64 now)
```

src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -5609,7 +5679,7 @@ SendChunkUDPIFC(ChunkTransportState *transportStates,
 	ic_control_info.lastPacketSendTime = 0;
 	conn->deadlockCheckBeginTime = now;
 
-	while (doCheckExpiration || (conn->curBuff = getSndBuffer(conn)) == NULL)
+	while (doCheckExpiration || (conn->curBuff = getSndBuffer(conn, now)) == NULL)
```
RekGRpth, line 5682: Now, `now` is stale inside the retry loop in `SendChunkUDPIFC`: `now` is calculated once before the loop, but the loop itself can spin for multiple iterations, blocking in `pollAcks(timeout)` with an increasing `timeout` (this is a backoff when waiting for acks/losing packets). At each iteration, `getSndBuffer(conn, now)` writes the `IC_OP_FREELIST_DELETE` value to the trace with the same, increasingly outdated `now` value — instead of the real time when the buffer was actually received. Previously, `getSndBuffer` called `getCurrentTime()` itself every time it was called, so the timestamp was accurate even inside such a loop. The irony is that the "wait/retry loop" is precisely the scenario (interconnect congestion) that all this tracing was done to diagnose. So, at just the right moment, the timestamp will become inaccurate. In other places, where `now` was also moved before the loop (`TeardownUDPIFCInterconnect_Internal`, `handleStopMsgs`), this is not a problem - there the loops are not blocked on the network between iterations, everything happens almost instantly.

src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -5270,6 +5339,7 @@ checkExpiration(ChunkTransportState *transportStates,
 
 		while ((curBuf = icBufferListPop(&unack_queue_ring.slots[unack_queue_ring.idx])) != NULL)
 		{
+			IC_TRACE(IC_OP_RING_DELETE, curBuf, now, curBuf->conn->route);
```
RekGRpth, line 5342: What if `curBuf->conn` is `NULL`?

src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -6027,6 +6097,7 @@ putIntoUnackQueueRing(UnackQueueRing *uqr, ICBuffer *buf, uint64 expTime, uint64
 
 	buf->unackQueueRingSlot = idx;
 	icBufferListAppend(&unack_queue_ring.slots[idx], buf);
+	IC_TRACE(IC_OP_RING_INSERT, buf, now, buf->conn->route);
```
RekGRpth, line 6102: What if `buf->conn` is `NULL`?

src/backend/cdb/motion/ic_udpifc.c
silent-observer: Can't be NULL here, or we will get segfault on line below (`computeExpirationPeriod` and `sendOnce` expect `conn` to not be NULL)

src/backend/cdb/motion/ic_udpifc.c
silent-observer: Can't be NULL, `putIntoUnackQueueRing` is either called right after setting `buf->conn = conn` (at line 4954), or on a buffer that was just taken out of the queue ring, so it already has a valid connection. I can add an assert here if needed.

src/backend/cdb/motion/ic_udpifc.c
silent-observer: Thanks, this is a genuine problem. Fixed by updating `now` variable, since it is not used after the loop anyway.

ShapatinAA:
```suggestion
	while (doCheckExpiration || (conn->curBuff = getSndBuffer(conn, getCurrentTime())) == NULL)
```
Maybe something like that will be enough? Or simply updating `now` inside this loop.

src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -2593,15 +2638,21 @@ icBufferListReturn(ICBufferList *list, bool inExpirationQueue)
 
 	while ((buf = icBufferListPop(list)) != NULL)
 	{
-		if (inExpirationQueue)	/* the buf is in also in the expiration queue */
+		if (isUnackQueue)
+			IC_TRACE(IC_OP_UNACK_QUEUE_DELETE, buf, now, route);
+
+		if (isUnackQueue && Gp_interconnect_fc_method == INTERCONNECT_FC_METHOD_LOSS)
```
ShapatinAA, line 2644: Why do we need this new condition `INTERCONNECT_FC_METHOD_LOSS`?

RekGRpth (approved):

src/backend/cdb/motion/ic_udpifc.c
silent-observer: It is not new, it is taken from existing calls to icBufferListReturn. Previously, `isExpirationQueue` was true when `list == conn->unackQueue` and `Gp_interconnect_fc_method != INTERCONNECT_FC_METHOD_CAPACITY` (so, `INTERCONNECT_FC_METHOD_LOSS`). Now I pass `isUnackQueue` (`list == conn->unackQueue`) as a separate condition instead, since tracing is done regardless of `Gp_interconnect_fc_method`, and the `Gp_interconnect_fc_method` condition is moved inside this function instead of the two callsites.

ShapatinAA (approved):

PR #552, opened 2026-07-30
Set up `dependabot.yml` for github actions

sobolevn:
I've noticed that you use deprecated action versions in your CI, I suggest auto-bumping them using the default dependabot feature of github.

Ideally, you should also harden your CI setup, currently I can see some security problems there.


magf:
Thank you for the suggestion, but Dependabot doesn't fit our CI versioning model.

Our reusable workflows are versioned with immutable tags intentionally. New CI versions may contain breaking changes required by the current `greengage` codebase, so older branches remain pinned to their existing CI version for compatibility and reproducibility.

Automatically updating these references with Dependabot would undermine this approach, so I don't think it would be useful here.

Regarding the security concerns, please point out the specific issues you noticed so we can address them separately.

PR #554, opened 2026-08-03
CI-5659: Build RPM-based Docker image for 7.x

magf:
## Build RPM-based Docker image for 7.x (#554)

Add a no-tests CI workflow and Rocky Linux 8/9 build support

Changes:
- Add new GitHub Actions workflow `Greengage CI (No Tests)` for Rocky 8/9.
  Builds Docker images but skips all tests. Not required for PRs,
  as Rocky Linux builds are not a mandatory target.
- Update `.github/workflows/README.md` to document CI changes.
- Refactor `README.Rhel-Rocky.bash`:
  - Install Python 3.12 and set as default via alternatives.
  - Build zstd static library from source (not available as package).
  - Separate bare-metal config (SELinux, firewall, sysctl, limits) from Docker.
  - Add OS version detection and conditional package lists for Rocky 8 and 9.
- Add multi-stage Dockerfile `ci/Dockerfile.rockylinux` for Rocky builds.
- Simplify `README.Linux.md` instructions for RHEL/Rocky.

Notes:
- New workflow limited to 7.x branch and targets Rocky 8 and 9.
- Can be used for OS targets not yet covered by main test matrix.

Task: [CI-5659](https://tracker.yandex.ru/CI-5659)
Parent task: [CI-5657](https://tracker.yandex.ru/CI-5657)


GirgerS:
Not a real review, just some things I want to share after working on a similar ticket:

---

Consider also adding these dependencies: 
```
systemd-devel
libicu-devel
tcl-devel 
```
Which are needed for the following configuration:
```
./configure --with-icu --with-tcl --with-systemd
```

Though I'm not sure if these `configure` options are useful. 

--- 

Our team [wants](https://tracker.yandex.ru/GG-407) to eventually separate [system configuration](https://github.com/GreengageDB/greengage/pull/554/changes#diff-2cf6db0e8fda0261c3d7860b0407d10be41a5e6693d38960d4af97049d038217R116) from `README.Rhel-Rocky.bash`, like it is done on the [6.x](https://github.com/GreengageDB/greengage/blob/6.x/README.linux.md?plain=1#L60-L113) branch. Do I understand correctly, that it would be safe to do it, as it is not being run in the docker image?

README.Rhel-Rocky.bash
```diff
@@ -1,38 +1,140 @@
...
+    libicu \
+    libkadm5 \
+    libtool \
+    libuuid-devel \
+    libuv-devel \
+    libxml2-devel \
+    libxslt-devel \
+    libyaml-devel \
+    libzstd-devel \
+    llvm-devel \
```
dimoffon, line 66: **Unversioned `llvm-devel` and `clang` (L40) resolve to LLVM 21.1.8 on Rocky 8 *and* 9** — the exact version this PR calls incompatible. Confirmed in this PR's own `build (rockylinux, 8)` CI log and by building the image locally for both versions.

This is load-bearing: `gpAux/Makefile:128-129` puts `--with-llvm` into `rocky{8,9}_x86_64_CONFIGFLAGS`, so JIT is compiled against it. The CI log shows the break:

```
checking whether /usr/bin/clang supports -Xclang -no-opaque-pointers... no
```

clang >= 17 removed that flag, which PG12-era `llvmjit` relies on to keep bitcode on typed pointers. It still compiles — hence green CI — and fails at query time under `jit=on`. No rockylinux test job exists to catch it.

EPEL ships `llvm14`/`llvm14-devel` for el8 and el9 but **no `clang14`**, so pinning to 14 is a decision, not a one-line package add.

dimoffon (commented):
## Review: RPM-based Docker image for 7.x

Build/CI-only change — nothing under `src/backend`, `src/include/catalog` or `gpcontrib`, so no MPP-correctness review applies. Reviewed as a build change.

**Verified by** building `ci/Dockerfile.rockylinux --target base` locally for `OS_VERSION=8` and `9` (both succeed), inspecting the resulting images, and cross-checking this PR's own `build (rockylinux, 8)` CI log. All findings are confirmed against head `01b29b5a`.

**5 comments — 2 high, 1 medium, 2 low.** The two LLVM comments are one defect seen from two files.

- **High** — unversioned `llvm-devel`/`clang` install LLVM **21.1.8** on both Rocky 8 and 9, not the 14 the PR description and `README.Linux.md` claim. `--with-llvm` is active for `rocky{8,9}_x86_64`, so JIT is built against it. CI is green because it still compiles, and no rockylinux test job exists to exercise it.
- **High** — `README.Rhel-Rocky.bash` aborts under `set -u` on the bare-metal path that `README.Linux.md` documents, after installing packages but before any host configuration runs.
- **Medium** — `psutil` / `pyyaml` / `psycopg2` are absent for the new default `python3` (3.12); `gppylib/commands/unix.py` imports psutil at module level.
- **Low** — `ci/readme.md` build and run sections no longer name the same image tag.

**Checked and clean:** `dnf config-manager` is absent from both base images, but `dnf -y install epel-release` pulls in `dnf-plugins-core` first, so `--set-enabled powertools`/`crb` works. The empty `target_os_version` on the ubuntu matrix leg is fine — `build (ubuntu)` passes and the `package` job already used that pattern. OS-version detection, python3.12 with `--with-python PYTHON=python3`, the `glibc-langpack-en` locale, the static `libzstd.a`, `pg_bsd_indent`, and the multi-stage `ARG`/`ENV` scoping all behave correctly. Both `base` stages build cleanly.

**Needs a cluster to settle:** every consequence of findings 1 and 3 lands at runtime, and no test job exercises these images. The PR scopes RPM test-environment validation out, which is a reasonable split — it is also why both defects can ship unnoticed. For finding 3 I traced the import chain rather than standing up a cluster.

---
_Generated with [greengage plugin](https://github.com/GreengageDB/gg-agent)_

README.Linux.md
```diff
@@ -1,14 +1,14 @@
...
+## For RHEL/Rocky (versions 8 or 9):
+
+- Install dependencies using README.Rhel-Rocky.bash script:
+
+  ```bash
+  sudo ./README.Rhel-Rocky.bash
+  ```
+
+- LLVM 21 provided by current Rocky/RHEL releases is not compatible with
+  Greengage 7. When building in the provided Docker environment, LLVM/Clang
```
dimoffon: **This is not true of the image this PR builds.** Nothing in the diff installs LLVM/Clang 14 — `README.Rhel-Rocky.bash` installs unversioned `llvm-devel`/`clang`, which resolve to **21.1.8** on both Rocky 8 and 9 (verified in this PR's CI log and by local build of `Dockerfile.rockylinux`).

As written the README states a guarantee the build does not provide. Either install 14, or describe what is actually installed. See the note on `README.Rhel-Rocky.bash:65` for why it matters.

README.Rhel-Rocky.bash
```diff
@@ -1,38 +1,140 @@
...
+# Build zstd with static library (not available as a package on Rocky)
+curl -Ls https://github.com/facebook/zstd/releases/download/v1.4.4/zstd-1.4.4.tar.gz | tar -xzf -
+make -j"$(nproc)" -C zstd-1.4.4
+make install PREFIX=/usr/local -C zstd-1.4.4
+rm -rf zstd-1.4.4
+
+#---------------------------------------------------------------------
+# Bare-metal only configuration
+#---------------------------------------------------------------------
+if [[ ! -f /.dockerenv && -z "$IS_DOCKER_BUILD" ]]; then
```
dimoffon: **The bare-metal path aborts here under `set -u` (L8)** — `IS_DOCKER_BUILD` is set only by `ci/Dockerfile.rockylinux`, so on a real host it is unbound.

Reproduced with `sudo ./README.Rhel-Rocky.bash`, which is exactly what `README.Linux.md:6` instructs:

```
+ [[ ! -f /.dockerenv ]]
README.Rhel-Rocky.bash: line 119: IS_DOCKER_BUILD: unbound variable
EXIT=1
```

All packages install, then the script dies before anything this guard protects: SELinux, sssd, firewalld, `/etc/sysctl.d/10-gpdb.conf`, `/etc/security/limits.d/10-nproc.conf`. The only path the guard exists to serve is the one that never runs.

Fix: `-z "${IS_DOCKER_BUILD:-}"`.

README.Rhel-Rocky.bash
```diff
@@ -1,38 +1,140 @@
...
+
+set -euxo pipefail
+
+dnf -y install epel-release
+
+# Detect OS version if not already set
+export OS_VERSION="${OS_VERSION:-$(grep -oP '(?<= release )\d+' /etc/redhat-release)}"
+
+perl_packages="perl-Env perl-ExtUtils-Embed perl-IPC-Run perl-JSON perl-Test-Base"
+python_packages="python3.12 python3.12-devel python3.12-pip python3.12-setuptools"
```
dimoffon, line 16: **Making 3.12 the default `python3` drops gpMgmt's runtime dependencies.** The old script installed `python3-psutil python3-pyyaml python3-psycopg2`; those are gone here, and the platform RPMs would not cover 3.12 anyway.

Verified in both built images — under `python3` (3.12.13), `import psutil`, `import yaml` and `import psycopg2` all fail. `python-dependencies.txt` still requires all three, and `gppylib/commands/unix.py:10` imports psutil at module level, pulled in by `gpstart:24` — so gpMgmt utilities raise `ModuleNotFoundError`. Ubuntu is unaffected: its apt packages land in jammy's 3.10.

Invisible to CI — `unittest-check` is backend cmockery, and no rockylinux test job runs. Suggest adding the three to the Dockerfile's pip install.

ci/readme.md
```diff
@@ -7,9 +7,17 @@ for Ubuntu:
 ```bash
 ```
-for Rocky Linux:
+
+for Rocky Linux 8:
+
+```bash
+docker build -t gpdb7_rockylinux8:latest -f ci/Dockerfile.rockylinux .
```
dimoffon: **The build section no longer produces `gpdb7_regress:latest`, which four sections below still run** — full regression suite, JIT regression, ORCA unit tests, and the manual demo cluster.

Replacing the old `-f ci/Dockerfile` build with these two tags leaves a reader following the file top-to-bottom with no instruction that builds the tag those sections invoke. Either point them at the new tags or restore a build line for the legacy image.

(The behave section is self-consistent — it carries its own `-f ci/Dockerfile` build.)

README.Linux.md
magf: the false promise about LLVM/Clang 14 removed

README.Rhel-Rocky.bash
magf: Fixed

ci/readme.md
magf: This is currently under discussion — there isn't a clear-cut answer yet on how to handle it correctly, so it's pending a decision before I make changes here.

magf: There's an old `ci/Dockerfile` (targeting `rockylinux:8.8`) that this PR doesn't touch — I added a separate parametrized `ci/Dockerfile.rockylinux` for 8/9 instead. `ci/readme.md` still references the old image in test-running instructions.

The scope here (CI-5657) doesn't require tests to pass, so I'm unsure whether to:
1. just swap references to the new Dockerfile/image without verifying the test commands actually run
2. verify the commands at least start (without caring about pass/fail) before swapping
3. leave `ci/readme.md` and the old Dockerfile untouched since it's specifically a test-running doc and out of scope.

magf: I've returned `ci/readme.md` to its original state for now.

README.Rhel-Rocky.bash
magf: This comment is beyond my understanding.
@Stolb27, I need help.

README.Rhel-Rocky.bash
magf: [fix(build): install psutil, pyyaml, psycopg2 for python3.12](https://github.com/GreengageDB/greengage/pull/554/commits/8248b42ca9bbf4286c2fc17433d4fbdc67b7a91e)

magf: And this packages not required for build and moved to install in Docker
[build: move psutil, pyyaml, psycopg2-binary from build script to test image](https://github.com/GreengageDB/greengage/pull/554/commits/cd992b16bc76967dd15533a61aa596f9c7f8eace)

magf:
> Not a real review, just some things I want to share after working on a similar ticket:
> 
> Consider also adding these dependencies:
> 
> ```
> systemd-devel
> libicu-devel
> tcl-devel 
> ```
> 
> Which are needed for the following configuration:
> 
> ```
> ./configure --with-icu --with-tcl --with-systemd
> ```
> 
> Though I'm not sure if these `configure` options are useful.

[fix(rockylinux): add systemd-devel, libicu-devel, tcl-devel rpm packages](https://github.com/GreengageDB/greengage/pull/554/commits/2905a59c3271dce02172035c275e6e8e53deb8df)
[2905a59](https://github.com/GreengageDB/greengage/pull/554/commits/2905a59c3271dce02172035c275e6e8e53deb8df)

> Our team [wants](https://tracker.yandex.ru/GG-407) to eventually separate [system configuration](https://github.com/GreengageDB/greengage/pull/554/changes#diff-2cf6db0e8fda0261c3d7860b0407d10be41a5e6693d38960d4af97049d038217R116) from `README.Rhel-Rocky.bash`, like it is done on the [6.x](https://github.com/GreengageDB/greengage/blob/6.x/README.linux.md?plain=1#L60-L113) branch. Do I understand correctly, that it would be safe to do it, as it is not being run in the docker image?

Yes, the script currently controls and doesn't run the bare-metal settings in Docker, but your solution from 6.x is much better.

Please do this in your task scope.

.github/workflows/README.md
```diff
@@ -1,45 +1,85 @@
-# Greengage CI Workflow
+# CI Workflows
```
bandetto, line 1: This file is obviously generated by agents and is better to be renamed to AGENTS.md.

The actual documentation could be much more compact if written manually without AI-cliches for easier read by humans. Word salad is unfortunately a downside of agents as of now.

bandetto, line 1: Also from what I noticed, for other projects, where tests take less than 3 hours to complete, full rebuild takes a lot of time.

Have we thought about caching docker images with all dependencies already baked-in, and building the components by pulling the cached image instead of reinstalling everything?

Example is to avoid building Greengage, keeping only the 'base' image from the Dockerfile and letting the CI script or the developer build the component.

magf: This is not AI-salad.
Previously it was README for [Greengage CI](https://github.com/GreengageDB/greengage/blob/CI-5659/.github/workflows/greengage-ci.yml#L2) only.
Now it common README for all CI workflows from [.github/workflows](https://github.com/GreengageDB/greengage/blob/CI-5659/.github/workflows/)

magf: This is a huge, important topic, and I've been circling around it since last fall.
It's easy on the surface, but there are many pitfalls.
This needs to be discussed as a separate, major issue.

GirgerS:
Do we expects other tests to be runnable on this image, or is it only `build` that should work?

For example, when following `ci/readme.md` locally, I get the [error](https://github.com/user-attachments/files/32002472/jit_tests.log).

GirgerS:
It seems that we starting to get syntax warning for unit tests.

Grep for `SyntaxWarning` in raw logs here: https://github.com/GreengageDB/greengage/actions/runs/34192514513/job/101953290606?pr=554

It is not the case for Ubuntus or Rocky images from 6.x. e.g.: https://github.com/GreengageDB/greengage/actions/runs/34339536116/job/102456639659?pr=588

It it expected?

.github/workflows/greengage-no-tests.yml
GirgerS, line 1: Just curious, why adding workflow here, not to [CI](https://github.com/GreengageDB/greengage-ci/tree/main/.github/workflows) repo?

magf:
> Do we expects other tests to be runnable on this image

No.

> is it only `build` that should work?

Yes.
Special workflow was added for separate run with no tests. For check builds and packaging only.
Our final goal here is clean build RPM package.

.github/workflows/greengage-no-tests.yml
magf: > Just curious, why adding workflow here, not to [CI](https://github.com/GreengageDB/greengage-ci/tree/main/.github/workflows) repo?

- In short:
A separate workflow described in a separate file is needed to localize it and manage it separately.
The idea is that we can enable or disable it independently of the main workflow.

- To answer the question:
A workflow is a set of jobs described in a separate file.
In this case, we created a new workflow that includes only the job `build`, but it is not stored here.
It is still located, as for the main workflow, in the `greengage-ci` repo you mentioned, **here we only call it** for execution.

In other words, we created **a new set of rules for calling processes (jobs) from the old location**, which became the new workflow.

magf:
> Grep for `SyntaxWarning` in raw logs

I didn't find`SyntaxWarning` matches in the raw log from your links.
Could you at least provide a couple of examples, pls?

> It it expected?

Also in general, our overall goal here is to ensure that the *build works and the RPM package builds correctly*
All dependencies are satisfied, there are no bad links, and all libraries are available.

Successful testing is not a requirement.

GirgerS:
> I didn't find`SyntaxWarning` matches in the raw log from your links. Could you at least provide a couple of examples, pls?

Yep, [here](https://github.com/GreengageDB/greengage/actions/runs/34192514513/job/101953290606?pr=554#step:5:3932)

I did search for this warning, it was [added/changed](https://docs.python.org/dev/whatsnew/3.12.html#other-language-changes) in Python3.12. It seems to indicate that this code will give an error in future versions, so we are fine, at least for now.

Overall, LGTM, let's wait for the other reviewer.

PR #557, opened 2026-08-05
GG-644: Unlink temporary AO base relfiles immediately

riogor:
Temp AO tables left base relfiles on disk until cluster restart on master and segments.
mdunlink_ao_base_relfile() truncates base relfile and registers a
SYNC_UNLINK_REQUEST, which later goes to mdunlinkfiletag.
But mdunlinkfiletag doesn't work with relfiles for temp tables.
Therefore, files remain.

Temp relation files can be safely removed immediately, as they're not 
saved in WAL. This matches mdunlinkfork() behvaior in md.c.

Pass relation temp status to mdunlink_ao_base_relfile().
Immediately unlink base relfile for temp table.
Truncate base relfile, for redo as well. This matches mdunlink_ao_perFile() 
and mdunlinkfork().

Add isolation2 test to cover base relfile removal.

Ticket: GG-644


Stolb27 (commented):
Did you check the same issue for columnar tables?

riogor:
> Did you check the same issue for columnar tables?

Yes. Fix works for both row and column oriented tables.

src/test/isolation2/expected/ao_temp.out
```diff
@@ -0,0 +1,44 @@
+-- This test checks for correct AO temp relfile deletion after session end.
+
+CREATE OR REPLACE FUNCTION relfile_test_cmd(tbl regclass) RETURNS text AS $$ SELECT format('[ -f %L ] && echo 1 || echo 0', (SELECT setting FROM pg_settings WHERE name = 'data_directory') || '/' || pg_relation_filepath(tbl)); $$ LANGUAGE sql;
+CREATE FUNCTION
+
+1: CREATE TEMP TABLE t2 (a int) WITH (APPENDONLY = TRUE, ORIENTATION = ROW);
```
red1452: Use name for table, which is not popular. Sometimes, parallel test may drop table from another test, because tables with the same names.

red1452 (commented):
Small typo in the description: `behvaior` ->`behavior`.
Please, format description by 80 symbols per line.

src/test/isolation2/expected/ao_temp.out
riogor: Done

dimoffon (commented):
## Review: GG-644 — unlink temp AO base relfiles immediately

Reviewed from a worktree at head `037f45f8` (7.x / PG 12.22), so the surrounding-code claims below are verified rather than inferred. CI is green and the new test does execute (`test ao_temp ... ok 10216 ms`).

**The C change is correct.** `RelFileNodeBackendIsTemp(rnode)` is exactly "relpersistence == TEMP" at every `mdunlink_ao` call site; no path resets the backend id to `InvalidBackendId` (the flag round-trips through `RelFileNodePendingDelete` into both the commit record and the 2PC state file); and the non-temp path is bit-for-bit unchanged. `ao_foreach_extent_file` never visits `(segno=0, filenum=0)`, so removing the base file first cannot truncate the extent walk.

**5 comments — 1 significant, 2 moderate, 2 minor.** None of them says the fix is wrong.

### Two corrections to the commit message

The stated mechanism hides how solid the fix actually is:

1. The `SYNC_UNLINK_REQUEST` is **not** "later canceled". `RememberSyncRequest()` only marks *fsync* entries canceled on a forget; `pendingUnlinks` is never touched. The real mechanism is a path mismatch: `FileTag` carries a bare `RelFileNode` with no backend id, and `mdunlinkfiletag()` computes `relpathperm(ftag->rnode, MAIN_FORKNUM)` (`md.c:1489`), i.e. `InvalidBackendId`. So the checkpointer unlinks `base/<db>/X`, which does not exist, while `base/<db>/t_X` survives to the next postmaster start. Silent ENOENT, no warning. This is the same invariant `md.c:1067` already asserts for the heap path — `Assert(!RelFileNodeBackendIsTemp(rnode));`, *"Should never be used with temp relations"* — that the AO path was quietly violating.
2. "matches behavior for unlogged tables" is inaccurate: `mdunlinkfork` treats unlogged MAIN forks exactly like permanent ones, and only the INIT fork goes immediately. No code depends on this misstatement.

### Why the test gap matters

The leak is worse on the segments than on the coordinator. The QD's temp AO base relfile is **zero-length**; each QE retained one `t_<relfilenode>` per temp AO table per session until that segment restarted, reached through `FinishPreparedTransaction` → `DropRelationFiles(..., false)` (`twophase.c:1757`) — live for temp relations per the Greengage note at `storage.c:556`, which explicitly contradicts the stale upstream comment just above the call. `gp_toolkit.gp_check_orphaned_files` never matched `t_*`, which is likely why this went unnoticed.

### Checked and clean

No plan or parse node field added or reordered (`outfast.c`/`readfast.c` not implicated); no motion or locus surface; no ORCA stake in `smgr_unlink`; no new catalog entry or GUC; no new static or global; no fault-injection point moved. AO auxiliary relations inherit the parent's persistence and already went through `mdunlinkfork`'s temp branch, so they were never leaking. Mirrors are unaffected — `t_X` never exists there, and `gp_replica_check` skips `t_`-prefixed files. `test: ao_temp` runs serially, and `packcore`'s ordering constraint still holds.

### Needs a running cluster to settle

That the test actually *fails* before the patch — from the code it must, but I did not build and run it. And whether 10 s is empirically sufficient on this CI hardware; that concern is structural, not measured.

Pre-existing and not this PR's doing: `src/backend/access/appendonly/test/aomd_test.c` calls `mdunlink_ao()` with two arguments against a three-argument signature. It is already stale and not in `TARGETS`, so it is not built.

---
_Generated with [greengage plugin](https://github.com/GreengageDB/gg-agent)_

src/test/isolation2/sql/ao_temp.sql
```diff
@@ -0,0 +1,32 @@
+-- This test checks for correct AO temp relfile deletion after session end.
+
+CREATE OR REPLACE FUNCTION relfile_test_cmd(tbl regclass) RETURNS text
+AS $$
+  SELECT format('[ -f %L ] && echo 1 || echo 0',
+         (SELECT setting FROM pg_settings WHERE name = 'data_directory')
+         || '/' || pg_relation_filepath(tbl)); $$ LANGUAGE sql;
```
dimoffon: **This checks the coordinator only — and the leak's real home is the segments.**

`data_directory` and `pg_relation_filepath()` both evaluate on the QD, so this tests exactly one file: `<coordinator datadir>/base/<db>/t_<QD relfilenode>` — a **zero-length** file that never holds user data. Relfilenodes are allocated independently per node, so this path says nothing about any segment.

The segments reach storage by a different route: `FinishPreparedTransaction` -> `DropRelationFiles(..., false)` (`twophase.c:1757`), live for temp rels per the Greengage note at `storage.c:556`. A regression that broke only the QE path would pass this test green.

`gp_orphaned_files.sql:110-116` has the idiom — a `getTableSegFiles()` declared `execute on all segments`, joined to `gp_segment_configuration.datadir`. Capture the per-segment paths while session 1 is still alive.

src/backend/access/appendonly/aomd.c
```diff
@@ -303,7 +303,7 @@ mdunlink_ao_base_relfile(void *ctx)
 	const char *baserel = unlinkFiles->segPath;
 
 	*unlinkFiles->segpathSuffixPosition = '\0';
-	if (unlinkFiles->isRedo)
+	if (unlinkFiles->isRedo || isRelFileNodeBackendTemp)
```
dimoffon, line 306: **This branch unlinks without truncating, unlike both things the PR cites.**

`mdunlinkfork` deliberately truncates for temp relations too — Greengage `#if 0`s out upstream's `if (!RelFileNodeBackendIsTemp(rnode))` guard (`md.c:380-393`). And `mdunlink_ao_perFile` truncates every AO extent before unlinking (`aomd.c:378`). After this patch the base relfile is the only AO segment file a temp drop unlinks untruncated.

Not a regression — the `isRedo` branch always had this shape — but this is the moment it starts applying to a hot path. Suggest `do_truncate(baserel);` before the `unlink` below; it also makes the "matches mdunlinkfork()" claim true.

(The predicate itself is right, and the non-temp path is unchanged — verified at every call site.)

src/test/isolation2/sql/ao_temp.sql
```diff
@@ -0,0 +1,32 @@
...
+
+1: CREATE TEMP TABLE temp_relfile_aoro (a int) WITH (APPENDONLY = TRUE, ORIENTATION = ROW);
+1: CREATE TEMP TABLE temp_relfile_aoco (a int) WITH (APPENDONLY = TRUE, ORIENTATION = COLUMN);
+
+1: @post_run 'echo "${RAW_STR}" | awk \'NR==3\' >> /tmp/check_ao_relfile_aoro.sh' :
```
dimoffon: **`>>` with no truncation makes a single failure stick.** The only `rm` is at the end of the file, so any run that dies earlier — failed assertion, aborted `make`, ^C — leaves a one-line script behind. The next run appends, `sh` emits two lines, and the test then fails *permanently* with `0\n0`: output that looks nothing like the real bug, so it gets debugged in the storage manager rather than in `/tmp`. Reproduced.

`gp_orphaned_files.sql:73,77` does this right — `>` for the first write, `>>` only for the second. Here there is one write per file, so plain `>` is correct. Same on L15.

src/test/isolation2/sql/ao_temp.sql
```diff
@@ -0,0 +1,32 @@
...
+     select * from relfile_test_cmd('temp_relfile_aoro');
+1: ! sh /tmp/check_ao_relfile_aoro.sh;
+1: @post_run 'echo "${RAW_STR}" | awk \'NR==3\' >> /tmp/check_ao_relfile_aoco.sh' :
+     select * from relfile_test_cmd('temp_relfile_aoco');
+1: ! sh /tmp/check_ao_relfile_aoco.sh;
+
+1q:
+
+-- Wait some time until file cleanup
+2: select pg_sleep(10);
```
dimoffon: **A fixed sleep in front of an unbounded wait.** `1q:` waits only for the driver-side Python child. The file removal happens afterwards, server-side and asynchronously — and because `RemoveTempRelationsCallback` forces a session reset, the DROP must build a *new gang* (fork a QE on every segment) before it can dispatch. On a loaded CI box that is not reliably under 10 s.

It also adds a flat 10 s to every isolation2 run — CI shows `ao_temp ... ok 10216 ms`, almost entirely this line.

`gp_orphaned_files.sql:93` shows the deterministic form: a bounded `while [ ... ]; do sleep 1; done` poll on the actual condition.

src/test/isolation2/sql/ao_temp.sql
```diff
@@ -0,0 +1,32 @@
+-- This test checks for correct AO temp relfile deletion after session end.
+
+CREATE OR REPLACE FUNCTION relfile_test_cmd(tbl regclass) RETURNS text
+AS $$
+  SELECT format('[ -f %L ] && echo 1 || echo 0',
+         (SELECT setting FROM pg_settings WHERE name = 'data_directory')
+         || '/' || pg_relation_filepath(tbl)); $$ LANGUAGE sql;
+
+1: CREATE TEMP TABLE temp_relfile_aoro (a int) WITH (APPENDONLY = TRUE, ORIENTATION = ROW);
+1: CREATE TEMP TABLE temp_relfile_aoco (a int) WITH (APPENDONLY = TRUE, ORIENTATION = COLUMN);
```
dimoffon, line 42: **Neither table is ever populated**, so the test only ever checks zero-byte files. For AO row, segment 0 is never written at all — writes start at segno 1.

An `INSERT` would cover a non-empty base file for AOCO and, more usefully, prove the `.1` / `.129` extents are gone too. Those go through the other callback (`mdunlink_ao_perFile`), which this test never exercises.

GirgerS:
> The SYNC_UNLINK_REQUEST is not "later canceled"

How convenient, I was about to say the same thing. 

The request can't be [cancelled](https://github.com/GreengageDB/greengage/blob/7b82d328986846a5055adec5b548b176937f15a8/src/backend/access/appendonly/aomd.c#L372), because these forgets will have different segnos.

What happens is that our sync request that is generated [here](https://github.com/GreengageDB/greengage/pull/557/changes#diff-e424d1e539ca052f4364115026f953083c6f966c543558a2b27f1f282a0619fdR332), goes to [checkpointer](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/storage/sync/sync.c#L618), where it will be [processed](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/storage/sync/sync.c#L194-L251) later.

And, as whatever AI this is pointed out, [this](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/storage/sync/sync.c#L218-L220) line gets expanded to  [mdunlinkfiletag](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/storage/sync/sync.c#L106) function, which never [expects](https://github.com/GreengageDB/greengage/blob/3318438b7da602bd78413e096276d222221bcf9d/src/backend/storage/smgr/md.c#L1489) temporary files, as [relpathperm](https://github.com/GreengageDB/greengage/blob/3318438b7da602bd78413e096276d222221bcf9d/src/include/common/relpath.h#L94-L96) always inserts `InvalidBackendId` as the backend id.

To see it ourself, you can follow the steps below (without this patch!). I use `lldb` though, so some steps should be adjusted to other debuggers. 
```bash
# Open three terminals
# terminal-1
psql postgres
create temp table tttest2 (a int) with (appendonly=true);
SELECT pg_backend_pid(); # copy this pid
# terminal-2
lldb 
attach #the_pid_that_was_returned_by_psql
b aomd.c:332
c
# terminal-1
\q
# terminal-2
# We should hit the breakpoint 
p tag # Visually remember the struct, it should match the one below
p CheckpointerShmem->checkpointer_pid # Copy this pid

# terminal-3
lldb
attach #the_pid_that_was_returned_by_lldb
b SyncPostCheckpoint
process handle SIGINT -p true -s false #checkpoint command sends an SIGINT to checkpointer, and this signal is intercepted by lldb by default. We don't want this
c

# terminal-2
c

# terminal-1
psql postgres
checkpoint
 
#terminal-3
# And how we should hit the breakpoint! You probably know what to do next, compare the handled entry and step into the line 218)
```

Overall, the fix for the root table seems to be correct, the description, on the other hand, doesn't.

GirgerS:
> This branch unlinks without truncating, unlike both things the PR cites.

And this thing might also be worth looking into, as we had some [problems](6f0a2c14350642400b695e9d87eb809ba23d6a98) with it before. Not sure how applicable these concerns to temporary tables, though.

src/test/isolation2/sql/ao_temp.sql
riogor: It doesn't matter. Patch and ticket only address base relfiles.

src/test/isolation2/sql/ao_temp.sql
riogor: Good point, implemented

src/test/isolation2/sql/ao_temp.sql
riogor: Removed

src/test/isolation2/sql/ao_temp.sql
riogor: Added

src/backend/access/appendonly/aomd.c
riogor: Added truncation

riogor:
> And, as whatever AI this is pointed out, [this](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/storage/sync/sync.c#L218-L220) line gets expanded to [mdunlinkfiletag](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/storage/sync/sync.c#L106) function, which never [expects](https://github.com/GreengageDB/greengage/blob/3318438b7da602bd78413e096276d222221bcf9d/src/backend/storage/smgr/md.c#L1489) temporary files, as [relpathperm](https://github.com/GreengageDB/greengage/blob/3318438b7da602bd78413e096276d222221bcf9d/src/include/common/relpath.h#L94-L96) always inserts InvalidBackendId as the backend id.

Interesting point. I never got it so far - couldn't fire a checkpointer and see any unlinking actions, so i gave up and made a fix. I'll go and check checkpointer.
Thanks for insight!

> And this thing might also be worth looking into, as we had some [problems](https://github.com/GreengageDB/greengage/pull/6f0a2c14350642400b695e9d87eb809ba23d6a98) with it before. Not sure how applicable these concerns to temporary tables, though.

Added it. Guess that it follow comment in [md.c](https://github.com/GreengageDB/greengage/blob/7.x/src/backend/storage/smgr/md.c#L376-L382) (which AI also found). Tests will show.

riogor:
Also, AI found some weird quirk with `pg_stat_file` - when getting distributed table data it could push function call on segments, therefore filecheck occurs in a wrong place - on segment instead of a coordinator.
It didn't happen to me when i was testing (via explain - it always gave a proper plan with function call not on segments).

As AI suggested adding a function that accepts a text array in between helps to mitigate it.
In any way, this shouldn't break overall behavior.

There is a small lead on GP7 [docs page](https://techdocs.broadcom.com/us/en/vmware-tanzu/data-solutions/tanzu-greenplum/7/greenplum-database/ref_guide-sql_commands-CREATE_FUNCTION.html) (in "Using EXECUTE ON attributes" section) which could explain it, although its not the same as my case.

riogor:
I'we looked a little more, found that pg_stat_file is volatile, which, according to comment in [pathnode.c](https://github.com/GreengageDB/greengage/blob/08988a468ce4ac96b65862e302f979d78ba13a29/src/backend/optimizer/util/pathnode.c#L2951) will always execute on coordinator.  Extra function was obsolete.

GirgerS:
LGTM, let's wait for the other reviewer.

red1452 (approved):

GirgerS (approved):

riogor:
@andreyaksenov can you check commit title?

andreyaksenov:
> @andreyaksenov can you check commit title?

LGTM. Optional suggestion (temp -> temporary):

> Unlink temporary AO base relfiles immediately

riogor:
> > @andreyaksenov can you check commit title?
> 
> LGTM. Optional suggestion (temp -> temporary):
> 
> > Unlink temporary AO base relfiles immediately


Done, thanks!

src/test/isolation2/sql/ao_temp.sql
dimoffon: does it mean that such files potentially will be orphaned?

riogor: No, these files are unlinking correctly before patch. `mdunlink_ao_perFile` does the job.

PR #558, opened 2026-08-06
GG-453: Correct exception handling for gpactivatestandby

ShapatinAA:
In some cases, the `gpactivatestandby` utility exits with status code 2, which
indicates that an exception was caught and handled. This can be confusing
because, depending on the exception, the standby may or may not have been
successfully promoted.

To avoid this ambiguity, handle certain exceptions more gracefully and report
the actual cluster state to the user. To do this, check whether the standby is
active and has been promoted, then return the appropriate exit code.

The exit codes now have the following meanings:

-   0 - Standby activated successfully
-   1 - Command encountered an unexpected internal failure that couldn't be
handled by the normal error-handling path
-   2 - Standby activation failed
-   3 - Standby activated successfully but exceptions were encountered

This patch also removes the unused `warnings_generated` variable. Its value was
never used because an exception was always raised before it could affect the
exit status.
It also reworks how port and hostname of standby are got: from now on we're not
getting them from global variables: port is got from `postgresql.conf` file
and hostname is always `localhost` as this utility is always launched from
corresponding host. These things have been changed in this patch because
first: other utilities that are being used inside `gpactivatestandby` take port
and other parameters not from global variables, but from files
(`postgresql.conf`, `postmaster.pid`), and second: we can not determine if
standby was promoted successfully if we're using port and hostname from
env. variable (as it was mentioned, underlying utils for promotion, i.e.
`gpstart`, use port from configuration file, not env. variable, and hostname is
always `localhost` for them for obvious reasons).


gpMgmt/bin/gpactivatestandby
```diff
@@ -361,8 +394,10 @@ setup_tool_logging(EXECNAME, unix.getLocalHostname(), unix.getUserName())
 if options_.logfile:
     setup_tool_logging(EXECNAME, unix.getLocalHostname(), unix.getUserName(), logdir=options_.logfile)
 
+read_standby_port_from_postgresqlconf(options_)
```
bimboterminator1: Is there any exception handling for postgresql.conf read? If the file does not exist or similar, the rc wil be 1

In fact, if we change the contract (1 for succeesfull cases with handled exceptions), any error leading to rc 1 must accurately correspond to the new contract.

I.e we should either check other similar functions which may throw exeception with rc 1 and make them throw rc 2, or do not change the contract

gpMgmt/bin/gpactivatestandby
```diff
@@ -376,13 +411,13 @@ try:
...
     if not requires_restart:
-        res = promote_standby(options_.coordinator_data_dir)
+        res = promote_standby(options_.coordinator_data_dir, options_.standby_port)
         if not res:
             raise GpActivateStandbyException('Timed out waiting for promoted coordinator to accept database connections.')
 
     # now we can access the catalog.  promote action has already updated
     # catalog, so array.coordinator is the old (promoted) standby at this point.
-    array_ = get_config()
+    array_ = get_config(options_.standby_port)
```
bimboterminator1, line 452: get_config internally calls initFromCatalog(), which may raise an exception. But your catch block does not cover it

gpMgmt/bin/gpactivatestandby
ShapatinAA: Yes, you totally right, we should also consider exceptions form outside of try/catch block, thou not much happens here.
According to task, we want to have return codes suitable for distinguishing states of `gpactivatestandby`, so rcs should be fixed.
Unhandled exceptions in python return 1, so this rc should be free of other meanings (thou `gpstart` uses it for signaling partially started cluster). 
As we don't have clear convention between applications and their return codes I would suggest using rc 

- 0 for successful standby activation 
- 1 for unexpected failure that couldn't be handled by normal error handler, with request to carefully examine error
- 2 for standby activation failure
- 3 for overall successful standby activation, but with some non-critical exceptions exceptions

What do you think?

gpMgmt/bin/gpactivatestandby
ShapatinAA: Thank you for noticing!
I consider that exception at this stage from this function should be handled by general exception with rc 2, as at this stage standby should be active, and unavailability would mean that something unexpected definitely happened.
Though I would like to add while cycle at this moment, as simple segment downtime detected by fts could trigger this error.

gpMgmt/bin/gpactivatestandby
bimboterminator1: Agree, this model covers most caveats

ShapatinAA: Fixed

gpMgmt/bin/gpactivatestandby
```diff
@@ -393,13 +438,17 @@ try:
...
+    if (check_standby_activation(options_, e)):
+        sys.exit(3)
     else:
-        sys.exit(0)
-    
+        logger.fatal('Error activating standby coordinator: %s' % str(e))
+        sys.exit(2)
 except Exception as e:
     logger.fatal('Error activating standby coordinator: %s' % str(e))
     sys.exit(2)
```
bimboterminator1: We cannot reliably tell whether the standby ended up promoted, so do not claim either outcome.

ShapatinAA: Why tho?

ShapatinAA: Btw new `for` loop in `get_config()` was excessive addition, because I totally forgot, that we're connecting in utility mode, so the segment downtime shouldn't be counted as the reason for it, but this addition can be left anyway, for reliability.

bimboterminator1: I think here should be exit code 1. since general Exception may mean anything, thus matches with your `Command encountered an unexpected internal failure that couldn't be handled by the normal error-handling path`

ShapatinAA: Yes, you right. I would say - we should consider exit code 2 for state where no promotion happened, and in case of `except (ExecutionError, GpActivateStandbyException) as e:` check it still could be promoted, just down for some reason.

ShapatinAA: Otherwise, we should get rid of rc 2 😅

bimboterminator1 (dismissed):

ShapatinAA:
Gonna add return codes as variables and add rc 2 for `pid>0 && isSegmentMaster == false`

gpMgmt/bin/gpactivatestandby
```diff
@@ -393,15 +453,26 @@ try:
...
+
+# Even with raised exceptions we still could have promoted standby.
+# User needs to be notified about this with separate return code.
+except ExecutionError as e:
+    return_code = check_standby_activation(options_, e)
+    sys.exit(return_code)
+
+except GpActivateStandbyException as e:
+    # We still can possibly get GPACTIVATESTANDBY_INTERNAL_FAILURE
+    if "the postmaster could not come up" in str(e):
```
bimboterminator1: in what case this branch fires? where the string "the postmaster could not come up" appears?

ShapatinAA: This branch fires when custom `GpActivateStandbyException` gets thrown, which is only inside this module. All this errors can safely be redirected to exit code 2 (as they mean that standby wasn't promoted), except for this one:
```python
    if not requires_restart:
        res = promote_standby(options_.master_data_dir, options_.standby_port)
        if not res:
            raise GpActivateStandbyException('Either the set port is incorrect or '
                                             'the postmaster could not come up.')
```
Here promotion could happen inside of `promote_standby()`, but some segments could be down, so `GpActivateStandbyException('Either the set port is incorrect or the postmaster could not come up.')` would be raised via
```python
def promote_standby(master_data_dir, port):
...
    for i in range(600):
        try:
            dburl = dbconn.DbURL(port=port)
            with dbconn.connect(dburl, utility=True, logConn=False) as conn:
                # When the new master is available for connections, we should
                # run a CHECKPOINT to force the new TimeLineID to be written to
                # the pg_control file.
                dbconn.execSQL(conn, 'CHECKPOINT')
            logger.info('Standby master is promoted')
            return True
        except pygresql.InternalError as e:
            pass
        time.sleep(1)
    return False
```

bimboterminator1: Ok, agree. But isn't checking an exception string an antipattern, is it?

ShapatinAA: Well, I'm not sure if it is an antipattern, maybe a bad practice. And I want left the things they are because:

- It would require re-implementation of `GpActivateStandbyException` exceptions (to be able to derive by their type what is recoverable and what is not), which I don't really want to include in this patch.
- And considering that every thrown exception of this type has unique message I think it is safe to assure that nothing bad can happen from derivation by message.

bimboterminator1 (dismissed):

gpMgmt/bin/gpactivatestandby
```diff
@@ -346,10 +363,51 @@ def promote_standby(coordinator_data_dir):
...
+    try:
+        array = get_config(options.standby_port)
+    except Exception as e:
+        logger.warning("Couldn't get cluster configuration: %s." % str(e))
+        logger.fatal('Internal error activating standby coordinator: %s' % str(err))
+        return GPACTIVATESTANDBY_INTERNAL_FAILURE
+    
+    if not array.coordinator.isSegmentCoordinator(current_role=True):
+        logger.fatal('Error activating standby coordinator: %s' % str(err))
+        return GPACTIVATESTANDBY_FAILED
```
hilltracer: Is it possible that initFromCatalog() successfully connects to the standby [get](https://github.com/GreengageDB/greengage/blob/ab129676775f977cb54d8784d5bd5303a69459ab/gpMgmt/bin/gppylib/gparray.py#L957) gp_segment_configuration,  but array.coordinator still refers to the old coordinator?

In this case GpArray.__init__() [selects](https://github.com/GreengageDB/greengage/blob/ab129676775f977cb54d8784d5bd5303a69459ab/gpMgmt/bin/gppylib/gparray.py#L807) array.coordinator based on isSegmentCoordinator(True), as you check. Should we also compare getSegmentPort() and getSegmentDataDirectory() with the expected standby values?

gpMgmt/bin/gpactivatestandby
```diff
@@ -376,13 +436,13 @@ try:
 
     # promote standby, only if the standby is running in recovery
     if not requires_restart:
-        res = promote_standby(options_.coordinator_data_dir)
+        res = promote_standby(options_.coordinator_data_dir, options_.standby_port)
         if not res:
             raise GpActivateStandbyException('Timed out waiting for promoted coordinator to accept database connections.')
```
hilltracer, line 448: Looks like promote_standby() can return False even after successful `pg_ctl promote` (so standby will be activate), when the later CHECKPOINT 600s loop will fail. In this case we will raise rc=2. Is it ok?

gpMgmt/test/behave/mgmt_utils/gpactivatestandby.feature
```diff
@@ -44,7 +44,7 @@ Feature: gpactivatestandby
         Then gpinitstandby should return a return code of 0
         And verify the standby coordinator entries in catalog
         And the user runs gpactivatestandby with options "-d invalid_directory"
-        Then gpactivatestandby should return a return code of 2
+        Then gpactivatestandby should return a return code of 1
```
hilltracer: It's ok that we change this error code to `unexpected internal failure`? Maybe it should be handled validation error with rc=2? Can we catch it?

gpMgmt/bin/gpactivatestandby
hilltracer: And the same for other places after promotion. For example if we have error in get_config() we return rc=2 but promotion already happened. Should we classify all unexpected errors after promotions as rc=3?

gpMgmt/test/behave/mgmt_utils/gpactivatestandby.feature
ShapatinAA: fixed

gpMgmt/bin/gpactivatestandby
ShapatinAA: Yes, we should, and we do that in these places. `GpActivateStandbyException` will be caught and checked for `check_standby_activation()` which will return rc 3 in case of promotion.

Same with `get_config()`. If it throws any error at this stage - it means something happened with system and we should study output - rc 1.

gpMgmt/bin/gpactivatestandby
ShapatinAA: Thank you a lot for mentioning this! Indeed this condition is a trivial one and we could have situation where it would pass, despite false activation...  Fixed

bimboterminator1: No, that's not possible.

https://github.com/GreengageDB/greengage/blob/7.x/src/backend/access/transam/xlog.c#L6456 calls gp_activate_standby() synchronously, in its own transaction, as part of the coordinator's post-recovery startup sequence - before the postmaster opens for normal connections. That function (segment_config_activate_standby() in src/backend/utils/gp/segadmin.c:715) does two things **atomically**:
1. Deletes the old coordinator's row from gp_segment_configuration entirely (CatalogTupleDelete) - not just leaves it stale.
2. Updates the standby's own row (same dbid) to role = PRIMARY, preferred_role = PRIMARY.

So by the time promote_standby()'s retry loop succeeds (which only happens once the instance accepts connections - necessarily after this startup-sequence step runs), there is no old-coordinator row left to accidentally match, and the standby's own row already correctly shows role='p'. array.coordinator can't resolve to the old coordinator because that row no longer exists.

ShapatinAA: On 7.x with hot standby on I've killed postmaster of coordinator with `kill <postmaster_pid>`, then connected to db via `psql -p 7001 -U gpadmin -h localhost -d template1` and got next:
```sql
template1=# SELECT dbid, content, role, preferred_role, mode, status,
            hostname, address, port, datadir
            FROM pg_catalog.gp_segment_configuration
            ORDER BY content, preferred_role DESC;
 dbid | content | role | preferred_role | mode | status |   hostname   |   address    | port |                                 datadir                                  
------+---------+------+----------------+------+--------+--------------+--------------+------+--------------------------------------------------------------------------
    1 |      -1 | p    | p              | n    | u      | 0f2027444a24 | 0f2027444a24 | 7000 | /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/qddir/demoDataDir-1
    8 |      -1 | m    | m              | s    | u      | 0f2027444a24 | 0f2027444a24 | 7001 | /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/standby
    2 |       0 | p    | p              | s    | u      | 0f2027444a24 | 0f2027444a24 | 7002 | /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/dbfast1/demoDataDir0
    5 |       0 | m    | m              | s    | u      | 0f2027444a24 | 0f2027444a24 | 7005 | /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/dbfast_mirror1/demoDataDir0
    3 |       1 | p    | p              | s    | u      | 0f2027444a24 | 0f2027444a24 | 7003 | /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/dbfast2/demoDataDir1
    6 |       1 | m    | m              | s    | u      | 0f2027444a24 | 0f2027444a24 | 7006 | /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/dbfast_mirror2/demoDataDir1
    4 |       2 | p    | p              | s    | u      | 0f2027444a24 | 0f2027444a24 | 7004 | /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/dbfast3/demoDataDir2
    7 |       2 | m    | m              | s    | u      | 0f2027444a24 | 0f2027444a24 | 7007 | /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/dbfast_mirror3/demoDataDir2
(8 rows)
```

UPD: To change this logic is good point nevertheless, as current condition checks it undoubtedly.

bimboterminator1: ok ok

gpMgmt/bin/gpactivatestandby
hilltracer: but you have 
https://github.com/GreengageDB/greengage/blob/ab814e5b90548b93b8b07da98047ba3b67738134/gpMgmt/bin/gpactivatestandby#L467
condition how `'Timed out waiting for promoted coordinator to accept database connections.'` will be parsed?

ShapatinAA: My bad, look at wrong PR. Fixed.

gpMgmt/bin/gpactivatestandby
```diff
@@ -376,8 +376,18 @@ def check_standby_activation(options, err):
...
         logger.fatal('Internal error activating standby coordinator: %s' % str(err))
         return GPACTIVATESTANDBY_INTERNAL_FAILURE
-    
-    if not array.coordinator.isSegmentCoordinator(current_role=True):
+
+    coordinator_hostname = array.coordinator.getSegmentHostName()
+    coordinator_data_dir = os.path.normpath(array.coordinator.getSegmentDataDirectory())
+    coordinator_port = array.coordinator.getSegmentPort()
+
+    standby_hostname = unix.getLocalHostname()
```
hilltracer: getLocalHostname() normalizes hostname to short form.
https://github.com/GreengageDB/greengage/blob/677398e45766110a32e318266f186cf0cbe720a5/gpMgmt/bin/gppylib/commands/unix.py#L39-L40
Can gp_segment_configuration contain hostname containing dots?
Maybe remove the `coordinator_hostname != standby_hostname` check to be safe.

gpMgmt/bin/gpactivatestandby
hilltracer: > Same with get_config(). If it throws any error at this stage - it means something happened with system and we should study output - rc 1.

So we can return rc=1 even when standby was activated, if an unexpected error happens in e.g. get_config()? Is that acceptable, or should we also catch this error and check whether promotion actually happened?

ShapatinAA: >So we can return rc=1 even when standby was activated, if an unexpected error happens in e.g. get_config()?

Possibly yes. In my understanding rc 1 is the situation when we neither can confirm or deny any kind of state of standby.

The only way for checking standby state is trying to connect to it and query currently visible cluster state. And if we can not do that, it means something went wrong.

gpMgmt/bin/gpactivatestandby
ShapatinAA: Painfully curious definition... Who needs to use "local" hostname in this form...

I think we shouldn't get rid of hostname check as port and datadir can be made the same on different hosts and this will be not enough to check.

gpMgmt/bin/gpactivatestandby
hilltracer: What if
https://github.com/GreengageDB/greengage/blob/fbd50df3eda527e76e55f5a754bdee5d8d03faa5/gpMgmt/bin/gpactivatestandby#L449
throws an exception? after successful promotion.
Do I understand right that it will caught by 
https://github.com/GreengageDB/greengage/blob/fbd50df3eda527e76e55f5a754bdee5d8d03faa5/gpMgmt/bin/gpactivatestandby#L478-L480
So we return 2 here instead 1
?

ShapatinAA: I assumed it should be pretty safe connection (localhost and utility mode), so if something went wrong here, then promotion clearly haven't happen, but general exceptions without any assumptions should have rc 1, there you are right...

gpMgmt/bin/gpactivatestandby
```diff
@@ -235,13 +238,22 @@ def check_original_coordinator_status(options):
...
+        options.standby_port = pgconf_dict.int('port')
+
 #-------------------------------------------------------------------------
-def get_config():
+def get_config(port):
     """Retrieves configuration information from the catalog."""
     
     logger.info('Reading current configuration...')
-    dburl = dbconn.DbURL()
+    dburl = dbconn.DbURL(port=port)
```
hilltracer: From AI: we now get port from postgresql.conf, which is good. But hostname is still taken from PGHOST if it is set.
https://github.com/GreengageDB/greengage/blob/492636ee893d92eb7b7d4fb25b0a918b2c4215f9/gpMgmt/bin/gppylib/db/dbconn.py#L86-L104
And it could redirect the connection to another host (for example `export PGHOST=old-coordinator`).
Maybe we should explicitly use localhost here?

ShapatinAA: Yes, It is also right...) Fixed

hilltracer (approved):

bimboterminator1 (approved):

PR #559, opened 2026-08-07
GG-452: Correct exception handling for gpactivatestandby

ShapatinAA:
In some cases, the `gpactivatestandby` utility exits with status code 2, which
indicates that an exception was caught and handled. This can be confusing
because, depending on the exception, the standby may or may not have been
successfully promoted.

To avoid this ambiguity, handle certain exceptions more gracefully and report
the actual cluster state to the user. To do this, check whether the standby is
active and has been promoted, then return the appropriate exit code.

The exit codes now have the following meanings:

-   0 - Standby activated successfully
-   1 - Command encountered an unexpected internal failure that couldn't be
handled by the normal error-handling path
-   2 - Standby activation failed
-   3 - Standby activated successfully but exceptions were encountered

This patch also removes the unused `warnings_generated` variable. Its value was
never used because an exception was always raised before it could affect the
exit status.
It also reworks how port and hostname of standby are got: from now on we're not
getting them from global variables: port is got from `postgresql.conf` file
and hostname is always `localhost` as this utility is always launched from
corresponding host. These things have been changed in this patch because
first: other utilities that are being used inside `gpactivatestandby` take port
and other parameters not from global variables, but from files
(`postgresql.conf`, `postmaster.pid`), and second: we can not determine if
standby was promoted successfully if we're using port and hostname from
env. variable (as it was mentioned, underlying utils for promotion, i.e.
`gpstart`, use port from configuration file, not env. variable, and hostname is
always `localhost` for them for obvious reasons).

### DIFF
--- a/gpMgmt/bin/gpactivatestandby
+++ b/gpMgmt/bin/gpactivatestandby
@@ -379,11 +379,12 @@ setup_tool_logging(EXECNAME, unix.getLocalHostname(), unix.getUserName())
 
 # parse args and options
 (options_, args_) = parseargs()
-read_standby_port_from_postgresqlconf(options_)
 
 # if we got a new log dir, we can now set it up.
 if options_.logfile:
     setup_tool_logging(EXECNAME, unix.getLocalHostname(), unix.getUserName(), logdir=options_.logfile)
+
+read_standby_port_from_postgresqlconf(options_)
 
 try:
     print_summary(options_)
--- a/gpMgmt/bin/gpactivatestandby
+++ b/gpMgmt/bin/gpactivatestandby
@@ -17,6 +17,7 @@ from builtins import range
 import os
 import sys
 import signal
+import socket
 import glob
 import time
 import shutil
@@ -374,11 +375,13 @@ def check_standby_activation(options, err):
     master_data_dir = os.path.normpath(array.master.getSegmentDataDirectory())
     master_port = array.master.getSegmentPort()
 
-    standby_hostname = unix.getLocalHostname()
+    standby_hostname = socket.gethostname()
+    standby_fqdn = socket.getfqdn()
     standby_data_dir = os.path.normpath(options.master_data_dir)
     standby_port = options.standby_port
 
-    if not (master_hostname == standby_hostname and
+    if not ((master_hostname == standby_hostname or
+            master_hostname == standby_fqdn) and
             master_data_dir == standby_data_dir and
             master_port == standby_port):
         logger.fatal('Error activating standby master: %s' % str(err))
--- a/gpMgmt/bin/gpactivatestandby
+++ b/gpMgmt/bin/gpactivatestandby
@@ -245,7 +245,7 @@ def read_standby_port_from_postgresqlconf(options):
         options.standby_port = pgconf_dict.int('port')
 
 #-------------------------------------------------------------------------
-def get_config(port=0):
+def get_config(port):
     """Retrieves configuration information from the catalog."""
     
     logger.info('Reading current configuration...')
@@ -342,9 +342,12 @@ def promote_standby(master_data_dir, port):
     for i in range(600):
         try:
             dburl = dbconn.DbURL(port=port)
-            conn = dbconn.connect(dburl, utility=True, logConn=False)
-            dbconn.execSQL(conn, 'CHECKPOINT')
-            conn.close()
+            with dbconn.connect(dburl, utility=True, logConn=False) as conn:
+                # When the new master is available for connections, we should
+                # run a CHECKPOINT to force the new TimeLineID to be written to
+                # the pg_control file.
+                dbconn.execSQL(conn, 'CHECKPOINT')
+            logger.info('Standby master is promoted')
             return True
         except pygresql.InternalError as e:
             pass
--- a/gpMgmt/bin/gpactivatestandby
+++ b/gpMgmt/bin/gpactivatestandby
@@ -249,19 +249,10 @@ def get_config(port):
     """Retrieves configuration information from the catalog."""
     
     logger.info('Reading current configuration...')
-    array = None
-    last_error = None
-    for i in range(600):
-        try:
-            dburl = dbconn.DbURL(port=port)
-            array = gparray.GpArray.initFromCatalog(dburl, utility=True)
-            return array
-        except (psycopg2.InternalError, psycopg2.OperationalError) as e:
-            last_error = e
-            pass
-        time.sleep(1)
-    # Shouldn't be able to get here in normal execution.
-    raise last_error
+    dburl = dbconn.DbURL(port=port)
+    array = gparray.GpArray.initFromCatalog(dburl, utility=True)
+    
+    return array
 
 #-------------------------------------------------------------------------
 def check_standby_running(options):
--- a/gpMgmt/bin/gpactivatestandby
+++ b/gpMgmt/bin/gpactivatestandby
@@ -253,7 +253,7 @@ def get_config(port):
     """Retrieves configuration information from the catalog."""
     
     logger.info('Reading current configuration...')
-    dburl = dbconn.DbURL(port=port)
+    dburl = dbconn.DbURL(port=port, hostname='localhost')
     array = gparray.GpArray.initFromCatalog(dburl, utility=True)
     
     return array
--- a/gpMgmt/bin/gpactivatestandby
+++ b/gpMgmt/bin/gpactivatestandby
@@ -337,22 +337,24 @@ def promote_standby(master_data_dir):
 
 
 #-------------------------------------------------------------------------
-def check_standby_activation(options_, array_=None):
+def check_standby_activation(options_):
     """
-    Check if coordinator postmaster is running.
+    Check if standby was promoted and running.
     """
 
-    # If array is empty, it means that we failed even before/in process of
-    # execution of get_config() and something nasty happened.
-    if not array_:
-        return False
-
-    # Here we check that gpstop accidentally hasn't stopped our promoted standby
+    # Here we check that gpstop or something else accidentally hasn't stopped
+    # our standby we're activating.
     pid = gp.get_postmaster_pid_locally(options_.coordinator_data_dir)
     if pid > 0:
-        return True
-    else:
-        return False
+        # We should wrap this in try/catch because we could get here after
+        # unsuccessful promotion and standby may not be able to accept queries
+        try:
+            array = get_config()
+        except Exception as e:
+            return False
+        if array.coordinator.isSegmentCoordinator(current_role=True):
+            return True
+    return False
 
 #-------------------------------------------------------------------------
 # Main
@@ -411,18 +413,17 @@ try:
     
     print_results(array_, unix.getLocalHostname(), options_)
 
-except ExecutionError as e:
-    errStr = str(e)
+except (ExecutionError, GpActivateStandbyException) as e:
 
-    # Even with ExecutionError we still can have promoted standby.
-    # User need to be notified about this with separate return code.  
-    if (errStr.startswith("ExecutionError: 'non-zero rc: 2'")):
-        if (check_standby_activation(options_, array_)):
-            logger.warning('Standby activated with exception', exc_info=e)
-            sys.exit(1)
-        else:
-            logger.fatal('Error activating standby coordinator: %s' % str(e))
-            sys.exit(2)
+    # Even with raised exceptions we still could have promoted standby.
+    # User need to be notified about this with separate return code.
+    if (check_standby_activation(options_)):
+        logger.warning('Encountered exception: ', exc_info=e)
+        logger.warning('The specified standby coordinator at directory %s acts as primary.' % options_.coordinator_data_dir)
+        sys.exit(1)
+    else:
+        logger.fatal('Error activating standby coordinator: %s' % str(e))
+        sys.exit(2)
 except Exception as e:
     logger.fatal('Error activating standby coordinator: %s' % str(e))
     sys.exit(2)
--- a/gpMgmt/bin/gpactivatestandby
+++ b/gpMgmt/bin/gpactivatestandby
@@ -142,12 +142,17 @@ def parseargs():
 
 
 #-------------------------------------------------------------------------
-def print_results(array, hostname, options):
+def print_results(array, hostname, options, successfully=False):
     """Prints out the summary of the operation."""
     
     logger.info('-----------------------------------------------------')
-    logger.info('The activation of the standby master has completed successfully.')
-    logger.info('%s is now the new primary master.' % hostname)
+    if successfully:
+        logger.info('The activation of the standby coordinator has completed successfully.')
+    else:
+        logger.info('The activation of the standby coordinator has completed.')
+        logger.warning('Some non-critical exceptions were encountered during activation process.')
+        logger.warning('Cluster state needs to be analyzed before proceeding with further instructions.')
+    logger.info('%s is now the new primary coordinator.' % hostname)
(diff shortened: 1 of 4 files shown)


gpMgmt/bin/gpactivatestandby
```diff
@@ -43,6 +45,11 @@ LOG_TIME_THRESHOLD_MINS = 120
 
 GP_RECOVERY_CONF_FILE = "recovery.conf"
 
+GPACTIVATESTANDBY_SUCCESS = 0
+GPACTIVATESTANDBY_INTERNAL_FAILURE = 1
```
whitehawk, line 50: I suggest adding these new return codes to the help output.

gpMgmt/bin/gpactivatestandby
```diff
@@ -324,13 +341,10 @@ def promote_standby(master_data_dir):
     logger.debug('Waiting for connection...')
     for i in range(600):
         try:
-            dburl = dbconn.DbURL()
-            with dbconn.connect(dburl, utility=True, logConn=False) as conn:
-                # When the new master is available for connections, we should
-                # run a CHECKPOINT to force the new TimeLineID to be written to
-                # the pg_control file.
-                dbconn.execSQL(conn, 'CHECKPOINT')
-            logger.info('Standby master is promoted')
```
whitehawk, line 333: Why was this part reworked? We've lost the 'with' clause. Now, if an exception happens at the `dbconn.execSQL(conn, 'CHECKPOINT')`, the connection will not be closed.

Also, the comment is lost now.

gpMgmt/bin/gpactivatestandby
```diff
@@ -237,14 +239,29 @@ def check_original_master_status(options):
...
     """Retrieves configuration information from the catalog."""
     
     logger.info('Reading current configuration...')
-    dburl = dbconn.DbURL()
-    array = gparray.GpArray.initFromCatalog(dburl, utility=True)
-    
-    return array
+    array = None
+    last_error = None
+    for i in range(600):
```
whitehawk: Where did the magic number 600 came from?

gpMgmt/bin/gpactivatestandby
```diff
@@ -237,14 +239,29 @@ def check_original_master_status(options):
...
  
 #-------------------------------------------------------------------------
-def get_config():
+def read_standby_port_from_postgresqlconf(options):
+        logger.info("Obtaining standby's port from standby data directory")
+        pgconf_dict = pgconf.readfile(options.master_data_dir + "/postgresql.conf")
+        options.standby_port = pgconf_dict.int('port')
+
+#-------------------------------------------------------------------------
+def get_config(port=0):
```
whitehawk: Is there any possibility that anybody will use the default `port=0`? This default looks not usable, I suggest to remove the default completely.

gpMgmt/bin/gpactivatestandby
```diff
@@ -237,14 +239,29 @@ def check_original_master_status(options):
...
+    array = None
+    last_error = None
+    for i in range(600):
+        try:
+            dburl = dbconn.DbURL(port=port)
+            array = gparray.GpArray.initFromCatalog(dburl, utility=True)
+            return array
+        except (psycopg2.InternalError, psycopg2.OperationalError) as e:
+            last_error = e
+            pass
```
whitehawk: do we need this `pass`?

gpMgmt/bin/gpactivatestandby
```diff
@@ -237,14 +239,29 @@ def check_original_master_status(options):
...
+        time.sleep(1)
+    # Shouldn't be able to get here in normal execution.
+    raise last_error
```
whitehawk: If we raise only the last error, and loose all preceding error, it means we can loose something important for debug.

gpMgmt/bin/gpactivatestandby
whitehawk: Why do we need the re-try loop here?

gpMgmt/bin/gpactivatestandby
```diff
@@ -354,8 +409,10 @@ setup_tool_logging(EXECNAME, unix.getLocalHostname(), unix.getUserName())
 if options_.logfile:
     setup_tool_logging(EXECNAME, unix.getLocalHostname(), unix.getUserName(), logdir=options_.logfile)
 
+read_standby_port_from_postgresqlconf(options_)
```
whitehawk: It sounds like the port handling is not directly related to the return codes. Why do we the port related changes here? I guess it can be a separate patch (if ever needed).

gpMgmt/bin/gpactivatestandby
ShapatinAA: As I see none of utilities do that. Would this be fine if only one of our utils will give such info to the client?

gpMgmt/bin/gpactivatestandby
ShapatinAA: Thank you for noticing! In process of cherry-picking from 453 I haven't notice this changes. Fixed.

gpMgmt/bin/gpactivatestandby
ShapatinAA: I just copied it from already existing loop in `promote_standby()` as it is seems that we have a lot of default timeouts as 10 minutes. And here I do not take assumption on what exactly could went wrong.

gpMgmt/bin/gpactivatestandby
ShapatinAA: Yeah, you right, I thought about making this function to be more robust, but it simply means that it will use `PGPORT` env variable value, which is not obvious behavior. Fixed.

gpMgmt/bin/gpactivatestandby
ShapatinAA: I think it makes code cleaner and intentions of skipping error are more obvious that way.

gpMgmt/bin/gpactivatestandby
ShapatinAA: Well, I think we shouldn't spam errors before this block or in the process of handling exceptions. I think if we're looping for 600 seconds in this block, we would be doing it by the same/obvious error(s).

ShapatinAA: Or do you suggest adding debug loggining?

gpMgmt/bin/gpactivatestandby
ShapatinAA: To make this function more robust. Initially I added this loop here because I haven't noticed `utility=true` connection type, and thought that segment downtime could brake it. But now it's here for additional layer of `defence`.

gpMgmt/bin/gpactivatestandby
ShapatinAA: Without handling port that way, we can not determine if standby was promoted or not, which is name of the game. It happens because `get_config` queries would fail (it uses `PGPORT` env variable). But standby would be promoted (as `pg_ctl` deals with configs not env variables)

whitehawk:
Did you face such behave test failure, is it related to your patch?

<details>

<summary>log</summary>

```
  Scenario: master can be made on dir with trailing slash
    Given the database is running                                            # 0.015s
    And the standby is not initialized                                       # 0.101s
    When the user runs gpinitstandby with options "-S /tmp/standby_data/"    # 19.821s
    Then gpinitstandby should return a return code of 0                      # 0.000s
    And verify the standby master entries in catalog                         # 0.168s
    When the master goes down                                                # 0.124s
    And the user runs gpactivatestandby with options "-d /tmp/standby_data/" # 3.578s
    Then gpactivatestandby should return a return code of 0                  # 0.000s
    And verify the standby master is now acting as master                    # 0.091s
    And clean up and revert back to original master                          # 17.061s
----
CAPTURED STDOUT: after_scenario
HOOK-ERROR in after_scenario: ExecutionError: ExecutionError: 'non-zero rc: 2' occurred.  Details: '$GPHOME/bin/gpstart -a'  cmd had rc=2 completed=True halted=False
  stdout='20260812:05:19:53:015088 gpstart:cdw:gpadmin-[INFO]:-Starting gpstart with args: -a
20260812:05:19:53:015088 gpstart:cdw:gpadmin-[INFO]:-Gathering information and validating the environment...
20260812:05:19:53:015088 gpstart:cdw:gpadmin-[CRITICAL]:-gpstart failed. (Reason='[Errno 2] No such file or directory: '/data/gpdata/master/gpseg-1/postgresql.conf'') exiting...
'
  stderr=''
----
----
CAPTURED STDOUT: before_scenario
HOOK-ERROR in before_scenario: ExecutionError: ExecutionError: 'non-zero rc: 2' occurred.  Details: '$GPHOME/bin/gpstart -a'  cmd had rc=2 completed=True halted=False
  stdout='20260812:05:19:54:015096 gpstart:cdw:gpadmin-[INFO]:-Starting gpstart with args: -a
20260812:05:19:54:015096 gpstart:cdw:gpadmin-[INFO]:-Gathering information and validating the environment...
20260812:05:19:54:015096 gpstart:cdw:gpadmin-[CRITICAL]:-gpstart failed. (Reason='[Errno 2] No such file or directory: '/data/gpdata/master/gpseg-1/postgresql.conf'') exiting...
'
  stderr=''
----
----
CAPTURED STDOUT: after_scenario
HOOK-ERROR in after_scenario: ExecutionError: ExecutionError: 'non-zero rc: 2' occurred.  Details: '$GPHOME/bin/gpstart -a'  cmd had rc=2 completed=True halted=False
  stdout='20260812:05:19:54:015104 gpstart:cdw:gpadmin-[INFO]:-Starting gpstart with args: -a
20260812:05:19:54:015104 gpstart:cdw:gpadmin-[INFO]:-Gathering information and validating the environment...
20260812:05:19:54:015104 gpstart:cdw:gpadmin-[CRITICAL]:-gpstart failed. (Reason='[Errno 2] No such file or directory: '/data/gpdata/master/gpseg-1/postgresql.conf'') exiting...
'
  stderr=''
----

  @concourse_cluster
  Scenario: tablespaces work on a multi-host environment
    Given the database is running
    And the standby is not initialized
    And a tablespace is created with data
    When the user runs gpinitstandby with options " "
    And verify the standby master entries in catalog
    When the master goes down
    Then the user runs gpactivatestandby with options " "
    And gpactivatestandby should return a return code of 0
    And verify the standby master is now acting as master
    And the tablespace is valid on the standby master
    And clean up and revert back to original master


Errored scenarios:
  test/behave/mgmt_utils/gpactivatestandby.feature:102  master can be made on dir with trailing slash
  test/behave/mgmt_utils/gpactivatestandby.feature:120  tablespaces work on a multi-host environment

0 features passed, 0 failed, 1 error, 24 skipped
6 scenarios passed, 0 failed, 2 hook_error, 661 skipped
91 steps passed, 0 failed, 8661 skipped, 12 untested
Took 6min 33.194s

```

</details>

gpMgmt/bin/gpactivatestandby
whitehawk: I think it is a good time to start. As we have different error return codes, better to document them.

ShapatinAA:
>Did you face such behave test failure, is it related to your patch?

I have been facing this issues (`after_scenario` step failure) in process of creating this patch, but I got rid of them.
And this issue exactly (with before_scenario and after_scenario simultaneously) I haven't faced. Can you give context how did it happen?

P.S. `master can be made on dir with trailing slash` is one of my previous patches and it seems to leave the cluster in unstable condition, thou if it's last one - `after_scenario` and `And clean up and revert back to original master` steps pass ok.

whitehawk:
> Can you give context how did it happen?

I've build docker image by:
```
docker build -t gpdb6_regress:gg_452 -f ci/Dockerfile.ubuntu .
```
launched cluster with this docker image (coordinator + 3 segments) and started the test via:
```
source /usr/local/greengage-db-devel/greengage_path.sh
cd ~/gpdb_src/
./configure --disable-gpcloud
make -C ./gpMgmt -f Makefile.behave behave tags=gpactivatestandby
```

ShapatinAA:
>Did you face such behave test failure, is it related to your patch?

I got it. `concourse_cluster` tagged test was being executed after `master can be made on dir with trailing slash`, which, as I said, left the cluster in somehow unstable state (is it even possible to execute `concourse_cluster` tagged scenarios on demo-cluster?). That is why all errors popped up. It has nothing to do with this patch and should be made into separate issue.

UPD: If you remove @concourse_cluster scenario, it will pass fine. Or if you launch tests as readme in ci suggests

gpMgmt/bin/gpactivatestandby
ShapatinAA: done

gpMgmt/bin/gpactivatestandby
whitehawk: `pass` is usually used as a stub to please python when otherwise it would throw a syntax error. So, I think it should be remove here.

gpMgmt/bin/gpactivatestandby
whitehawk: This sounds like not related to the scope of the PR, unless a scenario exists that can show why this re-try is needed for the exit code selection logic. Does such exist?

gpMgmt/bin/gpactivatestandby
whitehawk: In this case please add these details into the PR description

whitehawk:
>  It has nothing to do with this patch and should be made into separate issue.

Ok, please create a separate ticket in this case.

whitehawk:
Please correct the PR description to fit into 80-char limit.

gpMgmt/bin/gpactivatestandby
ShapatinAA: Added.

gpMgmt/bin/gpactivatestandby
ShapatinAA: You right, the only thing I came up with is connection problems, but on localhost they doubtable. Fixed.

gpMgmt/bin/gpactivatestandby
ShapatinAA: Okay, fixed.

gpMgmt/bin/gpactivatestandby
ShapatinAA: removed this logic

gpMgmt/bin/gpactivatestandby
ShapatinAA: removed this logic

whitehawk:
In the description:
> underlying utils for promotion, i.e. pg_ctl, use port from configuration file, not env.
variable, and hostname is always localhost for them for obvious reasons

Strictly speaking, it is not so - refer to [test_postmaster_connection()](https://github.com/GreengageDB/greengage/blob/4128bfd238cce7cfc996133943cff46a27e5c3e1/src/bin/pg_ctl/pg_ctl.c#L573), it gets the port from postmaster.pid file. Maybe we should follow this path as well?

ShapatinAA:
>Maybe we should follow this path as well?

Unfortunately, we can not rely on `postmaster.pid` file, as standby could be down, and we're activating it with `-f` flag. In this way `gpstart` launches it by taking parameters from configuration file, so I think it should be okay to also do it this way.

whitehawk:
> > Maybe we should follow this path as well?
> 
> Unfortunately, we can not rely on `postmaster.pid` file, as standby could be down, and we're activating it with `-f` flag. In this way `gpstart` launches it by taking parameters from configuration file, so I think it should be okay to also do it this way.

In this case please fix the part of description with `pg_ctl`, as `pc_ctl` doesn't get parameters directly from postgresql.conf.

bimboterminator1 (approved):

whitehawk (approved):

PR #561, opened 2026-08-10
GG-633: Add script to automate resource group v2 tests

hilltracer:
Resource group v2 tests were not automated, although they are required for 
Greengage 7.x deployments using cgroup v2.

Adapt concourse/scripts/ic_gpdb_resgroup_v2.bash for the Docker-based test
environment. Use concourse/scripts/ic_gpdb.bash as a base and reuse setup
ideas from ci/scripts/run_resgroup_test.bash, adjusted for resource group v2.
Add the required cgroup v2 and IO_LIMIT test environment checks.

Run installcheck-resgroup-v2 directly in src/test/isolation2. Do not use
top-level target yet, because it recurses into gpcontrib, while
gg_wait_sampling does not provide compatible resource group v2 target.

Ticket: GG-633


hilltracer:
### Permission changing checks
Before `docker run ... ic_gpdb_resgroup_v2.bash`:
```sh
den@dgarsh-u22-1:~$ stat -c '%A %a %n' /sys/fs/cgroup /sys/fs/cgroup/cgroup.procs /sys/fs/cgroup/gpdb 2>/dev/null
dr-xr-xr-x 555 /sys/fs/cgroup
-rw-r--r-- 644 /sys/fs/cgroup/cgroup.procs
```
After `docker run ... ic_gpdb_resgroup_v2.bash`:
```sh
den@dgarsh-u22-1:~$ stat -c '%A %a %n' /sys/fs/cgroup /sys/fs/cgroup/cgroup.procs /sys/fs/cgroup/gpdb 2>/dev/null
dr-xr-xr-x 555 /sys/fs/cgroup
-rw-rw-rw- 666 /sys/fs/cgroup/cgroup.procs
drwxrwxrwx 777 /sys/fs/cgroup/gpdb
```

Script intentionally does not change permissions on `/sys/fs/cgroup` itself. It only prepares GPDB subtree and the parent `cgroup.procs` file required by cgroup v2:
```sh
chmod a+rw "$basedir/cgroup.procs"
```
Change `/sys/fs/cgroup/cgroup.procs` from `644` to `666`. GPDB needs write access to the common parent `cgroup.procs` when moving backend processes into resource group cgroups.

```sh
mkdir -p "$gpdb_cgroup"
```
Create `/sys/fs/cgroup/gpdb` if it does not already exist.

```sh
chmod a+rwx "$gpdb_cgroup"
chmod -R a+rwX "$gpdb_cgroup"
```
Make only `/sys/fs/cgroup/gpdb` and its contents writable for the test user.

hilltracer:
### Problem with resgroup_io_limit_alter_in_transaction test
Current run lead to the problem (which is not related with the patch):

<details>
<summary>regression diff</summary>

```diff
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/expected/resgroup/resgroup_io_limit_alter_in_transaction.out /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/results/resgroup/resgroup_io_limit_alter_in_transaction.out
--- /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/expected/resgroup/resgroup_io_limit_alter_in_transaction.out	2026-08-07 06:32:52.524769936 +0000
+++ /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/results/resgroup/resgroup_io_limit_alter_in_transaction.out	2026-08-07 06:32:52.536770254 +0000
@@ -165,24 +186,24 @@
 (1 row)
 
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'reset', dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault 
 ------------
- Success:
+ Success:        
 (1 row)
 
 -- 5 Subtransaction rollback followed by the same io_limit value.
 -- The rolled back callback and the real callback carry the same value, the
 -- stored value compare keeps one and only one apply should matter.
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'reset', dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault 
 ------------
- Success:
+ Success:        
 (1 row)
 
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'skip', '', '', '', 1, 100, 0, dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault 
 ------------
- Success:
+ Success:        
 (1 row)
 
 BEGIN;
@@ -199,30 +220,30 @@
 COMMIT
 
 SELECT * FROM rg_alter_tran_status_io_limit;
- groupname       | io_limit
+ groupname       | io_limit                                     
 -----------------+----------------------------------------------
- rg_alter_tran   | 1663:rbps=5000,wbps=5000,riops=max,wiops=max
- rg_alter_tran_b | 1663:rbps=2000,wbps=2000,riops=max,wiops=max
+ rg_alter_tran   | 1663:rbps=5000,wbps=5000,riops=max,wiops=max 
+ rg_alter_tran_b | 1663:rbps=2000,wbps=2000,riops=max,wiops=max 
 (2 rows)
 
 SELECT check_cgroup_io_max('rg_alter_tran', 'pg_default', 'rbps=5242880000 wbps=5242880000 riops=max wiops=max');
- check_cgroup_io_max
+ check_cgroup_io_max 
 ------------
- t
+ t                   
 (1 row)
 
 -- Should be one in `num times hit`
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'status', dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault                                                                                                                                                                                                                 
 ------------
-
- Success: fault name:'resgroup_alter_on_commit' fault type:'skip' ddl statement:'' database name:'' table name:'' start occurrence:'1' end occurrence:'100' extra arg:'0' fault injection state:'triggered'  num times hit:'1'
+ 
+ Success: fault name:'resgroup_alter_on_commit' fault type:'skip' ddl statement:'' database name:'' table name:'' start occurrence:'1' end occurrence:'100' extra arg:'0' fault injection state:'triggered'  num times hit:'1' 
 (1 row)
 
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'reset', dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault 
 ------------
- Success:
+ Success:        
 (1 row)
 
 -- 6 ALTER changes io_limit and then changes it back in one transaction.
@@ -238,101 +259,103 @@
 COMMIT
 
 SELECT * FROM rg_alter_tran_status_io_limit;
- groupname       | io_limit
+ groupname       | io_limit                                     
 -----------------+----------------------------------------------
- rg_alter_tran   | 1663:rbps=5000,wbps=5000,riops=max,wiops=max
- rg_alter_tran_b | 1663:rbps=2000,wbps=2000,riops=max,wiops=max
+ rg_alter_tran   | 1663:rbps=5000,wbps=5000,riops=max,wiops=max 
+ rg_alter_tran_b | 1663:rbps=2000,wbps=2000,riops=max,wiops=max 
 (2 rows)
 
 SELECT check_cgroup_io_max('rg_alter_tran', 'pg_default', 'rbps=5242880000 wbps=5242880000 riops=max wiops=max');
- check_cgroup_io_max
+ check_cgroup_io_max 
 ------------
- t
+ t                   
 (1 row)
 
 -- 7 Two ALTERs set the same io_limit across two tablespaces in a different
 -- element order. The order-independent compare treats the two values as equal,
 -- so the duplicate is dropped and only one apply happens.
 select mkdir('/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_1') from gp_dist_random('gp_id') intersect select mkdir('/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_1') from gp_id;
- mkdir
+ mkdir 
 -------
- t
+ t     
 (1 row)
-
 select mkdir('/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_2') from gp_dist_random('gp_id') intersect select mkdir('/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_2') from gp_id;
- mkdir
+ mkdir 
 -------
- t
+ t     
 (1 row)
 
 CREATE TABLESPACE rg_io_limit_ts_1 LOCATION '/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_1';
 CREATE TABLESPACE
 CREATE TABLESPACE rg_io_limit_ts_2 LOCATION '/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_2';
 CREATE TABLESPACE
+
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'reset', dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault 
 ------------
- Success:
+ Success:        
 (1 row)
 
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'skip', '', '', '', 1, 100, 0, dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault 
 ------------
- Success:
+ Success:        
 (1 row)
 
 BEGIN;
 BEGIN
 ALTER RESOURCE GROUP rg_alter_tran SET io_limit 'rg_io_limit_ts_1:rbps=1000,wbps=1000;rg_io_limit_ts_2:rbps=2000,wbps=2000';
-ALTER RESOURCE GROUP
+ERROR:  io limit: tablespaces of io limit must locate at different disks, tablespace 'rg_io_limit_ts_1' and 'rg_io_limit_ts_2' have the same disk identifier.
+HINT:  either omit these tablespaces from the IO limit or mount them separately
 ALTER RESOURCE GROUP rg_alter_tran SET io_limit 'rg_io_limit_ts_2:rbps=2000,wbps=2000;rg_io_limit_ts_1:rbps=1000,wbps=1000';
-ALTER RESOURCE GROUP
+ERROR:  current transaction is aborted, commands ignored until end of transaction block
 COMMIT;
-COMMIT
+ROLLBACK
 
 SELECT check_cgroup_io_max('rg_alter_tran', 'rg_io_limit_ts_1', 'rbps=1048576000 wbps=1048576000 riops=max wiops=max');
- check_cgroup_io_max
+ check_cgroup_io_max 
 ------------
- t
+ f                   
 (1 row)
 
 SELECT check_cgroup_io_max('rg_alter_tran', 'rg_io_limit_ts_2', 'rbps=2097152000 wbps=2097152000 riops=max wiops=max');
- check_cgroup_io_max
+ check_cgroup_io_max 
 ------------
- t
+ f                   
 (1 row)
 
 -- Should be one in `num times hit`
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'status', dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault                                                                                                                                                                                                           
 ------------
-
- Success: fault name:'resgroup_alter_on_commit' fault type:'skip' ddl statement:'' database name:'' table name:'' start occurrence:'1' end occurrence:'100' extra arg:'0' fault injection state:'triggered'  num times hit:'1'
+ 
+ Success: fault name:'resgroup_alter_on_commit' fault type:'skip' ddl statement:'' database name:'' table name:'' start occurrence:'1' end occurrence:'100' extra arg:'0' fault injection state:'set'  num times hit:'0' 
 (1 row)
 
 SELECT gp_inject_fault('resgroup_alter_on_commit', 'reset', dbid) FROM gp_segment_configuration WHERE content = -1 AND role = 'p';
- gp_inject_fault
+ gp_inject_fault 
 ------------
- Success:
+ Success:        
 (1 row)
 
 -- Remove the tablespace dependency before the tablespaces are dropped.
 ALTER RESOURCE GROUP rg_alter_tran SET io_limit '-1';
 ALTER RESOURCE GROUP
+
 DROP TABLESPACE rg_io_limit_ts_1;
 DROP TABLESPACE
 DROP TABLESPACE rg_io_limit_ts_2;
 DROP TABLESPACE
+
 select rmdir('/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_1') from gp_dist_random('gp_id') intersect select rmdir('/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_1') from gp_id;
- rmdir
+ rmdir 
 -------
- t
+ t     
 (1 row)
-
 select rmdir('/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_2') from gp_dist_random('gp_id') intersect select rmdir('/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace/rg_io_limit_ts_2') from gp_id;
- rmdir
+ rmdir 
 -------
- t
+ t     
 (1 row)
 
 -- clean

```


</details>

As WA we can temporally use removing of wrong scenarios
```sh
sed -i '/^-- 4 Several ALTERs of the same limit type in one transaction/,/^-- clean$/{/^-- clean$/!d}' \
src/test/isolation2/input/resgroup/resgroup_io_limit_alter_in_transaction.source \
src/test/isolation2/output/resgroup/resgroup_io_limit_alter_in_transaction.source
```

hilltracer:
### Main test:
```bash
mkdir -p gpdemo-datadirs testtablespace logs
chmod -R 777 gpdemo-datadirs testtablespace logs

docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash  
```
All tests passed and we should see 0 in logs/.exitcode :
```sh
======================
 All 31 tests passed. 
======================

make: Leaving directory '/home/gpadmin/gpdb_src/src/test/isolation2'

real    11m7.127s
user    13m4.961s
sys     1m19.191s
+ record_exitcode
+ local exitcode=0
+ '[' -d /logs ']'
+ '[' -w /logs ']'
+ echo 0
+ exit 0
den@dgarsh-u22-1:~/greengage$ cat logs/.exitcode 
0
```

hilltracer:
### Security checks

<details>
<summary>spoiler</summary>

- Invalid `TEST_OS`:
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=debian \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
```
result:
```sh
FATAL: TEST_OS is set to an invalid value: debian; configure TEST_OS to be centos or ubuntu

den@dgarsh-u22-1:~/greengage$ cat logs/.exitcode 
1
```

- Non-root container user:
```bash
sudo groupadd gpadmin
sudo useradd -m -s /bin/bash -g gpadmin gpadmin
sudo usermod -aG sudo gpadmin
sudo passwd gpadmin


docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  --user 1000:1000 \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
```

```sh
FATAL: resource group v2 tests must run as root to configure cgroup v2
```
Note: use a numeric UID here. Using a username requires that the user already exists in the Docker image.

- Fake cgroup path:
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  -e CGROUP_BASEDIR=/tmp/not-cgroup \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  bash -lc 'mkdir -p /tmp/not-cgroup; /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash'
```
result:
```sh
FATAL: /tmp/not-cgroup is not a cgroup v2 mount
```

- Read-only cgroup mount:
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:ro \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
```
result
```sh
chmod: changing permissions of '/sys/fs/cgroup/cgroup.procs': Read-only file system
```

- No host cgroup namespace (why `--cgroupns=host` is required?):
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
```
result
```sh
============== running regression test queries        ==============                                                                                                                           
test setup                        ... ok         7548 ms                                                                                                                                       
test resgroup/resgroup_auxiliary_tools_v2 ... FAILED (test process exited with exit code 1)    22402 ms (diff  133 ms)                                                                         
psql: error: could not connect to server: No such file or directory                                                                                                                            
        Is the server running locally and accepting                                                                                                                                            
        connections on Unix domain socket "/tmp/.s.PGSQL.7000"?                                                                                                                                
...
+GP_IGNORE:20260810:04:59:45:016193 gpstart:7f925ad3f601:gpadmin-[INFO]:-CoordinatorStart pg_ctl cmd is env GPSESSID=0000000000 GPERA=None $GPHOME/bin/pg_ctl -D /home/gpadmin/gpdb_src/gpAux/g
pdemo/datadirs/qddir/demoDataDir-1 -l /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs/qddir/demoDataDir-1/log/startup.log -w -t 600 -o " -c gp_role=utility " start
+GP_IGNORE:20260810:04:59:45:016193 gpstart:7f925ad3f601:gpadmin-[CRITICAL]:-Failed to start Coordinator instance in admin mode
+GP_IGNORE:20260810:04:59:45:016193 gpstart:7f925ad3f601:gpadmin-[CRITICAL]:-Error occurred: non-zero rc: 1
```

- No `gpdemo-datadirs` bind mount:
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
```
result:
```sh
FATAL: /home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs is on overlayfs; bind mount it from a regular host filesystem for IO_LIMIT tests
```

- No `testtablespace` bind mount:
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
```
result:
```sh
FATAL: /home/gpadmin/gpdb_src/src/test/isolation2/testtablespace is on overlayfs; IO_LIMIT tests require a regular filesystem
```

- cgroup v2 controller '$controller' is not available in
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  -e CGROUP_BASEDIR=/tmp/fake-cgroup \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  bash -lc '
    mkdir -p /tmp/fake-cgroup/gpdb
    touch /tmp/fake-cgroup/cgroup.controllers
    touch /tmp/fake-cgroup/cgroup.procs
    touch /tmp/fake-cgroup/cgroup.subtree_control
    chmod 444 /tmp/fake-cgroup/cgroup.subtree_control
    /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
  '
```
result
```sh
FATAL: cgroup v2 controller 'cpu' is not available in /tmp/fake-cgroup/cgroup.controllers
```

- failed to enable cgroup v2 controller '$controller'
Idea: `cgroup.subtree_control` passes `-w`, but the `echo +cpu > ...` entry fails. The easiest way is to make `cgroup.subtree_control` a directory.
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  -e CGROUP_BASEDIR=/tmp/fake-cgroup \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  bash -lc '
    mkdir -p /tmp/fake-cgroup/gpdb
    echo "cpu cpuset io memory pids" > /tmp/fake-cgroup/cgroup.controllers
    touch /tmp/fake-cgroup/cgroup.procs
    mkdir -p /tmp/fake-cgroup/cgroup.subtree_control
    /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
  '
```
result:
```sh
FATAL: failed to enable cgroup v2 controller 'cpu';
```

- gpdb is not ready for resource group v2
```bash
docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=ubuntu \
  -e CGROUP_BASEDIR=/tmp/fake-cgroup \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  bash -lc '
    mkdir -p /tmp/fake-cgroup/gpdb
    echo "cpu cpuset io memory pids" > /tmp/fake-cgroup/cgroup.controllers
    touch /tmp/fake-cgroup/cgroup.procs
    touch /tmp/fake-cgroup/cgroup.subtree_control
    /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
  '
```
result:
```sh
FATAL: /tmp/fake-cgroup/gpdb is not ready for resource group v2
```

</details>

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,181 @@
 
 set -eox pipefail
 
-./ccp_src/scripts/setup_ssh_to_cluster.sh
+CWDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${CWDIR}/common.bash"
```
KnightMurloc, line 6: Why not just:
```bash
source gpdb_src/concourse/scripts/common.bash
```
We can expect that the script would work in a specific environment.

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,181 @@
...
+    cd /home/gpadmin
+    assert_real_filesystem "$GPDB_DEMO_DATADIRS"
+    assert_real_filesystem "$ISOLATION2_TESTTABLESPACE"
+    setup_cgroup_v2
+
+    time install_and_configure_gpdb
+    time setup_gpadmin_user
+    time make_cluster
+    time gen_env
+    time run_test
```
KnightMurloc, line 126: Why do we need time?

red1452:
Why tests failed on my PC?
```diff
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/expected/resgroup/resgroup_views.out /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/results/resgroup/resgroup_views.out
--- /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/expected/resgroup/resgroup_views.out    2026-08-10 14:56:32.605454173 +0000
+++ /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/results/resgroup/resgroup_views.out     2026-08-10 14:56:32.606280779 +0000
@@ -14,7 +14,7 @@
 select groupname , groupid , cpu_usage , memory_usage from gp_toolkit.gp_resgroup_status_per_host s join gp_segment_configuration c on s.hostname=c.hostname and c.content=-1 and role='p' where groupname='default_group';
  groupname     | groupid | cpu_usage | memory_usage 
 ---------------+---------+-----------+--------------
- default_group | 6437    | 0.00      | 0.00         
+ default_group | 6437    | 0.00      | 1.00         
 (1 row)
 
 select groupname , groupid , segment_id , vmem_usage from gp_toolkit.gp_resgroup_status_per_segment where groupname='default_group' order by segment_id;
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/gpadmin/gpdb_src/src/test/isolation2/expected/resgroup/resgroup_cpu_max_percent.out /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/results/resgroup/resgroup_cpu_max_percent.out
--- /home/gpadmin/gpdb_src/src/test/isolation2/expected/resgroup/resgroup_cpu_max_percent.out   2026-08-10 15:01:56.558741030 +0000
+++ /home/gpadmin/gpdb_src/src/test/isolation2/resgroup/results/resgroup/resgroup_cpu_max_percent.out   2026-08-10 15:01:56.564741131 +0000
@@ -244,7 +244,7 @@
 SELECT verify_cpu_usage('rg1_cpu_test', 90, 10);
  verify_cpu_usage 
 ------------
- t                
+ f                
 (1 row)
 
 GP_IGNORE:-- start_ignore
```

hilltracer:
> Why tests failed on my PC?

Please specify how did you run it?

And please show your PC parameters. The parameters something like:

<details>
<summary>spoiler</summary>

```sh
#!/bin/bash
set -e

echo "=== OS & Kernel Info ==="
uname -a
echo
lsb_release -a 2>/dev/null || cat /etc/os-release
echo
systemctl status 2>&1 | head -5 || echo "systemd info not available"
echo

echo "=== CPU Info ==="
lscpu
echo "CPU Cores: $(nproc)"
echo

echo "=== Memory Info ==="
free -h
echo
cat /proc/meminfo | grep -E "MemTotal|MemFree|MemAvailable|SwapTotal|SwapFree"
```

update: if its possible show please:

```sh
cat /sys/fs/cgroup/gpdb/cpuset.cpus.effective
cat /sys/fs/cgroup/gpdb/cpu.max
```

</details>

p.s.:
The first test expects default_group to have no memory usage at the time of the check (memory_usage = 0). Maybe some backend or other activity was running in default_group at that moment, so it reported 1 MB.

The second test generates CPU load and expects around 90% CPU usage. Maybe on your particular machine the test was not able to generate enough load to reach that level.

red1452:
> > Why tests failed on my PC?
> 
> Please specify how did you run it?

I got your branch, added `SKIP_UNITTESTS=1` and ran commands from readme:
```
mkdir -p gpdemo-datadirs testtablespace logs
chmod -R 777 gpdemo-datadirs testtablespace logs

docker run --name gpdb7_resgroup_v2 --rm -it -e TEST_OS=ubuntu \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
```

After the second run, tests failed with the same result.

> And please show your PC parameters. The parameters something like:

<details>
<summary>spoiler</summary>

```
=== OS & Kernel Info ===
Linux evgeniy-pc 7.0.0-28-generic #28~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jul  1 15:50:57 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

Distributor ID: Ubuntu
Description:    Ubuntu 24.04.3 LTS
Release:        24.04
Codename:       noble

● evgeniy-pc
    State: degraded
    Units: 669 loaded (incl. loaded aliases)
     Jobs: 0 queued
   Failed: 2 units

=== CPU Info ===
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             46 bits physical, 48 bits virtual
  Byte Order:                Little Endian
CPU(s):                      20
  On-line CPU(s) list:       0-19
Vendor ID:                   GenuineIntel
  Model name:                13th Gen Intel(R) Core(TM) i9-13900H
    CPU family:              6
    Model:                   186
    Thread(s) per core:      2
    Core(s) per socket:      14
    Socket(s):               1
    Stepping:                2
    CPU(s) scaling MHz:      14%
    CPU max MHz:             5400,0000
    CPU min MHz:             400,0000
    BogoMIPS:                5990,40
    Flags:                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopo
                             logy nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c
                              rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb ssbd ibrs ibpb stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid rdseed adx smap clflushopt cl
                             wb intel_pt sha_ni xsaveopt xsavec xgetbv1 xsaves split_lock_detect user_shstk avx_vnni dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp hwp_pkg_req hfi vnmi umip pku ospke waitpkg gfni vaes vpc
                             lmulqdq rdpid movdiri movdir64b fsrm md_clear serialize pconfig arch_lbr ibt flush_l1d arch_capabilities
Virtualization features:     
  Virtualization:            VT-x
Caches (sum of all):         
  L1d:                       544 KiB (14 instances)
  L1i:                       704 KiB (14 instances)
  L2:                        11,5 MiB (8 instances)
  L3:                        24 MiB (1 instance)
NUMA:                        
  NUMA node(s):              1
  NUMA node0 CPU(s):         0-19
Vulnerabilities:             
  Gather data sampling:      Not affected
  Ghostwrite:                Not affected
  Indirect target selection: Not affected
  Itlb multihit:             Not affected
  L1tf:                      Not affected
  Mds:                       Not affected
  Meltdown:                  Not affected
  Mmio stale data:           Not affected
  Old microcode:             Not affected
  Reg file data sampling:    Mitigation; Clear Register File
  Retbleed:                  Not affected
  Spec rstack overflow:      Not affected
  Spec store bypass:         Mitigation; Speculative Store Bypass disabled via prctl
  Spectre v1:                Mitigation; usercopy/swapgs barriers and __user pointer sanitization
  Spectre v2:                Mitigation; Enhanced / Automatic IBRS; IBPB conditional; PBRSB-eIBRS SW sequence; BHI BHI_DIS_S
  Srbds:                     Not affected
  Tsa:                       Not affected
  Tsx async abort:           Not affected
  Vmscape:                   Mitigation; IBPB before exit to userspace
CPU Cores: 20

=== Memory Info ===
               total        used        free      shared  buff/cache   available
Mem:            31Gi       8,5Gi       2,5Gi       3,3Gi        23Gi        22Gi
Swap:          511Mi       608Ki       511Mi

MemTotal:       32576644 kB
MemFree:         2574652 kB
MemAvailable:   23708308 kB
SwapTotal:        524284 kB
SwapFree:         523676 kB

=== Cgroup Info ===
0-19
1800000 100000
```

</details>

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,181 @@
 
 set -eox pipefail
 
-./ccp_src/scripts/setup_ssh_to_cluster.sh
+CWDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${CWDIR}/common.bash"
 
-CLUSTER_NAME=$(cat ./cluster_env_files/terraform/name)
+CGROUP_BASEDIR=${CGROUP_BASEDIR:-/sys/fs/cgroup}
```
KnightMurloc: Can control groups have a different path?

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,181 @@
...
 
-enable_cgroup_subtree_control() {
-    local gpdb_host_alias=$1
-    local basedir=$CGROUP_BASEDIR
+record_exitcode() {
+    local exitcode=$?
+
+    if [ -d /logs ] && [ -w /logs ]; then
+        echo "$exitcode" > /logs/.exitcode || true
+    fi
```
KnightMurloc: Can we make the script exit with a non-zero exit code in case of error as other our scripts do?

hilltracer:
> After the second run, tests failed with the same result.

I tried to run the test on my local machine with 16 CPUs and have repro for test1:

```sql
  groupname     | groupid | cpu_usage | memory_usage
 ---------------+---------+-----------+--------------
- default_group | 6437    | 0.00      | 0.00
+ default_group | 6437    | 0.00      | 0.03
```

On a clean yandex cloud host with 8 CPUs I cannot reproduce it. On your 20 CPU machine it reproduces consistently.

test 1 expects used memory to be exactly zero, but in cgroup v2 this value comes from live memory.current. A small backend allocation during the query is enough to make it non-zero.

test 2 creates fixed CPU workload and expects the resource group to reach around 90 percent CPU usage. This is also environment-sensitive: on machines with more CPUs, hybrid laptop CPUs etc. the fixed workload may not saturate enough CPU to fit the expected range.

I do not think we can reliably stabilize these tests only in the runner script. We can try workarounds, like running on very clean host, limiting Docker to a smaller CPU set, or trying to reduce background memory usage, but these are still workarounds. The proper fix should be in the tests smth like: use range instead of exact zero for memory_usage, and make CPU test scale with available CPUs or check cgroup configuration instead of expecting fixed live CPU percentage.

So my suggestion is to leave this PR as it is, and if it starts to work unstable on CI, create the task to stabilize flaky test.
@red1452, @KnightMurloc WDYT?

red1452 (dismissed):

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,181 @@
...
+        fi
+
+        for controller in cpu cpuset io memory pids; do
+            enable_cgroup_controller "$controller"
+        done
+
+        chmod -R a+rwX "$gpdb_cgroup"
+    fi
+
+    assert_gpdb_cgroup_ready
```
KnightMurloc: Why so complicated? Why steps from documentation is not enough?
```bash
mkdir -p /sys/fs/cgroup/gpdb.service
echo "+cpuset +io +cpu +memory" | tee -a /sys/fs/cgroup/cgroup.subtree_control
chown -R gpadmin:gpadmin /sys/fs/cgroup/gpdb.service

chmod a+w /sys/fs/cgroup/cgroup.procs
```

KnightMurloc (commented):
If the script has been completely rewritten, is there any point in keeping it in `concourse/scripts/`? Maybe we should move it to our other ci scripts?

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: I've used ic_gpdb.bash as template. With source gpdb_src/concourse/scripts/common.bash the script would only work when it is started from specific directory. I think it's not bad to keep ic_gpdb.bash style.

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: I've used ic_gpdb.bash as template. I think it's not bad to see in the output how long each command took.

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: Usually no.

But I used it CGROUP_BASEDIR for diagnostics in my `security checks`, see `-e CGROUP_BASEDIR=/tmp/fake-cgroup`.

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: It already exits with the original non-zero code. This only writes code to /logs/.exitcode when /logs is mounted and writable. The trap function ends with:
https://github.com/GreengageDB/greengage/blob/25851acb75e73b943dc9accdf8ad432bb480e74d/concourse/scripts/ic_gpdb_resgroup_v2.bash#L28

For example:

<details>
<summary>spoiler</summary>

```sh
den@dgarsh-u22-1:~/greengage$ docker run --rm --name gpdb7_resgroup_v2 -it -e TEST_OS=debian \
  --sysctl "kernel.sem=500 1024000 200 4096" \
  --privileged \
  --cgroupns=host \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
  -v "$PWD/gpdemo-datadirs":/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs:rw \
  -v "$PWD/testtablespace":/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace:rw \
  -v "$PWD/logs":/logs:rw \
  gpdb7_u22:latest \
  /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
+++ dirname /home/gpadmin/gpdb_src/concourse/scripts/ic_gpdb_resgroup_v2.bash
++ cd /home/gpadmin/gpdb_src/concourse/scripts
++ pwd
+ CWDIR=/home/gpadmin/gpdb_src/concourse/scripts
+ source /home/gpadmin/gpdb_src/concourse/scripts/common.bash
+ CGROUP_BASEDIR=/sys/fs/cgroup
+ OPTIMIZER=off
+ RESGROUP_STATEMENT_MEM=125MB
+ TEST_OS=debian
+ GPDB_DEMO_DATADIRS=/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs
+ ISOLATION2_TESTTABLESPACE=/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace
+ trap record_exitcode EXIT
+ _main
+ assert_root
++ id -u
+ '[' 0 -ne 0 ']'
+ '[' -z debian ']'
+ [[ debian != centos* ]]
+ [[ debian != ubuntu* ]]
+ fatal 'TEST_OS is set to an invalid value: debian; configure TEST_OS to be centos or ubuntu'
+ echo 'FATAL: TEST_OS is set to an invalid value: debian; configure TEST_OS to be centos or ubuntu'
FATAL: TEST_OS is set to an invalid value: debian; configure TEST_OS to be centos or ubuntu
+ exit 1
+ record_exitcode
+ local exitcode=1
+ '[' -d /logs ']'
+ '[' -w /logs ']'
+ echo 1
+ exit 1
den@dgarsh-u22-1:~/greengage$ echo $?
1
```

</details>

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: Still do not see this doc instructions. Thanks.

I used +pids because GPDB also enables it here:
https://github.com/GreengageDB/greengage/blob/ab129676775f977cb54d8784d5bd5303a69459ab/src/backend/utils/resgroup/cgroup-ops-linux-v2.c#L193-L197
But looks like it's not necessary.

I also moved cgroup setup after gpadmin setup. This lets the script use chown gpadmin:gpadmin.

I removed several extra checks as well to keep the setup simpler and closer to the documentation.
I tested the solution the same way as I [described](https://github.com/GreengageDB/greengage/pull/561#issuecomment-5238564831) in earlier message.

hilltracer:
> If the script has been completely rewritten, is there any point in keeping it in `concourse/scripts/`? Maybe we should move it to our other ci scripts?

I would keep it in concourse/scripts for now. It still follows same pattern as ic_gpdb.bash, so keeping them side by side makes sense for me.

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: Removed  CGROUP_BASEDIR since most of checks was removed.

concourse/scripts/ic_gpdb_resgroup_v2.bash
KnightMurloc: But what point in writing it in the file then?

hilltracer: I've done it because ci/scripts/run_resgroup_test.bash do the same. It was used on our CI. But looks like now its not necesarry because it [was removed](https://github.com/GreengageDB/greengage-ci/commit/edae83117d28bbeb5d886cf2337073a94db69c48#diff-904229bad61a6e24c566f69fd3a164576686e65dddf9df410454d2ec5392b8bcR69-L177). From CI processing. So I removed it.

KnightMurloc:
> [remove logs](https://github.com/GreengageDB/greengage/pull/561/commits/7a8908cf7c544e94be8d63a364cafda4a3dcfe43)

But what about other logs? Cluster logs and tests logs. Should they be saved after the test by a script, or will it be done in some other way?

hilltracer:
> But what about other logs? Cluster logs and tests logs. Should they be saved after the test by a script, or will it be done in some other way?

we can consider current CI v1 example here:
https://github.com/GreengageDB/greengage/blob/8cd6e588836c7001cc7731369fc9f33db8bce4fb/.github/workflows/greengage-ci.yml#L98

It has artifact [collect](https://github.com/GreengageDB/greengage-ci/blob/edb87b04590440a7892a46acc27bfa018b46a15a/.github/workflows/greengage-reusable-tests-resgroup.yml#L237) and [upload](https://github.com/GreengageDB/greengage-ci/blob/edb87b04590440a7892a46acc27bfa018b46a15a/.github/workflows/greengage-reusable-tests-resgroup.yml#L252
) steps.

So instead of using [run_resgroup_test.bash](https://github.com/GreengageDB/greengage-ci/blob/edb87b04590440a7892a46acc27bfa018b46a15a/.github/workflows/greengage-reusable-tests-resgroup.yml#L235) we can run `docker run -v "$PWD/logs:/logs" ... ic_gpdb_resgroup_v2.bash`. But I think it's not related with current script.

I'm going to ask CI team review for this patch after your review will finished.

KnightMurloc (dismissed):

red1452 (dismissed):

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,119 @@
...
-    local gpdb_master_alias=$1
+gen_env() {
+    cat > /opt/run_test.sh <<-EOF
+		trap look4diffs ERR
 
-    if [ "${TEST_BINARY_SWAP}" != "true" ]; then
-        return 0
-    fi
+		function look4diffs() {
+		    diff_files=\`find .. -name regression.diffs\`
```
magf:
```suggestion
		    diff_files=\$(find .. -name regression.diffs)
```
Two reasons:
- backticks are the legacy quoting style (deprecated in favor of `$(...)`)
- mixing backticks with `\$(...)` elsewhere in this same heredoc is inconsistent

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,119 @@
...
+        fatal "failed to enable cgroup v2 controllers"
     fi
 
-    ssh $gpdb_master_alias mkdir -p /tmp/local/greengage-db-devel
-    ssh $gpdb_master_alias tar -zxf - -C /tmp/local/greengage-db-devel \
-        < binary_swap_gpdb/bin_gpdb.tar.gz
-    ssh $gpdb_master_alias sed -i -e "s@/usr/local@/tmp/local@" \
-        /tmp/local/greengage-db-devel/greengage_path.sh
+    chown -R gpadmin:gpadmin /sys/fs/cgroup/gpdb
+    chmod a+w /sys/fs/cgroup/cgroup.procs
```
magf, line 64: Save original proc cgroup mode (will be restored before exit):
```suggestion
    chown -R gpadmin:gpadmin /sys/fs/cgroup/gpdb
    ORIG_CGROUP_PROCS_MODE=$(stat -c %a /sys/fs/cgroup/cgroup.procs)
    chmod a+w /sys/fs/cgroup/cgroup.procs
```

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,119 @@
...
 }
 
-setup_binary_swap_test() {
-    local gpdb_master_alias=$1
+# Create GPDB cgroup, enable parent controllers, and grant gpadmin access.
+setup_cgroup_v2() {
+    if [ ! -f /sys/fs/cgroup/cgroup.controllers ]; then
+        fatal "/sys/fs/cgroup is not a cgroup v2 mount"
+    fi
+
```
magf, line 50: Register a cleanup function to run on exit before touching host cgroups
```suggestion

trap cleanup_cgroup_v2 EXIT

```

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -2,113 +2,119 @@
...
+    fs_type=$(stat -f -c %T "$path")
+
+    if [ "$fs_type" = "overlayfs" ] || [ "$fs_type" = "tmpfs" ]; then
+        fatal "$path is on $fs_type; IO_LIMIT tests require a regular filesystem"
+    fi
 }
 
-setup_binary_swap_test() {
-    local gpdb_master_alias=$1
+# Create GPDB cgroup, enable parent controllers, and grant gpadmin access.
```
magf, line 45: Cleanup function for host cgroup changes:

```suggestion
# Restore cgroup.procs mode and remove the gpdb cgroup on exit.
cleanup_cgroup_v2() {
    local rc=$?
    [ -n "$ORIG_CGROUP_PROCS_MODE" ] && chmod "$ORIG_CGROUP_PROCS_MODE" /sys/fs/cgroup/cgroup.procs 2>/dev/null || true
    rmdir /sys/fs/cgroup/gpdb 2>/dev/null || true
    return "$rc"
}

# Create GPDB cgroup, enable parent controllers, and grant gpadmin access.
```

`|| true` on each step: cleanup must not fail even if a step errors, otherwise a failed cleanup would trip `set -e` again inside the trap.

ci/readme.md
```diff
@@ -72,6 +72,41 @@ for Rocky Linux:
 
 To use gdb inside the container, add the `--privileged` flag to the run command.
 
+## Resource group v2 isolation tests
+
+Resource group v2 tests require a Linux host with cgroup v2 enabled.
+`gpdemo-datadirs` and `testtablespace` must be directories on a regular host
+filesystem, not Docker overlay or tmpfs. Note that the script changes host
+cgroup settings.
```
magf:
```suggestion
cgroup settings; these changes are reverted automatically once the script
exits.
```
Accept only together with the cleanup trap suggestion.

ci/readme.md
```diff
@@ -72,6 +72,41 @@ for Rocky Linux:
...
+
+For Ubuntu:
+```bash
+mkdir -p gpdemo-datadirs testtablespace
+chmod -R 777 gpdemo-datadirs testtablespace
+
+docker run --name gpdb7_resgroup_v2 --rm -it -e TEST_OS=ubuntu \
```
magf:
```suggestion
docker run --name gpdb7_resgroup_v2 --rm -e TEST_OS=ubuntu \
```

I don't think `-it` makes much sense here since this isn't run interactively

magf (changes requested):
LGTM, but few things:

- non-critical: `-it` not necessary for a non-interactive `docker run`, and replacing backticks with `$(...)` for consistency;
- suggest adding cleanup for host `cgroup` state (`cgroup.procs` perms, `gpdb` dir) after script exit to avoid leftovers.

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: fixed

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: added

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: added

concourse/scripts/ic_gpdb_resgroup_v2.bash
hilltracer: I decided not to add rmdir /sys/fs/cgroup/gpdb, because it doesn't work reliably without properly stopping demo cluster and then recursively removing all child cgroups. Also, cgroup/gpdb may have already existed on host before  test, and we don't currently track that, so removing it could delete pre-existing state. There may also be another GPDB running on the host, which would prevent removing this cgroup anyway.

We also don't revert `echo "+cpuset +io +cpu +memory" > ...` change, and restoring that correctly require know original state.

So, I prefer to keep the script simple and only restore original permissions of cgroup.procs.

ci/readme.md
hilltracer: ok

ci/readme.md
hilltracer: I've changed note to `script can create host /sys/fs/cgroup/gpdb and enable cgroup controllers.`

hilltracer:
@andreyaksenov, could you please review patch title.

concourse/scripts/ic_gpdb_resgroup_v2.bash
magf: > I decided not to add rmdir /sys/fs/cgroup/gpdb, because it doesn't work reliably without properly stopping demo cluster and then recursively removing all child cgroups. Also, cgroup/gpdb may have already existed on host before test, and we don't currently track that, so removing it could delete pre-existing state. There may also be another GPDB running on the host, which would prevent removing this cgroup anyway.
> 
> We also don't revert `echo "+cpuset +io +cpu +memory" > ...` change, and restoring that correctly require know original state.
> 
> So, I prefer to keep the script simple and only restore original permissions of cgroup.procs.

OK

magf (approved):

andreyaksenov:
> @andreyaksenov, could you please review patch title.

The current wording looks good:

> Provide script to automate resource group v2 tests

Optional suggestion - replace Provide -> Add:

> Add script to automate resource group v2 tests

KnightMurloc (approved):

red1452 (approved):

PR #564, opened 2026-08-14
GG-699: Fix 'gpinitstandby' test step ignoring the '-S' standby directory

ShapatinAA:
Previously trailing slash test on `gpactivatestandby` utility could have left
the cluster in unstable state. It was happening because standby's datadir could
be specified with `-S` option during the process of standby initialization, but
wouldn't be captured in context of the test (`gpinitstandby` step didn't parse
this option at all, and putted in context of the test coordinator's  directory
or randomly generated one). This was leading to incorrect `clean up and revert
back to original coordinator` step execution, which had no verification of
return codes at the end.

To fix it parsing of this flag `-S` was added, to account in context if user
specifies custom datadir. And additionally verification of return code was
added for `clean up and revert back to original coordinator`, so the errors of
execution wouldn't go unnoticed.

### DIFF
--- a/gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
+++ b/gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
@@ -5,6 +5,7 @@ import json
 import os
 import re
 import pipes
+import shlex
 import shutil
 import socket
 import tempfile
@@ -98,6 +99,42 @@ def _cluster_contains_standard_demo_segments():
         actual_segments.add( (seg.content, seg.role, seg.preferred_role) )
 
     return expected_segments == actual_segments
+
+def _options_as_list(options):
+    """
+    Normalize options into a list of tokens.
+    """
+    if options is None:
+        return []
+
+    if isinstance(options, str):
+        return shlex.split(options)
+
+    return list(options)
+
+def _get_option_value(options, flag):
+    """
+    Return the value following a flag.
+
+    Example:
+      _get_option_value(["-S", "/data/standby"], "-S")
+    """
+
+    if flag not in options:
+        return None
+
+    reverse_idx = options[::-1].index(flag)
+    idx = len(options) - 1 - reverse_idx
+
+    try:
+        value = options[idx + 1]
+    except IndexError:
+        raise ValueError("Option %s requires a value" % flag)
+
+    if not value or value.startswith("-"):
+        raise ValueError("Option %s requires a value" % flag)
+
+    return value
 
 @given('a standard local demo cluster is running')
 def impl(context):
@@ -1188,7 +1225,11 @@ def init_standby(context, coordinator_hostname, options, segment_hostname):
     remote = (coordinator_hostname != segment_hostname)
     # -n option assumes gpinitstandby already ran and put standby in catalog
     if "-n" not in options:
-        if remote:
+        if "-S" in options:
+            opts = _options_as_list(options)
+            standby_data_dir = _get_option_value(opts, '-S')
+            context.standby_data_dir = standby_data_dir
+        elif remote:
             context.standby_data_dir = coordinator_data_dir
         else:
             context.standby_data_dir = tempfile.mkdtemp() + "/standby_datadir"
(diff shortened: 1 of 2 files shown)


gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -1188,7 +1225,11 @@ def init_standby(context, coordinator_hostname, options, segment_hostname):
     remote = (coordinator_hostname != segment_hostname)
     # -n option assumes gpinitstandby already ran and put standby in catalog
     if "-n" not in options:
```
silent-observer, line 1227: Maybe put `opts = _options_as_list(options)` on the top level, and replace checks with `if "-n" not in opts:` and `if "-S" in opts:`? That would avoid the edge case of -n or -S appearing as part of an argument, not as a flag.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -1386,7 +1427,9 @@ def impl(context):
     coordinator.run()
 
     cmd = "gpactivatestandby -a -d %s" % coordinator_data_dir
-    run_gpcommand(context, cmd)
+    rc, error, _ = run_gpcommand(context, cmd)
+    if (rc != 0):
```
silent-observer, line 1431: Why not `check_return_code(context, 0)`?

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
```diff
@@ -1188,7 +1225,11 @@ def init_standby(context, coordinator_hostname, options, segment_hostname):
     remote = (coordinator_hostname != segment_hostname)
     # -n option assumes gpinitstandby already ran and put standby in catalog
     if "-n" not in options:
-        if remote:
+        if "-S" in options:
+            opts = _options_as_list(options)
+            standby_data_dir = _get_option_value(opts, '-S')
+            context.standby_data_dir = standby_data_dir
```
silent-observer, line 1231: This introduces a new possible edge case when `standby_data_dir` ends with a /, like `"/tmp/standby_data/"` in the tests, and if `remote == True` (concourse cluster). [Here](https://github.com/GreengageDB/greengage/blob/a0aeae17629cf97d879e0c215549fcbfcd572105/gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py#L1192) standby_data_dir is used in
```py
create_dir(hostname, os.path.dirname(standby_data_dir))
```
which is supposed to create the *parent* directory (`/tmp` in the example). However, if `standby_data_dir == "/tmp/standby_data/"`, then it would create full `/tmp/standby_data`, in which case gpinitstandby would actually [fail](https://github.com/GreengageDB/greengage/blob/a0aeae17629cf97d879e0c215549fcbfcd572105/gpMgmt/bin/gpinitstandby#L613). I think here we should handle trailing slashes the same way gpinitstandby does:

```suggestion
            context.standby_data_dir = os.path.normpath(standby_data_dir)
```

This does not actually cause any issues right now, but it could if we had any tests on concourse cluster that use this

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
silent-observer, line 1239: Maybe handle `-P` here too, just in case?

ShapatinAA: Yep, we can. Fixed.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
ShapatinAA: Thank you for noticing! Fixed.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
ShapatinAA: Didn't know it exists. Fixed.

gpMgmt/test/behave/mgmt_utils/steps/mgmt_utils.py
ShapatinAA: Fixed

silent-observer:
As far as I can tell, the problem was unnoticed before because `gpinitstandby on dir with trailing slash` was the last test in the feature. Is there some other patch that adds a test after it? If not, maybe we can swap it with `gpinitstandby should create pg_hba standby entry to segment primary and mirror` for example, so that `gpinitstandby on dir with trailing slash` is not last anymore. This effectively adds a test to this patch.

ShapatinAA:
> so that gpinitstandby on dir with trailing slash is not last

Done!

silent-observer (approved):

KnightMurloc (approved):

PR #565, opened 2026-08-17
GG-685: Skip checks for removed objects in case of support function absence

GirgerS:
When upgrading from Greengage 6 to Greengage 7, pg_upgrade
can fail on pre-upgrade checks, if the Greengage 6 cluster has 
an older version that doesn't have required support functions.

This patch makes sure that we handle this case gracefully,
by skipping the checks in case of support function absence.

Also, fix a typo.

### DIFF
--- a/src/bin/pg_upgrade/greengage/check_gp.c
+++ b/src/bin/pg_upgrade/greengage/check_gp.c
@@ -20,7 +20,6 @@
 #include "access/nbtree.h"
 
 #define RELSTORAGE_EXTERNAL	'x'
-#define ERRCODE_UNDEFINED_FUNCTION_STRING "42883"
 
 static void check_external_partition(void);
 static void check_covering_aoindex(void);
@@ -851,6 +850,7 @@ check_views_with_removed_operators()
 	char  output_path[MAXPGPATH];
 	FILE *script = NULL;
 	bool  found = false;
+	bool  support_function_is_present;
 	int   dbnum;
 	int   i_viewname;
 
@@ -870,6 +870,20 @@ check_views_with_removed_operators()
 		conn = connectToServer(&old_cluster, active_db->db_name);
 		PQclear(executeQueryOrDie(conn, "SET search_path TO 'public';"));
 
+		/* Install check support function */
+		support_function_is_present = createSupportFunctionOrDie(conn,
+																 "CREATE OR REPLACE FUNCTION "
+																 "view_has_removed_operators(OID) "
+																 "RETURNS BOOL "
+																 "AS '$libdir/pg_upgrade_support' "
+																 "LANGUAGE C STRICT;");
+
+		if (!support_function_is_present)
+		{
+			PQfinish(conn);
+			break;
+		}
+
 		/*
 		 * Disabling track_counts results in a large performance improvement of
 		 * several orders of magnitude when walking the views. This is because
@@ -879,13 +893,6 @@ check_views_with_removed_operators()
 		 */
 		PQclear(executeQueryOrDie(conn, "SET track_counts TO off;"));
 
-		/* Install check support function */
-		PQclear(executeQueryOrDie(conn,
-								  "CREATE OR REPLACE FUNCTION "
-								  "view_has_removed_operators(OID) "
-								  "RETURNS BOOL "
-								  "AS '$libdir/pg_upgrade_support' "
-								  "LANGUAGE C STRICT;"));
 		res = executeQueryOrDie(conn,
 								"SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname) AS badviewname "
 								"FROM pg_class c JOIN pg_namespace n on c.relnamespace=n.oid "
@@ -915,7 +922,11 @@ check_views_with_removed_operators()
 		PQfinish(conn);
 	}
 
-	if (found)
+	if (!support_function_is_present)
+	{
+		check_skipped_due_to_missing_function();
+	}
+	else if (found)
 	{
 		pg_log(PG_REPORT, "fatal\n");
 		gp_fatal_log(
@@ -938,6 +949,7 @@ check_views_with_removed_functions()
 	char  output_path[MAXPGPATH];
 	FILE *script = NULL;
 	bool  found = false;
+	bool  support_function_is_present;
 	int   dbnum;
 	int   i_viewname;
 
@@ -957,6 +969,20 @@ check_views_with_removed_functions()
 		conn = connectToServer(&old_cluster, active_db->db_name);
 		PQclear(executeQueryOrDie(conn, "SET search_path TO 'public';"));
 
+		/* Install check support function */
+		support_function_is_present =  createSupportFunctionOrDie(conn,
+																  "CREATE OR REPLACE FUNCTION "
+																  "view_has_removed_functions(OID) "
+																  "RETURNS BOOL "
+																  "AS '$libdir/pg_upgrade_support' "
+																  "LANGUAGE C STRICT;");
+
+		if (!support_function_is_present)
+		{
+			PQfinish(conn);
+			break;
+		}
+
 		/*
 		 * Disabling track_counts results in a large performance improvement of
 		 * several orders of magnitude when walking the views. This is because
@@ -966,13 +992,6 @@ check_views_with_removed_functions()
 		 */
 		PQclear(executeQueryOrDie(conn, "SET track_counts TO off;"));
 
-		/* Install check support function */
-		PQclear(executeQueryOrDie(conn,
-								  "CREATE OR REPLACE FUNCTION "
-								  "view_has_removed_functions(OID) "
-								  "RETURNS BOOL "
-								  "AS '$libdir/pg_upgrade_support' "
-								  "LANGUAGE C STRICT;"));
 		res = executeQueryOrDie(conn,
 								"SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname) AS badviewname "
 								"FROM pg_class c JOIN pg_namespace n on c.relnamespace=n.oid "
@@ -1003,7 +1022,11 @@ check_views_with_removed_functions()
 		PQfinish(conn);
 	}
 
-	if (found)
+	if (!support_function_is_present)
+	{
+		check_skipped_due_to_missing_function();
+	}
+	else if (found)
 	{
 		pg_log(PG_REPORT, "fatal\n");
 		gp_fatal_log(
@@ -1026,6 +1049,7 @@ check_views_with_removed_types()
 	char  output_path[MAXPGPATH];
 	FILE *script = NULL;
 	bool  found = false;
+	bool  support_function_is_present;
 	int   dbnum;
 	int   i_viewname;
 
@@ -1045,6 +1069,19 @@ check_views_with_removed_types()
 		conn = connectToServer(&old_cluster, active_db->db_name);
 		PQclear(executeQueryOrDie(conn, "SET search_path TO 'public';"));
 
+		/* Install check support function */
+		support_function_is_present = createSupportFunctionOrDie(conn,
+																 "CREATE OR REPLACE FUNCTION "
+																 "view_has_removed_types(OID) "
+																 "RETURNS BOOL "
+																 "AS '$libdir/pg_upgrade_support' "
+																 "LANGUAGE C STRICT;");
+		if (!support_function_is_present)
+		{
+			PQfinish(conn);
+			break;
+		}
+
 		/*
 		 * Disabling track_counts results in a large performance improvement of
 		 * several orders of magnitude when walking the views. This is because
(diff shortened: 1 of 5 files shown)


src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1571,19 +1573,39 @@ check_views_with_removed_columns()
...
+			}
+			dieOnQueryFailure(conn, res, create_support_function_query);
+		}
+		PQclear(res);
+
 		/* track_counts is disables for the same reason as above */
 		PQclear(executeQueryOrDie(conn, "SET track_counts TO off;"));
 
-		/* Install check support function */
-		PQclear(executeQueryOrDie(conn,
```
bimboterminator1, line 1581: what about `check_views_with_removed_operators()` `check_views_with_removed_functions` `check_views_with_removed_types`?

bimboterminator1, line 1581: Additionally, we drop `pg_log(PG_VERBOSE, "executing: %s\n", query) `  call,  without executeQueryOrDie. logging behaviour differs after the patch.

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1742,7 +1790,12 @@ check_views_with_removed_relations()
 	if (script)
 		fclose(script);
 
-	if (found)
+	if (no_support_function)
+	{
+		pg_log(PG_REPORT, "skipped (required function was not found)\n");
+		fflush(stdout);
+	}
+	else if (found)
```
bimboterminator1, line 1815: Shouldn't we check found first? Is the case when found = true from previous database comes together with no_support_function?

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Do we ship a version of Greengage 6 without these functions? These where added more than two years ago. From what I've seen, this is the last one:

b5815b14d4dcc4ccd273fd770d284e7b57da6750

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Can such case happen though?

hilltracer:
Since PR doesn't contain tests, could you please provide detailed manual testing results, covering following scenarios:

1. Basic scenario: run upgrade with recent 6.x where required functions are available and dump. Upgrade completes successfully and checks are not skipped.

2. Old 6.x scenario: run upgrade with version where these support functions are absent. Corresponding checks are reported as skipped and the upgrade completes successfully.

3. Objects detected scenario: create objects that reference removed relations/columns and verify that checks detect them, fail as expected, and produce tcorresponding list of problem objects.

src/bin/pg_upgrade/pg_upgrade.h
```diff
@@ -533,6 +533,7 @@ void		init_tablespaces(void);
 /* server.c */
 
 PGconn	   *connectToServer(ClusterInfo *cluster, const char *db_name);
+void        dieOnQueryFailure(PGconn *conn, PGresult *result, const char *query);
```
hilltracer: maybe add `pg_attribute_noreturn();`? Since we always do exit(1).

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1638,7 +1660,12 @@ check_views_with_removed_columns()
 	if (script)
 		fclose(script);
 
-	if (found)
+	if (no_support_function)
+	{
+		pg_log(PG_REPORT, "skipped (required function was not found)\n");
+		fflush(stdout);
```
hilltracer: pg_log already [do](https://github.com/GreengageDB/greengage/blob/ab129676775f977cb54d8784d5bd5303a69459ab/src/bin/pg_upgrade/util.c#L225) fflush. The same for 1796

hilltracer (commented):
Since we skip relations also, not only columns. Maybe rename PR to `Skip removed relation and column...`?

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Thanks! Should be fixed

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Thanks! Not sure why it is done [here](https://github.com/GreengageDB/greengage/blob/3318438b7da602bd78413e096276d222221bcf9d/src/bin/pg_upgrade/util.c#L258) though.

src/bin/pg_upgrade/pg_upgrade.h
GirgerS: Yeah, without the attribute gcc thinks that this function returns, so it's a good idea. I've added it.

GirgerS:
> Do we ship a version of Greengage 6 without these functions? These where added more than two years ago. From what I've seen, this is the last one:
> 
> https://github.com/GreengageDB/greengage/commit/b5815b14d4dcc4ccd273fd770d284e7b57da6750

Well, I'll think about this question more, because I can't tell for sure if we will always have these functions or not.

GirgerS:
> Since PR doesn't contain tests, could you please provide detailed manual testing results

To test the checks themselves, install Greengage 6 from a branch that has c45d3fa4364dcbee8653d006df6bfd80030d7a13 commit, `6.x` will suffice. 

1)  Checks fired, no failing objects. Here SQL_SCHEMA is empty. 
     [test_success.txt](https://github.com/user-attachments/files/31186863/test_success.txt)

2) Checks fired, a view references a removed tables. SQL_SCHEMA:
    `CREATE VIEW break_relations_check AS (SELECT * FROM pg_partitions);`
     [test_failure_relations.txt](https://github.com/user-attachments/files/31187064/test_failure_relations.txt)      
     [test_failure_relations.log](https://github.com/user-attachments/files/31187023/test_failure_relations.log)

3) Checks fired, a view references a removed column. SQL_SCHEMA:
    `CREATE VIEW break_columns_check AS (SELECT * FROM pg_class);`
     [test_failure_columns.txt](https://github.com/user-attachments/files/31187129/test_failure_columns.txt)
     [test_failure_columns.log](https://github.com/user-attachments/files/31187135/test_failure_columns.log)

Now, to test older versions, checkout to 45abc189575e10932672c089402eb9428c8a5fd5 commit, right before these functions were added, and reinstall the database.

4) Checks skipped, no failing objects. SQL_SCHEMA is empty.
    [test_skipping_success.txt](https://github.com/user-attachments/files/31187211/test_skipping_success.txt)

5) Checks skipped, but there are views referencing removed objects. SQL_SCHEMA is:
    ```
    CREATE VIEW break_relations_check AS (SELECT * FROM pg_partitions);
    CREATE VIEW break_columns_check AS (SELECT * FROM pg_class);
    ```
    [test_skipping_failure.txt](https://github.com/user-attachments/files/31187268/test_skipping_failure.txt)
    [test_skipping_failure.log](https://github.com/user-attachments/files/31187269/test_skipping_failure.log)

6) These is also the case when a query that is used to create support functions fails for some other reasons, not because the function is absent. But it is hard to show this case here, because it requires some shenanigans with a debugger.

    First of all, change the testing script similar to [this](https://github.com/user-attachments/files/31188096/debugger.patch).

    Then, run `ci/scripts/pg_upgrade_run_6X_to_7X_migration.bash`, and when the debugger pops open, break on                   
    `check_gp.c:1718` or `check_gp.c:1587` and continue.

    After breakpoint hit, `pkill postgres` in a separate shell and `continue` in the debugger. Instead of saying that the check was   
    skipped, pg_upgrade should fail.

src/bin/pg_upgrade/greengage/check_gp.c
```diff
@@ -1638,7 +1661,11 @@ check_views_with_removed_columns()
 	if (script)
 		fclose(script);
 
-	if (found)
+	if (no_support_function)
+	{
+		pg_log(PG_REPORT, "skipped (required function was not found)\n");
```
hilltracer:
```
Performing Consistency Checks
-----------------------------
...
Checking for views with removed types                       ok
Checking for views with removed columns                     skipped (required function was not found)
Checking for views with removed relations                   skipped (required function was not found)
```

Maybe add instructions for user? Smth like: `Hint: Upgrade source cluster to enable this check.`

GirgerS: Maybe `skipped (source cluster version is too old)`? `required function was not found` seems to give no answers to the user anyway

hilltracer: Hm.. maybe smth like: `skipped (unsupported by source cluster; upgrade to enable)`?

bimboterminator1:
> if we will always have these functions or not.

i think it will be redundant. it's not a big deal for customers to update 6x

GirgerS:
> > if we will always have these functions or not.

> it's not a big deal for customers to update 6x

The whole point of this patch is that our architect thinks that it is. Plus many inconveniences for releases.

> 
> i think it will be redundant.

Yeah, I though the same. But I believe that not having an edge-case is just always better, plus the refactoring was not that hard.

Tests go as follows:

1) Check latest 6.x branch, no schema:
     
    [test_success.txt](https://github.com/user-attachments/files/31219236/test_success.txt)

2) Testing that checks still work:

   Operators, SQL_SCHEMA is 
   `CREATE VIEW break_operator_check AS (SELECT '1 2'::int2vector = '3 4'::int2vector);`;
   
   [test_failure_operators.txt](https://github.com/user-attachments/files/31219118/test_failure_operators.txt)
   [test_failure_operators.log](https://github.com/user-attachments/files/31219122/test_failure_operators.log)
   
   Functions, SQL_SCHEMA is 
   
   `CREATE VIEW break_function_check AS (SELECT pg_catalog.int2vectoreq('1 2'::int2vector, '3 4'::int2vector));`
 
   [test_failure_functions.txt](https://github.com/user-attachments/files/31219132/test_failure_functions.txt)
   [test_failure_functions.log](https://github.com/user-attachments/files/31219146/test_failure_functions.log)
  
  
   Types, SQL_SCHEMA is 
   
   `CREATE VIEW break_types_check AS (SELECT '["May 10, 1947 23:59:12" "Jan 14, 1973 03:14:21"]'::tinterval);`

   
    [test_failure_types.txt](https://github.com/user-attachments/files/31219486/test_failure_types.txt)
    [test_failure_types.log](https://github.com/user-attachments/files/31219488/test_failure_types.log)

   
   Functions with changed signatures, SQL_SCHEMA is
   
   `CREATE VIEW break_changed_functions_check AS (SELECT * FROM pg_show_all_settings());
   
   [test_failure_changed_functions.txt](https://github.com/user-attachments/files/31219158/test_failure_changed_functions.txt)
   [test_failure_changed_functions.log](https://github.com/user-attachments/files/31219172/test_failure_function_signatures.log)
   
   Relations, SQL_SCHEMA is: 
   
   `CREATE VIEW break_relations_check AS (SELECT * FROM pg_partitions);`

   [test_failure_relations.txt](https://github.com/user-attachments/files/31219189/test_failure_relations.txt)
   [test_failure_relations.log](https://github.com/user-attachments/files/31219199/test_failure_relations.log)
   
   Columns, SQL_SCHEMA is:
   
   `CREATE VIEW break_columns_check AS (SELECT * FROM pg_class);`

   [test_failure_columns.txt](https://github.com/user-attachments/files/31219205/test_failure_columns.txt)
   [test_failure_columns.log](https://github.com/user-attachments/files/31219211/test_failure_columns.log)


3) To test that everything is skipped correctly, nuke the entirety of Greengage 6 pg_upgrade_support from any branch and reinstall the database: 

    
    [pg_upgrade_support.patch](https://github.com/user-attachments/files/31219618/pg_upgrade_support.patch)
    [test_skipping_success.txt](https://github.com/user-attachments/files/31219361/test_skipping_success.txt)



4) Failure when skipping is probably not that interesting
5) And the last one still requires a debugger.

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: Sorry for the delay, refactoring and testing took some time 

I ended up with `skipped (unsupported by source cluster; install a later version to enable)`, because proposing an upgrade seems to be ambiguous in this context (we are in the middle of the one!). Whether it is good workdng or not is still up to a debate

bimboterminator1 (approved):

hilltracer (approved):
LGTM. Because now we skip checks not for only `columns and tables` but for other objects also. Please fix PR title and description. 
E.g. unsupported objects can be before 85f0547.

GirgerS:
@andreyaksenov, please review the title.

andreyaksenov:
> @andreyaksenov, please review the title.

My suggestion:

> Skip checks for removed objects in case of support function absence

src/bin/pg_upgrade/greengage/check_gp.c
GirgerS: I guess that I can resolve this one.

PR #566, opened 2026-08-18
CI-6154: Implement major upgrade test with pg_upgrade for 7.x

magf:
## Implement major upgrade test with pg_upgrade for 7.x

New job `pg_upgrade-test`:
- runs on pull_request, after `build`
- calls greengage-ci reusable workflow `tests-pg_upgrade@v56`
- Ubuntu 22.04 only, no matrix

What the workflow does (see reusable workflow's [README](https://github.com/GreengageDB/greengage-ci/blob/v56/README/REUSABLE-TESTS-PG_UPGRADE.md)):
- builds a combined 6.x/7.x image via `ci/Dockerfile.pg_upgrade`
- fetches a pre-generated Greengage 6.x SQL dump and runs the
  migration test against it
- collects runtime logs on every run, and diagnostic dumps/diffs
  only on failure, uploading them as _one_ artifact

Update workflows README accordingly.

Task: [CI-6154](https://tracker.yandex.ru/CI-6154)
Parent task: [CI-6130](https://tracker.yandex.ru/CI-6130)


uaqq (dismissed):

GirgerS:
I'm not sure that it got communicated properly (I bet that it wasn't), but the test is intended to be run with the latest regression dump from 6.x, i.e. the CI job needs to download it, and start the test as it is described [here](https://github.com/GreengageDB/greengage/blob/7.x/ci/readme.md?plain=1#L291-L302). Otherwise it is mostly pointless.

GirgerS:
SQL dumps before/after upgrade can be interesting two, as the [diffs](https://github.com/GreengageDB/greengage-ci/blob/main/.github/workflows/greengage-reusable-tests-pg_upgrade.yml#L69-L75) alone can be quite difficult to navigate on their own.

GirgerS:
https://github.com/GreengageDB/greengage-ci/blob/main/.github/workflows/greengage-reusable-tests-pg_upgrade.yml#L71 (and the same thing for `partitions_regression.diffs`) is not entirely correct, as:

1) There could be `regression.diffs` but not `partitions_regression.diffs`, and vise versa
2) pg_upgrade itself could fail, making the script terminate before it is able to collect diffs. So the test fails, but these shouldn't be any logs.

And we don't provide enough information to distinguish these cases (my bad), so until it is fixed (and I doubt that it would happen), it's difficult to tell anything about the logs.

magf:
@GirgerS, the task was completed according to the specified requirements:

> As a developer, I need additional test job to test pg_upgrade, so that avoid breaking upgrading logic by Greengage 7 changes.
> More details here: https://github.com/GreengageDB/greengage/tree/7.x/ci#running-pg_upgrade-test-from-the-docker-image

The task doesn't specify any initial parameters, such as whether a dump needs to be loaded before the process; the task doesn't specify a DoD.
Please describe all the conditions that must be met in the task description.

magf:
> SQL dumps before/after upgrade can be interesting two, as the [diffs](https://github.com/GreengageDB/greengage-ci/blob/main/.github/workflows/greengage-reusable-tests-pg_upgrade.yml#L69-L75) alone can be quite difficult to navigate on their own.

Is this only interesting in case of a fail or in case of a normal completion? These dumps are very large.

magf:
> https://github.com/GreengageDB/greengage-ci/blob/main/.github/workflows/greengage-reusable-tests-pg_upgrade.yml#L71 (and the same thing for `partitions_regression.diffs`) is not entirely correct, as:
> 
>     1. There could be `regression.diffs` but not `partitions_regression.diffs`, and vise versa

`partitions_regression.diffs` copied [below](https://github.com/GreengageDB/greengage-ci/blob/main/.github/workflows/greengage-reusable-tests-pg_upgrade.yml#L72-L74)

>     2. pg_upgrade itself could fail, making the script terminate before it is able to collect diffs. So the test fails, but these shouldn't be any logs.

If the script managed to create diffs, they will be copied because the next command will be executed regardless of the script's exit status.

The exception is if the entire container is killed.

GirgerS:
Sorry about the delay, forgot that I don't get github notifications.

> @GirgerS, the task was completed according to the specified requirements:

I see, the description wasn't rich for details.)

> The task doesn't specify any initial parameters, such as whether a dump needs to be loaded before the process; the task doesn't specify a DoD.
Please describe all the conditions that must be met in the task description.

Sql dump thing is the only one, I updated the ticket.

GirgerS:
> > SQL dumps before/after upgrade can be interesting two, as the [diffs](https://github.com/GreengageDB/greengage-ci/blob/main/.github/workflows/greengage-reusable-tests-pg_upgrade.yml#L69-L75) alone can be quite difficult to navigate on their own.
> 
> Is this only interesting in case of a fail or in case of a normal completion? These dumps are very large.

Mostly in case of failure.

GirgerS:
> If the script managed to create diffs, they will be copied because the next command will be executed regardless of the script's exit status.

> The exception is if the entire container is killed.

I worded it poorly, I meant that telling `regression.diffs not found (ok if test passed)` may be misleading in the above cases. E.g., a developer looks of the test failed at the pg_upgrade stage, these are no `diffs` collected, and the log says `regression.diffs not found (ok if test passed)`, which hints at the possibility that the test is broken. But everything is fine.

magf:
> I worded it poorly, I meant that telling `regression.diffs not found (ok if test passed)` may be misleading in the above cases. E.g., a developer looks of the test failed at the pg_upgrade stage, these are no `diffs` collected, and the log says `regression.diffs not found (ok if test passed)`, which hints at the possibility that the test is broken. But everything is fine.

Got it, I'll think about how to improve it

magf:
> I updated the ticket

Thanks! On my way.

GirgerS (approved):
Looks great!

uaqq (approved):

PR #568, opened 2026-08-19
GG-701: Fix gpload tests with python 3 on dual-stack host

dkovalev1:
get_ip() in TEST.py and TEST_local_base.py used getaddrinfo results stored in 
a set, so the loop order was non-deterministic and could return either an IPv4 
or IPv6 address for the same host on different runs. On dual-stack hosts this 
made gpload tests flaky/fail depending on which address type won.

This commit rewrites get_ip() to query getaddrinfo with AF_UNSPEC/SOCK_STREAM, 
collect the first IPv4 and first IPv6 address separately, and 
deterministically prefer IPv4 (falling back to IPv6 only if no IPv4 address 
exists).

GG-701

### DIFF
--- a/gpMgmt/bin/gpload_test/gpload2/TEST_local_base.py
+++ b/gpMgmt/bin/gpload_test/gpload2/TEST_local_base.py
@@ -62,17 +62,16 @@ def get_port():
 def get_ip(hostname=None):
     if hostname is None:
         hostname = socket.gethostname()
-    else:
-        hostname = hostname
-    hostinfo = socket.getaddrinfo(hostname, None)
-    ipaddrlist = list(set([(ai[4][0]) for ai in hostinfo]))
-    for myip in ipaddrlist:
-        if myip.find(":") > 0:
-            ipv6 = myip
-            return ipv6
-        elif myip.find(".") > 0:
-            ipv4 = myip
-            return ipv4
+    hostinfo = socket.getaddrinfo(hostname, None, socket.AF_UNSPEC, socket.SOCK_STREAM)
+    ipv4 = None
+    ipv6 = None
+    for family, _, _, _, sockaddr in hostinfo:
+        if family == socket.AF_INET and ipv4 is None:
+            ipv4 = sockaddr[0]
+            break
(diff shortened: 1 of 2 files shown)


gpMgmt/bin/gpload_test/gpload2/TEST_local_base.py
```diff
@@ -62,17 +62,16 @@ def get_port():
...
-        elif myip.find(".") > 0:
-            ipv4 = myip
-            return ipv4
+    hostinfo = socket.getaddrinfo(hostname, None, socket.AF_UNSPEC, socket.SOCK_STREAM)
+    ipv4 = None
+    ipv6 = None
+    for family, _, _, _, sockaddr in hostinfo:
+        if family == socket.AF_INET and ipv4 is None:
+            ipv4 = sockaddr[0]
+            break
```
whitehawk, line 71: Why such `break` is not presented in the 1st file?

dkovalev1: Its not strictly necessary, rather an optimization. But added to keep both in sync.

whitehawk (approved):

KnightMurloc:
What about other uses of `getaddrinfo`? Why do we want to prioritize IPv4? The `getaddrinfo` documentation recommends selecting the first address received. As I understand it, their order is deterministic and can be configured via `/etc/gai.conf`.

dkovalev1:
> What about other uses of `getaddrinfo`? 

The scope of this change is test only, it shall not affect production setups. And the problem was "order of interfaces is unpredictable in the test environment".

> Why do we want to prioritize IPv4? The `getaddrinfo` documentation recommends selecting the first address received. As I understand it, their order is deterministic and can be configured via `/etc/gai.conf`.

The idea for v4 precedence comes from `Dockerfile.ubuntu`, and the point is to make test environment more stable over other environments as well. `/etc/gai.conf` is not directly defines order, the rules are bit more complicated (RFC 3484/6724). In the test I had to add subnet as well to get v6 records to come first.

KnightMurloc (approved):

PR #570, opened 2026-08-20
GG-712: Fix HAVE_LIBZSTD being lost when configure is regenerated

deart2k:
### Fix HAVE_LIBZSTD being lost when configure is regenerated

The zstd probe in the committed configure was spliced in by hand from an autoconf 2.71 run without updating configure.in, so the two disagree. configure.in calls AC_CHECK_LIB(:libzstd.a, ...) with an empty action-if-found, which makes autoconf derive HAVE_LIB_LIBZSTD_A, while the committed configure defines HAVE_LIBZSTD -- the macro the tree actually tests, in 12 files.  Running `make -C gpAux autoconf` would therefore have silently disabled zstd support.

Spell out the action-if-found so configure.in reproduces the behaviour the committed configure already had, and regenerate with autoconf 2.69.

Regenerating also syncs pre-existing drift in pg_config.h.in: a duplicated USE_DEV_URANDOM entry, a missing HAVE_DECL_GSS_STORE_CRED_INTO template whose define was being dropped, and two HAVE__BUILTIN_BSWAP* templates that no configure run defines.  None of these change behaviour.

Task: [[GG-712](https://tracker.yandex.ru/GG-712)]

### DIFF
--- a/configure
+++ b/configure
@@ -269,10 +269,10 @@ fi
     $as_echo "$0: be upgraded to zsh 4.3.4 or later."
   else
     $as_echo "$0: Please tell bug-autoconf@gnu.org and
-$0: support@greengagedb.org about your system, including any
-$0: error possibly output before this message. Then install
-$0: a modern shell, or manually run the script under such a
-$0: shell if you do have one."
+$0: support@greengagedb.org about your system, including
+$0: any error possibly output before this message. Then
+$0: install a modern shell, or manually run the script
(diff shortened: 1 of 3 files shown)


bimboterminator1 (approved):
We also should fix 7.x. Additionally, we should guard the config generation from wrong autoconf version.

whitehawk (approved):

PR #572, opened 2026-08-20
Decide 2PC by permanent-relation WAL writes

bimboterminator1:
Decide 2PC by permanent-relation WAL writes
A distributed transaction normally commits in two phases: the QD
broadcasts PREPARE to every participant segment, waits until all of
them have durably flushed their prepared state, records the commit
decision and then broadcasts COMMIT PREPARED.  This is what makes the
commit atomic across segments, but it costs an extra dispatch
round-trip and extra WAL (prepare and commit-prepared records) on
every segment.

That price is only justified when segments hold durable changes whose
commit must be atomic.  If no segment changed anything durable, no
commit/abort divergence between segments can ever be observed, so the
QD can skip PREPARE and finish with a single cheap one-phase commit
broadcast.

This optimization was associated with the "no QE wrote WAL under its xid"
condition, which over-counts durable work: a transaction whose only WAL
is lock records (e.g. XLOG_STANDBY_LOCK from LOCK TABLE or from
creating unlogged/temp tables) changed nothing durable, yet was forced
into a full two-phase commit.

This patch changes  the optimization "some QE wrote WAL for durable object"
instead.  A QE now reports that it wrote WAL only if:

- it logged a record carrying registered block references under an
  assigned xid.  Temp relations don't WAL-log data pages at all, so
  block-carrying records mean permanent-relation changes.  The xid is
  required because read-only scans also emit block-carrying records
  (page pruning, hint-bit FPIs) without an xid, and those must not
  force 2PC either;

- or it wrote an AO/AOCS record (xlog_ao_insert/xlog_ao_truncate).
  AO segfiles are not buffer pages, so these records carry no block
  references and are marked explicitly instead of relying on the
  pg_aoseg heap update that accompanies every AO write today.

Lock/invalidation/commit-only transactions now take one-phase commit.

The patch introduced the xactWroteWalForPermanentRel xact state flag, which
stands for "current transaction emits WAL record that modifies a
permanent relation". It is set after a transaction with assigned xid
writes the WAL record related to registered block references. The isolation2
test dtx_onephase_commit is added.

Co-Authored-By: Egor Ivkov <e.o.ivkov@gmail.com>


bimboterminator1:
version of #543

src/test/isolation2/expected/dtx_onephase_commit.out
```diff
@@ -0,0 +1,189 @@
...
+-----------------
+ Success:        
+(1 row)
+1: begin;
+BEGIN
+1: lock table dtx_phase_heap in access exclusive mode;
+LOCK TABLE
+1: commit;
+COMMIT
+select gp_wait_until_triggered_fault('start_performDtxProtocolCommitOnePhase', 1, dbid) from gp_segment_configuration where role = 'p' and content = 0;
```
whitehawk: In case the functionality is broken (for example for demonstration purposes we can just revert the changes is the `src/backend`), the test will not fail, it will get stuck here. Can we make the test failing instead of getting stuck, when the target functionality is broken?

RekGRpth:
> This patch changes gates the optimization

Two verbs in a row, choose one.

RekGRpth:
> The patch indroduced

The patch introduced

src/backend/cdb/cdbappendonlyxlog.c
```diff
@@ -53,6 +54,18 @@ xlog_ao_insert(RelFileNode relFileNode, int32 segmentFileNum,
...
+	 * Append-optimized segfiles are not buffer pages, so this record carries
+	 * no block references and would not be noticed by the block-reference
+	 * check in XLogInsert_Internal().  Mark the durable permanent-relation
+	 * write explicitly so the two-phase commit decision does not silently
+	 * depend on the pg_aoseg/pg_aocsseg heap update that currently
+	 * accompanies every AO write in the same transaction.  All callers
+	 * already check RelationNeedsWAL, so this is never reached for
+	 * non-permanent relations.
+	 */
+	MarkWalWriteForPermanentRel();
```
RekGRpth, line 72: Is it possible to add an assertion here and below in a similar place?
```c
Assert(TransactionIdIsValid(GetTopTransactionIdIfAny()));
```

RekGRpth:
> Co-Authored-By: Alexander Kondakov [a.kondakov@arenadata.io](mailto:a.kondakov@arenadata.io)

I think it would be easier to do PR on your behalf by adding a third-party author as a co-author; it's not like cherry-picking.

src/test/isolation2/sql/dtx_onephase_commit.sql
```diff
@@ -0,0 +1,102 @@
...
+select gp_wait_until_triggered_fault('before_xlog_xact_prepare', 1, dbid)
+  from gp_segment_configuration where role = 'p' and content = 0;
+select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
+  from gp_segment_configuration where role = 'p' and content = 0;
+
+1q:
+drop table dtx_phase_heap;
+drop table dtx_phase_ao;
+drop table dtx_phase_aocs;
+drop table dtx_phase_prune;
```
RekGRpth, line 279: Claude suggested adding a few more uncovered scenarios to the tests:
```suggestion
-- Case 7: a transaction that inserts into an already-existing UNLOGGED table,
-- and touches nothing else, takes the one-phase path. Unlogged relations only
-- WAL-log their init fork; ordinary data writes to them are never WAL-logged
-- (RelationNeedsWAL is false), so no block-carrying record is ever produced.
create unlogged table dtx_phase_unlogged(a int) distributed by (a);
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: insert into dtx_phase_unlogged select i from generate_series(1, 100) i;
1: commit;
select gp_wait_until_triggered_fault('start_performDtxProtocolCommitOnePhase', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

-- Case 8: same for a TEMP table -- temp relations never WAL-log data pages at
-- all. The table is created in the same session so it is visible to the
-- later statements dispatched over connection "1".
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: create temp table dtx_phase_temp(a int) distributed by (a);
1: begin;
1: insert into dtx_phase_temp select i from generate_series(1, 100) i;
1: commit;
select gp_wait_until_triggered_fault('start_performDtxProtocolCommitOnePhase', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

-- Case 9: a transaction that only takes a lock and then performs the first
-- read of an already-committed, unrelated table still takes the one-phase
-- path. LOCK TABLE assigns an xid (see Case 4); the concern this guards
-- against is a later block-carrying maintenance record in the same
-- transaction (e.g. an XLOG_FPI_FOR_HINT from hint-bit setting) picking up
-- that xid and being misclassified as durable transaction work. autovacuum
-- is disabled on the target so autoanalyze cannot set the hint bits first.
create table dtx_phase_hint_lock(a int) distributed by (a);
create table dtx_phase_hint_target(a int) with (autovacuum_enabled = false) distributed by (a);
insert into dtx_phase_hint_target select i from generate_series(1, 100) i;
checkpoint;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: lock table dtx_phase_hint_lock in access exclusive mode;
1: select count(*) from dtx_phase_hint_target;
1: commit;
select gp_wait_until_triggered_fault('start_performDtxProtocolCommitOnePhase', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

-- Case 10: TRUNCATE of an append-optimized column-oriented table, with no
-- other statement in the transaction, still takes the two-phase path. AOCS
-- truncation produces no block-carrying record (AO segfiles are not buffer
-- pages); this exercises the explicit MarkWalWriteForPermanentRel() call in
-- xlog_ao_truncate(), distinct from Case 3's xlog_ao_insert() call.
select gp_inject_fault('before_xlog_xact_prepare', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: truncate dtx_phase_aocs;
1: commit;
select gp_wait_until_triggered_fault('before_xlog_xact_prepare', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

-- Case 11: cross-segment aggregation. A segment that itself only takes a lock
-- (no durable write of its own) still receives PREPARE: the QD broadcasts
-- PREPARE to every participating segment once any one of them reports a
-- durable permanent-relation write, not only to the segment that wrote it.
-- The LOCK touches every segment; the single inserted row lands on exactly
-- one of them (whichever one, by hash), so at least one other segment here
-- only ever sees the lock and still must be asked to PREPARE.
create table dtx_phase_agg_lock(a int) distributed by (a);
create table dtx_phase_agg_data(a int) distributed by (a);
select gp_inject_fault('before_xlog_xact_prepare', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content >= 0;
1: begin;
1: lock table dtx_phase_agg_lock in access exclusive mode;
1: insert into dtx_phase_agg_data values (1);
1: commit;
select gp_wait_until_triggered_fault('before_xlog_xact_prepare', 1, dbid)
  from gp_segment_configuration where role = 'p' and content >= 0;
select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content >= 0;

-- Case 12: COPY takes the two-phase path exactly like INSERT does. COPY has
-- its own dispatch/reporting path (cdbcopy.c) separate from ordinary query
-- dispatch (cdbdisp_async.c); this checks that path also reports durable
-- permanent-relation writes correctly.
select gp_inject_fault('before_xlog_xact_prepare', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: copy dtx_phase_heap from program 'seq 1 100';
1: commit;
select gp_wait_until_triggered_fault('before_xlog_xact_prepare', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

-- Case 13: TRUNCATE of a row-oriented append-optimized table, with no other
-- statement in the transaction, still takes the two-phase path. Same
-- xlog_ao_truncate() code path as Case 10's AOCS variant, exercised here for
-- the row-oriented storage format too.
select gp_inject_fault('before_xlog_xact_prepare', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: truncate dtx_phase_ao;
1: commit;
select gp_wait_until_triggered_fault('before_xlog_xact_prepare', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

-- Case 14: the durable-work flag correctly resets between transactions on
-- the same, reused QE backend. T1 does a real write and takes the two-phase
-- path; T2, immediately after in the same session/connection, is read-only
-- and must take the one-phase path rather than inherit T1's flag.
select gp_inject_fault('before_xlog_xact_prepare', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: insert into dtx_phase_heap select i from generate_series(201, 210) i;
1: commit;
select gp_wait_until_triggered_fault('before_xlog_xact_prepare', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: select count(*) from dtx_phase_heap;
1: commit;
select gp_wait_until_triggered_fault('start_performDtxProtocolCommitOnePhase', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

-- Case 15: stickiness across two separately-dispatched statements within one
-- explicit transaction. A lock-only statement followed by a real write in a
-- later statement of the same transaction must still force the two-phase
-- path -- not just within a single XLogInsert call as in Case 1, but across
-- statement/dispatch boundaries.
select gp_inject_fault('before_xlog_xact_prepare', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: lock table dtx_phase_agg_lock in access exclusive mode;
1: insert into dtx_phase_heap select i from generate_series(211, 220) i;
1: commit;
select gp_wait_until_triggered_fault('before_xlog_xact_prepare', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

-- Case 16: a multi-statement explicit transaction with no real write at any
-- point -- a read, a lock, then another read -- stays on the one-phase path.
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: select count(*) from dtx_phase_heap;
1: lock table dtx_phase_agg_lock in access exclusive mode;
1: select count(*) from dtx_phase_agg_lock;
1: commit;
select gp_wait_until_triggered_fault('start_performDtxProtocolCommitOnePhase', 1, dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;

1q:
drop table dtx_phase_heap;
drop table dtx_phase_ao;
drop table dtx_phase_aocs;
drop table dtx_phase_prune;
drop table dtx_phase_unlogged;
drop table dtx_phase_hint_lock;
drop table dtx_phase_hint_target;
drop table dtx_phase_agg_lock;
drop table dtx_phase_agg_data;
```

src/test/isolation2/expected/dtx_onephase_commit.out
bimboterminator1: Reworked. Factually the wait was not unbounded - `gp_wait_until_triggered_fault` times out after 10 minutes (faultinjector.c, 3000 × 0.2s retries) - but that is still effectively stuck for CI purposes.

The test now asserts synchronously, with no waits at all. 
Each case now arms two faults:
- `skip` on the expected path, checked immediately after the commit with `status` - `num times hit:'1'` proves the path was taken;
- `error` on the wrong path - if the decision logic is broken, the COMMIT itself fails instantly with the injected error.

src/backend/cdb/cdbappendonlyxlog.c
bimboterminator1: Added in `xlog_ao_insert` — validated empirically under a cassert build across plain AO/AOCS inserts, segfile-creation records, and the `ALTER TABLE SET TABLESPACE` copy path (aomd.c).

Not added in `xlog_ao_truncate`, because it does not hold there: the vacuum drop phase truncates awaiting-drop segfiles *before* it touches pg_aoseg - `AppendOnlyCompaction_DropSegmentFile()` runs ahead of `ClearFileSegInfo()` - so the truncate record can be emitted while `GetTopTransactionIdIfAny()` is still invalid.

The explicit `MarkWalWriteForPermanentRel()` there is correct regardless of the xid- the truncate is durable work, and the same transaction assigns an xid moments later via the pg_aoseg update. Documented this ordering in a comment at the call site instead of the assert.

src/test/isolation2/sql/dtx_onephase_commit.sql
bimboterminator1: Adopted.

bimboterminator1:
> I think it would be easier to do PR on your behalf by adding a third-party author as a co-author; it's not like cherry-picking.

ok, let's squash it then

src/backend/cdb/cdbappendonlyxlog.c
```diff
@@ -141,17 +141,7 @@ void xlog_ao_truncate(RelFileNode relFileNode, int32 segmentFileNum, int64 offse
...
-	/*
-	 * See xlog_ao_insert: no block references, mark the write explicitly.
-	 *
-	 * Unlike xlog_ao_insert, do not assert an assigned xid here: the vacuum
-	 * drop phase truncates awaiting-drop segfiles before it touches
-	 * pg_aoseg (AppendOnlyCompaction_DropSegmentFile runs ahead of
-	 * ClearFileSegInfo), so this record can be the transaction's first
-	 * xid-assigning-free WAL.  The mark is correct regardless: the truncate
-	 * is durable work, and the same transaction assigns an xid moments later
-	 * when it updates the segfile catalog entry.
```
RekGRpth: why removed?

src/test/isolation2/sql/dtx_onephase_commit.sql
```diff
@@ -0,0 +1,256 @@
...
+1: lock table dtx_phase_lock_only in access exclusive mode;
+1: select count(*) from dtx_phase_lock_only;
+1: commit;
+select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'status', dbid)
+  from gp_segment_configuration where role = 'p' and content = 0;
+select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'reset', dbid)
+  from gp_segment_configuration where role = 'p' and content = 0;
+select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
+  from gp_segment_configuration where role = 'p' and content = 0;
+
```
RekGRpth, line 249: how about
```sql
-- Case 14: LOCK TABLE assigns an xid, but a later statement's first read of
-- an unrelated, already-committed table still stays on the one-phase path --
-- verified empirically, no XLOG_FPI_FOR_HINT is produced for the read table
-- in this ordering. This holds regardless of data_checksums/wal_log_hints:
-- with both off, XLogHintBitIsNeeded() is false and MarkBufferDirtyHint()
-- never calls XLogSaveBufferForHint(), so no hint-bit WAL exists to pick up
-- the xid in the first place; with data_checksums on (the gpinitsystem
-- default, and this cluster's setting), the mechanism is active but still
-- does not emit a record for this table in this ordering, for reasons not
-- fully traced beyond this test. autovacuum is disabled on the target so
-- autoanalyze cannot set the hint bits before this transaction runs.
create table dtx_phase_hint_target(a int) with (autovacuum_enabled = false) distributed by (a);
insert into dtx_phase_hint_target select i from generate_series(1, 100) i;
checkpoint;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'skip', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('before_xlog_xact_prepare', 'error', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
1: begin;
1: lock table dtx_phase_lock_only in access exclusive mode;
1: select count(*) from dtx_phase_hint_target;
1: commit;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'status', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('start_performDtxProtocolCommitOnePhase', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
select gp_inject_fault('before_xlog_xact_prepare', 'reset', dbid)
  from gp_segment_configuration where role = 'p' and content = 0;
```

bimboterminator1: done

src/backend/cdb/cdbappendonlyxlog.c
bimboterminator1: Decided to simplify. Restored partially.

RekGRpth (approved):

whitehawk (approved):

PR #574, opened 2026-08-23
GG-647: Port PostgreSQL security fixes to Greengage 7

Stolb27:
Greengage 7 is based on PostgreSQL 12.22, a branch that is EOL.
Security fixes released for the supported PostgreSQL branches never reach it, so
known CVEs remain live in our tree.

This PR ports fixes for seven vulnerabilities. Nine commits are cherry-picks from
upstream; two are Greengage-specific fixes for analogous bugs found during the
audit, which no upstream patch covers.

## 7 vulnerabilities fixed

| CVE | Issue | Commits |
|---|---|---|
| CVE-2026-6478 | Timing side channel when comparing authentication secrets (SCRAM/MD5) | `283d98c1456`, `476ded13f9e` |
| CVE-2026-6637 | SQL injection and buffer overruns in `contrib/spi/refint` | `799bb0aee11` |
| CVE-2026-2005 | Buffer overflow in `pgp_pub_decrypt_bytea()` (pgcrypto) | `eef58b69cc1` |
| CVE-2026-2003 | Out-of-bounds read on unexpected `oidvector`/`int2vector` dimensions | `2f954ad3964`, `034cf1b636c`&nbsp;\* |
| CVE-2025-12817 | Missing `CREATE` privilege check on the schema in `CREATE STATISTICS` | `60806182d15` |
| CVE-2025-8715 | SQL injection at restore time via newline in a dump comment | `68e3222e5e4` |
| CVE-2026-6474 | Format-string injection via timezone abbreviation in `pg_strftime()` | `f1934bd7010`, `60764e63e9b`, `8f167848680`&nbsp;\* |

\* — Greengage-specific fixes.

## Greengage-specific findings

For each ported patch we checked whether Greengage-specific code contains an
analogous vulnerability that the upstream fix cannot cover by definition. Two
such cases were found.

**`034cf1b636c` — `cdblegacyhash_oidvector()`**
`src/backend/cdb/cdblegacyhash.c` has the same shape as upstream's
`hashoidvector()`: it reads `dim1`/`values` assuming the array is
one-dimensional and contains no nulls. The function does not exist in
PostgreSQL, so the CVE-2026-2003 patch does not touch it. Added the same
`check_valid_oidvector()` call used by the functions upstream fixed.

**`8f167848680` — `write_log()`**
`src/backend/cdb/cdbthreadlog.c` passed a prefix derived from `log_timezone` to
`vfprintf()` *as the format string*. Timezone abbreviations are not validated for
stray `%` characters, so a misconfigured `log_timezone` makes `vfprintf()`
consume more variadic arguments than were supplied. This is a structural analog
of the `timeofday()` bug upstream fixed. Applied the same remedy: render the
caller's message into its own buffer, then combine using a literal `"%s%s%s"`
format. The sink requires no privileges to reach — it is called from
unconditional error paths in `ic_udpifc.c`.

## Divergence from upstream

For eight of the nine cherry-picks, the set of changed lines is **byte-identical**
to the upstream patch. One exception:

- **`68e3222e5e4`** — the test in `src/bin/pg_dump/t/003_pg_dump_with_server.pl`
  was not ported: the file does not exist in this tree, having been introduced in
  PostgreSQL 13 (commit `2f9eb31320948b9`). The companion test in
  `t/002_pg_dump.pl` was ported and verified. The reason is recorded in the
  commit message.

Both Greengage-specific fixes are standalone commits without a cherry-pick
trailer, referencing the corresponding CVE in the message body.

## Note on refint (CVE-2026-6637)

The `contrib/spi/refint` module does not work on distributed tables in
Greengage: its SPI queries execute on a segment and are rejected by
`querytree_safe_for_qe()`. This is baked into the expected output in
`src/test/regress/expected/triggers.out` — 27 errors, not a single successful
invocation.

The vulnerable path is still reachable in utility mode (the check only applies
when `Gp_role == GP_ROLE_EXECUTE`) and when the referenced table is replicated,
so the fix was ported — but the practical risk is lower than in PostgreSQL.
Whether to keep shipping the module is worth a separate ticket.


RekGRpth:
Could you please add `postgres/postgres@` prefix before upstream commit hashes when specifying cherry-pick?

Stolb27:
> Could you please add postgres/postgres@

I'll add them with rebase when all possible question from reviewers are solved.

RekGRpth (dismissed):

contrib/spi/refint.c
```diff
@@ -462,43 +469,24 @@ check_foreign_key(PG_FUNCTION_ARGS)
...
 
-						/*
-						 * is_char_type =1 i set ' ' for define a new value
-						 */
-						snprintf(sql + strlen(sql), sizeof(sql) - strlen(sql),
-								 " %s = %s%s%s %s ",
-								 args2[k], (is_char_type > 0) ? "'" : "",
-								 nv, (is_char_type > 0) ? "'" : "", (k < nkeys) ? ", " : "");
+						appendStringInfo(&sql, " %s = %s ",
+										 args2[k], quote_literal_cstr(nv));
```
whitehawk: SPI_getvalue() returns NULL when the column's new value is SQL NULL. quote_literal_cstr() calls strlen(rawstr) with no NULL check, so a CASCADE UPDATE that sets a tracked key column to NULL crashes the backend.
  
```
postgres=# DROP TABLE IF EXISTS a, b CASCADE;
CREATE EXTENSION IF NOT EXISTS refint;

CREATE TABLE a (
        id      int4
) DISTRIBUTED REPLICATED;
CREATE UNIQUE INDEX ai ON a (id);

CREATE TABLE b (
        refb    int4
) DISTRIBUTED REPLICATED;
CREATE INDEX bi ON b (refb);

CREATE TRIGGER at BEFORE DELETE OR UPDATE ON a FOR EACH ROW
EXECUTE PROCEDURE
check_foreign_key (1, 'cascade', 'id', 'b', 'refb');

CREATE TRIGGER bt BEFORE INSERT OR UPDATE ON b FOR EACH ROW
EXECUTE PROCEDURE
check_primary_key ('refb', 'a', 'id');

INSERT INTO a VALUES (10);
INSERT INTO b VALUES (10);

SELECT * FROM a;
SELECT * FROM b;
NOTICE:  table "a" does not exist, skipping
NOTICE:  table "b" does not exist, skipping
DROP TABLE
CREATE EXTENSION
CREATE TABLE
CREATE INDEX
CREATE TABLE
CREATE INDEX
CREATE TRIGGER
CREATE TRIGGER
INSERT 0 1
INSERT 0 1
 id
----
 10
(1 row)

 refb
------
   10
(1 row)

postgres=# UPDATE a SET id = NULL WHERE id = 10;
ERROR:  Error on receive from seg2 127.0.1.1:7004 pid=607299: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.

```

Stolb27: Well, seems to be fixed by postgres/postgres@611756948ee. I'll add it too

Stolb27: Fixed in [a3f8647](https://github.com/GreengageDB/greengage/pull/574/commits/a3f8647f971b8dce51db9e1d3f10bacd0db026f1)

Stolb27:
> Could you please add postgres/postgres@

@RekGRpth, fixed too

RekGRpth (approved):

whitehawk (approved):

PR #579, opened 2026-08-27
Add support for local (coordinator-only) extensions

whitehawk:
Add support for local (coordinator-only) extensions

This patch adds a local extension mode that lets an ordinary, MPP-incompatible
PostgreSQL extension (e.g. pg_cron-style extensions that assume a single-node
Postgres and use plain heap tables plus a persistent background worker) install
and run correctly on the coordinator only, instead of requiring every table it
creates to be dispatched and distributed across segments.

To make an extension local, a special flag should be added to the .control file
of the extension - 'gg_local = true'.

While that extension's install/upgrade script is executing, every plain
CREATE TABLE it issues is given POLICYTYPE_ENTRY (no gp_distribution_policy
row — i.e. the table lives only on the coordinator) instead of the usual
hash/random distribution a table gets when created via CREATE EXTENSION in
dispatch mode.

How entry policy is assigned:
- creating_extension_local (extension.c) is a new global, set from the control
file's local flag for the duration of execute_extension_script(), mirroring the
existing creating_extension flag. It's only set outside GP_ROLE_EXECUTE, so it
never leaks onto a QE.
- When creating_extension_local is set, `transformDistributedBy()` returns
NULL (entry policy) for a plain CREATE TABLE with no explicit distribution.

'transformDistributedBy()' now also returns NULL (entry policy) if
IsBinaryUpgrade is set and no explicit policy is supplied. It is needed for
the '--binary-upgrade' scenario.
'pg_upgrade/pg_restore --binary-upgrade' restores everything through a
coordinator-only postmaster running in utility mode. Before this change,
transformDistributedBy()'s utility-mode bypass for IsBinaryUpgrade computed
a real (hash/random) policy for every restored table, including ones that were
actually entry-distributed on the source — since dump never carries a
DISTRIBUTED BY clause for those. That produced tables with a hash policy.
Now we only compute a policy when the restore actually carries a distribution
clause; fall through to entry policy otherwise. This is safe because pg_dump
always emits an explicit DISTRIBUTED BY/RANDOMLY/REPLICATED clause for any
table that has a real policy — omitting the clause is only ever the
entry-table case.

pg_dump is updated to support config tables (the tables marked with
'pg_extension_config_dump()') of the local extensions. 'dumpTableData()' now
forces INSERT-based dumping (not COPY) for a table with no distribution policy,
since COPY would otherwise dispatch as a writer-gang operation across segments —
which doesn't work for a coordinator-only table. -o/--oids option of pg_dump
combined with such a table is rejected, matching existing pg_dump's check
for INSERT-mode dumps.

gpexpand is also updated. A segment newly added by gpexpand is bootstrapped
by cloning the coordinator's data directory, so it can inherit data for any
extension-owned table that physically exists there — most notably a local
extension's coordinator-only tables, which have real data on the coordinator's
own storage (unlike a normally-distributed table, where the coordinator never
holds row data to begin with). cleanup_new_segments() now discovers every
extension-member table (via pg_depend/pg_extension, not scoped to local
extensions specifically) and issues a DELETE for each.

Current limitations:
- Partitioned tables are rejected inside a local extension script, as
supporting them would require reworking assumptions baked into the legacy
GPDB partitioning code.
- CREATE TABLE AS / CREATE MATERIALIZED VIEW are rejected: their
distribution-policy assignment goes through a different path
(setQryDistributionPolicy()/apply_motion()) that isn't aware of
creating_extension_local, and even suppressing that isn't sufficient on its
own — the populate phase still dispatches as a normal writer-gang operation.
Use CREATE TABLE + INSERT INTO ... SELECT instead.
- If the table does carry an explicit distribution — a plain DISTRIBUTED BY
clause, LIKE ... INCLUDING DISTRIBUTION - such cases are rejected as well.
As the feature is targeting native PG extensions, that is unlikely to happen in
real life.
- Still there can be SQL commands not supported by the Greengage. For ex.
queries with multiple writing CTEs.

New tests are not included in this patch, as yet there is no real extension
in our src code tree, that we can use as a local one.

### DIFF
--- a/src/backend/parser/parse_utilcmd.c
+++ b/src/backend/parser/parse_utilcmd.c
@@ -456,8 +456,8 @@ transformCreateStmt(CreateStmt *stmt, const char *queryString, bool createPartit
 		if (creating_extension_local &&
 			(stmt->partitionBy != NULL || stmt->is_part_child))
 			ereport(ERROR,
-					(errcode(ERRCODE_INVALID_TABLE_DEFINITION),
-					 errmsg("cannot create partition table in inside a local extension script")));
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("cannot create partition table inside a local extension script")));
 
 		AssertImply(stmt->is_part_parent,
 					stmt->distributedBy == NULL);
--- a/src/backend/parser/parse_utilcmd.c
+++ b/src/backend/parser/parse_utilcmd.c
@@ -43,6 +43,7 @@
 #include "catalog/pg_type_encoding.h"
 #include "commands/comment.h"
 #include "commands/defrem.h"
+#include "commands/extension.h"
 #include "commands/tablecmds.h"
 #include "commands/tablespace.h"
 #include "miscadmin.h"
@@ -444,6 +445,19 @@ transformCreateStmt(CreateStmt *stmt, const char *queryString, bool createPartit
 			ereport(ERROR,
 					(errcode(ERRCODE_INVALID_TABLE_DEFINITION),
 					 errmsg("cannot create partition table in utility mode")));
+
+		/*
+		 * Partitioned tables aren't supported inside a local extension
+		 * script: transformDistributedBy() unconditionally nulls out
+		 * distributedBy for creating_extension_local below, but partition
+		 * children require it non-NULL (see the numsegments handling
+		 * further down).
+		 */
+		if (creating_extension_local &&
+			(stmt->partitionBy != NULL || stmt->is_part_child))
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_TABLE_DEFINITION),
+					 errmsg("cannot create partition table in inside a local extension script")));
 
 		AssertImply(stmt->is_part_parent,
 					stmt->distributedBy == NULL);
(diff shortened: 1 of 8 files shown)


whitehawk:
Testing

Verified manually with a standalone `gg_local = true` test extension (contrib/gp_local_test, not included in this PR — used only for local verification, can be taken from the patch [test_local_extension.patch](https://github.com/user-attachments/files/31500046/test_local_extension.patch)):

- Manual CREATE EXTENSION / ALTER EXTENSION ... UPDATE / DROP scenarios: tables land with entry policy, correct data on the coordinator only.
- Manual pg_dump to a plain file, and restore from that file (note: there is existing issue GG-700 that will cause errors on the restore - it is a known issue and not related to this patch)
- Full ggupgrade 6.x→6.x initialize → execute → finalize cycle.
- gpcheckcat -A: no catalog issues on either postgres or template1 post-upgrade.

src/bin/pg_dump/pg_dump.c
```diff
@@ -15884,6 +15892,14 @@ processExtensionTables(Archive *fout, ExtensionInfo extinfo[],
...
 						if (strlen(extconditionarray[j]) > 0)
 							configtbl->dataObj->filtercond = pg_strdup(extconditionarray[j]);
+
+						/*
+						 * A config table is coordinator-only (entry policy)
+						 * if it has no distribution policy row.
+						 */
+						configtbl->dataObj->isCoordOnly =
+							(configtbl->distclause == NULL ||
+							 configtbl->distclause[0] == '\0');
```
RekGRpth:
```suggestion
						Assert(configtbl->distclause);
						configtbl->dataObj->isCoordOnly =
							 configtbl->distclause[0] == '\0';
```

RekGRpth:
> a standalone `gg_local = true` test extension (contrib/gp_local_test, not included in this PR

Why can't it be included as a test?

whitehawk:
> > a standalone `gg_local = true` test extension (contrib/gp_local_test, not included in this PR
> 
> Why can't it be included as a test?

I thought that introducing a completely new extension which is actually a throwaway sample (with the sole purpose of testing), would be a bad idea.

On the other hand, pretty soon we are going to introduce the `pg_cron`, that is going to be the first extension to utilize this feature. So, I guess we can setup regression testing with `pg_cron` soon.

RekGRpth (approved):

whitehawk:
@andreyaksenov can you please check PR title/description?

andreyaksenov:
> @andreyaksenov can you please check PR title/description?

Suggestion:

```
Add support for local (coordinator-only) extensions
```

whitehawk:
> > @andreyaksenov can you please check PR title/description?
> 
> Suggestion:
> 
> ```
> Add support for local (coordinator-only) extensions
> ```

Updated

src/backend/parser/parse_utilcmd.c
```diff
@@ -445,6 +446,19 @@ transformCreateStmt(CreateStmt *stmt, const char *queryString, bool createPartit
...
+		 * script: transformDistributedBy() unconditionally nulls out
+		 * distributedBy for creating_extension_local below, but partition
+		 * children require it non-NULL (see the numsegments handling
+		 * further down).
+		 */
+		if (creating_extension_local &&
+			(stmt->partitionBy != NULL || stmt->is_part_child))
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("cannot create partition table inside a local extension script")));
```
bimboterminator1, line 460: Shouldn't we reject writable external tables as well? Currently, they are created silently with random distribution.

whitehawk: I think it is not a problem:
1. standard Postgres doesn't support writable external tables (they are Greengage-specific), so we shouldn't meet them in the native Postgres extensions that we are targeting;
2. even if there were such - the data will be anyway external, and if needed, the queries that involve external tables in combination with QD-only tables should be planned properly.

We reject specifically partitioned tables and external tables with explicit 'distibuted by' as they may lead to null-dereference if used.

bimboterminator1 (approved):

PR #580, opened 2026-09-01
GG-738: Avoid dumping removed column references for heterogeneous partitions

GirgerS:
When fixing a heterogeneous partitioned hierarchy, 
ggupgrade removes references to removed columns 
from the partitions in the hierarchy, but not from the 
root table, expecting pg_dump to account for this 
case and not dump removed columns.

However, this was true only for 6.x; this logic 
was absent from 7.x, resulting in data loss.

As a fix, port the logic from 6.x. One caveat is that 
it expects to have check_heterogeneous_partition
upgrade pre-check, which is still missing. But it is 
needed only when different child partitions in the 
hierarchy have different removed column references, 
and this case wouldn't work nor before the patch, 
nor after. So it hardly counts as a regression, 
meaning that this check can be added separately.

Also, adjust the query so it doesn't collide 
with inheritance.

Regarding tests, this patch itself is a fix of a test in 
ggupgrade repository. Additionally, it is not 
currently possible to test this particular functionality
in Greengage repository.


GirgerS:
To test the changes:

```
cd # wherever greengage source code is located
git checkout GG-738

cd # wherever gggupgrade source code is located
git checkout GG-738
export IMAGE=ggdb7_ggupgrade:latest
docker build --build-arg GGDB7_IMAGE=gpdb7_u22:latest -t "$IMAGE" -f ci/new/Dockerfile .
docker run --rm -v "$(pwd)/logs:/logs" "$IMAGE" bash ggupgrade/ci/new/run_test_in_demo_cluster.bash make pg-upgrade-tests
```

riogor (dismissed):
LGTM

riogor:
> However, this was **trus** only for 6.x; this logic

Typo:

> However, this was **true** only for 6.x; this logic

src/bin/pg_dump/pg_dump.h
```diff
@@ -402,6 +402,8 @@ typedef struct _tableInfo
 	char	*distclause; /* distributed by clause */
 	char	*partclause;	/* partition definition, if table is partition parent */
 	char	*parttemplate;	/* subpartition template */
+	bool	 ignoreRootPartDroppedAttr /* true if need to ignore root partition's dropped columns */
+;
```
ShapatinAA: typo

ShapatinAA:
Shouldn't we add some kind of test for this? In greengage repo, I mean.

src/bin/pg_dump/pg_dump.h
GirgerS: Thanks

GirgerS:
> Shouldn't we add some kind of test for this? In greengage repo, I mean.

Ah, I was kinda tired yesterday, so I mistakenly thought that we can test this easily. 

We can't: adding a test would require to making a broken object in 6.x branch (so it can appear in the regression dump), which is fine by itself, but the caveat it that it would be fixed by pg_dump from 6.x when collecting regression dump.

I.e, if you try to create a similar table in the 6.x cluster:
```
-- Heterogeneous partition table with dropped column, constraint, and default
-- The root and only a subset of children have the dropped column reference.
CREATE TABLE dropped_column (
    a int CONSTRAINT positive_int CHECK (b > 0),
    b int DEFAULT 1,
    c char,
    d varchar(50)
) DISTRIBUTED BY (c)
PARTITION BY RANGE (a)
(
    PARTITION part_1 START(1) END(5),
    PARTITION part_2 START(5)
);

ALTER TABLE dropped_column DROP COLUMN d;

-- Splitting the subpartition leads to its rewrite, eliminating its dropped column
-- reference. So, after this, only part_2 and the root partition will have a
-- dropped column reference.
ALTER TABLE dropped_column SPLIT PARTITION FOR (1) AT (2)
INTO (
    PARTITION split_part_1,
    PARTITION split_part_2
);

INSERT INTO dropped_column SELECT i, i, 'a' FROM generate_series(1,99) i;
```

And then run `pg_dump postgres`, you would get just a regular table in the dump:

```
SET default_tablespace = '';

SET default_with_oids = false;

--
-- Name: dropped_column; Type: TABLE; Schema: public; Owner: regrig; Tablespace: 
--

CREATE TABLE public.dropped_column (
    a integer,
    b integer DEFAULT 1,
    c character(1),
    CONSTRAINT positive_int CHECK ((b > 0))
)
 DISTRIBUTED BY (c) PARTITION BY RANGE(a) 
          (
          PARTITION split_part_1 START (1) END (2) WITH (tablename='dropped_column_1_prt_split_part_1', appendonly='false'), 
          PARTITION split_part_2 START (2) END (5) WITH (tablename='dropped_column_1_prt_split_part_2', appendonly='false'), 
          PARTITION part_2 START (5) WITH (tablename='dropped_column_1_prt_part_2', appendonly='false')
          );
 ;
```

The whole thing with removed columns is `--binary-upgrade specific` (and even in this case it would be fixed or rejected by pg_dump from 6.x, thats the logic we are porting in this patch).

So we are out if luck this time. 

(P.S. these is src/test/regress/sql/gp_upgrade_cornercases.sql that was loaded directly into pg_upgrade tests when this was a greenplum repo. We can go this route, but it is a lengthy endeavour that would require changing the pg_upgrade test itself yet again, so it is a story for a separate ticket)

GirgerS:
@andreyaksenov, could you please review the title.

andreyaksenov:
> @andreyaksenov, could you please review the title.

suggestion:

```
Avoid dumping removed column references for heterogeneous partitions
```

ShapatinAA:
What about checking logs on "suppressing dropped column" statement?

GirgerS:
> What about checking logs on "suppressing dropped column" statement?

Can you please clarify it more, where in particular should it appear?

ShapatinAA:
>Can you please clarify it more, where in particular should it appear?

I think I finally figured it out... Do I understand right, that this code is suppose to be used upon 6.x cluster?

ShapatinAA (approved):

riogor (approved):

riogor:
Also, can you add to pr & commit message why we don't have tests in this patch?

GirgerS:
> Also, can you add to pr & commit message why we don't have tests in this patch?

Thanks for reminding, done

PR #582, opened 2026-09-03
Add support for local (coordinator-only) extensions

whitehawk:
Add support for local (coordinator-only) extensions

This patch adds a local extension mode that lets an ordinary, MPP-incompatible
PostgreSQL extension (e.g. pg_cron-style extensions that assume a single-node
Postgres and use plain heap tables plus a persistent background worker) install
and run correctly on the coordinator only, instead of requiring every table it
creates to be dispatched and distributed across segments.

To make an extension local, a special flag should be added to the .control file
of the extension - 'gg_local = true'.

While that extension's install/upgrade script is executing, every plain
CREATE TABLE it issues is given POLICYTYPE_ENTRY (no gp_distribution_policy
row — i.e. the table lives only on the coordinator) instead of the usual
hash/random distribution a table gets when created via CREATE EXTENSION in
dispatch mode.

How entry policy is assigned:
- creating_extension_local (extension.c) is a new global, set from the control
file's local flag for the duration of execute_extension_script(), mirroring the
existing creating_extension flag. It's only set outside GP_ROLE_EXECUTE, so it
never leaks onto a QE.
- When creating_extension_local is set, `transformDistributedBy()` returns
NULL (entry policy) for a plain CREATE TABLE with no explicit distribution.

'transformDistributedBy()' now also returns NULL (entry policy) if
IsBinaryUpgrade is set and no explicit policy is supplied. It is needed for
the '--binary-upgrade' scenario.
'pg_upgrade/pg_restore --binary-upgrade' restores everything through a
coordinator-only postmaster running in utility mode. Before this change,
transformDistributedBy()'s utility-mode bypass for IsBinaryUpgrade computed
a real (hash/random) policy for every restored table, including ones that were
actually entry-distributed on the source — since dump never carries a
DISTRIBUTED BY clause for those. That produced tables with a hash policy.
Now we only compute a policy when the restore actually carries a distribution
clause; fall through to entry policy otherwise. This is safe because pg_dump
always emits an explicit DISTRIBUTED BY/RANDOMLY/REPLICATED clause for any
table that has a real policy — omitting the clause is only ever the
entry-table case.

pg_dump is updated to support config tables (the tables marked with
'pg_extension_config_dump()') of the local extensions. 'dumpTableData()' now
forces INSERT-based dumping (not COPY) for a table with no distribution policy,
since COPY would otherwise dispatch as a writer-gang operation across segments —
which doesn't work for a coordinator-only table. -o/--oids option of pg_dump
combined with such a table is rejected, matching existing pg_dump's check
for INSERT-mode dumps.

gpexpand is also updated. A segment newly added by gpexpand is bootstrapped
by cloning the coordinator's data directory, so it can inherit data for any
extension-owned table that physically exists there — most notably a local
extension's coordinator-only tables, which have real data on the coordinator's
own storage (unlike a normally-distributed table, where the coordinator never
holds row data to begin with). cleanup_new_segments() now discovers every
extension-member table (via pg_depend/pg_extension, not scoped to local
extensions specifically) and issues a DELETE for each.

Current limitations:
- Partitioned tables are rejected inside a local extension script, as
supporting them would require reworking assumptions baked into the legacy
GPDB partitioning code.
- CREATE TABLE AS / CREATE MATERIALIZED VIEW are rejected: their
distribution-policy assignment goes through a different path
(setQryDistributionPolicy()/apply_motion()) that isn't aware of
creating_extension_local, and even suppressing that isn't sufficient on its
own — the populate phase still dispatches as a normal writer-gang operation.
Use CREATE TABLE + INSERT INTO ... SELECT instead.
- If the table does carry an explicit distribution — a plain DISTRIBUTED BY
clause, LIKE ... INCLUDING DISTRIBUTION - such cases are rejected as well.
As the feature is targeting native PG extensions, that is unlikely to happen in
real life.
- Still there can be SQL commands not supported by the Greengage. For ex.
queries with multiple writing CTEs.

New tests are not included in this patch, as yet there is no real extension
in our src code tree, that we can use as a local one.

(cherry picked from commit 3965f41ed8503d748e62bb8bb3e2f661770b0510)

Changes comparing to the original commit:
 - updated check for partitioned tables (as partitioned tables were completely
reworked in 7x to be more Postgres-like);
 - support for '--oids' option had been removed from pg_dump, so all checks
related to it were removed;
 - dumpTableData_insert() had been reworked in pg_dump, so some additional
changes were required to address these changes;
 - gpexpand was updated as 'execSQL()' doesn't return cursor anymore, so use
'query()' now;
 - pg_dump now rejects '--binary-upgrade' with '--no-gp-syntax': without a
distribution clause, every table - not just local-extension ones - would
restore with the wrong policy;
 - gpexpand's new-segment cleanup now truncates all extension tables in one
statement instead of one DELETE per table, and skips a database it can't
connect to instead of aborting;
 - changed the order of checks in transformDistributedBy() to keep local-extension
working in both dispatch and utility mode;
 - added test local extension and regression test.

### DIFF
--- a/gpMgmt/bin/gpexpand
+++ b/gpMgmt/bin/gpexpand
@@ -1459,6 +1459,7 @@ class gpexpand:
             where
             c.relkind = 'r' and
             d.deptype = 'e' and
+            d.objsubid = 0 and
             d.classid = 'pg_class'::regclass and
             d.refclassid = 'pg_extension'::regclass'''
 
@@ -1468,9 +1469,22 @@ class gpexpand:
             if database[0] == 'template0':
                 continue
             url.pgdb = database[0]
-            with closing (dbconn.connect(url, encoding='UTF8')) as conn:
-                cursor = dbconn.query(conn, extTablesSql)
-                db_statements_by_dbname[database[0]] = statements + ["delete from %s" % row[0] for row in cursor]
+            try:
+                with closing(dbconn.connect(url, encoding='UTF8')) as conn:
+                    ext_tables = [row[0] for row in dbconn.query(conn, extTablesSql)]
+            except Exception as e:
+                self.logger.warning(
+                    "Could not connect to database %s to look up extension-owned "
+                    "tables for new-segment cleanup, skipping: %s" % (database[0], str(e)))
+                ext_tables = []
+
+            db_statements = list(statements)
+            if ext_tables:
+                # A single TRUNCATE naming every extension table together, rather
+                # than one DELETE per table (order-dependent) or CASCADE (which
+                # could reach into unrelated, non-extension tables).
+                db_statements.append("truncate %s" % ", ".join(ext_tables))
+            db_statements_by_dbname[database[0]] = db_statements
 
         """
           Connect to each database in the new segments, and clean up the catalog tables.
--- a/gpMgmt/bin/gpexpand
+++ b/gpMgmt/bin/gpexpand
@@ -1432,6 +1432,47 @@ class gpexpand:
         statements = ["delete from pg_catalog.%s" % tab for tab in COORDINATOR_ONLY_TABLES]
 
         """
+        Purge every extension-member table unconditionally, regardless of
+        whether the owning extension is local or not. A brand-new segment's
+        pg_basebackup clone of the coordinator has no business holding
+        pre-existing rows for any table: for a local (coordinator-only)
+        extension the data is never supposed to live on a segment at all,
+        and for a normally-distributed extension table the coordinator's
+        own local storage never holds any row data to begin with, so the
+        clone can't have leaked anything real - the delete is a no-op
+        there. Any legitimate rows a normally-distributed table needs on
+        the new segments come from the standard gpexpand redistribution
+        phase.
+
+        The table list is looked up once per database, via the coordinator
+        (self.dburl), and reused for every new segment below, rather than
+        being re-queried per (segment, database) pair: a new segment's
+        catalog at this point is an untouched clone of the coordinator's -
+        the same assumption the MASTER_ONLY_TABLES cleanup above already
+        relies on - so the coordinator's answer is exactly what each
+        segment's own catalog would give too.
+        """
+        extTablesSql = '''select quote_ident(n.nspname) || '.' || quote_ident(c.relname) as fq_name from
+            pg_class c
+            join pg_namespace n on c.relnamespace = n.oid
+            join pg_depend d on c.oid = d.objid
+            where
+            c.relkind = 'r' and
+            d.deptype = 'e' and
+            d.classid = 'pg_class'::regclass and
+            d.refclassid = 'pg_extension'::regclass'''
+
+        db_statements_by_dbname = {}
+        url = copy.deepcopy(self.dburl)
+        for database in databases:
+            if database[0] == 'template0':
+                continue
+            url.pgdb = database[0]
+            with closing (dbconn.connect(url, encoding='UTF8')) as conn:
+                cursor = dbconn.query(conn, extTablesSql)
+                db_statements_by_dbname[database[0]] = statements + ["delete from %s" % row[0] for row in cursor]
+
+        """
           Connect to each database in the new segments, and clean up the catalog tables.
         """
         for seg in newSegments:
@@ -1444,11 +1485,14 @@ class gpexpand:
                                      , port=seg.getSegmentPort()
                                      , dbname=database[0]
                                      )
+
+                db_statements = db_statements_by_dbname[database[0]]
+
                 name = "gpexpand execute segment cleanup commands. seg dbid = %s, command = %s" % (
-                    seg.getSegmentDbId(), str(statements))
+                    seg.getSegmentDbId(), str(db_statements))
                 execSQLCmd = ExecuteSQLStatementsCommand(name=name
                                                          , url=dburl
-                                                         , sqlCommandList=statements
+                                                         , sqlCommandList=db_statements
                                                          )
                 self.pool.addCommand(execSQLCmd)
 
--- a/src/backend/parser/analyze.c
+++ b/src/backend/parser/analyze.c
@@ -58,6 +58,7 @@
 #include "cdb/cdbendpoint.h"
 #include "catalog/gp_distribution_policy.h"
 #include "commands/defrem.h"
+#include "commands/extension.h"
 #include "access/htup_details.h"
 #include "optimizer/clauses.h"
 #include "optimizer/tlist.h"
@@ -3050,6 +3051,22 @@ transformCreateTableAsStmt(ParseState *pstate, CreateTableAsStmt *stmt)
 {
 	Query	   *result;
 	Query	   *query;
+
+	/*
+	 * CREATE TABLE AS / CREATE MATERIALIZED VIEW are not supported inside a
+	 * local extension's install/upgrade script.
+	 * Unlike plain CREATE TABLE, their distribution-policy assignment is
+	 * not aware of creating_extension_local, and even suppressing that
+	 * policy assignment is not sufficient on its own: the populate phase
+	 * still dispatches as a normal writer-gang operation across segments.
+	 * Until that is fixed, reject this outright.
+	 */
+	if (creating_extension_local)
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("CREATE TABLE AS and CREATE MATERIALIZED VIEW are not "
+						"supported inside a local extension script"),
+				 errhint("Use CREATE TABLE followed by INSERT INTO ... SELECT instead.")));
 
 	/* transform contained query, not allowing SELECT INTO */
 	query = transformStmt(pstate, stmt->query);
--- a/src/backend/parser/parse_utilcmd.c
+++ b/src/backend/parser/parse_utilcmd.c
@@ -45,6 +45,7 @@
 #include "catalog/pg_type.h"
 #include "commands/comment.h"
 #include "commands/defrem.h"
+#include "commands/extension.h"
 #include "commands/sequence.h"
 #include "commands/tablecmds.h"
 #include "commands/tablespace.h"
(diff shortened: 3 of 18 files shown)


whitehawk:
Test extension that can be used for experiments - [test_local_extension_7x.patch](https://github.com/user-attachments/files/31776741/test_local_extension_7x.patch)

gpMgmt/bin/gpexpand
```diff
@@ -1431,6 +1431,47 @@ class gpexpand:
         """
         statements = ["delete from pg_catalog.%s" % tab for tab in COORDINATOR_ONLY_TABLES]
```
RekGRpth, line 1432: ### gpexpand: the unconditional `DELETE` over every extension-owned table can abort the cleanup phase

`gpMgmt/bin/gpexpand:1431` — the new code builds a `delete from <schema.table>`
statement for *every* member table of *any* extension and runs the list against
each new segment.

- **Foreign keys between extension tables.** The `DELETE` order is the order the
  rows come back from `pg_class`/`pg_depend` (effectively OID order), with no
  dependency ordering. If a coordinator-only extension has two tables with a FK
  between them, `DELETE FROM parent` before `DELETE FROM child` raises a FK
  violation and `cleanup_new_segments()` fails. This never surfaces for
  normally-distributed tables (0 rows in the clone), but for local extensions the
  rows really are there. Needs either `SET session_replication_role = replica` /
  disabling triggers for the duration of the deletes, or `TRUNCATE ... CASCADE`.
- **New per-database connections to the coordinator.** The loop now does
  `dbconn.connect(url)` to the coordinator's copy of every non-`template0`
  database. A database with `datallowconn = false` (other than template0) will
  throw right here, whereas previously no such connect happened at this step.

Minor points in the same query: it filters on `c.relkind = 'r'` only
(partitioned members, relkind `'p'`, are missed — but they are forbidden inside a
local extension anyway, so this is harmless yet inconsistent); no
`d.objsubid = 0` predicate.

src/backend/parser/parse_utilcmd.c
```diff
@@ -2392,10 +2393,57 @@ transformDistributedBy(ParseState *pstate,
...
+	 * runs entirely in utility mode but still needs real policies computed
+	 * for tables that were actually distributed (pg_dump always emits an
+	 * explicit DISTRIBUTED BY/RANDOMLY/REPLICATED clause for those). But if
+	 * there is no such clause, the source table had no gp_distribution_policy
+	 * row at all, i.e. it was entry-distributed -- fall through to the
+	 * ordinary utility-mode behavior instead of guessing a default policy.
 	 */
-	if (Gp_role != GP_ROLE_DISPATCH && !IsBinaryUpgrade)
+	if (Gp_role != GP_ROLE_DISPATCH &&
+		(!IsBinaryUpgrade || distributedBy == NULL))
```
RekGRpth: ### `--binary-upgrade` semantics widened globally, with no safety net

`src/backend/parser/parse_utilcmd.c:2401`

```c
if (Gp_role != GP_ROLE_DISPATCH &&
    (!IsBinaryUpgrade || distributedBy == NULL))
    return NULL;
```

Now, under `--binary-upgrade`, any `CREATE TABLE` without an explicit
`DISTRIBUTED BY/RANDOMLY/REPLICATED` clause gets an ENTRY policy. For a normal
Greengage `pg_upgrade` this is correct (and even fixes a long-standing bug where
entry-distributed tables came out with a hash policy after upgrade), because
`pg_upgrade` invokes `pg_dump` without `--no-gp-syntax`, i.e. `dumpGpPolicy = true`
and every distributed table therefore carries an explicit clause.

But the assumption "no clause ⇒ entry table" is now load-bearing, and it breaks
if `--binary-upgrade` is combined with `--no-gp-syntax` (`dumpGpPolicy = false`
→ no table carries a clause → *every* table becomes coordinator-only). There is
neither an `Assert` nor a comment about this dependency on `dumpGpPolicy`. At a
minimum this should be documented / asserted.

src/backend/parser/parse_utilcmd.c
```diff
@@ -2392,10 +2393,57 @@ transformDistributedBy(ParseState *pstate,
 	int			numsegments;
 
 	/*
-	 * utility mode creates can't have a policy.  Only the QD can have policies
+	 * utility mode creates can't have a policy.  Only the QD can have policies.
```
RekGRpth: ### No tests for changes that affect all users

The PR itself notes the absence of tests ("no real extension in our src tree to
use as a local one"). But the changes in `transformDistributedBy` (the
`--binary-upgrade` path) and in `gpexpand` affect *everyone*, not just users of
local extensions. A regression in pg_upgrade / gpexpand here is entirely
uncovered. A test for the binary-upgrade branch is writable even without a "real"
local extension (a plain entry-distributed table + `pg_upgrade`).

src/backend/parser/parse_utilcmd.c
```diff
@@ -2392,10 +2393,57 @@ transformDistributedBy(ParseState *pstate,
...
+	 * there is no such clause, the source table had no gp_distribution_policy
+	 * row at all, i.e. it was entry-distributed -- fall through to the
+	 * ordinary utility-mode behavior instead of guessing a default policy.
 	 */
-	if (Gp_role != GP_ROLE_DISPATCH && !IsBinaryUpgrade)
+	if (Gp_role != GP_ROLE_DISPATCH &&
+		(!IsBinaryUpgrade || distributedBy == NULL))
+		return NULL;
+
+	if (creating_extension_local)
```
RekGRpth, line 2401: ### `creating_extension_local` in utility mode bypasses the new checks

`creating_extension_local` is set for any `Gp_role != GP_ROLE_EXECUTE`, i.e.
including `GP_ROLE_UTILITY` (`extension.c:938`). But in utility mode
`transformDistributedBy` returns `NULL` at the very top
(`Gp_role != GP_ROLE_DISPATCH && !IsBinaryUpgrade`) and never reaches the
`if (creating_extension_local)` block. Result: a local-extension script run in
utility mode silently accepts `CREATE TABLE ... PARTITION BY` and an explicit
`DISTRIBUTED BY` as entry, instead of raising the error it does in dispatch mode.
Not a regression (in utility mode that was already the default behavior), but the
new guard checks are incomplete. The PR description's claim — "It's only set
outside GP_ROLE_EXECUTE, so it never leaks onto a QE" — ignores utility mode.

src/backend/parser/analyze.c
```diff
@@ -3051,6 +3052,22 @@ transformCreateTableAsStmt(ParseState *pstate, CreateTableAsStmt *stmt)
...
+	/*
+	 * CREATE TABLE AS / CREATE MATERIALIZED VIEW are not supported inside a
+	 * local extension's install/upgrade script.
+	 * Unlike plain CREATE TABLE, their distribution-policy assignment is
+	 * not aware of creating_extension_local, and even suppressing that
+	 * policy assignment is not sufficient on its own: the populate phase
+	 * still dispatches as a normal writer-gang operation across segments.
+	 * Until that is fixed, reject this outright.
+	 */
+	if (creating_extension_local)
```
RekGRpth, line 3064: ### `analyze.c`: the CTAS/matview rejection also fires in utility mode

`src/backend/parser/analyze.c:3061` — the check is not tied to `Gp_role`, so
combined with `CREATE TABLE AS` / `CREATE MATERIALIZED VIEW` /
`SELECT INTO` inside a local extension are rejected even during a single-node
utility install, where there is no distribution at all. Acceptable as a
"documented limitation", but worth calling out.

src/backend/parser/parse_utilcmd.c
whitehawk: Actually `--binary-upgrade` + `--no-gp-syntax` didn't work together even before my changes. Binary backup just transplates the relation files as is on the restored segments, and without `distibuted by` clause all restored tables will fallback to the same default distribution policy. Therefore, with very high probability (unless all of our tables had default distribution at the backup time, which is almost impossible in a real-world installation), some part of tables will be broken.

Therefore, I've just forbidden `--binary-upgrade` + `--no-gp-syntax` combination in this patch.

gpMgmt/bin/gpexpand
whitehawk: > FK ordering

 switched from one DELETE per table to a single TRUNCATE tbl1, tbl2, ... naming every extension-owned table together — Postgres resolves dependencies among co-truncated tables itself, so ordering doesn't matter, and unlike CASCADE it can't reach outside the named set into unrelated tables. One correction to the original framing though: Greengage doesn't actually enforce FK constraints at all (refer to `validateForeignKeyConstraint()`). So the specific crash isn't reachable today; the TRUNCATE change is still worth keeping as more correct/atomic regardless"

> datallowconn = false

wrapped the per-database connect in try/except.

  Also added the d.objsubid = 0 predicate.

src/backend/parser/parse_utilcmd.c
whitehawk: > A test for the binary-upgrade branch is writable even without a "real"
local extension (a plain entry-distributed table + pg_upgrade).

Currently there is no way to produce a plain entry-distributed relation that's also **catalog-consistent across coordinator and segments**, without going through a real local extension - and I'd rather not add one just for this.
test_gpdb.sh already runs a real pg_upgrade over the regression database in CI and diffs the before/after schema dumps, which does exercise this same code path transformDistributedBy()'s --binary-upgrade branch) for every ordinarily-distributed table in that suite - it just can't reach the new no-clause/entry fallback specifically without an entry-distributed fixture there. Given the tradeoff, I think passing that existing coverage is enough for now.

src/backend/parser/parse_utilcmd.c
whitehawk: changed the order of checks

src/backend/parser/analyze.c
whitehawk: This limitation is documented in the PR description, which will go into the commit description. I think it is enough for now.

RekGRpth:
squash or rebase?

whitehawk:
> squash or rebase?

I think squash will be Ok here, as we do not need to preserve authorship.

RekGRpth (dismissed):

RekGRpth (approved):

red1452 (approved):

PR #584, opened 2026-09-03
GG-548: Remove references to unsupported distros from Greengage 6

GirgerS:
Remove distributions that has reached the end of their 
support cycle. These include: 
- Ubuntu 16, 18, 20
- Rhel 5, 6, 7
- Centos 6, 7
- Suse 10

Some caveats: 
- It addition to other distributions, remove oel7 as it had 
  some dependency on rhel7,  even though it has not 
  reached end of life yet.

- Instructions for MacOS were removed, because
  they relied on a CentOS image

- There are some leftover mentions of CentOS 
 (with no particular version), but it seems reasonable 
 as long as CentOS 9 and CentOS 10 exist.

- Mentions of the removed distributions are kept in 
 some comments where they may have historical value,
 and in the tests, where they have no available 
 substitutions (i.e. gppkg).

- Concourse pipelines and unused
  concourse scripts will be cleaned up in a separate
  patch

- Behavior of the code is kept the same.


GirgerS:
Some other notable remaining mentions of the distros:

`gpcontrib/credcheck/README.md` is not changed for purely arbitrary reason: the latest [credcheck](https://github.com/HexaCluster/credcheck) still has the same readme, so I decided not to create additional conflicts. This is arguable of course.


`src/test/binary_swap/expected/resgroup_cleanup.out` and `src/test/binary_swap/expected/resgroup_switch_group.out`
are left unchanged unlike other tests with similar vm, because it seems that they are not being run anywhere.

gpAux/gpperfmon/README.md
```diff
@@ -7,10 +7,7 @@ Find more information about the architecture on [the wiki page](https://github.c
...
 
 ### libsigar:
-	https://github.com/hyperic/sigar
-	For CentOS 6:
-		yum install sigar-devel
-	For macOS and CentOS 7 (and others):
+	For Rocky 8 (and others):
 	    Use this updated fork: https://github.com/boundary/sigar
 		to build:
 	    `mkdir build && cd build && cmake .. && make && make install`
```
KnightMurloc: Maybe just:
```
Use this updated fork: https://github.com/boundary/sigar
to build:
`mkdir build && cd build && cmake .. && make && make install`
```

concourse/scripts/ic_gpdb_resgroup.bash
```diff
@@ -5,16 +5,7 @@ set -eox pipefail
...
-if [ "$TEST_OS" = centos6 ]; then
-    CGROUP_BASEDIR=/cgroup
-else
-    CGROUP_BASEDIR=/sys/fs/cgroup
-fi
-
-if [ "$TEST_OS" = centos7 ]; then
-    CGROUP_AUTO_MOUNTED=1
-fi
+CGROUP_BASEDIR=/sys/fs/cgroup
```
KnightMurloc: Is this file is used at all? Do we want to modify files whose functionality we cannot test?

concourse/scripts/setup_gpadmin_user.bash
```diff
@@ -147,7 +141,7 @@ determine_os() {
 
 # Set the "Set-User-ID" bit of ping, or else gpinitsystem will error by following message:
 # [FATAL]:-Unknown host d6f9f630-65a3-4c98-4c03-401fbe5dd60b: ping: socket: Operation not permitted
-# This is needed in centos7, sles12sp5, but not for centos6, ubuntu18.04
+# This is needed at least for sles12sp5
 workaround_before_concourse_stops_stripping_suid_bits() {
   chmod u+s $(which ping)
 }
```
KnightMurloc, line 153: We don't use concourse. Do we still need this function?

src/test/isolation2/output/resgroup/enable_resgroup_validate.source
```diff
@@ -1,61 +1,69 @@
 -- start_ignore
-! chmod 755 @cgroup_mnt_point@/cpu/gpdb/*/;
+! chmod 755 /sys/fs/cgroup/cpu/gpdb/*/;
+chmod: cannot access '/sys/fs/cgroup/cpu/gpdb/*/': No such file or directory
```
KnightMurloc: > chmod: cannot access '/sys/fs/cgroup/cpu/gpdb/*/': No such file or directory

Why this line is added?

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -10,12 +10,12 @@ ERROR:  resource group "rg1_cpu_test" does not exist
 DROP RESOURCE GROUP rg2_cpu_test;
 ERROR:  resource group "rg2_cpu_test" does not exist
 DROP VIEW IF EXISTS busy;
-DROP VIEW
+DROP
```
KnightMurloc: Why this file is changed?

src/test/isolation2/output/resgroup/resgroup_memory_runaway.source
```diff
@@ -7,7 +7,7 @@ DROP RESOURCE GROUP rg2_memory_test;
 ERROR:  resource group "rg2_memory_test" does not exist
 -- end_ignore
 
-CREATE OR REPLACE FUNCTION resGroupPalloc(float) RETURNS int AS '@abs_builddir@/../../regress/regress@DLSUFFIX@', 'resGroupPalloc' LANGUAGE C READS SQL DATA;
+CREATE OR REPLACE FUNCTION resGroupPalloc(float) RETURNS int AS '/home/gpadmin/gpdb_src/src/test/isolation2/resgroup/../../regress/regress.so', 'resGroupPalloc' LANGUAGE C READS SQL DATA;
```
KnightMurloc: Why this file is changed?

src/test/regress/expected/DML_over_joins.out
```diff
@@ -69,9 +69,8 @@ explain update s set b = b + 1 where exists (select 1 from r where s.a = r.b);
                            ->  Hash  (cost=8.00..8.00 rows=100 width=18)
                                  ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..8.00 rows=100 width=18)
                                        ->  Seq Scan on s  (cost=0.00..4.00 rows=34 width=18)
- Planning time: 1.342 ms
  Optimizer: Postgres query optimizer
-(14 rows)
+(13 rows)
```
KnightMurloc, line 73: Why?

src/test/regress/expected/DML_over_joins.out
```diff
@@ -89,9 +88,8 @@ explain delete from s where exists (select 1 from r where s.a = r.b);
                            ->  Hash  (cost=8.00..8.00 rows=100 width=14)
                                  ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..8.00 rows=100 width=14)
                                        ->  Seq Scan on s  (cost=0.00..4.00 rows=34 width=14)
- Planning time: 1.130 ms
  Optimizer: Postgres query optimizer
-(14 rows)
+(13 rows)
```
KnightMurloc, line 92: Same

src/test/regress/expected/DML_over_joins.out
```diff
@@ -1062,6 +1060,7 @@ create table r (a int, b int) distributed by (a);
 create table s (a int, b int) distributed by (a);
 create table m ();
 NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause, and no column type is suitable for a distribution key. Creating a NULL policy entry.
+WARNING:  creating a table with no columns.
```
KnightMurloc, line 1063: Why?

src/test/regress/expected/DML_over_joins_optimizer.out
```diff
@@ -58,10 +58,10 @@ insert into s select generate_series(1, 100), generate_series(1, 100) * 4;
...
                                                         QUERY PLAN                                                         
 ---------------------------------------------------------------------------------------------------------------------------
- Update  (cost=0.00..868.53 rows=34 width=1)
-   ->  Split  (cost=0.00..862.80 rows=67 width=22)
-         ->  Result  (cost=0.00..862.80 rows=34 width=22)
-               ->  Hash Semi Join  (cost=0.00..862.80 rows=34 width=18)
+ Update  (cost=0.00..868.86 rows=34 width=1)
+   ->  Split  (cost=0.00..863.13 rows=67 width=22)
+         ->  Result  (cost=0.00..863.13 rows=34 width=22)
+               ->  Hash Semi Join  (cost=0.00..863.13 rows=34 width=18)
```
KnightMurloc, line 64: Why?

src/test/regress/expected/DML_over_joins_optimizer.out
```diff
@@ -77,18 +77,19 @@ update s set b = b + 1 where exists (select 1 from r where s.a = r.b);
...
+                     Filter: (NOT (a IS NULL))
                ->  Hash  (cost=431.15..431.15 rows=3334 width=4)
                      ->  Result  (cost=0.00..431.15 rows=3334 width=4)
                            ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.14 rows=3334 width=4)
                                  Hash Key: r.b
                                  ->  Seq Scan on r  (cost=0.00..431.07 rows=3334 width=4)
- Optimizer: Pivotal Optimizer (GPORCA) version 2.74.0
-(11 rows)
+ Optimizer: Pivotal Optimizer (GPORCA)
+(12 rows)
```
KnightMurloc, line 92: Do this test even run on our CI?

KnightMurloc (commented):
> The changes to concourse pipelines were never tested,
(as they are outdated and are not being used anywhere).
The reason for not removing them entirely is to make
this commit smaller.

Is there a point to changing them then?

GirgerS:
> Is there a point to changing them then?

I thought about it a bit before.

If we don't change them, then we will have references `rhel7` or other nonexistent images, which is not great.
Removing them is kinda outside of the scope of the ticket (though I have removed some files which depend entirely on removed images). 
And just changing these files wouldn't hurt anyone.

That's said, I like the idea of removing them the most. If you are no against it, I'll ask if we can just do it tomorrow.

KnightMurloc:
> That's said, I like the idea of removing them the most. If you are no against it, I'll ask if we can just do it tomorrow.

I just don't see the point in maintaining files about which we have no idea if they work at all.

gpAux/gpperfmon/README.md
GirgerS: Good idea

concourse/scripts/ic_gpdb_resgroup.bash
GirgerS: From what I've seen, not. 

I'll test it [here](https://github.com/GreengageDB/greengage/pull/585).

concourse/scripts/setup_gpadmin_user.bash
GirgerS: I'm not sure about this one, because I had the same exact error when installing Greengage locally. 

I'll test it [here](https://github.com/GreengageDB/greengage/pull/585).

Ah, and also, this file is used in the current CI (concourse/scripts/setup_gpadmin_user.bash -> ci/scripts/init_containers.sh -> ci/scripts/run_behave_tests.bash).

src/test/isolation2/output/resgroup/enable_resgroup_validate.source
GirgerS: The idea behind this change (and for running tests that had `centos` vm overall) is to replace them with results from the latest regression runs, instead of building a frankenstein. 

Having up-to-date regression test results (especially because we already have to change them) just seems right to me.

src/test/regress/expected/DML_over_joins_optimizer.out
GirgerS: [Yes](https://productionresultssa18.blob.core.windows.net/actions-results/80ac312d-c8d6-4137-9744-692c0f52bdf5/workflow-job-run-3455f97b-f5c9-537d-950e-236a6114d7a6/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-09-07T14%3A46%3A39Z&sig=e5HCW1K0GVFUw1Z7%2FKFCI4w9MIIpiCYGTHmTHW%2FjOOY%3D&ske=2026-09-07T17%3A49%3A31Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-09-07T13%3A49%3A31Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-09-07T14%3A36%3A34Z&sv=2025-11-05).

Grep for DML_over_joins.

src/test/isolation2/output/resgroup/enable_resgroup_validate.source
KnightMurloc: But the output still doesn't seem right to me. Is it OK that we get an error?

GirgerS: I don't know if my reasoning is right or not, but anyways. 

This part of the scripts seems like a cleanup of the previously existing cgroups (if any), i.e. they are finally removed [here](https://github.com/GreengageDB/greengage/blob/6.x/src/test/isolation2/output/resgroup/enable_resgroup_validate.source#L8-L18), so [this](https://github.com/GreengageDB/greengage/blob/6.x/src/test/isolation2/output/resgroup/enable_resgroup_validate.source#L22-L25) command can fail.

So it makes sense why it is ignored, and why we can get errors

Also, output is taken from the latest runs of resgroup tests.

src/test/regress/expected/DML_over_joins_optimizer.out
KnightMurloc: Then why it pass? without these changes.

src/test/isolation2/output/resgroup/enable_resgroup_validate.source
KnightMurloc: If it in the ignore block, then maybe remove it completely?

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -147,49 +147,49 @@ SET
...
- {"rg1_cpu_test": {"1": 89.900000000000006, "0": 89.569999999999993, "2": 89.650000000000006, "-1": 90.480000000000004}, "rg2_cpu_test": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}, "admin_group": {"1": 0.01, "0": 0.01, "2": 0.01, "-1": 0.10000000000000001}, "default_group": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}} 
+ fetch_sample                                                                                                                                                                                                                                            
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ {"default_group": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}, "admin_group": {"-1": 0.13, "0": 0.08, "1": 0.13, "2": 0.07}, "rg1_cpu_test": {"-1": 0.03, "0": 91.37, "1": 91.58, "2": 91.79}, "rg2_cpu_test": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}}
```
KnightMurloc: value of `-1` segment is significantly different.
Old: `90.480000000000004`
New: `0.03`

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -147,49 +147,49 @@ SET
...
- {"rg1_cpu_test": {"1": 90.109999999999999, "0": 90.209999999999994, "2": 90.090000000000003, "-1": 88.549999999999997}, "rg2_cpu_test": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}, "admin_group": {"1": 0.01, "0": 0.02, "2": 0.02, "-1": 0.10000000000000001}, "default_group": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}} 
+ fetch_sample                                                                                                                                                                                                                                            
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ {"default_group": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}, "admin_group": {"-1": 0.11, "0": 0.07, "1": 0.07, "2": 0.08}, "rg1_cpu_test": {"-1": 0.02, "0": 90.58, "1": 90.85, "2": 90.84}, "rg2_cpu_test": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}}
```
KnightMurloc: Same

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -147,49 +147,49 @@ SET
...
- {"rg1_cpu_test": {"1": 90.040000000000006, "0": 90.069999999999993, "2": 90.049999999999997, "-1": 87.859999999999999}, "rg2_cpu_test": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}, "admin_group": {"1": 0.02, "0": 0.02, "2": 0.01, "-1": 0.10000000000000001}, "default_group": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}} 
+ fetch_sample                                                                                                                                                                                                                                            
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ {"default_group": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}, "admin_group": {"-1": 0.12, "0": 0.07, "1": 0.07, "2": 0.08}, "rg1_cpu_test": {"-1": 0.03, "0": 89.85, "1": 89.88, "2": 89.89}, "rg2_cpu_test": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}}
```
KnightMurloc: Same

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -147,49 +147,49 @@ SET
...
- {"rg1_cpu_test": {"1": 90.030000000000001, "0": 90.260000000000005, "2": 90.140000000000001, "-1": 90.290000000000006}, "rg2_cpu_test": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}, "admin_group": {"1": 0.0, "0": 0.02, "2": 0.01, "-1": 0.080000000000000002}, "default_group": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}} 
+ fetch_sample                                                                                                                                                                                                                                         
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ {"default_group": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}, "admin_group": {"-1": 0.11, "0": 0.06, "1": 0.07, "2": 0.1}, "rg1_cpu_test": {"-1": 0.04, "0": 90.5, "1": 90.6, "2": 90.31}, "rg2_cpu_test": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}}
```
KnightMurloc: Same

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -147,49 +147,49 @@ SET
...
- {"rg1_cpu_test": {"1": 90.019999999999996, "0": 90.310000000000002, "2": 90.049999999999997, "-1": 90.700000000000003}, "rg2_cpu_test": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}, "admin_group": {"1": 0.02, "0": 0.02, "2": 0.02, "-1": 0.10000000000000001}, "default_group": {"1": 0.0, "0": 0.0, "2": 0.0, "-1": 0.0}} 
+ fetch_sample                                                                                                                                                                                                                                          
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ {"default_group": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}, "admin_group": {"-1": 0.12, "0": 0.07, "1": 0.05, "2": 0.1}, "rg1_cpu_test": {"-1": 0.02, "0": 89.79, "1": 90.5, "2": 90.34}, "rg2_cpu_test": {"-1": 0.0, "0": 0.0, "1": 0.0, "2": 0.0}}
```
KnightMurloc: Same

src/test/isolation2/output/resgroup/resgroup_memory_runaway.source
```diff
@@ -265,7 +265,40 @@ DROP
 
 -- start_ignore
 ! gpconfig -c runaway_detector_activation_percent -v 99;
+20260905:00:20:16:014704 gpconfig:cdw:gpadmin-[INFO]:-completed successfully with parameters '-c runaway_detector_activation_percent -v 99'
+
 ! gpstop -ari;
+20260905:00:20:16:014755 gpstop:cdw:gpadmin-[INFO]:-Starting gpstop with args: -ari
```
KnightMurloc: We don't need output in ignore block.

concourse/scripts/ic_gpdb_resgroup.bash
GirgerS: Yeah, it not being used, so I reverted the changes.

concourse/scripts/setup_gpadmin_user.bash
GirgerS: It seems to work either way, so I removed it.

src/test/isolation2/output/resgroup/enable_resgroup_validate.source
GirgerS: Great idea, I've done it

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
GirgerS: Overall, to remove mentions to centos similar to:

`20210405:09:27:22:015397 gpstop:hubert-gp-centos:huanzhang`

And I reverted this change because it is a part of start_ignore block.

src/test/regress/expected/DML_over_joins.out
GirgerS: Because in order for `Planning time` to be [printed](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/commands/explain.c#L629-L638), `ANALYZE` option should be [set](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/commands/explain.c#L219-L220), and it is not the case here.

src/test/regress/expected/DML_over_joins.out
GirgerS: Because up-to-date Greengage gives [an error](https://github.com/GreengageDB/greengage/blob/6.x/src/backend/commands/tablecmds.c#L777-L786) for a table with no columns.

src/test/regress/expected/DML_over_joins_optimizer.out
GirgerS: I don't know, any commit that has changed costing model in ORCA, like [this](810d51b4c5a2bf37b5a1f8f5b72b34a75f77f8a4). I doubt that tracing it is relevant.

src/test/regress/expected/DML_over_joins_optimizer.out
GirgerS: Because src/test/regress/gpdiff.pl reports no diff:

Try downloading recent test results (for example from [here](https://github.com/GreengageDB/greengage/actions/runs/34073195514/job/101597554181)), and then compare them with the unchanged file, e.g. for me:

```bash
/home/regrig/programming/greengage-6.x/src/test/regress/gpdiff.pl -I HINT: -I CONTEXT: -I GP_IGNORE: -gpd_init /home/regrig/programming/greengage-6.x/src/test/regress/init_file output_expected_orca.sql output_orca.sql
```

Init file comes from [here](https://github.com/GreengageDB/greengage/blob/6.x/src/test/regress/GNUmakefile#L205), `-I HINT: -I CONTEXT: -I GP_IGNORE:` comes from [here](https://github.com/GreengageDB/greengage/blob/6.x/src/test/regress/pg_regress.c#L84) and goes [here](https://github.com/GreengageDB/greengage/blob/6.x/src/test/regress/pg_regress.c#L1952-L1955)

src/test/isolation2/output/resgroup/resgroup_memory_runaway.source
GirgerS: Reverted these changes

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
GirgerS: Well, coordinator may have less work to do. But it is purely a speculation.

I reverted these changes anyway, as they are a part of ignore block

src/test/isolation2/output/resgroup/resgroup_memory_runaway.source
```diff
@@ -305,3 +270,8 @@ DROP
 DROP
 0q: ... <quitting>
 
+-- start_ignore
+! gpconfig -c runaway_detector_activation_percent -v 100;
+
+! gpstop -ari;
+-- end_ignore
```
KnightMurloc, line 276: Why this was added?

src/test/isolation2/output/resgroup/resgroup_cpu_rate_limit.source
```diff
@@ -489,39 +489,39 @@ ERROR:  canceling statement due to user request
 
 -- start_ignore
 ! gpconfig -c gp_resource_group_cpu_ceiling_enforcement -v on;
-20210405:09:27:21:015004 gpconfig:hubert-gp-centos:huanzhang-[INFO]:-completed successfully with parameters '-c gp_resource_group_cpu_ceiling_enforcement -v on'
+20260905:00:21:48:015758 gpconfig:cdw:gpadmin-[INFO]:-completed successfully with parameters '-c gp_resource_group_cpu_ceiling_enforcement -v on'
```
KnightMurloc: Command output in the ignore block in next file is removed. Or should we update them at all? I don't think mentioning CentOS in them is a big deal.

src/backend/gporca/libgpos/include/gpos/common/CDebugCounter.h
```diff
@@ -89,8 +89,9 @@
 #include "gpos/common/CTimerUser.h"
 
 #ifdef GPOS_DEBUG
-// for now, disable this on Linux, as it caused the CentOS 6 and 7 builds to fail
-// (for some not well-known cause, possibly due to an inconsistency in a test script)
+// for historical reasons, this was disabled on Linux, as it caused the CentOS 6 and 7 builds to fail
+// (for some not well-known cause, possibly due to an inconsistency in a test script).
+// they are no longer supported, can we enabled this again?
```
KnightMurloc: Maybe do it now? In my tests, it doesn't cause any problems.

src/test/isolation2/output/resgroup/resgroup_memory_runaway.source
GirgerS: Because [input](https://github.com/GreengageDB/greengage/blob/6.x/src/test/isolation2/input/resgroup/resgroup_memory_runaway.source#L168-L171) has it. It is an ignore block anyway, so obviously it is not necessary, but because other ignore blocks are in the file, why not adding this one too

GirgerS:
> Or should we update them at all? I don't think mentioning CentOS in them is a big deal.

Right, it is debatable, at least because mentioned centos has no version.

But I still decided to change them because:
- they pollute `grep` results for centos (of cource these files can be ignored, but still)
- and they have no value to do so, as a name of some machine of some developer from the past. I.e. test wouldn't benefit from this information.

So, as a cleanup, and because we already touch other tests, I thing it is reasonable

KnightMurloc:
What about:
https://github.com/GreengageDB/greengage/blob/f813182e95c85a5d00e0ac8870d51ec9324c7de2/src/interfaces/ecpg/compatlib/Makefile#L33-L36

GirgerS:
> What about:
> 
> https://github.com/GreengageDB/greengage/blob/f813182e95c85a5d00e0ac8870d51ec9324c7de2/src/interfaces/ecpg/compatlib/Makefile#L33-L36

Seemls like I've missed this one, thanks.

Also, it had duplicating `SHLIB_EXPORTS = exports.txt` lines, which became obvious after removing centos branch.

VoidZeroNull0:
In several places support for MacOS was removed. We do not support it too?

KnightMurloc:
@GirgerS can you review test failure? It's not looks like a flaky and can be related to the patch changes.

GirgerS:
> @GirgerS can you review test failure? It's not looks like a flaky and can be related to the patch changes.

Indeed.

First of all, C [function](https://github.com/GreengageDB/greengage/blob/d55cc9315a63eab6ee8de37ae360da23bbe34632/src/interfaces/gppc/test/gppc_demo.c#L268-L278) that was causing the error (the same thing for other functions) is marked as [immutable](https://github.com/GreengageDB/greengage/blob/d55cc9315a63eab6ee8de37ae360da23bbe34632/src/interfaces/gppc/test/sql/gppc_demo.sql#L175), so the optimizers try to [evaluate](https://github.com/GreengageDB/greengage/blob/d55cc9315a63eab6ee8de37ae360da23bbe34632/src/backend/optimizer/util/clauses.c#L4233-L4240) it at planning time. 

Second of all, debug counter code that we enabled fails on 
`SELECT NULL;` or similar queries, causing ORCA to fallback.

So, ORCA tries to optimize `SELECT errfunc1('The quick brown fox jumps over the lazy dog');`, calls `errfunc1` at planning time, we get `LOG: ...`, then fails [here](https://github.com/GreengageDB/greengage/blob/d55cc9315a63eab6ee8de37ae360da23bbe34632/src/backend/gporca/libgpopt/src/optimizer/COptimizer.cpp#L186) cause the function returns null, fallbacks to the regular planner, the regular planner calls the functions, and thats how we get the second `LOG: ...`.


Because of that, I don't want to enable this code without carefully reading thought it, and I don't thinks it is a good idea to do it here.


Regarding other tests:
- Behave tests worked
- [this](https://github.com/GreengageDB/greengage/actions/runs/35024749531/job/104576471122) is [flaky](https://tracker.yandex.ru/GG-730), it seems
- and also, a stange one-time failure of [rocky 9](https://github.com/GreengageDB/greengage/actions/runs/35062045058/job/104684286738?pr=584), not sure if it is interesting

P.S. I guess that Greengage CI [doesn't](https://github.com/GreengageDB/greengage/actions/runs/35062045058/job/104703138350?pr=584) want to turn green today

GirgerS:
> In several places support for MacOS was removed. We do not support it too?

I believe that we should support MacOS development through docker, though nobody tested it in some time.

About removed files, it is mostly [‎README.docker.md](https://github.com/GreengageDB/greengage/pull/584/changes#diff-97290646ec29806fd928fa2a5771dea2200480f4d3ddf71dff976fe76d939d85), right? It relied on removed CentOS image, I don't know of a way to test MacOS, and I don't want to make untested changes to user documentation.

VoidZeroNull0 (approved):

KnightMurloc (approved):

GirgerS:
@andreyaksenov, any suggestions about the title?

andreyaksenov:
> @andreyaksenov, any suggestions about the title?

suggestion:

```
Remove references to unsupported distros from Greengage 6
```

PR #586, opened 2026-09-08
GG-740: Fix resgroup_io_limit_alter_in_transaction test

riogor:
resgroup v2 resgroup_io_limit_alter_in_transaction test was broken.
Mostly, output tables miss spaces and identation.
But there is one case with two tablespaces on one resgroup.
When ALTER RESOURCE GROUP statement with both tablespaces is executed
it fails. Issue is that db prohibits having tablespaces with different io_limit parameters
on same block device in one query.

A way to mitigate it is to use loop devices and place tablespaces separately on them.
This solves the issue as they simulate block device good enough.

Add loop devices creation and swap tablespace folders with them.
Fix spaces.

Ticket: GG-740


riogor:
CI workflow was added temporarily to verify the fix. It will be deleted once pr is approved.
Taken from this PR https://github.com/GreengageDB/greengage/pull/581

magf:
> CI workflow was added temporarily to verify the fix.

A review request is automatically sent to the `CI` team as soon as you mark your PR as ready for review.
Please create a draft if your PR is not yet ready or just for tests.
This will not trigger a review request to the CI team.

riogor:
> > CI workflow was added temporarily to verify the fix.
> 
> A review request is automatically sent to the `CI` team as soon as you mark your PR as ready for review. Please create a draft if your PR is not yet ready or just for tests. This will not trigger a review request to the CI team.

PR is ready for review.

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -64,6 +64,42 @@ setup_cgroup_v2() {
...
 
+setup_loop_devices() {
+    local i dir img src prev=""
+    for i in 1 2; do
+        dir="$ISOLATION2_TESTTABLESPACE/rg_io_limit_loop_$i"
+        img="$ISOLATION2_TESTTABLESPACE/io_limit_fs_$i.img"
+
+        mkdir -p "$dir"
+        if ! mountpoint -q "$dir"; then
+            dd if=/dev/zero of="$img" bs=1M count=64
```
KnightMurloc: better use `truncate -s 1M "$img"`

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -64,6 +64,42 @@ setup_cgroup_v2() {
...
+    trap 'cleanup_cgroup_v2 || true; teardown_loop_devices' EXIT 
+}
+
+teardown_loop_devices() {
+    local i dir
+    for i in 1 2; do
+        dir="$ISOLATION2_TESTTABLESPACE/rg_io_limit_loop_$i"
+        mountpoint -q "$dir" && umount "$dir"
+    done
+    return 0
```
KnightMurloc: Why is always zero? The `cleanup_cgroup_v2` preserve exit code.

concourse/scripts/ic_gpdb_resgroup_v2.bash
riogor: Done

concourse/scripts/ic_gpdb_resgroup_v2.bash
riogor: Rewrote it with general cleanup function, which captures retcode and runs each cleanup function.

KnightMurloc (dismissed):

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -60,8 +70,44 @@ setup_cgroup_v2() {
...
     chmod a+w /sys/fs/cgroup/cgroup.procs
+    
+    exit_handlers+=(cleanup_cgroup_v2)
+}
+
+cleanup_loop_devices() {
+    local i dir
+    for i in 1 2; do
+        dir="$ISOLATION2_TESTTABLESPACE/rg_io_limit_loop_$i"
+        mountpoint -q "$dir" && umount "$dir"
```
ShapatinAA, line 82: Because of [this](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#The-Set-Builtin-1) (`-e` flag description, case of command before `&&` operator)...
```suggestion
        if ! mountpoint -q "$dir"; then
            fatal "the directory $dir is not a mountpoint, or device is not a block device"
        fi
        umount "$dir";
```

concourse/scripts/ic_gpdb_resgroup_v2.bash
```diff
@@ -12,6 +12,18 @@ TEST_OS=${TEST_OS:-ubuntu}
 GPDB_DEMO_DATADIRS=/home/gpadmin/gpdb_src/gpAux/gpdemo/datadirs
 ISOLATION2_TESTTABLESPACE=/home/gpadmin/gpdb_src/src/test/isolation2/testtablespace
 
+exit_handlers=()
+
+run_exit_handlers() {
+    local rc=$?
+    local h
+    for h in "${exit_handlers[@]}"; do
+        "$h" || true
```
ShapatinAA, line 21: Why always true is suitable? Btw if it's `set -eox`, it means, that `smth || true` will never return true, because of `-e` option.

concourse/scripts/ic_gpdb_resgroup_v2.bash
riogor: Good find, thanks

concourse/scripts/ic_gpdb_resgroup_v2.bash
KnightMurloc: > will never return true, because of -e option.

It's not work like that. If first part exit with non-zero exit code the second part will cancel it. Even with `-e`.

concourse/scripts/ic_gpdb_resgroup_v2.bash
KnightMurloc: Why do we want to exit from the script in case of the error in cleanup code?

concourse/scripts/ic_gpdb_resgroup_v2.bash
ShapatinAA: I was testing it out with this script and it fails, did I miss something?

<details><summary>Script</summary>
<p>

```bash
#!/bin/bash

set -eox pipefail

fatal() {
    echo "FATAL: $*" >&2
    exit 1
}

cleanup_loop_devices() {
    local i dir
    for i in 1 2; do
        dir="/tmp/haha"
        if ! mountpoint -q "$dir"; then
            fatal "the directory $dir is not a mountpoint, or device is not a block device"
        fi
        umount "$dir";
    done
}

exit_handlers+=cleanup_loop_devices

for h in "${exit_handlers[@]}"; do
        "$h" || true
done
```

</p>
</details> 

<details><summary>Output</summary>
<p>


```bash
artem@artem-CREFG-XX:~/Work/10.09.2026$ ./script.sh 
+ exit_handlers+=cleanup_loop_devices
+ for h in "${exit_handlers[@]}"
+ cleanup_loop_devices
+ local i dir
+ for i in 1 2
+ dir=/tmp/haha
+ mountpoint -q /tmp/haha
+ fatal 'the directory /tmp/haha is not a mountpoint, or device is not a block device'
+ echo 'FATAL: the directory /tmp/haha is not a mountpoint, or device is not a block device'
FATAL: the directory /tmp/haha is not a mountpoint, or device is not a block device
+ exit 1
```

</p>
</details>

concourse/scripts/ic_gpdb_resgroup_v2.bash
ShapatinAA: If we're doing cleanup it means we want to bring system in some state for future usage and if it is not the case, then maybe cleanup isn't needed at all?

concourse/scripts/ic_gpdb_resgroup_v2.bash
riogor: > Why always true is suitable?

We'd like to do all cleanups, even if some fail.

And yeah, `set -e` won't exit if `$h` fails. In || list only last command would trigger an exit ([docs](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#The-Set-Builtin-1)).

KnightMurloc: With `fatal` yes. Because `exit` is exiting from script completely.

concourse/scripts/ic_gpdb_resgroup_v2.bash
KnightMurloc: If one cleanup step is failed for some reason other may still be successful. In our case, if `dir` for some reason is not a mount point it's OK to just skip it.

riogor: Got carried away.
Yeah, that doesn't make much sense, i'll revert it.

ShapatinAA (dismissed):

concourse/scripts/ic_gpdb_resgroup_v2.bash
ShapatinAA: > In || list only last command would trigger

The very same thing I was referring to previously... Yep. Was totally blinded by `fatal` exit in upper example.

KnightMurloc (dismissed):

riogor:
I'll remove CI since the PR was approved. @KnightMurloc @ShapatinAA can you re-approve it?

KnightMurloc (approved):

riogor:
@andreyaksenov any suggestions on PR title?

andreyaksenov:
> @andreyaksenov any suggestions on PR title?

the current title is good

ShapatinAA (approved):

PR #592, opened 2026-09-10
Backport local (coordinator-only) extension improvements to 6.x

whitehawk:
Backport local (coordinator-only) extension improvements to 6.x

This patch backports some improvements from commit
3f3846f7db7aa511a987b6f57c002341cf295e0d

List of changes:
 - gpexpand's new-segment cleanup now truncates all extension tables in one
statement instead of one DELETE per table, and skips a database it can't
connect to instead of aborting;
 - changed the order of checks in transformDistributedBy() to keep
local-extension working in both dispatch and utility mode;
 - added test local extension and regression test.

### DIFF
--- a/contrib/Makefile
+++ b/contrib/Makefile
@@ -66,6 +66,7 @@ SUBDIRS += \
 		formatter_fixedwidth \
 		extprotocol \
 		indexscan \
+		gp_local_test
 
 ifeq ($(with_openssl),yes)
 SUBDIRS += sslinfo
--- a/contrib/gp_local_test/gp_local_test--1.0--1.1.sql
+++ b/contrib/gp_local_test/gp_local_test--1.0--1.1.sql
@@ -1,0 +1,41 @@
+-- gp_local_test--1.0--1.1.sql
+\echo Use "ALTER EXTENSION gp_local_test UPDATE TO '1.1'" to load this file. \quit
+
+ALTER TABLE @extschema@.state ADD COLUMN last_reset timestamptz;
+
+CREATE TABLE @extschema@.history (
+    id         serial PRIMARY KEY,
+    code       text NOT NULL,
+    old_value  integer NOT NULL,
+    new_value  integer NOT NULL,
+    changed_at timestamptz UNIQUE NOT NULL DEFAULT now()
+);
+
+CREATE OR REPLACE FUNCTION @extschema@.bump_counter(p_code text) RETURNS integer
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    v_old integer;
+    v_new integer;
+BEGIN
+    UPDATE @extschema@.state
+       SET value = value + 1,
+           updated_at = now()
+     WHERE code = p_code
+    RETURNING value - 1, value INTO v_old, v_new;
+
+    INSERT INTO @extschema@.history (code, old_value, new_value)
+    VALUES (p_code, v_old, v_new);
+
+    RETURN v_new;
+END;
+$$;
+
+CREATE FUNCTION @extschema@.reset_counter(p_code text) RETURNS void
+LANGUAGE sql
+AS $$
+    UPDATE @extschema@.state
+       SET value = 0,
+           last_reset = now()
+     WHERE code = p_code;
+$$;
--- a/contrib/gp_local_test/gp_local_test--1.1.sql
+++ b/contrib/gp_local_test/gp_local_test--1.1.sql
@@ -1,0 +1,78 @@
+-- gp_local_test--1.1.sql
+\echo Use "CREATE EXTENSION gp_local_test" to load this file. \quit
+
+-- Deliberately vanilla: no DISTRIBUTED BY, no Greengage-specific clauses.
+-- id is PRIMARY KEY *and* code is independently UNIQUE — two
+-- non-overlapping unique constraints, which fails under Greengage's
+-- default hash-distributed policy but is perfectly ordinary Postgres.
+CREATE TABLE @extschema@.state (
+    id         serial PRIMARY KEY,
+    code       text UNIQUE NOT NULL,
+    value      integer NOT NULL DEFAULT 0,
+    updated_at timestamptz NOT NULL DEFAULT now(),
+    last_reset timestamptz
+);
+
+INSERT INTO @extschema@.state (code, value) VALUES ('counter', 0);
+
+-- Genuine "config" table: DBA-tunable settings that must survive
+-- backup/restore even though the rest of the extension's schema is
+-- recreated fresh by re-running this script. Marked via
+-- pg_extension_config_dump() per the standard PostgreSQL convention
+-- so pg_dump/gpbackup include its row data instead of treating it as
+-- pure extension-owned reference data.
+CREATE TABLE @extschema@.settings (
+    key   text PRIMARY KEY,
+    value text NOT NULL
+);
+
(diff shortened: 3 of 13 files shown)


contrib/Makefile
```diff
@@ -66,6 +66,7 @@ SUBDIRS += \
 		formatter_fixedwidth \
 		extprotocol \
 		indexscan \
+		gp_local_test
```
RekGRpth, line 69: why not in `gpcontrib`?

whitehawk: Because it is a sample of `native` postgres extension, so I considered `contrib` would be a more suitable place. And it is the place in 7x where it already resides.

contrib/gp_local_test/gp_local_test--1.0--1.1.sql
```diff
@@ -0,0 +1,41 @@
+-- gp_local_test--1.0--1.1.sql
```
RekGRpth, line 1: ### `contrib/gp_local_test/gp_local_test--1.0--1.1.sql:4` — upgrade path and fresh install diverge

The `1.0 → 1.1` upgrade script never inserts the `('bump_interval_seconds', '5')` row into the `settings` table that a fresh `CREATE EXTENSION ... VERSION '1.1'` install gets from `gp_local_test--1.1.sql`.

**Verified live:**
```sql
CREATE EXTENSION gp_local_test VERSION '1.0';
ALTER EXTENSION gp_local_test UPDATE TO '1.1';
SELECT * FROM ext_local_test.settings;
-- (0 rows)

DROP EXTENSION gp_local_test;
CREATE EXTENSION gp_local_test VERSION '1.1';
SELECT * FROM ext_local_test.settings;
-- bump_interval_seconds | 5
```

Same nominal end version ('1.1'), different data depending on install path — an upgrade/fresh-install parity bug in the sample extension used to validate the whole local-extension feature.

contrib/gp_local_test/gp_local_test--1.1.sql
```diff
@@ -0,0 +1,78 @@
...
+INSERT INTO @extschema@.settings (key, value) VALUES ('bump_interval_seconds', '5');
+
+SELECT pg_catalog.pg_extension_config_dump('@extschema@.settings', '');
+
+CREATE TABLE @extschema@.history (
+    id         serial PRIMARY KEY,
+    code       text NOT NULL,
+    old_value  integer NOT NULL,
+    new_value  integer NOT NULL,
+    changed_at timestamptz UNIQUE NOT NULL DEFAULT now()
```
RekGRpth, line 38: ### `contrib/gp_local_test/gp_local_test--1.1.sql:38` — unique violation on `changed_at`

`history.changed_at` is declared `timestamptz UNIQUE NOT NULL DEFAULT now()`, but `now()` returns the same value for an entire transaction, so two `bump_counter()` calls in one transaction collide on the unique constraint.

**Verified live:**
```sql
BEGIN;
SELECT ext_local_test.bump_counter('counter');  -- 1
SELECT ext_local_test.bump_counter('counter');
-- ERROR:  duplicate key value violates unique constraint "history_changed_at_key"
-- DETAIL:  Key (changed_at)=(2026-09-14 09:34:59.396612+05) already exists.
COMMIT;
SELECT * FROM ext_local_test.history;
-- (0 rows)
```

Worse than initially assessed: the whole transaction aborts, so even the *first* `bump_counter()` call's effect is lost, not just the second.

gpMgmt/bin/gpexpand
```diff
@@ -1491,9 +1492,22 @@ class gpexpand(object):
...
-                cursor = dbconn.execSQL(conn, extTablesSql)
-                db_statements_by_dbname[database[0]] = statements + ["delete from %s" % row[0] for row in cursor]
+            try:
+                with closing(dbconn.connect(url, encoding='UTF8')) as conn:
+                    ext_tables = [row[0] for row in dbconn.execSQL(conn, extTablesSql)]
+            except Exception as e:
+                self.logger.warning(
+                    "Could not connect to database %s to look up extension-owned "
+                    "tables for new-segment cleanup, skipping: %s" % (database[0], str(e)))
+                ext_tables = []
```
RekGRpth, line 1502: ### `gpMgmt/bin/gpexpand:1495-1502` — connection failure and query failure are conflated

```python
try:
    with closing(dbconn.connect(url, encoding='UTF8')) as conn:
        ext_tables = [row[0] for row in dbconn.execSQL(conn, extTablesSql)]
except Exception as e:
    self.logger.warning(
        "Could not connect to database %s to look up extension-owned "
        "tables for new-segment cleanup, skipping: %s" % (database[0], str(e)))
    ext_tables = []
```

The `try` block wraps both `dbconn.connect(...)` and the subsequent `dbconn.execSQL(...)`. Any exception from either — not just a connection failure — is logged as "Could not connect to database ... skipping", and `ext_tables` is silently set to `[]`.

**Verified by reproduction** (same connect/execSQL/except pattern, against a live database, with a deliberately broken query in place of `extTablesSql`):
```
WARNING: Could not connect to database gplocaltest_verify to look up extension-owned
tables for new-segment cleanup, skipping: ERROR: relation "pg_class_typo_does_not_exist"
does not exist
ext_tables: []
```
The connection succeeded; only the query failed. The log message misattributes the failure, and the extension-owned tables on that new segment are silently never truncated — with no diagnostic pointing at the real cause (e.g. a permissions problem on `pg_depend`/`pg_class`, a transient timeout, or a future bug in `extTablesSql`).

contrib/Makefile
RekGRpth: okay

contrib/gp_local_test/gp_local_test--1.1.sql
whitehawk: The sole purpose of the extension is to be used in the tests for the local extensions feature. This scenario is not a target scenario for the tests. Therefore, I think that we should not try to fix all possible cases here, unless they affect the testing of the local extensions directly.

contrib/gp_local_test/gp_local_test--1.0--1.1.sql
whitehawk: The sole purpose of the extension is to be used in the tests for the local extensions feature. This scenario is not a target scenario for the tests. Therefore, I think that we should not try to fix all possible cases here, unless they affect the testing of the local extensions directly.

gpMgmt/bin/gpexpand
whitehawk: > The log message misattributes the failure

We have in the warning the exact log of the occured exception, which can be used for the diagnostic purposes. I think it is sufficient enough. And, disregarding the exact reason the exception occured (no connection, wrong query, smth else), if the error occured - we can't do truncation. Therefore, I think we do not need here any additional handling.

RekGRpth (approved):

whitehawk:
@andreyaksenov , please verify the PR description.

andreyaksenov:
> @andreyaksenov , please verify the PR description.

Suggestion:

```
Backport local (coordinator-only) extension improvements to 6.x
```

whitehawk:
> > @andreyaksenov , please verify the PR description.
> 
> Suggestion:
> 
> ```
> Backport local (coordinator-only) extension improvements to 6.x
> ```

Ok, updated.

red1452 (approved):

PR #593, opened 2026-09-11
Improve LASJ cardinality estimation for multi-predicate joins

whitehawk:
Improve LASJ cardinality estimation for multi-predicate joins

Problem description:
Orca did wrong estimates for LASJ (Left Anti-Semi Join) cardinality for
multi-predicate join condition. For example, for a query like:

t_left(x, y): x = y = i, for i in 1..100;
t_right(x2,y2): x2 = i, y2 = i+(i%2), for i in 1..100;

SELECT * FROM t_left LEFT JOIN t_right
ON t_left.x=t_right.x2 AND t_left.y=t_right.y2
WHERE t_right.x2 IS NULL;

where 1st predicate "t_left.x=t_right.x2" has 100% match, and 2nd predicate
"t_left.y=t_right.y2" has only 50% rows match, meaning that the result rows with
"t_right.x2 IS NULL" should be 50% of the total 't_left' rows, Orca estimated
output rows as 1.

Root cause:
The problem is actually a complex one, combined from several issues:
1. OR-logic when calculating final emptiness of the result in
'CJoinStatsProcessor::SetResultingJoinStats()'. It is suitable for usual joins,
but LASJ requires AND-logic: the output of LASJ is empty only if results of all
predicates are empty.
2. Scale factor for join cardinality is calculated wrongly for LASJ in
'CJoinStatsProcessor::CalcJoinCardinality()'. Cumulative join scale factor
was applied. While actually for LASJ, if a row fails to match on ANY AND-ed
single predicate of the original LEFT JOIN, it will survive in LASJ.
3. Error in histogram calculation in 'MakeLASJHistogram()', where we advance
'idx2' index of the other histogram always with no condition, while if the
'upper_split_bucket' was null, we need to re-examine the same bucket of the
other histogram against next bucket of 'this' histogram.

Fix:
1. Change to AND-logic when calculating final emptiness of the result in
'CJoinStatsProcessor::SetResultingJoinStats()' for LASJ.
2. Add logic of scale factor calculation for LASJ into
'CJoinStatsProcessor::CalcJoinCardinality()'. For LASJ the final scale factor
is bounded by the least restrictive predicate's own scale factor - i.e. the
minimum scale factor of the AND-ed predicates.
3. Update logic in 'MakeLASJHistogram()' - advance 'idx2' index only when we've
fully processed it.
4. Fixing 1-3 revealed another separate issue: in a deeply nested join-order
search, a LASJ histogram merge could underflow GetFrequency() to a subnormal
value, causing 1.0/GetFrequency() to blow up the normalization scale factor
and produce a structurally invalid histogram (resulting in
'GPOS_ASSERT(IsValid())' failure in Orca's unit tests). Fix by capping the
scale factor rather than the input frequency (so as not to affect ordinary
small-but-real selectivities), and emptying the histogram in that case so
that callers relying on IsEmpty() to detect a degenerate join (e.g.
'MakeJoinHistogramNormalize', 'JoinHistogramsLASJ') still get their intended
fallback instead of a stale, un-normalized histogram.

Known limitation: AND-of-marginals is necessary, but not sufficient 

The AND-combination fix (root cause 1) checks whether every AND-ed predicate 
individually shows the inner side fully covering the outer side, and concludes 
the LASJ is empty only when all of them do. This is a real improvement over the 
old OR-based logic — it can only misfire on a strict subset of what used to 
misfire — but it is not a complete fix for multi-predicate LASJ: each 
predicate's coverage is judged from its own per-column histogram, with no 
visibility into whether the same inner row satisfies every predicate at once. 
Two columns can each look fully covered in isolation without any single inner 
row covering both simultaneously, e.g.: 

t_left(x, y): x = y = i, for i in 1..100;
t_right2(x2,y2): x2 = i, y2 = 101-i, for i in 1..100;

t_left LEFT JOIN t_right2 ON x=x2 AND y=y2 WHERE x2 IS NULL;

Here x and x2 fully overlap, and y and y2 fully overlap, so both predicates 
independently look empty and the AND collapses the estimate to ~1 row — but no 
t_right2 row ever satisfies x2=i AND y2=i at once (that needs 101-i=i), so the 
true result is all 100 rows.
A sound general fix for it would require a joint (multi-column) statistic over 
the predicate columns rather than per-column marginals — a larger, separate 
change to Orca's statistics model, not something this patch attempts. We're 
documenting this as a known, accepted residual gap.

Test Changes:
1. Add LASJMultiplePredicates.mdp under CAntiSemiJoinTest, capturing the
original reproduction sequence.
2. A bunch of minidump files are updated to reflect corrected cardinality/cost
estimates following the LASJ fix.
3. workfile_gp_toolkit isolation2 test: add 'analyze workfile_test;' right
after the table is populated. Root cause: without real statistics, GPORCA's
plan choice for the test's un-analyzed cross join was already non-deterministic
before this fix; the LASJ fix incidentally made a previously-random cost tie
resolve consistently to one specific plan which lacked a Materialize node.
ANALYZE gives the optimizer real cardinalities, stabilizing the plan and making
the intended spill-to-workfile behavior (which is what this test actually
verifies) independent of that fragility.
4. bfv_partition_plans_optimizer.out and bfv_planner_optimizer.out: update
answer file as the plans for some of the queries have changed. New plans are not
worse than previous ones. Some more details about bfv_planner_optimizer.out
are below:

Incidental fix: empty-relation cross joins 

Splitting the LASJ-specific AND check from the OR check used by other join types 
(in "CJoinStatsProcessor::SetResultingJoinStats()") moved the "outer side is 
empty -> join is empty" test out of the per-predicate loop into a single check 
applied after the loop. That check now runs even when a join has zero predicates 
(a plain CROSS JOIN), which it previously did not — the old code only 
evaluated it once per predicate, so it was silently skipped whenever the 
predicate list was empty. 

This surfaces in bfv_planner_optimizer.out: "t_hashdist cross join (select a, 
sum(random()) from generate_series(1, 10) a group by a) x", where t_hashdist is 
created empty and never analyzed. Before this fix, GPORCA never recognized the 
empty outer relation and estimated the cross join's cardinality via the default 
cartesian-product formula (1000 rows); after, it correctly collapses to 
"CStatistics::MinRows". That estimate change is enough to flip which side of the 
join gets broadcast versus gathered, producing the plan diff in the answer file. 
This has no connection to LASJ semantics itself — it is a side effect of the 
shared statistics-derivation code both paths go through, and it makes the 
affected estimate more accurate, not less.

### DIFF
--- a/src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
+++ b/src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
@@ -314,6 +314,15 @@ CJoinStatsProcessor::SetResultingJoinStats(
 	const ULONG num_join_conds = join_pred_stats_info->Size();
 
 	BOOL output_is_empty = false;
+	// For a LASJ, the anti-join produces no rows only if EVERY AND-ed
+	// predicate independently shows complete coverage (no outer value
+	// survives unmatched on that column) - a single predicate with
+	// surviving buckets is enough on its own to keep the anti-join
+	// non-empty, so predicates must be combined with AND here. This is the
+	// opposite of the OR used below for other join types, where a single
+	// AND-ed predicate with no possible matches is enough to make the whole
+	// (intersection-based) join empty.
+	BOOL lasj_all_preds_fully_covered = (num_join_conds > 0);
 	CDouble num_join_rows = 0;
 	// iterate over join's predicate(s)
 	for (ULONG i = 0; i < num_join_conds; i++)
@@ -366,9 +375,17 @@ CJoinStatsProcessor::SetResultingJoinStats(
 					   DoIgnoreLASJHistComputation);
 
 
-		output_is_empty = JoinStatsAreEmpty(
-			outer_stats->IsEmpty(), output_is_empty, outer_histogram,
-			inner_histogram, outer_histogram_after, join_type);
+		BOOL pred_is_empty = JoinStatsAreEmpty(outer_histogram, inner_histogram,
+											   outer_histogram_after);
+		if (IsLASJ)
+		{
+			lasj_all_preds_fully_covered =
+				lasj_all_preds_fully_covered && pred_is_empty;
+		}
+		else
+		{
+			output_is_empty = output_is_empty || pred_is_empty;
+		}
 
 		CStatisticsUtils::AddHistogram(mp, colid1, outer_histogram_after,
 									   result_col_hist_mapping);
@@ -420,6 +437,13 @@ CJoinStatsProcessor::SetResultingJoinStats(
 				local_scale_factor, mdid_pair, both_dist_keys));
 	}
 
+
+	// finalize output_is_empty: for LASJ, empty only if every AND-ed
+	// predicate independently showed complete coverage (AND across
+	// predicates); for other join types, keep the OR-accumulated
+	// per-predicate result plus the (unchanged) outer-empty rule.
+	output_is_empty = IsLASJ ? lasj_all_preds_fully_covered
+							 : (output_is_empty || (outer_stats->IsEmpty()));
 
 	num_join_rows = CStatistics::MinRows;
 	if (!output_is_empty)
@@ -477,17 +501,48 @@ CJoinStatsProcessor::CalcJoinCardinality(
 	CDouble limit_for_result_scale_factor(
 		std::max(left_num_rows.Get(), right_num_rows.Get()));
 
-	CDouble scale_factor = CScaleFactorUtils::CumulativeJoinScaleFactor(
-		mp, stats_config, join_conds_scale_factors,
-		limit_for_result_scale_factor);
+	const ULONG num_join_conds = join_conds_scale_factors->Size();
+
+	BOOL IsLASJ = (IStatistics::EsjtLeftAntiSemiJoin == join_type);
+
+	CDouble scale_factor(1.0);
+	if (IsLASJ && 0 < num_join_conds)
+	{
+		// For a LASJ, a row survives the AND-ed anti-join condition if it
+		// fails to match on ANY single predicate (survives(x) iff there is
+		// no y satisfying every predicate at once). A row that survives
+		// even one predicate alone therefore also survives the combined
+		// condition, so:
+		//     survivors(combined) >= max(survivors(P1), survivors(P2), ...)
+		// Since scale_factor = left_num_rows / survivors, that means the
+		// combined scale factor is bounded by the LEAST restrictive
+		// predicate's own scale factor - i.e. the minimum, not the
+		// damped product CumulativeJoinScaleFactor computes for
+		// match-fraction-style combination (inner/semi join). Additional
+		// AND-ed predicates can only grow the LASJ survivor set relative
+		// to the most permissive one, never shrink it below what that
+		// predicate alone already guarantees.
+		scale_factor = (*join_conds_scale_factors)[0]->m_scale_factor;
+		for (ULONG ul = 1; ul < num_join_conds; ul++)
+		{
+			scale_factor =
+				std::min(scale_factor.Get(),
+						 (*join_conds_scale_factors)[ul]->m_scale_factor.Get());
+		}
+	}
+	else
+	{
+		scale_factor = CScaleFactorUtils::CumulativeJoinScaleFactor(
+			mp, stats_config, join_conds_scale_factors,
+			limit_for_result_scale_factor);
+	}
 	CDouble cartesian_product_num_rows = left_num_rows * right_num_rows;
 
-	if (IStatistics::EsjtLeftAntiSemiJoin == join_type ||
-		IStatistics::EsjtLeftSemiJoin == join_type)
+	if (IsLASJ || IStatistics::EsjtLeftSemiJoin == join_type)
 	{
 		CDouble rows = left_num_rows;
 
-		if (IStatistics::EsjtLeftAntiSemiJoin == join_type)
+		if (IsLASJ)
 		{
 			rows = left_num_rows / scale_factor;
 		}
@@ -509,23 +564,25 @@ CJoinStatsProcessor::CalcJoinCardinality(
 
 
 
-// check if the join statistics object is empty output based on the input
-// histograms and the join histograms
+// check whether a single join predicate's histograms show that this
+// predicate alone cannot produce any matches. Note this says nothing on its
+// own about whether the overall (possibly multi-predicate) join is empty -
+// see SetResultingJoinStats for how this is combined across predicates.
 BOOL
-CJoinStatsProcessor::JoinStatsAreEmpty(BOOL outer_is_empty,
-									   BOOL output_is_empty,
-									   const CHistogram *outer_histogram,
+CJoinStatsProcessor::JoinStatsAreEmpty(const CHistogram *outer_histogram,
 									   const CHistogram *inner_histogram,
-									   CHistogram *join_histogram,
-									   IStatistics::EStatsJoinType join_type)
+									   CHistogram *join_histogram)
 {
 	GPOS_ASSERT(nullptr != outer_histogram);
 	GPOS_ASSERT(nullptr != inner_histogram);
 	GPOS_ASSERT(nullptr != join_histogram);
-	BOOL IsLASJ = IStatistics::EsjtLeftAntiSemiJoin == join_type;
-	return output_is_empty || (!IsLASJ && outer_is_empty) ||
-		   (!outer_histogram->IsEmpty() && !inner_histogram->IsEmpty() &&
-			join_histogram->IsEmpty());
+
+	BOOL pred_is_empty = !outer_histogram->IsEmpty() &&
+						 !inner_histogram->IsEmpty() &&
+						 join_histogram->IsEmpty();
+
+
+	return pred_is_empty;
 }
 
 // Derive statistics for join operation given array of statistics object
--- a/src/backend/gporca/data/dxl/minidump/Blocking-Spool-Parallel-Union-All.mdp
+++ b/src/backend/gporca/data/dxl/minidump/Blocking-Spool-Parallel-Union-All.mdp
@@ -742,7 +742,7 @@
     <dxl:Plan Id="0" SpaceSize="8">
       <dxl:GatherMotion InputSegments="0,1,2" OutputSegments="-1">
         <dxl:Properties>
-          <dxl:Cost StartupCost="0" TotalCost="1356744856.482021" Rows="3002438.000000" Width="12"/>
+          <dxl:Cost StartupCost="0" TotalCost="1356694004.617104" Rows="1.000000" Width="12"/>
         </dxl:Properties>
         <dxl:ProjList>
           <dxl:ProjElem ColId="0" Alias="a">
@@ -759,7 +759,7 @@
         <dxl:SortingColumnList/>
         <dxl:NestedLoopJoin JoinType="Inner" IndexNestedLoopJoin="false" OuterRefAsParam="false">
           <dxl:Properties>
-            <dxl:Cost StartupCost="0" TotalCost="1356744722.212994" Rows="3002438.000000" Width="12"/>
+            <dxl:Cost StartupCost="0" TotalCost="1356694004.617059" Rows="1.000000" Width="12"/>
           </dxl:Properties>
           <dxl:ProjList>
             <dxl:ProjElem ColId="0" Alias="a">
@@ -820,7 +820,7 @@
           </dxl:BroadcastMotion>
           <dxl:Append IsTarget="false" IsZapped="false">
             <dxl:Properties>
-              <dxl:Cost StartupCost="0" TotalCost="1324032.335660" Rows="3002438.000000" Width="4"/>
+              <dxl:Cost StartupCost="0" TotalCost="1324032.286926" Rows="3002429.000000" Width="4"/>
             </dxl:Properties>
             <dxl:ProjList>
               <dxl:ProjElem ColId="9" Alias="a">
@@ -830,7 +830,7 @@
             <dxl:Filter/>
             <dxl:Materialize Eager="true">
               <dxl:Properties>
-                <dxl:Cost StartupCost="0" TotalCost="1324032.335660" Rows="11.000000" Width="4"/>
+                <dxl:Cost StartupCost="0" TotalCost="1324032.286926" Rows="2.000000" Width="4"/>
               </dxl:Properties>
               <dxl:ProjList>
                 <dxl:ProjElem ColId="9" Alias="a">
@@ -840,7 +840,7 @@
               <dxl:Filter/>
               <dxl:RedistributeMotion InputSegments="0,1,2" OutputSegments="0,1,2">
                 <dxl:Properties>
-                  <dxl:Cost StartupCost="0" TotalCost="1324032.335645" Rows="11.000000" Width="4"/>
+                  <dxl:Cost StartupCost="0" TotalCost="1324032.286923" Rows="2.000000" Width="4"/>
                 </dxl:Properties>
                 <dxl:ProjList>
                   <dxl:ProjElem ColId="9" Alias="a">
@@ -856,7 +856,7 @@
                 </dxl:HashExprList>
                 <dxl:NestedLoopJoin JoinType="Left" IndexNestedLoopJoin="false" OuterRefAsParam="false">
                   <dxl:Properties>
-                    <dxl:Cost StartupCost="0" TotalCost="1324032.335645" Rows="11.000000" Width="4"/>
+                    <dxl:Cost StartupCost="0" TotalCost="1324032.286923" Rows="2.000000" Width="4"/>
                   </dxl:Properties>
                   <dxl:ProjList>
                     <dxl:ProjElem ColId="9" Alias="a">
--- a/src/backend/gporca/data/dxl/minidump/CompositeTypeRelMdId.mdp
+++ b/src/backend/gporca/data/dxl/minidump/CompositeTypeRelMdId.mdp
@@ -324,7 +324,7 @@
     <dxl:Plan Id="0" SpaceSize="4">
       <dxl:Sort SortDiscardDuplicates="false">
         <dxl:Properties>
-          <dxl:Cost StartupCost="0" TotalCost="882758.081967" Rows="1.000000" Width="8"/>
+          <dxl:Cost StartupCost="0" TotalCost="882730.298799" Rows="1.000000" Width="8"/>
         </dxl:Properties>
         <dxl:ProjList>
           <dxl:ProjElem ColId="0" Alias="rngfuncid">
@@ -343,7 +343,7 @@
         <dxl:LimitOffset/>
         <dxl:Result>
           <dxl:Properties>
-            <dxl:Cost StartupCost="0" TotalCost="882758.081876" Rows="1.000000" Width="8"/>
+            <dxl:Cost StartupCost="0" TotalCost="882730.298708" Rows="1.000000" Width="8"/>
           </dxl:Properties>
           <dxl:ProjList>
             <dxl:ProjElem ColId="0" Alias="rngfuncid">
@@ -366,7 +366,7 @@
               </dxl:ParamList>
               <dxl:Result>
                 <dxl:Properties>
-                  <dxl:Cost StartupCost="0" TotalCost="0.040900" Rows="400.000000" Width="4"/>
+                  <dxl:Cost StartupCost="0" TotalCost="0.040900" Rows="1.000000" Width="4"/>
                 </dxl:Properties>
                 <dxl:ProjList>
                   <dxl:ProjElem ColId="10" Alias="f2">
--- a/src/backend/gporca/data/dxl/minidump/DPv2GreedyOnly.mdp
+++ b/src/backend/gporca/data/dxl/minidump/DPv2GreedyOnly.mdp
@@ -5885,7 +5885,7 @@
     <dxl:Plan Id="0" SpaceSize="18">
       <dxl:HashJoin JoinType="Left">
         <dxl:Properties>
-          <dxl:Cost StartupCost="0" TotalCost="228773.635646" Rows="2010720827.932932" Width="32"/>
+          <dxl:Cost StartupCost="0" TotalCost="115829.503451" Rows="1005039812.733285" Width="32"/>
         </dxl:Properties>
         <dxl:ProjList>
           <dxl:ProjElem ColId="0" Alias="a1">
@@ -5923,7 +5923,7 @@
         </dxl:HashCondList>
         <dxl:GatherMotion InputSegments="0,1,2" OutputSegments="-1">
           <dxl:Properties>
-            <dxl:Cost StartupCost="0" TotalCost="2339.169012" Rows="2013597.069984" Width="24"/>
+            <dxl:Cost StartupCost="0" TotalCost="2220.960112" Rows="1007049.894372" Width="24"/>
           </dxl:Properties>
           <dxl:ProjList>
             <dxl:ProjElem ColId="0" Alias="a1">
@@ -5949,7 +5949,7 @@
           <dxl:SortingColumnList/>
           <dxl:HashJoin JoinType="Left">
             <dxl:Properties>
-              <dxl:Cost StartupCost="0" TotalCost="2159.072890" Rows="2013597.069984" Width="24"/>
+              <dxl:Cost StartupCost="0" TotalCost="2130.889570" Rows="1007049.894372" Width="24"/>
             </dxl:Properties>
             <dxl:ProjList>
               <dxl:ProjElem ColId="0" Alias="a1">
--- a/src/backend/gporca/data/dxl/minidump/DPv2MinCardOnly.mdp
+++ b/src/backend/gporca/data/dxl/minidump/DPv2MinCardOnly.mdp
@@ -5885,7 +5885,7 @@
     <dxl:Plan Id="0" SpaceSize="18">
       <dxl:HashJoin JoinType="Left">
         <dxl:Properties>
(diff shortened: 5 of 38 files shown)


src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -509,23 +566,25 @@ CJoinStatsProcessor::CalcJoinCardinality(
...
-// histograms and the join histograms
+// check whether a single join predicate's histograms show that this
+// predicate alone cannot produce any matches. Note this says nothing on its
+// own about whether the overall (possibly multi-predicate) join is empty -
+// see SetResultingJoinStats for how this is combined across predicates.
 BOOL
-CJoinStatsProcessor::JoinStatsAreEmpty(BOOL outer_is_empty,
-									   BOOL output_is_empty,
-									   const CHistogram *outer_histogram,
+CJoinStatsProcessor::JoinStatsAreEmpty(const CHistogram *outer_histogram,
```
RekGRpth, line 574: This function-level comment describes only the non-LASJ interpretation and does not warn that, for LASJ, the meaning of the returned flag is inverted with respect to "matches".

src/test/regress/expected/bfv_planner_optimizer.out
```diff
@@ -511,19 +511,17 @@ explain (costs off) select * from t_hashdist cross join (select random () from g
...
                ->  Seq Scan on t_hashdist
-         ->  Materialize
-               ->  Redistribute Motion 1:3  (slice2)
-                     ->  HashAggregate
-                           Group Key: generate_series.generate_series
-                           ->  Function Scan on generate_series
- Optimizer: Pivotal Optimizer (GPORCA)
-(14 rows)
+ Optimizer: GPORCA
+(9 rows)
```
RekGRpth, line 524: Where is LASJ here? Why was plan changed?

src/test/regress/expected/bfv_partition_plans_optimizer.out
```diff
@@ -554,7 +554,7 @@ INFO:  GPORCA failed to produce a plan, falling back to Postgres-based planner
 DETAIL:  Falling back to Postgres-based planner because GPORCA does not support the following feature: SIRV functions
  count_operator 
 ----------------
-              0
+              2
```
RekGRpth, line 557: Describe this change, please.

RekGRpth:
Could you please add tests for:

- `CHistogram::MakeLASJHistogram` using histograms specifically constructed to trigger the old `advance idx2` bug (requires a case with multiple buckets where `upper_split_bucket` is sometimes present and sometimes absent);
- `NormalizeHistogram()` with subnormal/near-zero frequency (at the 1e15 boundary).

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
whitehawk: The function level comments have a disclaimer: "Note this says nothing on its own about whether the overall (possibly multi-predicate) join is empty - see SetResultingJoinStats for how this is combined across predicates.". I think it is enough, as the LASJ-related logic is in the SetResultingJoinStats(), and we reference to it.

src/test/regress/expected/bfv_planner_optimizer.out
whitehawk: The plan on original code:
```
postgres=# drop table if exists t_hashdist;
create table t_hashdist(a int, b int, c int) distributed by (a);
explain analyze select * from t_hashdist cross join (select a, sum(random()) from generate_series(1, 10) a group by a) x;
DROP TABLE
CREATE TABLE
                                                                   QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..882898.13 rows=1000 width=24) (actual time=6.116..6.130 rows=0 loops=1)
   ->  Nested Loop  (cost=0.00..882898.04 rows=334 width=24) (actual time=0.000..5.650 rows=0 loops=1)
         Join Filter: true
         ->  Broadcast Motion 3:3  (slice3; segments: 3)  (cost=0.00..431.00 rows=1 width=12) (actual time=0.000..0.051 rows=0 loops=1)
               ->  Seq Scan on t_hashdist  (cost=0.00..431.00 rows=1 width=12) (actual time=0.000..0.058 rows=0 loops=1)
         ->  Materialize  (cost=0.00..0.17 rows=334 width=12) (actual time=4.985..4.985 rows=1 loops=1)
               ->  Redistribute Motion 1:3  (slice2)  (cost=0.00..0.16 rows=334 width=12) (actual time=4.967..4.975 rows=4 loops=1)
                     ->  HashAggregate  (cost=0.00..0.13 rows=1000 width=12) (actual time=0.324..0.407 rows=10 loops=1)
                           Group Key: generate_series.generate_series
                           Extra Text: hash table(s): 1; chain length 1.0 avg, 1 max; using 10 of 2048 buckets; total 0 expansions.

                           ->  Function Scan on generate_series  (cost=0.00..0.00 rows=1000 width=4) (actual time=0.076..0.080 rows=10 loops=1)
 Optimizer: GPORCA
 Planning Time: 12.651 ms
   (slice0)    Executor memory: 98K bytes.
   (slice1)    Executor memory: 36K bytes avg x 3 workers, 36K bytes max (seg0).  Work_mem: 17K bytes max.
 * (slice2)    Executor memory: 97K bytes (entry db).  Work_mem: 73K bytes max, 73K bytes wanted.
   (slice3)    Executor memory: 37K bytes avg x 3 workers, 37K bytes max (seg0).
 Memory used:  128000kB
 Memory wanted:  716kB
 Execution Time: 30.312 ms
(21 rows)

```

On updated code:
```
postgres=# drop table if exists t_hashdist;
create table t_hashdist(a int, b int, c int) distributed by (a);
explain analyze select * from t_hashdist cross join (select a, sum(random()) from generate_series(1, 10) a group by a) x;
DROP TABLE
CREATE TABLE
                                                             QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=0.00..882837.67 rows=1 width=24) (actual time=0.792..0.793 rows=0 loops=1)
   Join Filter: true
   ->  HashAggregate  (cost=0.00..0.13 rows=1000 width=12) (actual time=0.108..0.127 rows=10 loops=1)
         Group Key: generate_series.generate_series
         Extra Text: hash table(s): 1; chain length 1.0 avg, 1 max; using 10 of 2048 buckets; total 0 expansions.

         ->  Function Scan on generate_series  (cost=0.00..0.00 rows=1000 width=4) (actual time=0.016..0.020 rows=10 loops=1)
   ->  Materialize  (cost=0.00..431.00 rows=1 width=12) (actual time=0.060..0.060 rows=0 loops=11)
         ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=12) (actual time=0.649..0.650 rows=0 loops=1)
               ->  Seq Scan on t_hashdist  (cost=0.00..431.00 rows=1 width=12) (actual time=0.000..0.037 rows=0 loops=1)
 Optimizer: GPORCA
 Planning Time: 55.348 ms
 * (slice0)    Executor memory: 119K bytes.  Work_mem: 73K bytes max, 73K bytes wanted.
   (slice1)    Executor memory: 36K bytes avg x 3 workers, 36K bytes max (seg0).
 Memory used:  128000kB
 Memory wanted:  516kB
 Execution Time: 2.603 ms
(17 rows)

```

As it can be seen, in the previous plan the estimates for join results were:
```
Nested Loop  (cost=0.00..882898.04 rows=334 width=24) (actual time=0.000..5.650 rows=0 loops=1)
```
now they are
```
Nested Loop  (cost=0.00..882837.67 rows=1 width=24) (actual time=0.792..0.793 rows=0 loops=1)
```

They now are much more accurate.

The reason:
- `t_hashdist` is created empty and never analyzed in this test, so it has 0 actual rows.
- The cross join has `num_join_conds == 0`; there's no real predicate in join_pred_stats_info.
- Before the fix, the "outer side is empty -> join is empty" check lived inside the per-predicate loop:

```
output_is_empty = JoinStatsAreEmpty(outer_stats->IsEmpty(), output_is_empty, ...);
```

 Since the loop never run for 0 predicates, this check was silently skipped for cross joins — `output_is_empty` stayed false, and Orca fell back to 1000 rows estimate.
  
- After the fix, splitting LASJ's AND-semantics from the OR-semantics of other join types required pulling the outer-empty check out of the loop into one line that now runs
  unconditionally:

```
output_is_empty = IsLASJ ? lasj_all_preds_fully_covered
                         : (output_is_empty || outer_stats->IsEmpty());
```

  This now correctly fires even when num_join_conds == 0. Since `t_hashdist` really is empty, `output_is_empty` becomes true, and the join's row estimate collapses from 1000 down to
  CStatistics::MinRows (1) — the objectively more accurate estimate for a join against an empty table.

This more accurate cost estimate triggered different plan generation.

Though it is an incidental side effect, it is a correct improvement.

--- 
Updated PR description about this.

src/test/regress/expected/bfv_partition_plans_optimizer.out
whitehawk: This is a direct consequence of the LASJ scale-factor fix.

`p1 except all p2` compiles to a LeftAntiSemiJoin with three AND-ed predicates:
```
Hash Cond: ((NOT (p1.a IS DISTINCT FROM p2.a)) AND (NOT (p1.b IS DISTINCT FROM p2.b)) AND (NOT ((row_number() OVER (?)) IS DISTINCT FROM (row_number() OVER (?)))))
```

The update to scale-factor calculation changed the anti-join's row estimate from 9 to 1334 for this test's data.

That new estimate value changed the plan: with a ~9-row anti-join, redistributing that tiny intermediate result and hash-joining it with `t` directly is cheapest, and GPORCA skips Dynamic Partition Elimination. With ~1334 rows, broadcasting the much smaller `t` (100 rows) and pruning `p1`/`p2` via Partition Selectors becomes the cheaper plan — hence the two `Partition Selector` nodes `count_operator` now finds.

One can say the new estimate is worse than the old one, since the actual anti-join result for this test's data is 0 rows (`p1` and `p2` are populated identically), and 9 is numerically closer to 0 than 1334 is.

Sweeping the overlap between `p1`'s and `p2`'s key domains in 10% steps, at five table sizes from 1,000 to 500,000 rows, gives:

```
      N  ov%    true  old_est  new_est  old_err%  new_err%
   1000    0    1000        3      400      99.7      60.0
   1000   10     900        3      400      99.7      55.6
   1000   20     800        3      400      99.6      50.0
   1000   30     700        3      400      99.6      42.9
   1000   40     600        3      400      99.5      33.3
   1000   50     500        3      400      99.4      20.0
   1000   60     400        3      400      99.2       0.0
   1000   70     300        3      400      99.0      33.3
   1000   80     200        3      400      98.5     100.0
   1000   90     100        3      400      97.0     300.0
   1000  100       0        3      400      abs3    abs400
   5000    0    5000       13     2000      99.7      60.0
   5000   10    4500       13     2000      99.7      55.6
   5000   20    4000       13     2000      99.7      50.0
   5000   30    3500       13     2000      99.6      42.9
   5000   40    3000       13     2000      99.6      33.3
   5000   50    2500       13     2000      99.5      20.0
   5000   60    2000       13     2000      99.4       0.0
   5000   70    1500       13     2000      99.1      33.3
   5000   80    1000       13     2000      98.7     100.0
   5000   90     500       13     2000      97.4     300.0
   5000  100       0       13     2000     abs13   abs2000
  20000    0   20000       51     8000      99.7      60.0
  20000   10   18000       51     8000      99.7      55.6
  20000   20   16000       51     8000      99.7      50.0
  20000   30   14000       51     8000      99.6      42.9
  20000   40   12000       51     8000      99.6      33.3
  20000   50   10000       51     8000      99.5      20.0
  20000   60    8000       51     8000      99.4       0.0
  20000   70    6000       51     8000      99.2      33.3
  20000   80    4000       51     8000      98.7     100.0
  20000   90    2000       51     8000      97.5     300.0
  20000  100       0       51     8000     abs51   abs8000
 100000    0  100000      253    40000      99.7      60.0
 100000   10   90000      253    40000      99.7      55.6
 100000   20   80000      253    40000      99.7      50.0
 100000   30   70000      253    40000      99.6      42.9
 100000   40   60000      253    40000      99.6      33.3
 100000   50   50000      253    40000      99.5      20.0
 100000   60   40000      253    40000      99.4       0.0
 100000   70   30000      253    40000      99.2      33.3
 100000   80   20000      253    40000      98.7     100.0
 100000   90   10000      253    40000      97.5     300.0
 100000  100       0      253    40000    abs253  abs40000
 500000    0  500000     1265   200000      99.7      60.0
 500000   10  450000     1265   200000      99.7      55.6
 500000   20  400000     1265   200000      99.7      50.0
 500000   30  350000     1265   200000      99.6      42.9
 500000   40  300000     1265   200000      99.6      33.3
 500000   50  250000     1265   200000      99.5      20.0
 500000   60  200000     1265   200000      99.4       0.0
 500000   70  150000     1265   200000      99.2      33.3
 500000   80  100000     1265   200000      98.7     100.0
 500000   90   50000     1265   200000      97.5     300.0
 500000  100       0     1265   200000   abs1265 abs200000
```
(N = row count of `p1`/`p2`; ov% = fraction of `p1`'s key domain that also appears in `p2`; true = actual `except all` row count, confirmed via `count(*)`; estimates are GPORCA's own reported row counts; error% = relative error against true, undefined at ov%=100 since true=0.)

The error pattern is identical at every table size, from 1,000 to 500,000 rows — table size does not change the qualitative picture at all.

New's error is not uniformly small: it is a fixed estimate that happens to exactly match the true value around 60% overlap for this data family, and overshoots on the high-overlap side, reaching 300% error (a 4x overestimate) at 90% overlap. New is not uniformly conservative either — it swings from exact to overestimating depending on where the true value sits relative to its fixed point.

Old never gets close, anywhere on this grid. Even at its best (90% overlap) old is still 97-97.5% wrong, capturing roughly 1/40th of the true row count; at every overlap below that it is 99%+ wrong, i.e., functionally blind to the join's real size. New's worst case (a 4x overestimate) is a materially smaller mistake than old's typical case (a 100x-400x underestimate), and unlike old, new's error stays in the same rough order of magnitude as the truth across the whole range.

So the identical-data test case in this PR is the specific point on this grid where old's error happens to cancel toward the right answer by accident. Everywhere else — including patterns with real, common overlap — it underestimates by two orders of magnitude regardless of scale. That is an edge case and a trade-off we need to accept.

----
Regarding documentation in the PR description, we already have 
> bfv_partition_plans_optimizer.out and bfv_planner_optimizer.out: update
answer file as the plans for some of the queries have changed. New plans are not
worse than previous ones.

I don't think we need to dive deeper in the details here.

whitehawk:
> Could you please add tests for:
> 
> * `CHistogram::MakeLASJHistogram` using histograms specifically constructed to trigger the old `advance idx2` bug (requires a case with multiple buckets where `upper_split_bucket` is sometimes present and sometimes absent);

The new Orca unit test case should reproduce it, as without this `idx2` fix the estimate for the test query is wrong (100 rows):
```
                                                          QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.05 rows=100 width=16) (actual time=5.450..6.000 rows=50 loops=1)
   ->  Result  (cost=0.00..862.04 rows=34 width=16) (actual time=2.557..4.821 rows=20 loops=1)
         Filter: (r.id2 IS NULL)
         ->  Hash Left Join  (cost=0.00..862.04 rows=67 width=16) (actual time=2.552..4.811 rows=38 loops=1)
               Hash Cond: ((l.id1 = r.id2) AND (l.val1 = r.val2))
               Extra Text: (seg0)   Hash chain length 1.0 avg, 1 max, using 38 of 524288 buckets.
               ->  Seq Scan on t_left l  (cost=0.00..431.00 rows=34 width=8) (actual time=0.008..0.012 rows=38 loops=1)
               ->  Hash  (cost=431.00..431.00 rows=34 width=8) (actual time=0.104..0.105 rows=38 loops=1)
                     Buckets: 524288  Batches: 1  Memory Usage: 4098kB
                     ->  Seq Scan on t_right r  (cost=0.00..431.00 rows=34 width=8) (actual time=0.071..0.076 rows=38 loops=1)
 Optimizer: GPORCA
 Planning Time: 50.885 ms
   (slice0)    Executor memory: 43K bytes.
   (slice1)    Executor memory: 4159K bytes avg x 3 workers, 4159K bytes max (seg0).  Work_mem: 4098K bytes max.
 Memory used:  128000kB
 Execution Time: 7.119 ms
(16 rows)
```
with the fix it is correct (50 rows):
```
                                                          QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.04 rows=50 width=16) (actual time=7.428..8.326 rows=50 loops=1)
   ->  Result  (cost=0.00..862.04 rows=17 width=16) (actual time=2.623..6.202 rows=20 loops=1)
         Filter: (r.id2 IS NULL)
         ->  Hash Left Join  (cost=0.00..862.04 rows=50 width=16) (actual time=2.617..6.191 rows=38 loops=1)
               Hash Cond: ((l.id1 = r.id2) AND (l.val1 = r.val2))
               Extra Text: (seg0)   Hash chain length 1.0 avg, 1 max, using 38 of 524288 buckets.
               ->  Seq Scan on t_left l  (cost=0.00..431.00 rows=34 width=8) (actual time=0.012..0.017 rows=38 loops=1)
               ->  Hash  (cost=431.00..431.00 rows=34 width=8) (actual time=0.132..0.134 rows=38 loops=1)
                     Buckets: 524288  Batches: 1  Memory Usage: 4098kB
                     ->  Seq Scan on t_right r  (cost=0.00..431.00 rows=34 width=8) (actual time=0.098..0.104 rows=38 loops=1)
 Optimizer: GPORCA
 Planning Time: 58.042 ms
   (slice0)    Executor memory: 43K bytes.
   (slice1)    Executor memory: 4159K bytes avg x 3 workers, 4159K bytes max (seg0).  Work_mem: 4098K bytes max.
 Memory used:  128000kB
 Execution Time: 9.631 ms
(16 rows)
```

whitehawk:
> Could you please add tests for:
> 
> * `NormalizeHistogram()` with subnormal/near-zero frequency (at the 1e15 boundary).

No new tests are needed, this issue was reproduced (after applying all other fixes from this patch) with old tests `DPv2GreedyOnly` and `DPv2MinCardOnly` (with `CMAKE_BUILD_TYPE=Debug`).

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -366,9 +377,17 @@ CJoinStatsProcessor::SetResultingJoinStats(
...
+		BOOL pred_is_empty = JoinStatsAreEmpty(outer_histogram, inner_histogram,
+											   outer_histogram_after);
+		if (IsLASJ)
+		{
+			lasj_all_preds_fully_covered =
+				lasj_all_preds_fully_covered && pred_is_empty;
+		}
+		else
+		{
+			output_is_empty = output_is_empty || pred_is_empty;
```
KnightMurloc, line 389: Is there a point in calculating histograms for other predicates if one of them is empty? For the inner join case at least.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -477,17 +503,48 @@ CJoinStatsProcessor::CalcJoinCardinality(
...
-		limit_for_result_scale_factor);
+	const ULONG num_join_conds = join_conds_scale_factors->Size();
+
+	BOOL IsLASJ = (IStatistics::EsjtLeftAntiSemiJoin == join_type);
+
+	CDouble scale_factor(1.0);
+	if (IsLASJ && 0 < num_join_conds)
+	{
+		// For a LASJ, a row survives the AND-ed anti-join condition if it
+		// fails to match on ANY single predicate (survives(x) iff there is
```
KnightMurloc: A typo.
```suggestion
		// fails to match on ANY single predicate (survives(x) if there is
```

RekGRpth (dismissed):

whitehawk:
@andreyaksenov , please verify the PR description.

src/test/regress/expected/bfv_partition_plans_optimizer.out
KnightMurloc: But this query is planned by Postgres planner:
```
INFO:  GPORCA failed to produce a plan, falling back to Postgres-based planner
DETAIL:  Falling back to Postgres-based planner because GPORCA does not support the following feature: SIRV functions
```

andreyaksenov:
> @andreyaksenov , please verify the PR description.

Suggestion (close to the original):

```
Fix wrong LASJ cardinality estimation for multi-predicate joins
```

whitehawk:
> > @andreyaksenov , please verify the PR description.
> 
> Suggestion (close to the original):
> 
> ```
> Fix wrong LASJ cardinality estimation for multi-predicate joins
> ```

The solution here has a known limitation, documented in the PR description. Therefore, I used wording 'Improve', not 'Fix', as the patch doesn't fix all 100% possible cases. Having that said, do you confirm that "Fix wrong LASJ cardinality estimation for multi-predicate joins" would be still fine?

andreyaksenov:
> > > @andreyaksenov , please verify the PR description.
> > 
> > 
> > Suggestion (close to the original):
> > ```
> > Fix wrong LASJ cardinality estimation for multi-predicate joins
> > ```
> 
> The solution here has a known limitation, documented in the PR description. Therefore, I used wording 'Improve', not 'Fix', as the patch doesn't fix all 100% possible cases. Having that said, do you confirm that "Fix wrong LASJ cardinality estimation for multi-predicate joins" would be still fine?

Let's keep `Improve` given that the PR itself documents that the fix is partial. So the current title is good.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -421,6 +440,13 @@ CJoinStatsProcessor::SetResultingJoinStats(
 	}
 
 
+	// finalize output_is_empty: for LASJ, empty only if every AND-ed
+	// predicate independently showed complete coverage (AND across
+	// predicates); for other join types, keep the OR-accumulated
+	// per-predicate result plus the (unchanged) outer-empty rule.
+	output_is_empty = IsLASJ ? lasj_all_preds_fully_covered
+							 : (output_is_empty || (outer_stats->IsEmpty()));
```
KnightMurloc, line 448: Is there a point in running most of this function if `outer_stats` is empty? In that case we will fallback to the single row estimation anyway.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -477,17 +503,48 @@ CJoinStatsProcessor::CalcJoinCardinality(
...
+		// AND-ed predicates can only grow the LASJ survivor set relative
+		// to the most permissive one, never shrink it below what that
+		// predicate alone already guarantees.
+		scale_factor = (*join_conds_scale_factors)[0]->m_scale_factor;
+		for (ULONG ul = 1; ul < num_join_conds; ul++)
+		{
+			scale_factor =
+				std::min(scale_factor.Get(),
+						 (*join_conds_scale_factors)[ul]->m_scale_factor.Get());
+		}
```
KnightMurloc, line 533: I understand correctly that we finding the most strict predicate and invert its scale factor (since we pass rows that do not match the condition)? Invert in general terms, not exactly in this piece of code.

src/test/regress/expected/bfv_partition_plans_optimizer.out
whitehawk: There is query in query:
```
select count_operator('select * from (select * from p1 except all select * from p2) as p_all, t where p_all.b=t.b;','Partition Selector');
```
where `select count_operator` is indeed planned by Postgres planner, while `select * from (select * from p1 except all select * from p2) as p_all, t where p_all.b=t.b;` is planned by Orca

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
whitehawk: Updated comments

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
whitehawk: Valid observation. For non-LASJ joins, `output_is_empty` is OR-accumulated, so once one predicate is found empty the final answer is empty regardless of what's left. The complication is that the loop also populates the per-column histogram map for the output statistics object as it goes, so breaking early would leave later predicates' columns without an entry there. Therefore, it might have side-effects, and I consider it out of scope of this patch. It may be a separate improvement, but so far I don't think that it will bring good benefits (planning is one-time operation, saving several loops here without actual plan improvement) that will justify the time we'll need to invest in it.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
whitehawk: Good point. Digging it revealed pre-existing bug - I've created GG-800 ticket. It is reproduced on the pre-fix baseline. So it will be handled there, separately from this PR.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
whitehawk: In general - yes

KnightMurloc (approved):

RekGRpth (approved):

PR #597, opened 2026-09-14
GG-796 Add support for --with-product-name to ci/Dockerfile…

uaqq:
## Add support for --with-product-name to ci/Dockerfile.ubuntu

#573 added `./configure --with-product-name=<name>`, but ci/Dockerfile.ubuntu
hardcoded CONFIGURE_FLAGS, so the Docker-based CI build had no way to exercise
it - every image came out as the default "Greengage Database" build regardless
of what the caller wanted.

Adds an `EXTRA_CONFIGURE_FLAGS` build arg and threads it into
`CONFIGURE_FLAGS` via Docker's `${VAR:+...}` expansion, so callers can pass
any `./configure` flag(s) at build time, e.g.:

    docker build --build-arg EXTRA_CONFIGURE_FLAGS="--with-product-name=<name>" ...

Task: GG-796


GirgerS (approved):

ci/Dockerfile.ubuntu
```diff
@@ -33,7 +33,9 @@ RUN set -eux; \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /var/cache/man/
 
 ENV LANG=en_US.UTF-8
-ENV CONFIGURE_FLAGS="--enable-debug-extensions --with-gssapi --enable-cassert --enable-debug --enable-depend"
+
+ARG EXTRA_CONFIGURE_FLAGS
```
red1452, line 37: May be move this line to the top of file?

uaqq: That's a bad idea because we should keep ARG right before the line where it's actually used. Moving it to the top of the file will break the build cache logic. Each time we pass a new EXTRA_CONFIGURE_FLAGS, Docker will invalidate the cache and reinstall all the heavy dependencies again

red1452 (approved):

PR #601, opened 2026-09-16
Fix LASJ cardinality estimate for a pushed-down unsupported predicate

whitehawk:
Fix LASJ cardinality estimate for a pushed-down unsupported predicate

Problem description:

A LEFT JOIN whose WHERE clause filters for '<col> IS NULL' on the joined table
is implemented internally by Orca as a LASJ (Left Anti-Semi Join). Orca produced
a wrong LASJ cardinality estimate for such a query whenever its ON-clause
combines a normal equi-join predicate with a second, single-table predicate that
Orca cannot model precisely - for example a modulo expression, a scalar array
comparison whose left-hand side is an expression rather than a plain column, or
a LIKE comparison between two columns. Orca pushes that single-table predicate
down to a filter on the referenced table's own scan, ahead of the join, whether
the predicate stands alone or appears as one branch of a disjunction, or touches
one column or several. For example, for a query like:

t_left (id1, val1): id1 = i, for i in 1..100;
t_right(id2, val2): id2 = i, for i in 1..100;

select * from t_left l left join t_right r
on l.id1 = r.id2 and r.id2 % 2 = 1
where r.id2 is NULL;

"r.id2 % 2 = 1" is pushed down to a filter on t_right's own scan, rather than
evaluated as part of the join condition.

Root cause:

The pushed-down filter is a predicate Orca can't model precisely, so
'CFilterStatsProcessor' applies its selectivity by scaling the affected column's
row count down by a default guess, without narrowing the column's histogram
bucket content. The resulting histogram still spans the same value range as
before the filter, even though only part of it actually survived. Several
independent code paths could reach this state without recording which column, or
columns, it affected: the conjunction path records them, but the disjunction
path never carried that record over to the backfilled per-column histograms, and
the two sites that directly constructed an unsupported predicate for a two-
column LIKE comparison and for an unsupported array comparison never recorded
any columns at all. When 'CLeftAntiSemiJoinStatsProcessor::JoinHistogramsLASJ()'
later used one of these un-narrowed-but-unflagged histograms for its per-
predicate LASJ coverage check, it wrongly trusted the bucket content as if it
reflected the true, narrowed value range of the filtered rows.

Fix:

1. Give histograms a way to record that their row count was scaled down by a
guess without their value range actually being narrowed, so a histogram like
that can be told apart from one that genuinely reflects the filtered rows.
2. Whenever a predicate can't be modeled precisely, work out every column its
expression actually touches - not just a single column - and mark each of those
columns' histograms accordingly. Apply this consistently whether the predicate
stands alone, sits inside a conjunction or a disjunction, or takes the shape of
a two-column LIKE comparison or an array comparison whose left-hand side isn't a
plain column.
3. When the anti-semi join's cardinality logic checks whether the two join
columns' value ranges overlap, treat a marked histogram as unreliable for that
check and fall back to the same conservative default already used when the
comparison itself can't be reasoned about precisely, instead of trusting bucket
content that was never actually narrowed.

Note on supported pushed-down predicates:

A supported predicate pushed down to a pre-join filter (e.g. a plain range
comparison like 'r.id2 <= 50') is unaffected by this change. The new marking is
only applied on the predicate-construction paths described above, so a supported
filter's histogram is genuinely narrowed rather than merely scaled, and the LASJ
coverage check already trusts it correctly, both before and after this fix.

Test Changes:

1. 'LASJPredicatePushdown.mdp' - a single-table ON-clause predicate that gets
pushed down to a pre-join filter on the inner table, where that predicate is a
plain expression Orca can't model precisely (a modulo).
2. 'LASJPredicatePushdownDisjunction.mdp' - the same shape, but the pushed-down
predicate is a disjunction with a modulo branch, not just a plain modulo
expression by itself.
3. 'LASJPredicatePushdownArrayCmp.mdp' - the pushed-down predicate is a scalar
array comparison (ANY/ALL) whose left-hand side is an expression, not a plain
(possibly cast) column reference.
4. 'LASJPredicatePushdownLike.mdp' - the pushed-down predicate is a LIKE
comparison between two columns, rather than a column and a constant pattern.
5. 'LASJPredicatePushdownMultiColumnA.mdp' and
'LASJPredicatePushdownMultiColumnB.mdp' - a single pushed-down predicate that
touches two different columns; the two tests each build a LASJ keyed on a
different one of the two touched columns, verifying that the fix applies to both
columns, not just one.


src/backend/gporca/libnaucrates/src/statistics/CFilterStatsProcessor.cpp
```diff
@@ -361,6 +361,28 @@ CFilterStatsProcessor::MakeHistHashMapConjFilter(
...
+					ULONG used_colid = *(*used_colids)[uli];
+					CHistogram *used_col_histogram =
+						result_histograms->Find(&used_colid);
+					if (nullptr != used_col_histogram)
+					{
+						used_col_histogram->SetUnsupportedPredDerived();
+					}
+				}
+			}
+
```
RekGRpth, line 385: `IsUnsupportedPredDerived` flag gets wiped out within the same conjunction

**File:** `src/backend/gporca/libnaucrates/src/statistics/CFilterStatsProcessor.cpp`, method `MakeHistHashMapConjFilter` (new marking block added in hunk `@@ -361,6 +361,28 @@`, roughly lines 364–385 post-patch; the overwrite happens in the pre-existing call to `CStatisticsUtils::AddHistogram(..., true /* fReplaceOld */)` at line 398).

**Mechanism** (traced through the actual code, not speculative):

1. `CStatsPredUtils::IsUnsupportedPredOnDefinedCol()` (`CStatsPredUtils.cpp:1492`) returns true for **any** `CStatsPredUnsupported` with `GetColId() == ulong_max` — which covers every new case that carries `used_colids` (`CreateStatsPredUnsupported`, the unsupported branch of `GetStatsPredLike`, the unsupported branch of `ProcessArrayCmp`).
2. Conjunction predicates are sorted via `CStatsPred::StatsPredSortCmpFunc` (`CStatsPred.h:115`, untouched by this PR): `(INT) colid1 - (INT) colid2`. Casting `ulong_max` to `INT` yields **-1**, so predicates with no single associated column (`colid == ulong_max`) always sort **first** — ahead of any predicate with a real (positive) colid.
3. So for a query like `WHERE some_expr(a,b) = 'x' AND a > 5`: the unsupported predicate on `(a,b)` is processed first — column `a`'s histogram in `result_histograms` gets correctly marked via `SetUnsupportedPredDerived()` (this PR's new logic). Then `a > 5` is processed: `hist_before = result_histograms->Find(&a)->CopyHistogram()` correctly copies the flag, but the chain `MakeHistSimpleFilter → MakeHistPointFilter → CHistogram::MakeHistogramFilterNormalize → MakeHistogramFilter` (**none of these are touched by this PR**) builds a **brand-new** `CHistogram` object via `GPOS_NEW` rather than via `CopyHistogram()` — so the flag defaults back to `false`.
4. That fresh, now-untainted histogram replaces the marked one in `result_histograms` via the existing `AddHistogram(..., fReplaceOld=true)` call — erasing the flag the PR just set moments earlier.

**Why this defeats the PR's purpose:** if column `a` later becomes a LASJ join key, `CLeftAntiSemiJoinStatsProcessor::JoinHistogramsLASJ` will see `IsUnsupportedPredDerived() == false` and skip the new conservative fallback — reproducing the exact cardinality-estimation bug this PR is meant to fix, for the very plausible case of "an unsupported predicate plus an ordinary filter on the same column in the same AND."

Separately verified: when the unsupported predicate is the *only* predicate touching that column in the conjunction (no further filter on it), the fix does work correctly.

src/backend/gporca/libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -113,6 +113,36 @@ CStatsPredUtils::StatsCmpType(IMDId *mdid)
...
+		if (nullptr != outer_refs)
+		{
+			local_col_refs->Exclude(outer_refs);
+		}
+		local_col_refs->ExtractColIds(mp, used_colids);
+		local_col_refs->Release();
+	}
+
+	return used_colids;
+}
```
RekGRpth, line 144: **`CStatsPredUtils.cpp`, `ExtractUsedColIds`** — when `outer_refs == nullptr`, it still deep-copies the `CColRefSet` (`GPOS_NEW(mp) CColRefSet(mp, *used_col_refs)`), even though `CColRefSet::ExtractColIds()` is `const` and needs no copy — the copy is only required in the `Exclude(outer_refs)` branch. A small avoidable allocation on a potentially hot statistics path.

src/backend/gporca/libnaucrates/src/statistics/CFilterStatsProcessor.cpp
```diff
@@ -361,6 +361,28 @@ CFilterStatsProcessor::MakeHistHashMapConjFilter(
...
+				{
+					ULONG used_colid = *(*used_colids)[uli];
+					CHistogram *used_col_histogram =
+						result_histograms->Find(&used_colid);
+					if (nullptr != used_col_histogram)
+					{
+						used_col_histogram->SetUnsupportedPredDerived();
+					}
+				}
+			}
```
KnightMurloc, line 384: > their row count was
			// just reduced by the scale factor above without their
			// histogram's value range being narrowed

But can we narrow histogram by this scale factor?

KnightMurloc (commented):
Can such an incorrect histogram be used in other join types. Won't it cause problems there?

src/backend/gporca/libnaucrates/src/statistics/CStatsPredUtils.cpp
whitehawk: Updated.

src/backend/gporca/libnaucrates/src/statistics/CFilterStatsProcessor.cpp
whitehawk: Scaling the row count wouldn't help - the scale factor only tells us how many rows survived, not which values. The bucket boundaries would stay unchanged, so the histogram would still show the full original value range, just with a smaller count. LASJ's coverage check looks at value-range overlap, not row count, so it would still see full-range coverage and mis-estimate the same way. We have no way to know which sub-range actually survived, therefore we can only fall back to a conservative default here.

src/backend/gporca/libnaucrates/src/statistics/CFilterStatsProcessor.cpp
whitehawk: Updated.

whitehawk:
> Can such an incorrect histogram be used in other join types. Won't it cause problems there?

1. Other join types are less affected because the row-count estimate comes from NDV, not bucket content. Every join type besides LASJ estimates equi-join cardinality through CHistogram::MakeJoinHistogramNormalize(), using (rows1*rows2)/max(NDV1,NDV2) as the scale factor - it only cares how many distinct values each side reports, not whether the buckets were genuinely narrowed. A stale NDV from an un-narrowed histogram just biases that ratio proportionally, therefore the introduced estimation error is not so significant (empirically, the same predicate that collapsed LASJ's estimate over 10x only produces a ~20% underestimate on a regular inner join).
2. It may be possible that applying a similar mechanism could improve other joins' estimation too, but it would touch much more code - MakeJoinHistogramNormalize() is shared by every non-anti join type - without a currently evident approach for what a flagged histogram should fall back to there, which would need prior research. I think that's out of scope for this PR.

src/backend/gporca/libnaucrates/src/statistics/CFilterStatsProcessor.cpp
RekGRpth, line 547: ### 3. Nested unsupported predicate inside an OR-branch's conjunction is silently dropped from marking

**File:** `src/backend/gporca/libnaucrates/src/statistics/CFilterStatsProcessor.cpp`,
`MakeHistHashMapDisjFilter` (line 518, `unsupported_pred_used_colids` population)

The disjunction-level `CBitSet unsupported_pred_used_colids` is populated only when an unsupported
predicate is a **direct child** of the OR's predicate array (gated by
`IsUnsupportedPredOnDefinedCol(child_pred_stats)`). An unsupported multi-column predicate nested
one level deeper — inside a conjunction that is itself one OR-branch — never reaches that bitset.

Traced mechanism for `(b.k LIKE b.m AND b.j = 3) OR (b.n = 7)`:

1. `CStatisticsUtils::GetColsNonUpdatableHistForDisj` (untouched by this PR) sees that
   `b.k`/`b.m`/`b.j` are touched by only one OR-branch and marks them **non-updatable**.
2. `CStatisticsUtils::MergeHistogramMapsForDisjPreds` then *skips* non-updatable columns when
   merging, so `b.k`'s histogram — correctly marked inside the nested
   `MakeHistHashMapConjFilter` call by this PR's own conjunction-path logic — never makes it into
   `disjunctive_result_histograms`.
3. The trailing `CHistogram::AddHistograms(mp, input_histograms, disjunctive_result_histograms)`
   backfills `b.k` straight from the raw, pre-filter `input_histograms` map — unmarked.
4. The PR's own top-level backfill loop (re-marking via `unsupported_pred_used_colids`) can't
   rescue this, since that bitset was never populated for the nested case.

KnightMurloc:
> Other join types are less affected because the row-count estimate comes from NDV, not bucket content. Every join type besides LASJ estimates equi-join cardinality through CHistogram::MakeJoinHistogramNormalize(), using (rows1*rows2)/max(NDV1,NDV2) as the scale factor - it only cares how many distinct values each side reports, not whether the buckets were genuinely narrowed. A stale NDV from an un-narrowed histogram just biases that ratio proportionally, therefore the introduced estimation error is not so significant (empirically, the same predicate that collapsed LASJ's estimate over 10x only produces a ~20% underestimate on a regular inner join).

This is true only for:
https://github.com/GreengageDB/greengage/blob/3ad2a28203c491955ef106e7b33cf0d0865b7804/src/backend/gporca/libnaucrates/src/statistics/CHistogram.cpp#L1130-L1134

Do we need to take into account unsupported pred in the `NeedsNDVBasedCardEstimationForEq`?

KnightMurloc:
Can we just make empty histogram for unsupported preds instead of mark it someway?

KnightMurloc:
I found the following case. Looks like the result histogram of the inner join is broken:
```
explain analyze with cte as (
    select * from t_left l where l.id1 in (select id2 from t_right where id2 > 50)
)
select * from t1 left join cte on t1.a = cte.id1 where id1 is NULL;
                                                               QUERY PLAN                                                                
-----------------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..1293.04 rows=1 width=16) (actual time=3.030..4.338 rows=50 loops=1)
   ->  Result  (cost=0.00..1293.04 rows=1 width=16) (actual time=3.059..3.968 rows=18 loops=1)
         Filter: (l.id1 IS NULL)
         ->  Hash Left Join  (cost=0.00..1293.04 rows=34 width=16) (actual time=3.056..3.955 rows=38 loops=1)
               Hash Cond: (t1.a = l.id1)
               Extra Text: (seg0)   Hash chain length 1.0 avg, 1 max, using 20 of 262144 buckets.
               ->  Seq Scan on t1  (cost=0.00..431.00 rows=34 width=8) (actual time=0.017..0.025 rows=38 loops=1)
               ->  Hash  (cost=862.01..862.01 rows=34 width=8) (actual time=2.164..2.165 rows=20 loops=1)
                     Buckets: 262144  Batches: 1  Memory Usage: 2049kB
                     ->  Hash Semi Join  (cost=0.00..862.01 rows=34 width=8) (actual time=0.928..2.150 rows=20 loops=1)
                           Hash Cond: (l.id1 = t_right.id2)
                           Extra Text: (seg0)   Hash chain length 1.0 avg, 1 max, using 20 of 262144 buckets.
                           ->  Seq Scan on t_left l  (cost=0.00..431.00 rows=34 width=8) (actual time=0.012..0.021 rows=38 loops=1)
                           ->  Hash  (cost=431.00..431.00 rows=17 width=4) (actual time=0.064..0.064 rows=20 loops=1)
                                 Buckets: 262144  Batches: 1  Memory Usage: 2049kB
                                 ->  Seq Scan on t_right  (cost=0.00..431.00 rows=17 width=4) (actual time=0.039..0.049 rows=20 loops=1)
                                       Filter: (id2 > 50)
                                       Rows Removed by Filter: 18
```

RekGRpth:
### 1. Mark not propagated through the ordinary (non-LASJ) join path

**File:** `src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp:83` (not touched by this PR)

`CJoinStatsProcessor::JoinHistograms` only special-cases `IStatistics::EsjtLeftAntiSemiJoin`,
routing it to `CLeftAntiSemiJoinStatsProcessor::JoinHistogramsLASJ` (the function this PR patches
to check `IsUnsupportedPredDerived()`). The ordinary equi-join / left-semi-join path instead calls
`histogram1->MakeJoinHistogramNormalize(...)`, which builds a **brand-new** `CHistogram` and never
checks or propagates `IsUnsupportedPredDerived()` on either input.

**Failure scenario:** a multi-way join where a column's histogram was marked
`IsUnsupportedPredDerived()` (e.g. by a pushed-down modulo/LIKE/array-cmp filter) flows through an
intervening supported inner or left-semi equi-join before reaching a deeper LASJ. The intermediate
join silently trusts the stale, un-narrowed bucket content, and since its output histogram never
inherits the mark, the deeper LASJ no longer sees it as marked either — reproducing the exact
wrong-cardinality bug this PR set out to fix.

RekGRpth:
### 2. Unsupported predicates in the join's own ON-clause are never marked

**File:** `src/backend/gporca/libnaucrates/src/statistics/CStatsPredUtils.cpp:1416` (not touched by this PR)

`ExtractJoinStatsFromExpr` extracts the unsupported predicate's `CStatsPred` and immediately
`CRefCount::SafeRelease()`s it without marking any histogram — matches the pre-existing
`// TODO: handle unsupported predicates for LASJ, LOJ and LS joins` comment.

**Failure scenario:** `a.k = b.k AND ...` where an unsupported predicate lives directly in the
join's own ON-clause rather than as a pre-join `Select` filter (the only path this PR wires marking
into, via `CFilterStatsProcessor`). The join-key histogram is never flagged, so
`JoinHistogramsLASJ` has nothing to check and the original wrong cardinality estimate can still
occur.

RekGRpth:
### 4. `MergeMCVHist` fallback loses track of which side was unsupported-derived (low risk)

**File:** `src/backend/gporca/libnaucrates/src/statistics/CStatisticsUtils.cpp:210,213`

In the uncomparable-datum fallback, `MergeMCVHist` returns `mcv_histogram->CopyHistogram()` or
`histogram->CopyHistogram()` depending on which side has more weight — `CopyHistogram()` now also
copies `m_is_unsupported_pred_derived` (this PR's change to `CHistogram.cpp`), so the returned
histogram's mark reflects only the one side that was kept. Currently low risk since `MergeMCVHist`'s
only caller today is a unit test (`CMCVTest.cpp`); worth confirming there's no plan to feed it a
mix of marked/unmarked histograms from production code later.

whitehawk:
> Can we just make empty histogram for unsupported preds instead of mark it someway?

Marking a histogram as effectively empty routes it into a different code path with a much harsher fallback. Replaying our existing test cases will give Rows=2.44 instead of the correct Rows=50 - close to the same magnitude of collapse as the
original bug. So I think it is better keep the explicit flag than reuse empty histogram for this.

KnightMurloc:
> Marking a histogram as effectively empty routes it into a different code path with a much harsher fallback. Replaying our existing test cases will give Rows=2.44 instead of the correct Rows=50 - close to the same magnitude of collapse as the
original bug. So I think it is better keep the explicit flag than reuse empty histogram for this.

I mean use original histogram parameters but strip buckets. In my test I get opposite result, 140 result rows estimation. Which is not the best result I think...

